Issue #46, opened 2018-12-04
Publish System.Xaml as a separate nuget package.


kekekeks:
There is a lot of code in production that uses System.Xaml for non-UI purposes. It's worth as a separate package and will be useful right now. I believe it could be shipped for `netstandard2.0` to make it useful in .NET Standard libraries.

kekekeks:
#51

karelz:
Just FYI: Having System.Xaml as separate package (while it is part of Desktop UI shared framework shipping in the .NET Core platform) brings a set of non-trivial challenges (CoreFX repo had plenty of good & bad experience with them over last few years).
CoreFX team and WPF team will work closely together to fully understand all the challenges and additional costs associated with them. That will help us make the best decision if/when/how best to address this ask.

taori:
I originally thought this is what was supposed to happen [here](https://github.com/Microsoft/xaml-standard/issues), but since wpf is no open sources i guess that's a different site.

WinInsider:
This has been an ask for long time, by many... Please do consider System.Xaml target .netstandard 2.0

dotMorten:
> There is a lot of code in production that uses System.Xaml separetely from WPF on server-side for purposes that aren't related to UI in any way, shape or form.

I'm curious if someone could list a set of detailed real-world scenarios for this. It might help sell the team on it.

kekekeks:
@dotMorten 
Mostly custom workflow engines. 
There is also Workflow Foundation open-source attempt: https://github.com/dmetzgar/corewf that would greatly benefit from having System.Xaml as a separate package.

Kation:
For now Net Core 3.0  was released.
Is there any information about this issues?

idea-zone:
How can I use System.Xaml.dll in .netcore or .net5 ? I need "XmlnsDefinitionAttribute" and “XmlnsPrefixAttribute” 。Thanks。

lindexi:
@idea-zone Reference WPF framework.

Adding the code to your csproj file:

```
  <PropertyGroup>
    <UseWPF>true</UseWPF>
  </PropertyGroup>
```

idea-zone:
> @idea-zone参考WPF框架。
> 
> 将代码添加到您的 csproj 文件中：
> 
> ```
>   <PropertyGroup>
>     <UseWPF>true</UseWPF>
>   </PropertyGroup>
> ```

My project needs to be published to linux. It's will here an error and ask add "-windows" to TargetFramework .

[...]

PR #51, opened 2018-12-04
Add netstandard2.0 target framework support for System.Xaml

wieslawsoltes:
The `ValueSerializerAttribute` is taken from:
https://github.com/dotnet/corefx/blob/a10890f4ffe0fadf090c922578ba0e606ebdd16c/src/System.ObjectModel/src/System/Windows/Markup/ValueSerializerAttribute.cs

This is part of work needed to fix https://github.com/dotnet/wpf/issues/46


stevenbrix:
We're in the process of setting up a new CI build now that the project is public on GitHub. You'll need to close and re-open this PR in order to kick things off once it is set up. I'll update you when this is complete, thanks!

stevenbrix:
@wieslawsoltes this should be up and running now. can you please try closing and re-opening this pull request? Thanks!

wieslawsoltes:
@stevenbrix rebased the PR to kick the CI 😄

wieslawsoltes:
CI fails because of this:
```
018-12-04T19:06:44.9809062Z System\Xaml\Schema\TypeReflector.cs(1126,26): error CS0109: The member 'TypeReflector.ThreadSafeDictionary<K, V>.TryAdd(K, V)' does not hide an accessible member. The new keyword is not required. [F:\vsagent\61\s\src\Microsoft.DotNet.Wpf\src\System.Xaml\System.Xaml.csproj]
2018-12-04T19:06:44.9810505Z System\Xaml\Schema\TypeReflector.cs(1126,26): error CS0109: The member 'TypeReflector.ThreadSafeDictionary<K, V>.TryAdd(K, V)' does not hide an accessible member. The new keyword is not required. [F:\vsagent\61\s\src\Microsoft.DotNet.Wpf\src\System.Xaml\System.Xaml.csproj]
```

vatsan-madhavan:
@wieslawsoltes, FYI - just running build.cmd locally after cloning [netstandard2.0](https://github.com/wieslawsoltes/wpf) results in those same failures.

wieslawsoltes:
I have pushed fix for CI errors.

> @wieslawsoltes, FYI - just running build.cmd locally after cloning and checking out [netstandard2.0](https://github.com/wieslawsoltes/wpf) results in those same failures.

I get this error when run build.cmd:
```
The property 'installationPath' cannot be found on this object. Verify that the property exists.
System.Management.Automation.PropertyNotFoundException: The property 'installationPath' cannot be found on this object. Verify that the property exists.
   w System.Management.Automation.ExceptionHandlingOps.CheckActionPreference(FunctionContext funcContext, Exception exception)
   w System.Management.Automation.Interpreter.ActionCallInstruction`2.Run(InterpretedFrame frame)
   w System.Management.Automation.Interpreter.EnterTryCatchFinallyInstruction.Run(InterpretedFrame frame)
   w System.Management.Automation.Interpreter.EnterTryCatchFinallyInstruction.Run(InterpretedFrame frame)
at InitializeVisualStudioBuild, C:\DOWNLOADS\GitHub-Forks\wpf\eng\common\tools.ps1: line 93
at InitializeTools, C:\DOWNLOADS\GitHub-Forks\wpf\eng\common\tools.ps1: line 181
at <ScriptBlock>, C:\DOWNLOADS\GitHub-Forks\wpf\eng\common\Build.ps1: line 72
```

dotMorten:
Would it be better to target 2.1, so System.Xaml can start being optimized with `Span<T>` ? (considering there's lots of string parsing going on in there)

wieslawsoltes:
> Would it be better to target 2.1, so System.Xaml can start being optimized with `Span<T>` ? (considering there's lots of string parsing going on in there)

@dotMorten I think using netstandard2.0 is universal (and necessary for library developers) and you get `Span<T>` anyway when your app targets `netcoreapp2.1`.

dotMorten:
> you get Span<T> anyway when your app targets netcoreapp2.1.

But that's not what you're targeting. For `System.Xaml` itself to take advantage of the `Span<T>` APIs, it needs to target the version that adds `Span<T>`

vatsan-madhavan:
@wieslawsoltes 
> The property 'installationPath' cannot be found on this object. Verify that the property exists.

Do you have a full Visual Studio installation (2019 preview) on your development machine?

wieslawsoltes:
> @wieslawsoltes
> 
> > The property 'installationPath' cannot be found on this object. Verify that the property exists.
> 
> Do you have a full Visual Studio installation (2019 preview) on your development machine?

@vatsan-madhavan No, I use VS2017

dotMorten:
@vatsan-madhavan did you enable .NET Core preview SDK? 
image

vatsan-madhavan:
The way you have added `ValueSerializerAttribute` is perfect - the codebase already had the affordance to do what you have done - you found just the right approach!

I want to note some considerations. 

- We would want to build the same sources (modulo `ValueSerializerAttribute` and minor nits like variants of `ThreadSafeDictionary.TryAdd`) for all the targets. 
- We would like to clean up our sources and get rid of old code related to partial-trust etc. (esp. if we don't use it for netcoreapp3.0, we don't want to keep it around for other targets either). 
  - Changes (large or small) will have to result in corresponding updates to tests as well. 
  - The tests we have onboarded in our repo at this time (DrtXaml) are only a small percentage of total Xaml tests. We are working on continuously onboarding more tests 
 - There are no plans on building and shipping non-netcore products (targets) out of this repo. The goal of this project is to bring WPF to .NET Core 3.0. 
   - The practical concern is that this repo is currently set up to publish its outputs into the _.NET Core 3.0 shared framework_ through our CI system. Any attempt to build additional targets will require a reworking of packaging and creation of a new publishing pipeline to enable a separate System.Xaml package.
    - Even to enable multitargeted builds nominally, we have to do some work to ensure that those non-netcoreapp3.0 targets do not make their way into the _.NET Core 3.0 shared framework/SDK_ packages. 

In general, I think we can continue looking at this PR, but since it will require some additional work to do builds and packaging differently, I've marked it as a _future_ change. Our immediate/short-term goal is onboarding of additional WPF sources to the repo - for now that takes precedence over changes to build/packaging to enable System.Xaml on additional targets. 

cc @dotnet/wpf-developers

src/Microsoft.DotNet.Wpf/src/System.Xaml/System.Xaml.csproj
```diff
@@ -2,7 +2,7 @@
   <PropertyGroup>
     <ProjectGuid>{9AC36357-34B7-40A1-95CA-FE9F46D089A7}</ProjectGuid>
     <AssemblyName>System.Xaml</AssemblyName>
-    <TargetFrameworks>netcoreapp3.0</TargetFrameworks>
+    <TargetFrameworks>netstandard2.0;net461;netcoreapp2.0;netcoreapp2.1;netcoreapp3.0</TargetFrameworks>
     <Platforms>AnyCPU;x64</Platforms>
```
ericstj: Targeting net461 means you are trying to replace the System.Xaml which is inbox in .NETFramework.  This can be done, but would require that this assembly be versioned to 4.1.0.0 to break out of unification.  It'd also put this assembly into the "oob" state which we've been trying to avoid for a number of reasons on desktop (search for System.Net.Http in the corefx repository to get an idea).  /cc @karelz 

Targeting NETStandard2.0 and maintaining identical surface area to desktop with a placeholder in the package for desktop is another option.  This is what we do for things like Microsoft.CSharp and other assemblies.  The downside is you can only add API to "new" frameworks like NETCoreApp3.0 and netstandard2.1.

There are other options that involve renaming the assembly and using facades, but I think I'd want the WPF folks to weigh in on the direction here.

karelz: This is a non-trivial decision (experts in area need to understand all implications which are fairly complicated). A bad decision can easily bite the community, the maintainers and everyone.
We should be extra careful with these things - see my comment https://github.com/dotnet/wpf/issues/46#issuecomment-444293310
I will mark it NO MERGE until we have clear understanding here on direction. I expect it may take days/weeks (I am around CoreFX out-of-band releases for quite some time and the key thing I learned is "be scared around them").

jkoritzinsky: One option would be something similar to what @ericstj suggested with a ref-assembly and placeholder package for netstandard2.0/netcoreapp2.0 and only adding the new APIs to netcoreapp3.0 and netstandard2.1 surfaces. That would satisfy the use cases of @wieslawsoltes as well as other groups currently using 3rd-party solutions for XAML parsing off-.NET Framework.

And then the WPF team could use the .NET Standard 2.1+ and .NET Core 3.0+ targets for new API surface area, which people could move up to as they desire.

kekekeks: >ref-assembly and placeholder package for netstandard2.0

Xamarin frameworks doesn't have a proper version of System.Xaml at the moment, so it would be nice to have a proper version for netstandard2.0 build with placeholder for net461.

wieslawsoltes: > I don't get why you'd have to compile for both Netcoreapp 2.0, 2.1 and 3.0 as well as netstandard 2.0. Netstandard2.0 would cover all of these (and also as mentioned earlier I think it should be at least netstandard 2.1).

Because of issues with `ValueSerializerAttribute` type missing in those frameworks I had to add additional targets for compile and use `System.Xaml` in my tests.

src/Microsoft.DotNet.Wpf/src/System.Xaml/System/Xaml/XamlSchemaContext.cs
```diff
@@ -646,7 +646,7 @@ internal XamlType GetXamlType(Type type, string alias)
             return (XamlValueConverter<TConverterBase>)result;
         }
 
-        internal virtual XamlMember GetProperty(PropertyInfo pi)
+        public virtual XamlMember GetProperty(PropertyInfo pi)
```
karelz: All public API additions have to go through API review process first: https://github.com/dotnet/wpf/blob/master/Documentation/issue-guide.md#high-quality-features-and-api-suggestions

wieslawsoltes: Reverted changes.

Issue #57, opened 2018-12-04
Command dotnet build for WPF app fails with 'File can not be found' error on Mac


atverma:
* .NET Core Version: SDK 3.0.100-preview-009812
* OS version: macOS Mojave v 10.14.1
* Does the bug reproduce also in WPF for .NET Framework 4.8?: No

 
 
 **Problem description: After creating WPF project by running dotnet -new wpf command, dotnet build command fails with file not found errors.**
 
 **Actual behavior:** 
A sample error is displayed below. The file exists however it seems that path separator is not correct which may be causing this build failure e.g. for root path and App.xaml.

/usr/local/share/dotnet/sdk/3.0.100-preview-009812/Sdks/Microsoft.NET.Sdk.WindowsDesktop/targets/Microsoft.WinFX.targets(243,9): error BG1002: File '/Users/atverma/Documents/GitHub/DotNetCoreWPFApp/MyWPFApp\App.xaml' cannot be found. [/Users/atverma/Documents/GitHub/DotNetCoreWPFApp/MyWPFApp/MyWPFApp.csproj]
/usr/local/share/dotnet/sdk/3.0.100-preview-009812/Sdks/Microsoft.NET.Sdk.WindowsDesktop/targets/Microsoft.WinFX.targets(243,9): error BG1002: File 'MainWindow.xaml' cannot be found. [/Users/atverma/Documents/GitHub/DotNetCoreWPFApp/MyWPFApp/MyWPFApp.csproj]
/usr/local/share/dotnet/sdk/3.0.100-preview-009812/Sdks/Microsoft.NET.Sdk.WindowsDesktop/targets/Microsoft.WinFX.targets(243,9): error BG1003: The project file contains a property value that is not valid. [/Users/atverma/Documents/GitHub/DotNetCoreWPFApp/MyWPFApp/MyWPFApp.csproj]

Build FAILED.

 
 **Expected behavior:**
Expected dotnet build to succeed.
 
 **Minimal repro:**
1. On a Mac, I have installed lastest sdk from
https://dotnet.microsoft.com/download/dotnet-core/3.0
2. Executed dotnet -new wpf -o SampleWPFApp command
3. Changed current folder in terminal to SampleWPFApp
3. Executed dotnet build however it fails

cdesewell:
I don't think WPF has been/will be ported to MacOS

Does the bug reproduce also in WPF for .NET Framework 4.8 **on MacOS**?:

dotMorten:
[...]

PR #61, opened 2018-12-04
Add OS to bug report md

cdesewell:


cdesewell:
In light of [this](https://github.com/dotnet/wpf/issues/57) issue, specifying the OS may be helpful.

dotMorten:
@cdesewell ??? .NET Framework is only available on Windows.

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -5,7 +5,7 @@ about: Create a report to help us improve
 
 * .NET Core Version: (e.g. 3.0 Preview1, or daily build number, use `dotnet --info`)
 * Windows version: (`winver`)
-* Does the bug reproduce also in WPF for .NET Framework 4.8?: Yes/No
+* Does the bug reproduce also in WPF for .NET Framework 4.8 on Windows?: Yes/No
```
karelz: @cdesewell thanks, but I don't quite see how it adds more clarity.

The previous line already hints Windows-only (asking for Windows version / `winver`).
Moreover, .NET Framework 4.8 is only on Windows, so I am not sure that this will help ...

Note that [README.md](https://github.com/dotnet/wpf/blob/master/README.md#windows-presentation-framework-wpf) also clarifies we are Windows only at this point.

If we see continuous stream of misunderstandings from contributors past today, I would be ok with making some changes to the template clarifying Windows-only even more ... let's wait for a few such reports.
What do you think?

dotMorten: @karelz I do feel we need a clearer way for dependencies and nuget packages to clearly state "Windows only" which would generate a very helpful obvious error message

karelz: @dotMorten do you mean in general? Or is it specific to WPF?

dotMorten: Anything relying on the Desktop SDK. I'd also prefer that nuget packages instantly display "sure this is .NET Core 3.0 compatible, but beware that it'll only work on windows desktop"

karelz: @terrajobst any thoughts on that? We have the same problem already with some Windows Compat Pack packages.

dotMorten: This might be a bigger issue wrt to nuget that should be discussed elsewhere - being able to limit packages and frameworks to certain runtimes, and have those dependencies become viral up the dependency stack

PR #65, opened 2018-12-04
Use Auto-Properties and nameof

thomasclaudiushuber:
Cleaning up the coding style of some XAML attributes like XmlnsDefinitionAttribute.

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/System.Xaml/System/Windows/Markup/XmlnsPrefixAttribute.cs
+++ b/src/Microsoft.DotNet.Wpf/src/System.Xaml/System/Windows/Markup/XmlnsPrefixAttribute.cs
@@ -11,8 +11,6 @@
 //  Created:   04/28/2005 Microsoft
 //
 
-using System;
-using System.ComponentModel;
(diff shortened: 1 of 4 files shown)


src/Microsoft.DotNet.Wpf/src/System.Xaml/System/Windows/Markup/XmlnsPrefixAttribute.cs
```diff
@@ -49,49 +47,19 @@ public sealed class XmlnsPrefixAttribute : Attribute
...
-
-            if (prefix == null)
-            {
-                throw new ArgumentNullException("prefix");
-            }
-
-            _xmlNamespace = xmlNamespace;
-            _prefix= prefix;
+            XmlNamespace = xmlNamespace ?? throw new ArgumentNullException(nameof(xmlNamespace));
+            Prefix= prefix ?? throw new ArgumentNullException(nameof(prefix));
```
karelz: Nit:
```suggestion
            Prefix = prefix ?? throw new ArgumentNullException(nameof(prefix));
```

PR #67, opened 2018-12-04
Fields which are used in GetHashCode() should be readonly

AndreyAkinshin:


thomasclaudiushuber:
Jep, true. But I think in some places we could use Auto Properties with only a getter, like for example in [ContentWrapperAttribute.cs](https://github.com/dotnet/wpf/blob/master/src/Microsoft.DotNet.Wpf/src/System.Xaml/System/Windows/Markup/ContentWrapperAttribute.cs)
I did this in #66

karelz:
Thanks @AndreyAkinshin! Looking forward to more PRs from you 😉

PR #69, opened 2018-12-04
Better TimeSpan comparison with zero

AndreyAkinshin:
It's a bad practice to compare double values with zero.
A better way to compare TimeSpan with zero is using the Ticks property.


omariom:
Using comparison operator is even better
```C#
if (dateTime.TimeOfDay == TimeSpan.Zero)
```
https://github.com/dotnet/corefx/blob/7a37cba0884c97d7c1b1f29b79d937d986bc7492/src/Common/src/CoreLib/System/TimeSpan.cs#L534-L537

AndreyAkinshin:
@omariom, good idea! I updated the PR.

rladuca:
Looks good to me, thanks!

PR #102, opened 2018-12-06
We use language keywords instead of BCL types

lindexi:


MaherJendoubi:
Hi friends,

What's the purpose/benefit of using language keyword instead of BCL types?

Thanks,
Maher

lindexi:
Thx for @karelz .  This is the point of optimization I found during the reading process.

PR #104, opened 2018-12-06
Remove redundant explicit array type specifications

AndreyAkinshin:


src/Microsoft.DotNet.Wpf/src/System.Xaml/System/Xaml/Replacements/DateTimeOffsetConverter2.cs
```diff
@@ -55,7 +55,7 @@ public override object ConvertTo(ITypeDescriptorContext context, CultureInfo cul
 
                 Type intType = typeof(int);
                 ConstructorInfo constructor = typeof(DateTimeOffset).GetConstructor(
-                    new Type[] {
+                    new[] {
```
lindexi: Do not make the `{` in the lastest character in the line

AndreyAkinshin: @lindexi, it was the original formatting. Should I fix it as well in the current pull request?

AndreyAkinshin:
> coding-style.md

It's the code style from the CoreFX repo. I guess it makes sense to add this link to the README in dotnet/wpf.

karelz:
FYI: It is linked already from [Contributing guide](https://github.com/dotnet/wpf/blob/master/Documentation/contributing.md) - via link to [CoreFX contributing doc](https://github.com/dotnet/coreclr/blob/master/Documentation/project-docs/contributing.md):
* DO follow our coding style (C# code-specific)

WPF developer docs just started, I expect them to grow over next few months with community help and feedback.

karelz:
I am not sure I like this change. I don't think it improves code readability. Is that something recommended in our coding guidelines? Does it help perf-wise?
IMO this is on the very border of personal preference (I actually personally prefer the explicit type for readability). Personally, I would not take this PR.

mikedn:
Does it make sense to make such changes "manually"? It would make more sense to run some formatter tool on the code, once it's all out and call it day. Especially that AFAIR WPF source formatting is pretty dubious, or at least that's how it was last time I looked at the reference source code.

src/Microsoft.DotNet.Wpf/src/Shared/MS/Internal/Xaml/Parser/SpecialBracketCharacters.cs
```diff
@@ -20,7 +20,7 @@ internal class SpecialBracketCharacters : ISupportInitialize
     {
         private string _startChars;
         private string _endChars;
-        private readonly static ISet<char> _restrictedCharSet = new SortedSet<char>((new char[] { '=', ',', '\'', '"', '{', '}', '\\' }));
+        private readonly static ISet<char> _restrictedCharSet = new SortedSet<char>((new[] { '=', ',', '\'', '"', '{', '}', '\\' }));
```
walterlv:
```suggestion
        private static readonly ISet<char> _restrictedCharSet = new SortedSet<char>(new[] { '=', ',', '\'', '"', '{', '}', '\\' });
```

1. You can remove redundant parentheses.
1. You can reorder the keywords.

ryalanms:
Thank you for the contribution, @AndreyAkinshin.  Apologies for the delay in reviewing.  

I would like to keep these explicit type specifiers in for readability.  Please reopen if you feel strongly otherwise, and we can revisit the PR.  Thanks.

PR #106, opened 2018-12-06
Improve XamlType.LookupIsConstructible with the help of LINQ

AndreyAkinshin:


batzen:
I think the maintainer team should publish guidance on when to use LINQ and when not to use LINQ.
I am also a bit worried about perf when LINQ comes in.

batzen:
I also didn't want to advocate against linq. I only thought we should be careful when using linq as it might impact perf in certain scenarios.

rrelyea:
Pulling in the assembly has other costs. On phone...so hard to read the above.
Historically, wpf has worked to avoid at all levels.

Think carefully

EgorBo:
LINQ is faster in this case because the origin code should be what Any() actually does:
```csharp
using (var e = GetConstructors().GetEnumerator())
    return e.MoveNext();
```
🙂 and in this case it'd be faster than Any (no null-check, Any() won't be inlined).

stephentoub:
Hmm, actually, I misread the code.  GetConstructors is returning an enumerable rather than an array, so the allocation will still occur with the foreach.

Still, the rest of the reasons stand.

omariom:
May be `return GetConstructors().GetEnumerator().MoveNext();`

batzen:
@stephentoub I guess every WPF app uses System.Linq.dll. At least i haven't seen a single app not using Linq. So avoiding it because the assembly would have to be loaded does not make much sense from my point of view.
I did not want to start an argument about to use Linq or not to use Linq, i was just trying to point out that it shouldn't be used with thinking about it.
And as AndreyAkinshin pointed out by doing a benchmark, it doesn't hurt perf here.

stephentoub:
Does WPF itself already use Enumerable.Any? Does every WPF already use Enumerable.Any? It may not impact throughout, but it might add an additional method with an additional generic instantiation to be compiled. That's small, of course, but so is the benefit of this change.

AndreyAkinshin:
@stephentoub @rrelyea
> Second, as Rob suggests, if WPF isn't already using System.Linq.dll, this might potentially bring in a whole assembly and require code generation for it that wouldn't otherwise be needed.

It's a very good point; I didn't know that WPF doesn't use System.Linq.
If we don't want to have any references to System.Linq in the future, we need an automatic check that prevents similar changes from merging. For example, another PR from me with such changes were accepted (see https://github.com/dotnet/wpf/blob/a31b5f15bdc66ff83d65b871395d1bf26d06e84a/src/Microsoft.DotNet.Wpf/src/System.Xaml/System/Xaml/XamlObjectReader.cs#L2122), so we have a reference to System.Linq in `XamlObjectReader` (by the way, should I send another PR with reverting of these changes?)

AndreyAkinshin:
Great, thanks!

PR #120, opened 2018-12-07
Readme WPF name consistency fix

SparroHawc:
This is the only location that the full name of WPF is written out in the codebase as it currently stands.  The correct name, as it happens, is Windows Presentation Foundation.


jzebedee:
It looks like Foundation is the right usage. +1

rladuca:
@SparroHawc Oops.  Thanks!

PR #132, opened 2018-12-08
Adding editorconfig from corefx

### LINKED ISSUE #115: Add .editorconfig to clarify code style
I think it would be nice if we could get an .editorconfig in this repo so no one has to worry about how code should be formatted and tools can take care of adhering to the desired formatting.

I guess the most consistent way would be to sync this across all dotnet repos or at least use the same as in corefx for example.
https://github.com/dotnet/corefx/blob/master/.editorconfig

batzen:
This PR adds an nearly exact copy of .editorconfig taken from https://github.com/dotnet/corefx

The only things i changed are the usage of var and this.
I made those changes because, after having a look at some code samples from this repo, i noticed that those seem to be the desired configs.

Fixes #115


karelz:
Also note that `var` rules are explicitly called out in coding style (in CoreFX, linked from WPF): https://github.com/dotnet/corefx/blob/master/Documentation/coding-guidelines/coding-style.md (see no. 10)

batzen:
I just tried to not cause studio showing suggested changes everywhere. Will change it back to the corefx default.

What about "this"?

karelz:
Hmm, turns out I overlooked your comment for why you changed `var` in your PR description (it is based on status of current code base) 🙄 ... sorry for that @batzen!
I let WPF team members comment if it is intentional coding style in their code base and they want your full changes, or if they want just copy from CoreFX (changing my review)

karelz:
@batzen 

> I just tried to not cause studio showing suggested changes everywhere. Will change it back to the corefx default.

That is very good point which I missed (see above). Very sorry for that!
I suggest to wait for WPF team members comment. My suggestions might be wrong and I don't want you to change the code multiple times, just because of me 😊.

batzen:
Ok, then just ask them and if they prefer `var` they can change it.
I guess it's currently more important to at least have an editorconfig in place.

lindexi:
What do you think needs to add space after cast?

```diff
-   var f = (int)o;
+   var f = (int) o;
```

karelz:
@lindexi it is not explicitly listed in the coding style, but I have never seen space in that situation (in CoreFX and similar repos at least).
Same as no space between function name and parenthesis in function call, or function definition. And similar to always having space between `if`/`while`/`for` and parenthesis.
I assume those are treated as non-controversial defaults. Maybe VS defaults?

lindexi:
@karelz For Resharper will add the space after `cast` in default.

tannergooding:
I think, for everything not explicitly called out, we tend to follow the VS defaults (as stated in the CoreFX style guidelines).

tannergooding:
You can also see a good list of the options and their default values in the roslyn source code. Such as: http://source.roslyn.io/#Microsoft.CodeAnalysis.CSharp.Workspaces/CodeStyle/CSharpCodeStyleOptions.cs,7430792e3866637a,references and http://source.roslyn.io/#Microsoft.CodeAnalysis.Workspaces/CodeStyle/CodeStyleOptions.cs,bfdff4f8bcba6a30

walterlv:
@lindexi @tannergooding Visual Studio is a first-party IDE and ReSharper is a third-party one. I recommend following Microsoft's original coding style instead of the JetBrains' team.

batzen:
@karelz What did the team say about their "var" preference?
I'd really like to get this PR merged. If there is anything i should change, please tell me so.

vatsan-madhavan:
@batzen

Re `var`: We use it where it enhances readability and maintainability, and refrain from using it when it doesn’t add much value. 

Your current settings for `var` seem fine to me. We can tweak it later if needed. We will almost certainly modify or remove  C++ settings. 

@karelz - Do you see any concerns wrt modifying this file periodically?

stevenbrix:
Well, this is my bad. I wasn't aware of this PR and had created the .editorconfig (I too copied it directly from corefx) in our internal repository. It was pushed out when we brought the Arcade infrastructure updates to GitHub. @batzen can you resolve the conflicts so we can complete this PR? Once again, I'm really sorry about this!

batzen:
No worries.
Will resolve the conflicts during the weekend.

lindexi:
The file is conflicting.

batzen:
Totally forget that I wanted to resolve the conflicts. Will take care of that today.

batzen:
I don't know what the team really prefers as the conflicts are caused by differences in the "this" style preferences. If the team decided to prefer the setting currently being in the editorconfig we can just close thie PR as everything else is the same, aside from one comment from my editorconfig.

vatsan-madhavan:
Generally, I don’t see much use of `this` prefix in the codebase. Where we do use this style,  it is out of necessity, or it happens to be be a localized break in convention.

stevenbrix:
I agree with Vatsan, and I think how the editorconfig is now makes the most sense. I don't think there is much need to be overly concerned with it. Use `this` only where it really provides value.

rladuca:
@stevenbrix If you haven't already, can you mirror this to our internal repo?

stevenbrix:
@rladuca it already is

PR #139, opened 2018-12-10
 Remove dead code from ReflectionHelper

hughbe:
Above, we have the code
```
string[] nameFrags = typeName.Split(new char[] { ',' }, 2);
```

This guarantees that `nameFrags.Length` is maximum 2/

We know that the string can not be empty and check for length one before, so therefore this check if irrelevant

Also put some minor code cleanup in there

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/Shared/System/Windows/Markup/ReflectionHelper.cs
+++ b/src/Microsoft.DotNet.Wpf/src/Shared/System/Windows/Markup/ReflectionHelper.cs
@@ -79,60 +79,45 @@ namespace System.Xaml
 
 #region Type
         /// <summary>
-        /// Parse and get the type of the passed in string
+        /// Parse and get the type of the passed-in string.
         /// </summary>
         internal static Type GetQualifiedType(string typeName)
         {
-            // ISSUE: we only parse the assembly name and type name
-            // all other Type.GetType() type fragments (version, culture info, pub key token etc) are ignored!!!
-            string[] nameFrags = typeName.Split(new Char[] { ',' }, 2);
-            Type type = null;
+            // We only parse the assembly name and type name.
+            // All other Type.GetType() type fragments (version, culture info, public
+            // key token etc) are ignored.
+            string[] nameFrags = typeName.Split(new char[] { ',' }, 2);
             if (nameFrags.Length == 1)
             {
-                // treat it as an absolute name
-                type = Type.GetType(nameFrags[0]);
+                // Treat this as an absolute name.
+                return Type.GetType(nameFrags[0]);
             }
-            else
+
+            Assembly a = null;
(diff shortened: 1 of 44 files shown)


src/Microsoft.DotNet.Wpf/src/Shared/System/Windows/Markup/ReflectionHelper.cs
```diff
@@ -78,65 +78,48 @@ internal static Type GetQualifiedType(string typeName)
...
-            string[] nameFrags = typeName.Split(new Char[] { ',' }, 2);
-            Type type = null;
+            string[] nameFrags = typeName.Split(new char[] { ',' }, 2);
             if (nameFrags.Length == 1)
             {
                 // treat it as an absolute name
-                type = Type.GetType(nameFrags[0]);
+                return Type.GetType(nameFrags[0]);
             }
             else
```
kevingosse: Since you added a `return` above, you could remove this `else` to reduce nesting

src/Microsoft.DotNet.Wpf/src/Shared/System/Windows/Markup/ReflectionHelper.cs
```diff
@@ -72,71 +72,55 @@ static ReflectionHelper()
...
-            Type type = null;
+            // We only parse the assembly name and type name.
+            // All other Type.GetType() type fragments (version, culture info, public
+            // key token etc) are ignored.
+            string[] nameFrags = typeName.Split(new char[] { ',' }, 2);
             if (nameFrags.Length == 1)
             {
-                // treat it as an absolute name
-                type = Type.GetType(nameFrags[0]);
+                // Treat his as an absolute name.
```
stevenbrix: typo

src/Microsoft.DotNet.Wpf/src/Shared/System/Windows/Markup/ReflectionHelper.cs
```diff
@@ -193,11 +178,6 @@ internal static Type GetFrameworkType(string assemblyName, Type type)
             return type;
         }
 
-        internal static Type GetMscorlibType(Type type)
```
stevenbrix: can you add this back? it looks like this is still in use by other WPF binaries

stevenbrix: not that you could know at this point, we'll get there!

hughbe: No probs, restored with an `#if SYSTEM_XAML` check

hughbe:
Have rebased to fix the merge conflcits

PR #141, opened 2018-12-10
Remove dead checks from TypeConverterHelper

hughbe:
- `GetCoreConverterFromCoreType` is only called from `GetTypeConverter` which is only called from `ValueSerializer.GetSerializerFor`. This special-cases `typeof(string)`

```cs
public static ValueSerializer GetSerializerFor(Type type)
{
    if (type == null)
        throw new ArgumentNullException("type");
    object value = _valueSerializers[type];
    if (value != null)
        // This uses _valueSerializersLock's instance as a sentinal for null  (as opposed to not attempted yet).
        return value == _valueSerializersLock ? null : value as ValueSerializer;

    AttributeCollection attributes = TypeDescriptor.GetAttributes(type);
    ValueSerializerAttribute attribute = attributes[typeof(ValueSerializerAttribute)] as ValueSerializerAttribute;
    ValueSerializer result = null;

    if (attribute != null)
        result = (ValueSerializer)Activator.CreateInstance(attribute.ValueSerializerType);

    if (result == null)
    {
        if (type == typeof(string))
        {
            result = new StringValueSerializer();
        }
        else
        {
            // Try to use the type converter
            TypeConverter converter = TypeConverterHelper.GetTypeConverter(type);

            // DateTime is a special-case.  We can't use the DateTimeConverter, because it doesn't
            // support anything other than user culture and invariant culture, and we need to specify
            // en-us culture.
            if (converter.GetType() == typeof(DateTimeConverter2))
            {
                result = new DateTimeValueSerializer();
            }
            else if (converter.CanConvertTo(typeof(string)) && converter.CanConvertFrom(typeof(string)) &&
                        !(converter is ReferenceConverter))
            {
                result = new TypeConverterValueSerializer(converter);
            }
        }
    }
    lock (_valueSerializersLock)
    {
        // This uses _valueSerializersLock's instance as a sentinal for null (as opposed to not attempted yet).
        _valueSerializers[type] = result == null ? _valueSerializersLock : result;
    }

    return result;
}
```

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/Shared/System/Windows/Markup/TypeConverterHelper.cs
+++ b/src/Microsoft.DotNet.Wpf/src/Shared/System/Windows/Markup/TypeConverterHelper.cs
@@ -312,10 +312,12 @@ namespace System.Windows.Markup
             {
                 typeConverter = new System.ComponentModel.GuidConverter();
             }
+#if !SYSTEM_XAML
             else if (type == typeof(String))
             {
                 typeConverter = new System.ComponentModel.StringConverter();
             }
+#endif
             else if (type == typeof(CultureInfo))
             {
                 typeConverter = new System.ComponentModel.CultureInfoConverter();


vatsan-madhavan:
I'd prefer to avoid changes to /src/Shared. Every one of these changes is going to require manual fixup to closed-source sources, as well as additional code-reviews and validation to ensure that modules that depend on these sources aren't inadvertently affected. For some changes, correctness could be reasoned about locally without regard to other modules, but it still creates coherence problems. 

The closed-source version of these files are likely to be changed in the interim, and they will overwrite the versions here (unless we work on merging them across repos somehow, which I don't think we will undertake to do just-yet except for critical bug-fixes). 

We do need to document someplace what sources to avoid merging PR's for just-yet. I would categorize this PR as something to not merge for now. 

cc @dotnet/wpf-developers

karelz:
Marked as "NO MERGE" for now.
We should add the blurb into contributing docs. Although, I expect most people will miss it ... it's still worth having it there.

richlander:
Why not just rely on git merge semantics? That can be used in both directions. In fact, merging in these sources sooner will reduce the cost and risk of a larger merge later.

src/Microsoft.DotNet.Wpf/src/Shared/System/Windows/Markup/TypeConverterHelper.cs
```diff
@@ -455,10 +452,14 @@ internal static TypeConverter GetCoreConverterFromCustomType(Type type)
         /// <returns>A TypeConverter for the Type type if found. Null otherwise.</returns>
         internal static TypeConverter GetTypeConverter(Type type)
         {
+#if SYSTEM_XAML
```
stevenbrix: what does it mean now that System.Xaml will no longer throw this exception?

src/Microsoft.DotNet.Wpf/src/Shared/System/Windows/Markup/TypeConverterHelper.cs
```diff
@@ -455,10 +452,14 @@ internal static TypeConverter GetCoreConverterFromCustomType(Type type)
...
+            Debug.Assert(type != null, "Null passed for type to GetTypeConverter");
+#else
             if (type == null)
             {
                 throw new ArgumentNullException(nameof(type));
```
stevenbrix: What does it mean now that System.Xaml will no longer throw this exception?

hughbe: This is dead code in `System.Xaml.csproj` and is never reached. I didn't remove the code as it is shared. This way we explicitly acknowledge the dead code, but don't break anyone. For reasoning as to why it is dead, see description of PR

hughbe: Hi - backed out this change

hughbe:
This should be a simple PR to review now - let me know if you think the code change is needed or not. I'm not too fussed, so we can close this PR if you think.

PR #186, opened 2018-12-17
Set IsShipping as per Arcade guidance for XLIFF generation

### LINKED ISSUE #171: Start generating XLIFF data from System.Xaml
We are not currently setup to generate XLIFF data.

We need to configure the repo to do so as a first step in the localization pipeline.

rladuca:
Removing custom localization targets and setting IsShipping where app…ropriate to enable XLIFF generation.

See [Arcade issue](https://github.com/dotnet/arcade/issues/1591) for reasoning.

Fixes #171


rladuca:
After talking with @stevenbrix, removing IsShipping overrides from the build yml and checking CI build output.  The packages are binplaced into a different directory now, so we will have to check the packaging and publishing steps.

PR #199, opened 2018-12-21
Backing out problematic shared source changes

rladuca:
Certain changes to shared sources proved to be a problem when integrated into our unpublished sources.  This is due to the fact that analyzers do not have a full picture of what is needed based on the fact that the sources are not published.

@AndreyAkinshin I had to revert the changes to shared sources caused by #103.  I should have caught this prior to merging, but I am fixing them up now.  The cleanup done in System.Xaml is perfectly fine as no unpublished mirror exists for those files.  However, the Shared directory still contains files used by PresentationBuildTasks (which multi-targets with 4.7.2) and WindowsBase.  Due to this, we can cause build breaks of various kinds in our unpublished sources.  It's more expedient to revert these for now and wait until all code that requires these shared sources can be published.  Apologies for not catching this during the PR review.

@karelz and @vatsan-madhavan FYI.  We're working on a clarifying statement regarding the shared sources.

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/Shared/MS/Internal/CriticalExceptions.cs
+++ b/src/Microsoft.DotNet.Wpf/src/Shared/MS/Internal/CriticalExceptions.cs
@@ -2,10 +2,7 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
-using System.Reflection;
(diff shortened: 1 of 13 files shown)


AndreyAkinshin:
@rladuca no problem, thanks for the explanation.
Are you going to add some automatic checks for the Shared directory? I guess I'm not the last one who is trying to make any changes there.

PR #200, opened 2018-12-22
Null check missing before calling a method marked with 'NotNull' attribute

deanchalk:
Was looking through the code and noticed a missing null check where the method to be called is marked as NotNull


hughbe:
I don't see this as much of an improvement. It's also breaking:

We get the following error messages:
` System.NotSupportedException : 'DateTimeValueSerializer' ValueSerializer cannot convert from '(null)''`

Now we'd get an `ArgumentException`

Following tests confirm current behaviour:
```cs
[Theory]
[InlineData(null)]
[InlineData(1)]
public void ConvertFrom_InvalidObject_ThrowsNotSupportedException(object value)
{
    var type = new XamlType(typeof(DateTime), new XamlSchemaContext());
    TypeConverter converter = type.TypeConverter.ConverterInstance;
    Assert.Throws<NotSupportedException>(() => converter.ConvertFrom(value));
}
```

deanchalk:
community feels this isnt valid code, so closed the request

PR #203, opened 2018-12-22
Cleanup TypeConverter replacements

hughbe:
- Delete `DateTimeValueSerializerContext` as it is a) unused in the `DateTimeValueSerializer` base class and it has b) absolutely no functionality - passing `null` will suffice!
- Put `DateTimeOffsetConverter2` in the right namespace
- Light code modernization and cleanup

xref with https://github.com/dotnet/wpf/pull/200

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/System.Xaml/System/Xaml/Replacements/TypeTypeConverter.cs
+++ b/src/Microsoft.DotNet.Wpf/src/System.Xaml/System/Xaml/Replacements/TypeTypeConverter.cs
@@ -4,8 +4,8 @@
 
 using System.ComponentModel;
 using System.Globalization;
+using System.Windows.Markup;
 using System.Xaml.Schema;
-using XAML3 = System.Windows.Markup;
 
 namespace System.Xaml.Replacements
 {
@@ -15,22 +15,16 @@ namespace System.Xaml.Replacements
     internal class TypeTypeConverter : TypeConverter
     {
         public override bool CanConvertFrom(ITypeDescriptorContext context, Type sourceType)
-        {
-            return sourceType == typeof(string);
-        }
+            => sourceType == typeof(string);
 
         public override object ConvertFrom(ITypeDescriptorContext context, CultureInfo culture, object value)
         {
-            string typeName = value as string;
-
-            if (null != context && null != typeName)
+            if (context != null && value is string typeName)
             {
-                var typeResolver = GetService<XAML3.IXamlTypeResolver>(context);
-
-                if (null != typeResolver)
+                IXamlTypeResolver typeResolver = GetService<IXamlTypeResolver>(context);
+                if (typeResolver != null)
                 {
-                    Type type = typeResolver.Resolve(typeName);
-                    return type;
+                    return typeResolver.Resolve(typeName);
                 }
             }
 
@@ -38,15 +32,11 @@ namespace System.Xaml.Replacements
         }
 
         public override bool CanConvertTo(ITypeDescriptorContext context, Type destinationType)
-        {
-            return destinationType == typeof(string);
-        }
+            => destinationType == typeof(string);
 
         public override object ConvertTo(ITypeDescriptorContext context, CultureInfo culture, object value, Type destinationType)
         {
-            Type type = value as Type;
-
-            if (context != null && type != null && destinationType == typeof(string))
+            if (context != null && value is Type type && destinationType == typeof(string))
             {
                 string result = ConvertTypeToString(context, type);
                 if (result != null)
@@ -54,12 +44,13 @@ namespace System.Xaml.Replacements
                     return result;
                 }
             }
+
             return base.ConvertTo(context, culture, value, destinationType);
         }
 
         private static string ConvertTypeToString(ITypeDescriptorContext context, Type type)
         {
-            var schemaContextProvider = GetService<IXamlSchemaContextProvider>(context);
+            IXamlSchemaContextProvider schemaContextProvider = GetService<IXamlSchemaContextProvider>(context);
             if (schemaContextProvider == null)
             {
                 return null;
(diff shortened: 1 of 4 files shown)


src/Microsoft.DotNet.Wpf/src/System.Xaml/System/Xaml/Replacements/DateTimeOffsetConverter2.cs
```diff
@@ -64,11 +60,13 @@ public override object ConvertTo(ITypeDescriptorContext context, CultureInfo cul
                         intType,
                         intType,
                         typeof(TimeSpan)
+                    });
+
+                    if (constructor == null)
```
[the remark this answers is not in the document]
hughbe: @dipeshmsft what do you reckon?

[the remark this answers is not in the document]
dipeshmsft: I will take a look at this PR as well. I will ping you back once reviewed.

hughbe:
@dipeshmsft I will try to break this up to some more commits to provide more context :)

src/Microsoft.DotNet.Wpf/src/System.Xaml/System/Xaml/Replacements/TypeListConverter.cs
```diff
@@ -4,50 +4,26 @@
...
             {
-                // Consider HashMap(int, int), HashMap(int, int)
-                string[] tl = StringHelpers.SplitTypeList(typeList);
-                Type[] types = new Type[tl.Length];
-                for (int i = 0; i < tl.Length; i++)
-                {
-                    types[i] = (Type)typeTypeConverter.ConvertFrom(context, TypeConverterHelper.InvariantEnglishUS, tl[i]);
-                }
-                return types;
+                throw new NullReferenceException();
```
lindexi, line 23: I think this code may be controversial, why throw a null exception when the variable is not null?

src/Microsoft.DotNet.Wpf/src/WindowsBase/System/Windows/Markup/DateTimeConverter2.cs
```diff
@@ -45,27 +40,21 @@ public override bool CanConvertTo(ITypeDescriptorContext context, Type destinati
...
-
-
+            => _dateTimeValueSerializer.ConvertFromString( value as string, null );
 
         public override object ConvertTo(ITypeDescriptorContext context, CultureInfo culture, object value, Type destinationType)
         {
             if (destinationType != null && value is DateTime)
             {
-                _dateTimeValueSerializer.ConvertToString( value as string, _valueSerializerContext );
+                _dateTimeValueSerializer.ConvertToString(value as string, null);
```
hughbe, line 50: 🚩 this doesn't return the actual value?!

src/Microsoft.DotNet.Wpf/src/System.Xaml/System/Xaml/Replacements/DateTimeConverter2.cs
```diff
@@ -32,30 +26,27 @@ public override bool CanConvertFrom(ITypeDescriptorContext context, Type sourceT
...
-            
+
             return base.CanConvertTo(context, destinationType);
         }
 
         public override object ConvertFrom(ITypeDescriptorContext context, CultureInfo culture, object value)
-        {
-            return _dateTimeValueSerializer.ConvertFromString( value as string, _valueSerializerContext );
-        }
+            => _dateTimeValueSerializer.ConvertFromString(value as string, null);
```
hughbe, line 38: Since `DateTimeValueSerializerContext` doesn't have any functionality, we can just pass null and remove the class here and in the other `DateTimeConverter2`

src/Microsoft.DotNet.Wpf/src/System.Xaml/System/Xaml/Replacements/DateTimeOffsetConverter2.cs
```diff
@@ -27,42 +28,32 @@ public override bool CanConvertFrom(ITypeDescriptorContext context, Type sourceT
...
-                        intType,
-                        typeof(TimeSpan)
-                    }
-                    );
-
-                if (constructor != null)
+                else if (destinationType == typeof(InstanceDescriptor))
                 {
+                    // Use the year, month, day, hour, minute, second, millisecond, offset constructor
+                    ConstructorInfo constructor = typeof(DateTimeOffset).GetConstructor(new Type[]
```
hughbe, line 46: constructor will always exist, remove null check

src/Microsoft.DotNet.Wpf/src/System.Xaml/System/Xaml/Replacements/TypeListConverter.cs
hughbe: Yeah - basically this always used to throw `NullReferenceException`, because `StringHelpers.SplitTypeList` always returns null and then `tl.Length` would throw `NullReferenceException`.

Welcome comments here.

rchauhan18: Hey @hughbe ,
We are considering merging this PR. I'd like to offer a suggestion: rather than removing the code and returning ```null``` , we could set it up like this:
```public static string[] SplitTypeList(string typeList) => Array.Empty<string>(); ```

With this approach, ```StringHelpers.SplitTypeList``` will return an empty string, allowing the code to proceed. This setup also leaves the door open for us to potentially utilize this feature in the future.

Thank you for your consideration.

src/Microsoft.DotNet.Wpf/src/WindowsBase/System/Windows/Markup/DateTimeConverter2.cs
rchauhan18: @hughbe My apologies for missing this while reviewing. It appears that this should return:
    ``` return _dateTimeValueSerializer.ConvertToString(value as string, null);```
    
Regarding the build failure, we cannot define ```StringHelpers``` as private. 

If you would like, I can add a commit with the above changes, as they are very small.

rchauhan18:
Thanks @hughbe for your contribution.

PR #206, opened 2018-12-23
Remove dead code from XamlNamespace generic parsing

hughbe:
The method `TryGetXamlType(string typeName, Type[] typeArgs)` is only called from `GetXamlType(string typeName, params Type[] typeArgs()` which has the following code

```cs
public XamlType GetXamlType(string typeName, params XamlType[] typeArgs)
{
    if (!IsResolved)
    {
        return null;
    }

    string fallbackName = GetTypeExtensionName(typeName);
    if (typeArgs == null || typeArgs.Length == 0)
    {
        return TryGetXamlType(typeName) ?? TryGetXamlType(fallbackName);
    }
    else
    {
        Type[] clrTypeArgs = ConvertArrayOfXamlTypesToTypes(typeArgs);
        return TryGetXamlType(typeName, clrTypeArgs) ?? TryGetXamlType(fallbackName, clrTypeArgs);
    }
}
```

Therefore, the `typeArgs` parameter of `TryGetXamlType(string typeName, Type[] typeArgs)` is never null or empty. Therefore the code in `MangleGenericTypeName` is dead and we can collapse this function into the caller!

I've also snuck in some cleanups that make sense

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/System.Xaml/System/Xaml/Schema/XamlNamespace.cs
+++ b/src/Microsoft.DotNet.Wpf/src/System.Xaml/System/Xaml/Schema/XamlNamespace.cs
@@ -11,7 +11,7 @@ using MS.Internal.Xaml.Parser;
 
 namespace System.Xaml.Schema
 {
-    class XamlNamespace
+    internal class XamlNamespace
     {
         public readonly XamlSchemaContext SchemaContext;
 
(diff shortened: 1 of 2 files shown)


src/Microsoft.DotNet.Wpf/src/System.Xaml/System/Xaml/Schema/XamlNamespace.cs
```diff
@@ -98,43 +90,45 @@ private XamlType TryGetXamlType(string typeName)
...
-            // Can't get an array of open generic and then call MakeGenericType on it.
-            // So we need to process subscripts on generics ourselves.
+            Debug.Assert(typeArgs.Length > 0, "This method should only be called for generic types.");
+
+            // It is not possible to get an array of open generic and then call
+            // MakeGenericType on it so we need to process array subscripts.
             string subscript;
             typeName = GenericTypeNameScanner.StripSubscript(typeName, out subscript);
-            typeName = MangleGenericTypeName(typeName, typeArgs.Length);
+            typeName = typeName + KnownStrings.GraveQuote + typeArgs.Length;
```
stevenbrix: i love all these code cleanup changes! this is probably one i'd prefer to keep the MangleGenericTypeName method, as it is easier to read and get an understanding of what `typeName + KnownStrings.GraveQuote + typeArgs.Length` is doing. up to you though

hughbe: Sure. Fixed

dipeshmsft:
Thanks @hughbe for this contribution.

PR #211, opened 2018-12-26
Fix bug in `XamlTypeInvoker.GetAddMethod` if there is no such match

### LINKED ISSUE #210: XamlTypeInvoker.GetAddMethod throws ArgumentException if there is no matching type - it shouldn't
Example code:
```cs
[Fact]
public void GetAddMethod_NoSuchContentType_ReturnsNull()
{
    var invoker = new XamlTypeInvoker(new XamlType(typeof(List<int>), new XamlSchemaContext()));
    var type = new XamlType(typeof(string), new XamlSchemaContext());
    Assert.Null(invoker.GetAddMethod(type));
};
```

Expected: we return null
Actual: we throw an `ArgumentException` for duplicate value in a dictionary

### LINKED ISSUE #216: XamlTypeInvoker.UnknownInvoker.GetEnumeratorMethod throws NullReferenceException
Example:

```cs
[Fact]
public void GetEnumeratorMethod_Unknown_ThrowsNullReferenceException()
{
    XamlTypeInvoker invoker = XamlTypeInvoker.UnknownInvoker;
    Assert.Throws<NullReferenceException>(() => invoker.GetEnumeratorMethod());
}
```

- Expected: returns `null`, matching behaviour with `GetAddMethod()`
- Actual: throws `NullReferenceException`

hughbe:
## Fix 1: Use `Dictionary.TryAdd` instead of `Add` to prevent duplicates

Example code:
```cs
[Fact]
public void GetAddMethod_NoSuchContentType_ReturnsNull()
{
    var invoker = new XamlTypeInvoker(new XamlType(typeof(List<int>), new XamlSchemaContext()));
    var type = new XamlType(typeof(string), new XamlSchemaContext());
    Assert.Null(invoker.GetAddMethod(type));
};
```

Use `Dictionary.TryAdd` to avoid throwing if there is more than one value. I opted to change this to `TryAdd` rather than removing `addMethods.Add(_xamlType.ItemType, _xamlType.AddMethod);`, as firstly a user can override `LookupContentWrappers` so removing this value could be breaking, and secondly because `ContentWrappers` could contain duplicate types which could throw anyway previously

Expected: we return null
Actual: we throw an ArgumentException for duplicate value in a dictionary

Fixes #210

## Fix 2: Bail out of `GetEnumeratorMethod` if the type invoker is unknown

```
[Fact]
public void GetEnumeratorMethod_Unknown_ReturnsNull()
{
    XamlTypeInvoker invoker = XamlTypeInvoker.UnknownInvoker;
    Assert.Null(invoker.GetEnumeratorMethod());
}
```

Expected: returns null, matching behaviour with GetAddMethod()
Actual: throws NullReferenceException

Fixes #216

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/System.Xaml/System/Xaml/Schema/XamlTypeInvoker.cs
+++ b/src/Microsoft.DotNet.Wpf/src/System.Xaml/System/Xaml/Schema/XamlTypeInvoker.cs
@@ -208,6 +208,11 @@ namespace System.Xaml.Schema
 
         public virtual MethodInfo GetEnumeratorMethod()
         {
+            if (IsUnknown)
+            {
+                return null;
+            }
+
             return _xamlType.GetEnumeratorMethod;
         }
 
--- a/src/Microsoft.DotNet.Wpf/src/System.Xaml/System/Xaml/Schema/XamlTypeInvoker.cs
+++ b/src/Microsoft.DotNet.Wpf/src/System.Xaml/System/Xaml/Schema/XamlTypeInvoker.cs
@@ -180,7 +180,9 @@ namespace System.Xaml.Schema
                         _xamlType.UnderlyingType, type.UnderlyingType);
                     if (addMethod != null)
                     {
-                        addMethods.Add(type, addMethod);
+                        // Use TryAdd as AllowedContentTypes can contain
+                        // duplicate types.
+                        addMethods.TryAdd(type, addMethod);
                     }
                 }
                 _addMethods = addMethods;


src/Microsoft.DotNet.Wpf/src/System.Xaml/System/Xaml/Schema/XamlTypeInvoker.cs
```diff
@@ -42,47 +41,29 @@ public class XamlTypeInvoker
         [SecurityCritical]
         private ThreeValuedBool _isPublic;
 
-        // vvvvv---- Unused members.  Servicing policy is to retain these anyway.  -----vvvvv
```
stevenbrix: this might be fine, but i'd be a bit hesitant to remove these. @vatsan-madhavan or @rladuca do you have an idea what this servicing policy is and why these unused variables were kept in?

src/Microsoft.DotNet.Wpf/src/System.Xaml/System/Xaml/Schema/XamlTypeInvoker.cs
```diff
@@ -202,27 +189,28 @@ public virtual MethodInfo GetAddMethod(XamlType contentType)
...
                 {
-                    addMethod = CollectionReflector.GetAddMethod(
-                        _xamlType.UnderlyingType, type.UnderlyingType);
+                    addMethod = CollectionReflector.GetAddMethod(_xamlType.UnderlyingType, type.UnderlyingType);
                     if (addMethod != null)
                     {
-                        addMethods.Add(type, addMethod);
+                        // Use TryAdd as AllowedContentTypes can contain
+                        // duplicate types.
+                        addMethods.TryAdd(type, addMethod);
```
stevenbrix: are the add methods the same in this case?

src/Microsoft.DotNet.Wpf/src/System.Xaml/System/Xaml/Schema/XamlTypeInvoker.cs
```diff
@@ -241,6 +229,11 @@ public virtual MethodInfo GetAddMethod(XamlType contentType)
         /// </SecurityNote>
         public virtual MethodInfo GetEnumeratorMethod()
         {
+            if (IsUnknown)
```
stevenbrix: what's this change for?

src/Microsoft.DotNet.Wpf/src/System.Xaml/System/Xaml/Schema/XamlTypeInvoker.cs
```diff
@@ -241,6 +229,11 @@ public virtual MethodInfo GetAddMethod(XamlType contentType)
...
+            {
+                return null;
+            }
+
             return _xamlType.GetEnumeratorMethod;
```
stevenbrix:
```suggestion
            return _xamlType?.GetEnumeratorMethod;
```

src/Microsoft.DotNet.Wpf/src/System.Xaml/System/Xaml/Schema/XamlTypeInvoker.cs
stevenbrix: ahh i see, this is to fix a null reference exception. i think you could just add a null check.

src/Microsoft.DotNet.Wpf/src/System.Xaml/System/Xaml/Schema/XamlTypeInvoker.cs
hughbe: I would expect the add method to be the same when called on the same type, so yes this add method would be the same method for 
 the existing type as for the duplicate type

src/Microsoft.DotNet.Wpf/src/System.Xaml/System/Xaml/Schema/XamlTypeInvoker.cs
hughbe: I actively chose not to do this to match the implementation of `GetAddMethod` which checks for IsUnknown before calling a function on `_xamlType`

stevenbrix:
Thanks for this PR @hughbe! I also really appreciate you breaking up the commits the way you did so that we can see the bug fix vs. general code cleanup. I'm sorry that we've just left this sitting here, but there are few reasons why we are a bit hesitant on accepting these sorts of changes:

1. There are a lot of righteous syntax fixes here. While we really appreciate those, the entire WPF code base needs these improvements. In general, we want to adopt a workflow where these types of improvements are done via an automated tool of some sort, and ideally ones that use roslyn analyzers + codefixers to make these changes. The PR should then include how the tool was run. All your changes look great, so I'm not suggesting anything different be done here, but it's just an FYI. Many of these improvements can be subjective, so we'd rather have discussions on which analyzers+codefixers we are using on our codebase, rather then have developers go through the work and send out PR's, only to have it be rejected.
2. As I mentioned on one of your other PR's, we are working towards getting a more modern test infrastructure out there that new test cases like this could be added to. These PR's are great candidates to help us onboard that work once we are ready.

dipeshmsft (approved):

hughbe:
I might need to rebase this PR

dipeshmsft:
@hughbe, let's wait for the CI build to finish.

Issue #213, opened 2018-12-26
XamlLoadPermission.FromXml for element without `class` attribute throws NRE


hughbe:
```cs
[Fact]
public void FromXml_NoClassAttribute_ThrowsNullReferenceException()
{
    var permission = new XamlLoadPermission(PermissionState.None);
    var elem = new SecurityElement("IPermission");
    permission.FromXml(elem);
}
```

- Expected: throws `ArgumentException` to match other parse failures
- Actual: throws `NullReferenceException`!

## Cause

```cs

public override void FromXml(SecurityElement elem)
{
    if (elem == null)
    {
        throw new ArgumentNullException(nameof(elem));
    }
    if (elem.Tag != XmlConstants.IPermission)
    {
        throw new ArgumentException(SR.Get(SRID.SecurityXmlUnexpectedTag, elem.Tag, XmlConstants.IPermission), nameof(elem));
    }

    string className = elem.Attribute(XmlConstants.Class);
    if (!className.StartsWith(GetType().FullName, false, TypeConverterHelper.InvariantEnglishUS))
    {
        throw new ArgumentException(SR.Get(SRID.SecurityXmlUnexpectedValue, className, XmlConstants.Class, GetType().FullName), nameof(elem));
    }
```

Problem:
```cs
string className = elem.Attribute(XmlConstants.Class);
if (!className.StartsWith(GetType().FullName, false, TypeConverterHelper.InvariantEnglishUS))
```
The `className` can be null!

vatsan-madhavan:
We are going to remove `XamlLoadPermission`'s implementation from .NET Core 3.0 since CAS is no longer supported. Given this, I'm closing this issue. 

/c @ojhad

PR #214, opened 2018-12-26
Fix NRE in XamlLoadPermission.FromXml for element without `class` attribute

hughbe:
Fixes https://github.com/dotnet/wpf/issues/213

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/System.Xaml/System/Xaml/Permissions/XamlLoadPermission.cs
+++ b/src/Microsoft.DotNet.Wpf/src/System.Xaml/System/Xaml/Permissions/XamlLoadPermission.cs
@@ -17,6 +17,12 @@ namespace System.Xaml.Permissions
         private static IList<XamlAccessLevel> s_emptyAccessLevel;
         private bool _isUnrestricted;
 
+        private const string IPermissionTagName = "IPermission";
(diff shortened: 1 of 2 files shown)


src/Microsoft.DotNet.Wpf/src/System.Xaml/System/Xaml/Permissions/XamlLoadPermission.cs
```diff
@@ -17,6 +17,12 @@ public sealed class XamlLoadPermission : CodeAccessPermission, IUnrestrictedPerm
         private static IList<XamlAccessLevel> s_emptyAccessLevel;
         private bool _isUnrestricted;
 
+        private const string IPermissionTagName = "IPermission";
```
danmosemsft: Should all this CAS stuff be reduced to stubs, as we don't support CAS on .NET Core?

hughbe: Not sure. Probably not something for this PR, opened #241

stevenbrix:
Thanks @hughbe! Same situation applies here as the other PR's you've made, let's hold on this until we have better testing and we can add the test cases that you've used to validate your fix.

Issue #217, opened 2018-12-26
XamlType.Equals/XamlType.GetHashCode throws NullReferenceException for types without namespaces


hughbe:
Example:
```cs

public static IEnumerable<object[]> Equals_TestData()
{
    var type = new XamlType(typeof(int), new XamlSchemaContext());
    yield return new object[] { type, type, true };
    yield return new object[] { type, new XamlType(typeof(int), new XamlSchemaContext()), true };
    yield return new object[] { type, new XamlType(typeof(string), new XamlSchemaContext()), false };
    yield return new object[] { type, new XamlType("namespace", "name", null, new XamlSchemaContext()), false };
    yield return new object[] { type, new SubXamlType("name", null, new XamlSchemaContext()), false };

    yield return new object[] { new XamlType("namespace", "name", null, new XamlSchemaContext()), new XamlType("namespace", "name", null, new XamlSchemaContext()), true };
    yield return new object[] { new XamlType("namespace", "name", new XamlType[] { new XamlType(typeof(int), new XamlSchemaContext()) }, new XamlSchemaContext()), new XamlType("namespace", "name", new XamlType[] { new XamlType(typeof(int), new XamlSchemaContext()) }, new XamlSchemaContext()), true };
    yield return new object[] { new XamlType("namespace", "name", null, new XamlSchemaContext()), new XamlType("otherNamespace", "name", null, new XamlSchemaContext()), false };
    yield return new object[] { new XamlType("namespace", "name", null, new XamlSchemaContext()), new XamlType("namespace", "otherName", null, new XamlSchemaContext()), false };
    yield return new object[] { new XamlType("namespace", "name", null, new XamlSchemaContext()), new XamlType("namespace", "name", new XamlType[] { new XamlType(typeof(int), new XamlSchemaContext()) }, new XamlSchemaContext()), false };
    yield return new object[] { new XamlType("namespace", "name", new XamlType[] { new XamlType(typeof(int), new XamlSchemaContext()) }, new XamlSchemaContext()), new XamlType("namespace", "name", null, new XamlSchemaContext()), false };
    yield return new object[] { new XamlType("namespace", "name", new XamlType[] { new XamlType(typeof(int), new XamlSchemaContext()) }, new XamlSchemaContext()), new XamlType("namespace", "name", new XamlType[] { new XamlType(typeof(string), new XamlSchemaContext()) }, new XamlSchemaContext()), false };
[...]

PR #220, opened 2018-12-26
 Fix NullReferenceException in XamlType.Equals/XamlType.GetHashCode

hughbe:
Fixes https://github.com/dotnet/wpf/issues/217

The issue here is that `XamlType._namespaces` can actually be null, for example when a `XamlType` is constructed using the `XamlType(string typeName, IList<XamlType> typeArguments, XamlSchemaContext schemaContext` constructor


stevenbrix:
@hughbe in the issue you showed some great test code that validates this fix, we have some System.Xaml DRTs in this repo, are you able to add the tests there?

dipeshmsft (approved):

hughbe:
Thanks! No problem at all, really good to hear and I hope to contribute more soon :)

Issue #218, opened 2018-12-26
[Discussion/Cleanup] Can we remove ThreeValuedBool in favor of `bool?`


hughbe:
Simple really!

stevenbrix:
The comment above the type definition seems to indicate that it serves a purpose. However, it looks like race conditions still exist with the way it used, since no one is doing any sort of interlock compare/exchange. In fact, the comment is misleading since it seems to imply that the type itself is thread-safe. Unless I'm overlooking something obvious, I don't see any reason why not.

mikedn:
> However, it looks like race conditions still exist with the way it used, since no one is doing any sort of interlock compare/exchange. 

It doesn't need interlocked for that, at least not on x86/x64. But it does need the value to be atomic, something that nullable is not.

In general, if that code is truly used in a multithreaded manner such a change is not safe. Quite the opposite, you need to keep that enum and instead investigate adding volatile or interlocked so the code doesn't fail on ARM.

weltkante:
@mikedn but isn't the JIT/CPU allowed to reorder instructions, even on x86/x64, if you don't use volatile or interlocked (or some other kind of barrier)?

mikedn:
x86/x64 is pretty conservative in reordering loads and stores. You have a load, a store and then another load, all from the same location. Loads do not reorder, store followed by load to the same location does not reorder. So it seems like it should work fine.

If the value is atomic. Otherwise you have different locations involved and this means that some load/store reordering may occur.

The JIT is pretty conservative as well. It does load reordering and elimination but it cannot do this in this case due to the store (again, to the same location).

Of course, all this doesn't mean that not using volatile/interlocked is OK. They should be used, if only for the sake of clarity. Even on x86/64.

stevenbrix:
@mikedn thanks for pointing that out, I just meant that some sort of barrier/synchronization is needed, volatile would also potentially work. All this is speculative and depends on how it's being used. From what I see, it seems to just be used as a replacement for nullable bool, and even if there was a race condition, I don't believe there would be any side effect since both threads would come to the same answer.

stevenbrix:
thanks @mikedn for that little diagram, that makes. I had not thought about how Nullable actually stores values

hughbe:
Closing per discussion :)

PR #237, opened 2019-01-04
Remove ThreeValuedBool

hughbe:
Fixed https://github.com/dotnet/wpf/issues/218
/cc @stevenbrix

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/System.Xaml/System/Xaml/Schema/TypeBits.cs
+++ b/src/Microsoft.DotNet.Wpf/src/System.Xaml/System/Xaml/Schema/TypeBits.cs
@@ -41,14 +41,6 @@ namespace System.Xaml
         Directive    = Default,
         AllValid     = 0xFFFF << 16
     }
-
-    // Use this instead of a Nullable<bool> when a single-word read is needed for thread safety
-    internal enum ThreeValuedBool : byte
-    {
-        NotSet,
-        False,
-        True
-    }
 	
     // Thread safety: it's important that this structure remain word-sized, so that reads and
     // writes to it are atomic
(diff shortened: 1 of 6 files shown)


src/Microsoft.DotNet.Wpf/src/System.Xaml/System/Xaml/Schema/TypeBits.cs
```diff
@@ -41,14 +41,6 @@ internal enum BoolMemberBits
         Directive    = Default,
         AllValid     = 0xFFFF << 16
     }
-
-    // Use this instead of a Nullable<bool> when a single-word read is needed for thread safety
```
ptorr-msft: What is the impact on thread safety of deleting this?

hughbe: /cc @mikedn

ptorr-msft: It seems like the [most recent comment](https://github.com/dotnet/wpf/issues/218#issuecomment-452104173) implies this is a breaking change... am I reading that right? An explicit "this change is OK" or "this change is not OK" would be very helpful.

mikedn: Well, I'm not familiar with this code so I can only assume that the comment is not out of date or something and that it is indeed supposed to be thread safe. In that case, yes, this change is not safe to make.

ptorr-msft: Thank you for the additional clarification.

stevenbrix: @mikedn yeah thanks for the clarification. I agree, this does seem to be a breaking change.

hughbe: Closing :)

PR #253, opened 2019-01-10
removing specific version of microbuild so we don't timeout

### LINKED ISSUE #252: incorrect version of Microbuild plugin is being used
We are using outdated templates (tracked by #251). This issue is just to fix the version so that we use the correct microbuild and so our signed builds can complete

stevenbrix:
Fixes #252


stevenbrix:
These are changes to an Arcade file, and so should be wiped out once we get a successful drop of the latest Arcade changes (which also contain this fix)

PR #254, opened 2019-01-10
build against latest core fx packages

### LINKED ISSUE #248: WPF should build against latest CoreFx packages
Currently this repository is set up to get its NETCoreApp assemblies as an implicit framework reference from the SDK.

This project does not define a RuntimeFrameworkVersion, so it gets the default implicit framework references.
https://github.com/dotnet/wpf/blob/master/src/Microsoft.DotNet.Wpf/src/System.Xaml/System.Xaml.csproj#L5

And it's picking up the 3.0 dotnet cli + SDK that will have "latest" (that's made its way to the SDK repo).
https://github.com/dotnet/wpf/blob/b9a9a501b7118185ea31e48aa9d79b8a385e8a98/global.json#L3

However, its picking up much older versions of packages that build on top of NETCoreApp.
https://github.com/dotnet/wpf/blob/b9a9a501b7118185ea31e48aa9d79b8a385e8a98/eng/Versions.props#L7-L8

This will "work" in most cases, since we make compatible changes to the surface area, but it has a couple draw backs:
1. If the packages have implementation assemblies that aren't in the shared framework, your tests in this repo will be running with **much older** implementation than what gets bundled into the desktop shared framework.
2. When we move types down you'll get compiler errors.  See https://github.com/dotnet/corefx/issues/34447.

These dependencies need to change to be the latest packages published from CoreFx and you should have a subscription set up so that maestro auto updates them.

stevenbrix:
I've added the darc subscription for the ".NET Core 3 Dev" channel from the CoreFX repo to the WPF (this) repo. This change checks-in the files required by darc to allow Maestro to update these files and get the latest packages that come from the CoreFX build

I've also added other packages that are required for building WPF that come from the prototype branch.

Fixes #248

### DIFF
--- a/eng/Versions.props
+++ b/eng/Versions.props
@@ -4,10 +4,32 @@
     <VersionPrefix>4.8.0</VersionPrefix>
     <PreReleaseVersionLabel>prerelease</PreReleaseVersionLabel>
     <AssemblyVersion>4.0.0.0</AssemblyVersion>
-    <SystemSecurityVersion>4.5.0</SystemSecurityVersion>
-    <SystemReflectionEmitVersion>4.3.0</SystemReflectionEmitVersion>
-    <XUnitVersion>2.4.0</XUnitVersion>
+  </PropertyGroup>
+  <!-- NuGet Package Versions -->
+  <PropertyGroup>
+    <!-- Packages that come from https://github.com/dotnet/corefx -->
+    <MicrosoftWin32RegistryPackageVersion>4.6.0-preview.19060.1</MicrosoftWin32RegistryPackageVersion>
+    <SystemConfigurationConfigurationManagerPackageVersion>4.6.0-preview.19060.1</SystemConfigurationConfigurationManagerPackageVersion>
+    <SystemDiagnosticsEventLogPackageVersion>4.6.0-preview.19060.1</SystemDiagnosticsEventLogPackageVersion>
+    <SystemReflectionEmitPackageVersion>4.6.0-preview.19060.1</SystemReflectionEmitPackageVersion>
+    <SystemReflectionTypeExtensionsPackageVersion>4.6.0-preview.19060.1</SystemReflectionTypeExtensionsPackageVersion>
+    <SystemSecurityAccessControlPackageVersion>4.6.0-preview.19060.1</SystemSecurityAccessControlPackageVersion>
+    <SystemSecurityCryptographyXmlPackageVersion>4.6.0-preview.19060.1</SystemSecurityCryptographyXmlPackageVersion>
+    <SystemSecurityPermissionsPackageVersion>4.6.0-preview.19060.1</SystemSecurityPermissionsPackageVersion>
+    <SystemSecurityPrincipalWindowsPackageVersion>4.6.0-preview.19060.1</SystemSecurityPrincipalWindowsPackageVersion>
+    <SystemWindowsExtensionsPackageVersion>4.6.0-preview.19060.1</SystemWindowsExtensionsPackageVersion>
+    <!-- Packages that come from https://github.com/dotnet/corefxlab -->
+    <SystemReflectionTypeLoaderPackageVersion>0.1.0-preview2-181205-2</SystemReflectionTypeLoaderPackageVersion>
+    <!-- Maintain System.CodeDom PackageVersion at 4.4.0. See https://github.com/Microsoft/msbuild/issues/3627 -->
+    <SystemCodeDomPackageVersion>4.4.0</SystemCodeDomPackageVersion>
+    <!-- Other Packages that require manual updating-->
+    <AccessibilityPackageVersion>4.6.0-alpha-27128-8</AccessibilityPackageVersion>
+    <MicrosoftBuildFrameworkPackageVersion>15.9.20</MicrosoftBuildFrameworkPackageVersion>
+    <MicrosoftBuildUtilitiesCorePackageVersion>15.9.20</MicrosoftBuildUtilitiesCorePackageVersion>
+    <SystemReflectionPackageVersion>4.3.0</SystemReflectionPackageVersion>
+    <SystemSecurityPackageVersion>4.5.0</SystemSecurityPackageVersion>
     <XUnitRunnerConsoleVersion>2.4.0</XUnitRunnerConsoleVersion>
     <XUnitRunnerVisualStudioVersion>2.4.0</XUnitRunnerVisualStudioVersion>
+    <XUnitVersion>2.4.0</XUnitVersion>
   </PropertyGroup>
 </Project>
--- a/eng/Versions.props
+++ b/eng/Versions.props
@@ -32,4 +32,8 @@
     <XUnitRunnerVisualStudioVersion>2.4.0</XUnitRunnerVisualStudioVersion>
     <XUnitVersion>2.4.0</XUnitVersion>
   </PropertyGroup>
+  <PropertyGroup>
+    <!-- This is required to restore System.Reflection.Typeloader successfully -->
+    <RestoreSources>$(RestoreSources);https://dotnet.myget.org/F/dotnet-corefxlab/api/v3/index.json</RestoreSources>
+  </PropertyGroup>
 </Project>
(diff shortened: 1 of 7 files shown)


eng/Versions.props
```diff
@@ -4,10 +4,32 @@
     <VersionPrefix>4.8.0</VersionPrefix>
     <PreReleaseVersionLabel>prerelease</PreReleaseVersionLabel>
     <AssemblyVersion>4.0.0.0</AssemblyVersion>
-    <SystemSecurityVersion>4.5.0</SystemSecurityVersion>
-    <SystemReflectionEmitVersion>4.3.0</SystemReflectionEmitVersion>
```
ericstj: Looks like you removed this property entirely.  You need to add it back with a subscription.

```xml
    <SystemReflectionEmitVersion>4.6.0-preview.19060.1</SystemReflectionEmitVersion>
```

ericstj:
In addition to the fix I pointed out above, you may need to update the dotnet version you're building with as was mentioned here: https://github.com/dotnet/corefx/issues/34447#issue-397144351, since those new packages will depend on type in the newer Microsoft.NETCore.App version which your currently getting via implicit reference from the SDK.

ericstj:
For now you can manually update the version here: https://github.com/dotnet/wpf/blob/c921533675f8be922db448a8357adabb8a17f4cf/global.json#L9

> but if we don't take coresdk updates at the same time, it could break stuff.

It's a little worse than that.  You'll get corefx right away, but core-sdk will be a few steps behind (since at best it would be ingesting corefx at the same time you are, in reality it needs to go through core-setup first).  As a result the current way winforms and WPF are setup means you'll be using a newer set of CoreFx packages than the shared framework.  Mostly this will be OK but it could occasionally cause issues, in which case you may just wait to merge a corefx update until it flows into core-sdk and you get a build out.  Alternatively you could use RuntimeFrameworkVersion and subscribe to MS.NETCore.App out of core-setup, but this will require folks to install that when running tests (and can still be a day behind).

@livarcocc is core-sdk publishing to BAR?

eng/Versions.props
stevenbrix: I changed it to `SystemReflectionEmitPackageVersion` so that it will be properly updated by Maestro. Looks like the build failed because I forgot to update a few other files

AdamYoblick:
I checked the default channels with darc get-default-channels and I do see an entry for coresdk: https://github.com/dotnet/sdk @ refs/heads/master -> .NET Core 3 Dev

I checked what I *think* is the internal coresdk build (https://dnceng.visualstudio.com/internal/_build/results?buildId=70106&view=results). I DO see publishing of asset manifests to the build artifacts, but I do NOT see a Publish to BAR phase, so I think the answer is no.

Livar would be able to answer better :)

global.json
```diff
@@ -1,6 +1,6 @@
 {
   "tools": {
-    "dotnet": "3.0.100-preview-009764",
+    "dotnet": "3.0.100-preview-010024",
```
stevenbrix: it looks like this version of dotnet sdk requires v 16.0 of MSBuild but only 15.9.21.664 is available? @ericstj do you know why that would be? and how do i know what version to use that requires a specific version of MSBuild?

rladuca: Is that available on the build machines now?  Or will Arcade downgrade as needed?

eng/Versions.props
AdamYoblick: Also Steven, if you don't want to change the property names in your build, you can just make a property in Versions.props that matches the darc dependency name and just have your existing property wrap it. Like <MyProp>$PropFromDarc</MyProp>.

We did this for the winforms package in DotNet-Trusted to avoid breaking other stuff.

global.json
ericstj: @livarcocc / @nguerrera  how can folks build using the latest SDK + machine/wide MSBuild?  Are there different machines to use (or some property to suppress this check)?

vatsan-madhavan: There is a dev16 build machine pool that we can use AFAIK. It may not have the C++ compiler bits we WPF needs etc., but it surely has at least Preview 1. Regardless, WPF should bump vs.version to 16.0 because that's the min we support. This is separate from any other msbuild concerns.

ericstj: Indeed.  So long as you do that and you are able to ensure you're building with the locally acquired SDK (though using the machine-wide MSBuild) you should be back on track for getting updates to NETCore.App you build/test against.

eng/Versions.props
```diff
@@ -4,10 +4,36 @@
...
+    <SystemWindowsExtensionsPackageVersion>4.6.0-preview.19060.1</SystemWindowsExtensionsPackageVersion>
+    <!-- Packages that come from https://github.com/dotnet/corefxlab -->
+    <SystemReflectionTypeLoaderPackageVersion>0.1.0-preview2-181205-2</SystemReflectionTypeLoaderPackageVersion>
+    <!-- Maintain System.CodeDom PackageVersion at 4.4.0. See https://github.com/Microsoft/msbuild/issues/3627 -->
+    <SystemCodeDomPackageVersion>4.4.0</SystemCodeDomPackageVersion>
+    <!-- Other Packages that require manual updating-->
+    <AccessibilityPackageVersion>4.6.0-alpha-27128-8</AccessibilityPackageVersion>
+    <MicrosoftBuildFrameworkPackageVersion>15.9.20</MicrosoftBuildFrameworkPackageVersion>
+    <MicrosoftBuildUtilitiesCorePackageVersion>15.9.20</MicrosoftBuildUtilitiesCorePackageVersion>
+    <SystemReflectionPackageVersion>4.3.0</SystemReflectionPackageVersion>
```
ericstj: You shouldn't need these old versions.

eng/Versions.props
```diff
@@ -4,10 +4,36 @@
...
+    <SystemReflectionEmitPackageVersion>4.6.0-preview.19060.1</SystemReflectionEmitPackageVersion>
+    <SystemReflectionTypeExtensionsPackageVersion>4.6.0-preview.19060.1</SystemReflectionTypeExtensionsPackageVersion>
+    <SystemSecurityAccessControlPackageVersion>4.6.0-preview.19060.1</SystemSecurityAccessControlPackageVersion>
+    <SystemSecurityCryptographyXmlPackageVersion>4.6.0-preview.19060.1</SystemSecurityCryptographyXmlPackageVersion>
+    <SystemSecurityPermissionsPackageVersion>4.6.0-preview.19060.1</SystemSecurityPermissionsPackageVersion>
+    <SystemSecurityPrincipalWindowsPackageVersion>4.6.0-preview.19060.1</SystemSecurityPrincipalWindowsPackageVersion>
+    <SystemWindowsExtensionsPackageVersion>4.6.0-preview.19060.1</SystemWindowsExtensionsPackageVersion>
+    <!-- Packages that come from https://github.com/dotnet/corefxlab -->
+    <SystemReflectionTypeLoaderPackageVersion>0.1.0-preview2-181205-2</SystemReflectionTypeLoaderPackageVersion>
+    <!-- Maintain System.CodeDom PackageVersion at 4.4.0. See https://github.com/Microsoft/msbuild/issues/3627 -->
```
ericstj: You may want to avoid hardcoding this here and instead do it in the build-task project itself.  Then have your tests reference the latest.

vatsan-madhavan: Or perhaps define another property, $(SystemCodeDomPackageVersionForPresentationBuildTasks), and keep it constant. Then we can subscribe for updates to $(SystemCodeDomPackageVersion) - which will be the default used everywhere except PBT. This way we'd have all these versions defined in the same place, while also ensuring that the latest version System.CodeDom is used everywhere except in PBT.

eng/Versions.props
stevenbrix: sounds good, thanks!

stevenbrix:
Thanks everyone!

Closes #248

AdamYoblick:
@stevenbrix Did you remember to create the maestro subscription from corefx -> wpf so these changes will flow automatically?

global.json
```diff
@@ -1,8 +1,8 @@
...
     "vs": {
-      "version": "15.9"
+      "version": "16.0"
     }
   },
   "sdk": {
```
AdamYoblick: @stevenbrix Pretty sure you want the sdk version to match the dotnet version, no?

AdamYoblick:
Also @stevenbrix I spoke with @mmitche about where the dependencies should go (in Version.Details.xml) and he said they stop graph traversal under toolset. So if you want to be able to create a dependency graph for wpf, you should put all your dependencies (except things like coresdk and arcadesdk) under ProductDependencies instead of ToolsetDependencies.

Functionally (in terms of darc updates) it doesn't make any difference though, but it might later on.

PR #261, opened 2019-01-13
Add License Badge

alxnull:
Many popular repositories have a license badge at the top of the readme (e.g. https://github.com/dotnet/winforms/pull/258), so I thought we could also add one here.


vatsan-madhavan:
Can you help us understand your choice of badge color (blue)?

karelz:
WinForms uses green color: https://github.com/dotnet/winforms/blob/master/README.md

alxnull:
There is no special reason for choosing blue, MIT license badges mostly seem to be either green or blue (https://github.com/Microsoft/ChakraCore for example). I changed it to green to fit with the WinForms one.

Issue #268, opened 2019-01-16
A recent change in XamlObjectReader introduced the use of Linq, which is unnecessary


danzil:
* .NET Core Version: 3.0.100-preview-010024
* Windows version: 17763
* Does the bug reproduce also in WPF for .NET Framework 4.8?: No

**Problem description:**
A recent change in XamlObjectReader #73 introduced the use of Linq, which is unnecessary and ads extra dll loads.

karelz:
Fixed in #269

PR #269, opened 2019-01-16
Revert back to not using Linq, which ads unnecessary dll loads

danzil:
This addresses #268 by reverting back to the code that was there before the change.


karelz:
@danzil if you use "Fixes #268" in top post, GitHub will automatically close the associated issue upon merge - see [PR guidance](https://github.com/dotnet/wpf/blob/master/Documentation/issue-guide.md#pr-guidance) [4].

FYI @AndreyAkinshin as he was involved in the change and further discussion

danzil:
Thanks @karelz, good tip!

Issue #283, opened 2019-01-22
WPF's transport package has dependency on System.Security.Permissions


ericstj:
**Problem description:**
WPF's transport packages should be binaries only.  The version of the dependencies is meant to be decided in dotnet-trusted when the shared framework is built.

 **Actual behavior:** 
Restores WPF binaries and some CoreFx dependencies (the closure of System.Security.Permissions).
 
 **Expected behavior:**
Only restores WPF binaries
 
 **Minimal repro:**
create a project and reference:
```xml
<PackageReference Include="runtime.win-x64.Microsoft.DotNet.Wpf.Private" Version="4.8.0-prerelease.19064.4" />
```

The fix here is to set PrivateAssets=All on all PackageDependencies in the project which produces your transport package.   Since that happens to be System.Xaml today, it needs to be done there.

rladuca:
No longer applicable.

PR #284, opened 2019-01-22
Setting PrivateAssets=All in PackageReference

vatsan-madhavan:
Fixing #283

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/System.Xaml/System.Xaml.csproj
+++ b/src/Microsoft.DotNet.Wpf/src/System.Xaml/System.Xaml.csproj
@@ -99,6 +99,8 @@
     
   </ItemGroup>
   <ItemGroup>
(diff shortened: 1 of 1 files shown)


src/Microsoft.DotNet.Wpf/src/System.Xaml/System.Xaml.csproj
```diff
@@ -99,6 +99,8 @@
     
   </ItemGroup>
   <ItemGroup>
-    <PackageReference Include="System.Security.Permissions" Version="$(SystemSecurityPermissionsPackageVersion)" />
+    <PackageReference Include="System.Security.Permissions" Version="$(SystemSecurityPermissionsPackageVersion)">
+      <PrivateAssets>All</PrivateAssets>
```
ericstj: nit: you can set this using an attribute, or better make an item definition so future additions don't cause this problem.  May be a moot point if you plan to replace this with a different transport package project in near future.

PR #305, opened 2019-01-31
Test projects, and test documentation

### LINKED ISSUE #304: Add TestHost, test documentation, and a project to contribute unit tests to
In order to facilitate testing we need the following:

- A TestHost to run tests against
- A test project shell for tests to be added to
- Documentation on all of the above (ongoing)

This is related to #180 and is part of that work (though more needs doing there)

rladuca:
Adding System.Xaml unit test project for contributions
Adding some test documentation

Fixes #304


rladuca:
There is an issue with the extension point being used from Arcade and the CI Build.

The CI build is passing -test along with the rest of the params.  Due to this, we can't use configure-toolset.ps1 predicated on the test flag in order to setup the testhost for later use.  We'll have to find another way to insert ourselves into the build process prior to MSBuild being called so we can setup the correct dotnet tools.

rladuca:
Edited the build yml in order to split what used to be a single call to eng/common/cibuild.cmd into multiple steps.

These steps allow us to build the testhost as expected and then switch to using dotnet.exe as the build environment prior to running the test portion.  The rest of the steps still use MSBuild and it all seems to work as expected.

eng/configure-toolset.ps1
```diff
@@ -0,0 +1,58 @@
+# If we are running a test pass, ensure that we point at the TestHost and that
+# tests are run against it.
+if ($test)
```
stevenbrix: where is test defined/set? what exactly is "test" (is it a bool?). I believe this will return true if test is a Boolean set to $false

stevenbrix: the same goes for all checks you have that are like this. i think there are better/safer powershell symantex for checking if a variable is defined

rladuca: Test is a switch defined in eng/common/build.ps1.  It's always true or false and works as expected.  You can tell because this never runs when the build is not explicitly called with -test.  In fact, this running during the build phase was what the original error that the yml change fixes.  So we've validated this inside and out at this point.

rladuca: I think adding a comment here to say where $test comes from is a good idea though.  Perhaps a better discussion of how this is an extension script from Arcade's build.ps1.

rladuca: Let me clarify above, its always $true or $false, so we're not going to run into the situation where if ($test) evaluates to true just because it is defined.

src/Microsoft.DotNet.Wpf/src/testhost/testhost.csproj
```diff
@@ -0,0 +1,43 @@
+<!-- 
+    This project exists solely to manage the DotNet TestHost.
+    The TestHost is a local installation of .NET Core that houses everything needed to run a WPF application against
+    the assemblies building out of this repo.
+    
+    See TestHostBinPlacing.targets for information on how projects binplace into the TestHost.
+-->
+<Project Sdk="Microsoft.NET.SDK">
```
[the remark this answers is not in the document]
rladuca: Oh well that's interest, I'll update.

eng/configure-toolset.ps1
rladuca: In terms of the null checks, I should probably add `-ne $null` in those instances, I'll make that change.

rladuca: Made a few changes to params. comments, and usage throughout the script.  When there are vars that are null checked I made them explicit, otherwise I ensured that some vars are switches in order to more explicitly call out that they are always set.

src/Microsoft.DotNet.Wpf/src/testhost/testhost.csproj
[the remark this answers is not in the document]
rladuca: I'll have to check into this.  Adding this as the SDK gives errors that the SDK doesn't exist in the local toolset installation of .NET Core.  This persists even if I update the dotnet version to be the latest daily (though that adds other breaking changes for arcade).  I also tried just PackageReferencing this, but that doesn't work either.  It seems we need the SDK present in the local toolset installation as far as I can tell.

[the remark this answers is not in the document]
vatsan-madhavan: I’ve tried using it in one of our internal prototype repos successfully - FYI. No reason it shouldn’t work here.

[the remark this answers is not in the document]
rladuca: Hmm, what I am seeing is the following:

`
D:\wpf_open\src\Microsoft.DotNet.Wpf\src\testhost\testhost.csproj : error : D:\wpf_open\.dotnet\sdk\3.0.100-preview4 -010279\Sdks\Microsoft.Build.NoTargets\Sdk not found. Check that a recent enough .NET Core SDK is installed and/or i ncrease the version specified in global.json.
`

When I bump to latest daily via global.json, I still get this error and a whole host of others.  I even tried installing the latest daily to my machine and directly running this from MSBuild with the same results save pointing to the shared dotnet install.

Do you recall if there was some special install step you did to get this SDK to appear where it should?  It's definitely not in `C:\Program Files\dotnet\sdk\3.0.100-preview4-010289\Sdks` on my machine.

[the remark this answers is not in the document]
vatsan-madhavan: Add NoTargets Sdk to global.json first, in the Sdk section.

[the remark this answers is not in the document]
rladuca: Got it, wasn't thinking along those lines, all builds now, thanks!

rladuca:
We've marked this as NO MERGE because we have some internal engineering work to move to a new build system and this change has not been modified to fully support that.  We don't want to merge a system that may end up changing within the next month or so due to modifications that may need to occur for the new build system.  We're going to work on porting this over and then we'll bring this back around.

rladuca:
No longer relevant

Issue #170, opened 2018-12-14
Make MarkupExtension an interface


YannikZ12:
I believe that making MarkupExtension an interface enables developers to work with (Static/Dynamic)Resources more flexibly. In my case it has hurt the creation of ResourceDictionaries with simple types like Double (e.g. font sizes or the like) or fonts in general. 

The implementation of the class is already really close to that of an interface so it should not be too much work to make the change.

Example:
```
<TypeOfIMarkupExtension x:Key="GlobalFontSize" Value="{Binding BindingToMyClass}"/>

<TextBlock FontSize="{StaticResource GlobalFontSize}"/>
```

h82258652:
agree. ```MarkupExtension``` is not inherbit ```DependencyObject```, use interface is better than abstract class.

dotMorten:
I'm curious how you'd justify the breaking changes this would create? Aren't breaking API changes basically dead on arrival? Is there another way this could be accomplished without breaking anything? (otherwise I doubt this issue would get much support from the WPF team, and anyone else outside who relies on things not breaking on them).

From your PR it looks like the breaking change is isolated to `XamlSetMarkupExtensionEventArgs.cs`. Perhaps injecting a base class between this class and `XamlSetValueEventArgs` that adds the interface version, and leave `XamlSetMarkupExtensionEventArgs` alone (apart from now inheriting from the intermediate class)

YannikZ12:
First, I thouth this is a new Implementation, and their is no offical Release yet. Then it is not a breaking change. 
Second I would just keep the `MarkupExtension` Class, but use everywhere the `IMarkupExtension` interface.
@dotMorten  Other ways to accomplish that, is to inherit `MarkupExtension` from `DependencyObject`, but I think that is the worst way.

stevenbrix:
@YannikZ12, you are right that it isn't *technically* a breaking change, but we want to ensure that migration from .NET Framework to .NET Core is as easy as possible for everyone.  In that regard, we are considering anything that would change the runtime behavior and/or compile time behavior between .NET Framework and .NET Core a breaking change.  In theory, developers should be able to just retarget their applications for .NET Core and not need to change anything. I haven't fully thought about the implications of this change, are you sure that we can withold that guarantee here?

YannikZ12:
Hello @stevenbrix, 
my colleague @Dabbel created a Fork and made some Changes to ensure no Breaking change for any Developer.

[...]

PR #306, opened 2019-01-31
MarkupExtension use interface instead of abstract class

h82258652:
[issue](https://github.com/dotnet/wpf/issues/170)

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/System.Xaml/System/Windows/Markup/XamlSetMarkupExtensionEventArgs.cs
+++ b/src/Microsoft.DotNet.Wpf/src/System.Xaml/System/Windows/Markup/XamlSetMarkupExtensionEventArgs.cs
@@ -9,17 +9,17 @@ namespace System.Windows.Markup
     public class XamlSetMarkupExtensionEventArgs : XamlSetValueEventArgs
     {
         public XamlSetMarkupExtensionEventArgs(XamlMember member,
-            MarkupExtension value, IServiceProvider serviceProvider) :
+            IMarkupExtension value, IServiceProvider serviceProvider) :
             base(member, value)
         {
             ServiceProvider = serviceProvider;
         }
 
-        public MarkupExtension MarkupExtension { get { return Value as MarkupExtension; } }
+        public IMarkupExtension MarkupExtension { get { return Value as IMarkupExtension; } }
         public IServiceProvider ServiceProvider { get; private set; }
 
         internal XamlSetMarkupExtensionEventArgs(XamlMember member,
-            MarkupExtension value, IServiceProvider serviceProvider, object targetObject)
+            IMarkupExtension value, IServiceProvider serviceProvider, object targetObject)
             : this(member, value, serviceProvider)
         {
             TargetObject = targetObject;
--- a/src/Microsoft.DotNet.Wpf/src/System.Xaml/System/Xaml/RefOnly/LooseTypeExtensions.cs
+++ b/src/Microsoft.DotNet.Wpf/src/System.Xaml/System/Xaml/RefOnly/LooseTypeExtensions.cs
@@ -50,11 +50,11 @@ namespace System.Xaml
             AssemblyName name1, AssemblyName name2)
         {
             AssemblyName windowsBaseName = null;
-            if (name1.Name == WindowsBase && a2 == typeof(MarkupExtension).Assembly)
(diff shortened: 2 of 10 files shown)


walterlv:
That's a great improvement that we can make an existed class used like a MarkupExtension.

src/Microsoft.DotNet.Wpf/src/System.Xaml/System/Xaml/RefOnly/LooseTypeExtensions.cs
```diff
@@ -50,11 +50,11 @@ internal static bool AssemblyQualifiedNameEquals(Type t1, Type t2)
             AssemblyName name1, AssemblyName name2)
         {
             AssemblyName windowsBaseName = null;
-            if (name1.Name == WindowsBase && a2 == typeof(MarkupExtension).Assembly)
+            if (name1.Name == WindowsBase && a2 == typeof(IMarkupExtension).Assembly)
```
walterlv: Here only needs an assembly instance and there is no need to migrate `MarupExtension` to `IMarkupExtension`.

h82258652: Done. thanks for review.

YannikZ12:
Why u didn't use MarkupExtension now in all Classes? I think you need to check for IMarkupExtension everywhere. @h82258652

vatsan-madhavan:
This is a design change along with the  addition of a public type. I would recommend opening an issue for this and discussing the design, so that we could ensure that this is the approach we want to take. It will also allow interested folks to chime in with their input. 

Marking this as *NO MERGE* for now, until we have consensus on the approach.

src/Microsoft.DotNet.Wpf/src/System.Xaml/System/Windows/Markup/XamlSetMarkupExtensionEventArgs.cs
```diff
@@ -9,17 +9,17 @@ namespace System.Windows.Markup
...
             ServiceProvider = serviceProvider;
         }
 
-        public MarkupExtension MarkupExtension { get { return Value as MarkupExtension; } }
+        public IMarkupExtension MarkupExtension { get { return Value as IMarkupExtension; } }
         public IServiceProvider ServiceProvider { get; private set; }
 
         internal XamlSetMarkupExtensionEventArgs(XamlMember member,
-            MarkupExtension value, IServiceProvider serviceProvider, object targetObject)
+            IMarkupExtension value, IServiceProvider serviceProvider, object targetObject)
```
dotMorten: Breaking change

src/Microsoft.DotNet.Wpf/src/System.Xaml/System/Windows/Markup/XamlSetMarkupExtensionEventArgs.cs
```diff
@@ -9,17 +9,17 @@ namespace System.Windows.Markup
...
         public XamlSetMarkupExtensionEventArgs(XamlMember member,
-            MarkupExtension value, IServiceProvider serviceProvider) :
+            IMarkupExtension value, IServiceProvider serviceProvider) :
             base(member, value)
         {
             ServiceProvider = serviceProvider;
         }
 
-        public MarkupExtension MarkupExtension { get { return Value as MarkupExtension; } }
+        public IMarkupExtension MarkupExtension { get { return Value as IMarkupExtension; } }
```
dotMorten: Breaking change.
Perhaps injecting a base class between this class and XamlSetValueEventArgs that adds the interface version, and leave XamlSetMarkupExtensionEventArgs alone (apart from now inheriting from the intermediate class).

src/Microsoft.DotNet.Wpf/src/System.Xaml/System/Windows/Markup/XamlSetMarkupExtensionEventArgs.cs
```diff
@@ -9,17 +9,17 @@ namespace System.Windows.Markup
     public class XamlSetMarkupExtensionEventArgs : XamlSetValueEventArgs
     {
         public XamlSetMarkupExtensionEventArgs(XamlMember member,
-            MarkupExtension value, IServiceProvider serviceProvider) :
+            IMarkupExtension value, IServiceProvider serviceProvider) :
```
dotMorten: Breaking change

PR #341, opened 2019-02-11
adding back platform builds

### LINKED ISSUE #339: we are no longer building x64 packages
this was regressed by #264

stevenbrix:
when I made the change to move to the new jobs.yml arcade templates, I accidently removed the platform specific configurations. this pr adds them back

Fixes #339

### DIFF
--- a/eng/pipeline.yml
+++ b/eng/pipeline.yml
@@ -41,6 +41,8 @@ jobs:
           value: ''
         - name: _Platform
           value: x86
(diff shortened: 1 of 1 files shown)


vatsan-madhavan:
Please set platform explicitly for x86 builds

vatsan-madhavan:
> Please set platform explicitly for x86 builds

Never mind - I see that the default is already x86.

PR #353, opened 2019-02-14
Update .NET Core SDK link.

KexyBiscuit:


KexyBiscuit:
https://github.com/dotnet/winforms/pull/442

README.md
```diff
@@ -15,7 +15,7 @@ See the [WPF Roadmap](roadmap.md) to learn about project priorities, status and
 
 ## Getting started
 
-* [.NET Core 3.0 SDK Preview 2](https://www.microsoft.com/net/download)
+* [.NET Core SDK 3.0 Preview](https://dotnet.microsoft.com/download/dotnet-core/3.0)
```
vatsan-madhavan: If we want to read without a Preview version number, with still include a version for the SDK itself, the 3.0 version should attach to `.NET Core` - i.e., it should read as '.NET Core 3.0 SDK Preview'.

vatsan-madhavan: Core SDK is available here - https://github.com/dotnet/core-sdk

KexyBiscuit: In release notes, it's written like .NET Core SDK 2.2.200 Preview, see https://github.com/dotnet/core/blob/master/release-notes/2.2/preview/2.2.200-preview.md.

vatsan-madhavan: I think it reads better to write _.NET Core 3.0_ when writing about a non-specific major version (like _3.0_) that is also used commonly to refer to the product itself. 

As you have rightly pointed out, when specifying the version in detail like `3.0.100-preview4-010334`, it certainly works better write it as `.NET Core [SDK|Runtime] 3.0.100-preview4-010334`. 

In the release notes you have linked, we can see both kinds of usage.

PR #373, opened 2019-02-21
Set copyright and package license

tmat:
Prepares repo for change https://github.com/dotnet/arcade/pull/2003 by setting `Copyright` and `PackageLicenseExpression` properties. These values will be required to be set by each repository once https://github.com/dotnet/arcade/pull/2003 is merged.

In order to not break the current builds this change sets the properties conditionally. This condition can be removed once all repos switch to Arcade that has https://github.com/dotnet/arcade/pull/2003.

@markwilkie

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -1,15 +1,15 @@
 <?xml version="1.0" encoding="utf-8"?>
 <Project>
-  <PropertyGroup>
-    <ImportNetSdkFromRepoToolset>false</ImportNetSdkFromRepoToolset>
-    
-    <PublishWindowsPdb>true</PublishWindowsPdb>
-  </PropertyGroup>
-
-  <Import Project="$(MsBuildThisFileDirectory)SystemResources.props"/>
   <Import Project="Sdk.props" Sdk="Microsoft.DotNet.Arcade.Sdk" />
+  <Import Project="SystemResources.props"/>
+
+  <PropertyGroup Condition="'$(CopyrightNetFoundation)' != ''">
+    <Copyright>$(CopyrightNetFoundation)</Copyright>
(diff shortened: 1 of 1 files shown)


Directory.Build.props
```diff
@@ -1,15 +1,15 @@
 <?xml version="1.0" encoding="utf-8"?>
 <Project>
-  <PropertyGroup>
-    <ImportNetSdkFromRepoToolset>false</ImportNetSdkFromRepoToolset>
-    
-    <PublishWindowsPdb>true</PublishWindowsPdb>
-  </PropertyGroup>
-
-  <Import Project="$(MsBuildThisFileDirectory)SystemResources.props"/>
   <Import Project="Sdk.props" Sdk="Microsoft.DotNet.Arcade.Sdk" />
```
vatsan-madhavan: I don't recall whether SystemResources.props really needs to come before Arcade/Sdk.props  - perhaps we don't need to reorder the props just now.

tmat: The Arcade [recommendation](https://github.com/dotnet/arcade/blob/master/Documentation/ArcadeSdk.md#directorybuildprops) is to import Sdk.props first:

> The Sdk.props file sets various properties and item groups to default values. It is recommended to perform any customizations after importing the SDK.

Should I file an issue to do so instead?

tmat: Since SystemResporces.props only defines EmbeddedResource ItemDefinitionGroup, I don't see why  it would need to be before Sdk.props import.

vatsan-madhavan: Looking through the binlogs from the CI build, its clear that the change  has no adverse effect - you are absolutely right.

tmat:
Thanks for double checking!

PR #414, opened 2019-03-07
LOC CHECKIN: Update localization files to align with .NET Framework

ryalanms:
.NET Framework contains a number of internal localization files in different formats.  These files were combined, checked for consistency, translated to xlf format, and compared against the existing xlf files in .NET core.  This change contains the first updated language translation file (de) that had additional and/or conflicting content with existing language translation files in .NET core.


ryalanms:
xlfs must conform to shared strings.resx file.

vatsan-madhavan:
@ryalanms, Am I correct in understanding that the latest commit acheives the following: 

 -  Wherever possible, take the localized copy of a string from .NET Framework's localization db and bring it over here. 
  - If it were not possible to bring over a localized string from .NET Framework's localization db (presumably because it did not have a localized copy of the string, for historical reasons), then leave the localized string intact (i.e., it remains the same as the unlocalized neutral language (`en`, in this case) string.

ryalanms:
@vatsan-madhavan: Yes, that is correct.

ryalanms:
Also, strings that were not translated for any language have already been removed from strings.rex and placed directly in the source.

PR #423, opened 2019-03-08
LOC CHECKIN: Strings not intended to be translated should not be included in System.Xaml satellite assemblies

ryalanms:
Untranslated strings should not be included in System.Xaml satellite assemblies.  Remove strings from shared Strings.resx, add strings to XamlPullParser, and update XLF files for all languages.


vatsan-madhavan:
What do you mean by "untranslated strings" ? Do you mean "strings that happened to be untranslated because we couldn't find sufficient localization assitance to do the translation", or do you mean "strings that were never intended to be translated, and therefore should never have appeared in any of the resx files (and should be moved to code-behind as const strings" ?

ryalanms:
Updated title to more accurately reflect the intent of the PR.

PR #428, opened 2019-03-09
Update VS2019 link in README.md

Garfield550:
VS2019 is in RC now, no need for downloading preview version.


dotMorten:
There is a big difference though: with the preview version, .net core 3.0 "just works" whereas with the RC version you have to go to settings and enable .net core preview support.
It would be good to call this out.

Garfield550:
@dotMorten Thanks for pointing out, now the link points to VS Preview.

PR #485, opened 2019-04-02
Fallbacks for building internal WPF projects that use PresentationBuildTasks.dll and Microsoft.WinFX.targets for markup compilation

vatsan-madhavan:
Adding fallbacks for building WPF projects that use PresentationBuildTasks.dll and Microsoft.WinFX.targets for markup compilation

When PresentationBuildTasks (PBT) moves to [dotnet/wpf](https://github.com/dotnet/wpf) from [dotnet-wpf-int](https://dev.azure.com/dnceng/internal/_git/dotnet-wpf-int), internal builds will need to consume PBT via a transport package instead of using the local source copy of `Microsoft.WinFX.targets` and local build output of PresentationBuildTasks.dll. 

Since we already produce and MSBuild Sdk containing these files - `Microsoft.NET.Sdk.WindowsDesktop` (which will also move alongwith PBT to dotnet/wpf - they go together) - we will simply leverage this Sdk package as our _transport package_  for internal builds. 

===
This fix also contains an unrealted fixup to Packaging.targets that was inadvertently missing production of version.txt.

### DIFF
--- a/eng/WpfArcadeSdk/tools/Pbt.props
+++ b/eng/WpfArcadeSdk/tools/Pbt.props
@@ -1,11 +1,37 @@
 <Project>
   <PropertyGroup>
+    <LocalMicrosoftWinFXTargets>$(RepoRoot)src\Microsoft.DotNet.Wpf\src\PresentationBuildTasks\Microsoft.WinFX.targets</LocalMicrosoftWinFXTargets>
+  </PropertyGroup>
+
+  <PropertyGroup Condition="Exists('$(LocalMicrosoftWinFXTargets)')">
     <PbtTfm Condition="'$(MSBuildRuntimeType)'=='Core'">netcoreapp2.1</PbtTfm>
     <PbtTfm Condition="'$(MSBuildRuntimeType)'!='Core'">net472</PbtTfm>
   </PropertyGroup>
 
-  <PropertyGroup>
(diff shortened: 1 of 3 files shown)


ryalanms:
All the changes look good, and dev/vatsanm/pbt-fallbacks built cleanly for me locally.

eng/WpfArcadeSdk/tools/Pbt.props
```diff
@@ -1,11 +1,37 @@
...
+    <LocalMicrosoftWinFXTargets>$(RepoRoot)src\Microsoft.DotNet.Wpf\src\PresentationBuildTasks\Microsoft.WinFX.targets</LocalMicrosoftWinFXTargets>
+  </PropertyGroup>
+
+  <PropertyGroup Condition="Exists('$(LocalMicrosoftWinFXTargets)')">
     <PbtTfm Condition="'$(MSBuildRuntimeType)'=='Core'">netcoreapp2.1</PbtTfm>
     <PbtTfm Condition="'$(MSBuildRuntimeType)'!='Core'">net472</PbtTfm>
   </PropertyGroup>
 
-  <PropertyGroup>
+  <PropertyGroup Condition="Exists('$(LocalMicrosoftWinFXTargets)')">
```
ryalanms: nit: These PropertyGroups can be combined.

PR #501, opened 2019-04-05
LOC CHECKIN | dotnet/wpf master | 20190405

v-zbsail:


vatsan-madhavan:
@v-zbsail Please update your branch by rebuilding like this (substitute whatever values you usually use for `Platform`, `Configuration` etc.). 

`build.cmd -pack /p:UpdateXlfOnBuild=true /p:Platform=x86 /p:Configuration=Release`

This should get you past the build failures. 

/cc @ryalanms

nguerrera:
@vatsan-madhavan The loc folks don't usually build. We usually push to their PR branch for them to run the update if needed.

vatsan-madhavan:
Sorry I did not know that. We’ll follow up.

ryalanms:
It looks like the Strings.resx that was used to generate these xlf files doesn't match the Strings.resx that is checked-in.  @v-zbsail could you confirm which Strings.resx file was used?  It looks like several strings from the generated xlfs are missing.  I can also build the new xlfs locally to verify.  Thanks.

nguerrera:
The loc folks do not feel with resx directly, they operate on the xlf that's checked in. 

What probably happened is the xlf had different content when the handoff occurred. If you rebuild it will merge back changes from resx into xlf. In normal circumstances, this would be enough but you're in a unique circumstance where you have put translations yourselves into xlf. For that you will have to reapply them. I mentioned several times that nobody else does it this way and instead we let the mic team handle recycling, but you all insisted that you had to do the recycling. This makes merges harder with translations coming from both sides. I imagine this will settle down after a few loc cycles.

PR #509, opened 2019-04-06
Dev/stevenbrix/enable analyzers and code cleanup

### LINKED ISSUE #506: Setting EnableAnalyzers to true in .csproj doesn't pick up correct ruleset
This was broken when the ruleset was moved to the Wpf Arcade SDK. Also, the changes to System.Xaml that were made were lost as well.

stevenbrix:
Fixes #506 

When I had moved the WpfCodeAnalysis.ruleset file into the WpfArcadeSdk, putting it into a .props file was the wrong thing to do. Moving it into a .targets file allows projects to properly set the `EnableAnalyzers` project flag to begin seeing the issues show up. Most of these changes to System.Xaml had been made in our internal repo while we were working on the SDK, and now I'm moving them out here.  I've also onboarded the analyzers for PresentationBuildTasks.

Below is the logic behind how we should suppress rules:

1. Any rules we want suppressed globally, get suppressed in WpfCodeAnaysis.ruleset by setting Action to "None"
2. Any rules that are enabled but require certain suppressions, should be suppressed in GlobalSuppressions.cs
3. Any suppressions that we want to fix, but won't for the 3.0 release due to compat concerns, should be suppressed inline.


ryalanms:
Thanks, Steven!

Issue #508, opened 2019-04-05
Could not find assembly 'System.Xaml, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089'


fadimounir:
* .NET Core Version: (e.g. 3.0 Preview1, or daily build number, use `dotnet --info`)
* Windows version: (`17763.379`)
* Does the bug reproduce also in WPF for .NET Framework 4.8?: I haven't tried

 
 
 **Problem description:**
Using the latest core-sdk version, the WPF app from the https://github.com/ridomin/msix-catalog repo no longer builds.
 
 **Actual behavior:**  
`Microsoft.WinFX.targets(243,9): error MC1000: Unknown build error, 'Could not find assembly 'System.Xaml, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089'. Either explicitly load this assembly using a method such as LoadFromAssemblyPath() or use a MetadataAssemblyResolver that returns a valid assembly.'  [D:\GIT\msix-catalog-master\msix.catalog.app\msix.catalog.app.core.csproj]`

 **Expected behavior:**
Compilation should work.
 
 **Minimal repro:**
1) Download https://github.com/ridomin/msix-catalog
2) cd msix-catalog-master\msix.catalog.app
3) dotnet publish -r win-x64 msix.catalog.app.core.csproj

NOTE:
I'm observing that we're now setting _PresentationBuildTaskTfm to netcoreapp2.1 in Microsoft.WinFX.targets:
`<_PresentationBuildTasksTfm Condition="'$(MSBuildRuntimeType)' == 'Core'">netcoreapp2.1</_PresentationBuildTasksTfm>`
It used to be netcoreapp3.0. Is this intentional?

fadimounir:
Here's my .NET Core Version info:

```
.NET Core SDK (reflecting any global.json):
 Version:   3.0.100-preview4-011077
 Commit:    c32af26f6f

Runtime Environment:
 OS Name:     Windows
 OS Version:  10.0.17763
 OS Platform: Windows
 RID:         win10-x86
 Base Path:   D:\GIT\sdk\.dotnet\sdk\3.0.100-preview4-011077\

Host (useful for support):
  Version: 3.0.0-preview4-27605-02
  Commit:  70d56fffb1

.NET Core SDKs installed:
  3.0.100-preview4-011077 [D:\GIT\sdk\.dotnet\sdk]

.NET Core runtimes installed:
[...]

Issue #208, opened 2018-12-24
WPF assemblies are incompatible with each other in 3.0.0-alpha-27128-4


albahari:
* .NET Core Version: 3.0.0-alpha-27128-4
* Windows version: 10.0.17763.0
* Does the bug reproduce also in WPF for .NET Framework 4.8?: No

In dotnet\shared\Microsoft.WindowsDesktop.App\3.0.0-alpha-27128-4, PresentationCore.dll references System.Xaml, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089.

But the System.Xaml.dll assembly in that folder has a different public key token (31bf3856ad364e35).
 
So when you use Roslyn to compile a WPF app by referencing these assemblies, you end up with errors like this:

The type 'IQueryAmbient' is defined in an assembly that is not referenced. You must add a reference to assembly 'System.Xaml, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089'. 

I don't these errors in Visual Studio, so I presume Visual Studio must be relying on referencing assemblies drawn from some other place. But as I understand, reference assemblies aren't mandatory. It should still be possible to compile code against real assemblies.

karelz:
I believe that reference assemblies are in general mandatory. I don't think we support compiling against runtime assemblies in general (if reference assemblies are available). At minimum compiling against runtime assemblies causes some interesting problems in certain scenarios, so I would personally recommend to avoid that.
For general .NET and also for WPF or any other framework.

albahari:
Could you elaborate on what the interesting problems are? Right now I'm writing LINQPad for .NET Core 3. LINQPad uses Roslyn to compile queries against the runtime assemblies, as it has always done with .NET Framework. I would like to keep LINQPad.exe small, and have it work with whatever version of .NET Core is installed on the machine. It's a lightweight tool and I'd rather not have to embed 300+ reference assemblies (and then have it locked into a particular version of .NET Core).

Your statement that reference assemblies are mandatory seems to imply that Roslyn is effectively useless in .NET Core without a whole additional set of baggage. In other words, if we want to dynamically generate an assembly at runtime, it's not enough to reference and ship Microsoft.CodeAnalysis. We also need to (somehow) ship a set of reference assemblies. Is there any guidance on the procedure here? (The problem is compounded by not having Reflection.Emit as an alternative, because the option to save assemblies is not available in .NET Core.)

[...]

PR #515, opened 2019-04-06
Public key tokens of .NET core WPF assemblies and .NET Framework WPF assemblies need to match

ryalanms:
This is the public half of a two part change. The original fix (adding StrongNameKeyId=Ecma to WPF assembly projects) was insufficient. After help from @nguerrera and more investigation of which PKTs were contained in which NETFX assemblies, @vatsan-madhavan provided a much more comprehensive fix that also fixed other regressions introduced during Arcade onboarding. @vatsan-madhavan's complete fix is included here. 

- WPF assemblies are incompatible with each other in 3.0.0-alpha-27128-4 #208 (fixes issue #208)

- Could not find assembly 'System.Xaml, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089' #508 (fixes issue #508)

This internal half of this change will be committed after the darc update containing this change reaches dotnet-wpf-int.  Thanks.

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/PresentationBuildTasks/Microsoft/Build/Tasks/Windows/MarkupCompilePass1.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationBuildTasks/Microsoft/Build/Tasks/Windows/MarkupCompilePass1.cs
@@ -102,6 +102,8 @@ namespace Microsoft.Build.Tasks.Windows
         /// <returns></returns>
         public override bool Execute()
         {
+            System.Diagnostics.Debugger.Launch();
+
             TaskHelper.DisplayLogo(Log, SR.Get(SRID.MarkupCompilePass1Task));
 
             bool bSuccess = true;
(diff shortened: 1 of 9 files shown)


src/Microsoft.DotNet.Wpf/src/PresentationBuildTasks/Microsoft/Build/Tasks/Windows/MarkupCompilePass1.cs
```diff
@@ -102,6 +102,8 @@ public MarkupCompilePass1( ) : base(SR.SharedResourceManager)
         /// <returns></returns>
         public override bool Execute()
         {
+            System.Diagnostics.Debugger.Launch();
```
vatsan-madhavan: Is this an accidental addition ?

ryalanms: Yes. Let me update that.

PR #517, opened 2019-04-06
Dev/stevenki/port analyzers and system xaml

stevenbrix:
This PR fixes enabling roslyn analyzers and ports the changes made to System.Xaml that we had made in our internal repo out here into the public.

Here is the thought process on how we should handle suppressing rules:
1. Any rules we want suppressed across all projects, get suppressed in WpfCodeAnaysis.ruleset by setting Action to "None"
2. Any rules that are enabled but require certain suppressions on a project level bases, should be suppressed in that projects GlobalSuppressions.cs file.
3. Any suppressions that we want to fix, but won't for the 3.0 release due to compat concerns, should be suppressed inline.


src/Microsoft.DotNet.Wpf/src/PresentationBuildTasks/PresentationBuildTasks.csproj
```diff
@@ -32,6 +32,7 @@
     <AutoGenerateBindingRedirects>true</AutoGenerateBindingRedirects>
     <GenerateBindingRedirectsOutputType>true</GenerateBindingRedirectsOutputType>
     <Platforms>AnyCPU;x64</Platforms>
+    <!--<EnableAnalyzers>true</EnableAnalyzers>-->
```
ryalanms: nit: PBT file.

src/Microsoft.DotNet.Wpf/src/System.Xaml/System/Xaml/XamlMember.cs
```diff
@@ -147,10 +138,6 @@ internal XamlMember(EventInfo eventInfo, XamlSchemaContext schemaContext, XamlMe
         internal XamlMember(string attachablePropertyName, MethodInfo getter, MethodInfo setter,
             XamlSchemaContext schemaContext, XamlMemberInvoker invoker, MemberReflector reflector)
         {
-            if (attachablePropertyName == null)
```
ryalanms: ValidateGetter and ValidateSetter are now called before checking attachablePropertyName for null and throwing the exception.  If ValidateGetter/Setter fail before the new check on line 153, the behavior is different (ArgumentNullException vs ArgumentException).

eng/WpfArcadeSdk/Sdk/Sdk.targets
```diff
@@ -16,6 +16,7 @@
   <Import Project="$(WpfArcadeSdkToolsDir)ExtendedAssemblyInfo.targets" />
   <Import Project="$(WpfArcadeSdkToolsDir)Redist.targets" />
   <Import Project="$(WpfArcadeSdkToolsDir)WpfProjectReference.targets" />
+  <Import Project="$(WpfArcadeSdkToolsDir)CodeAnalysis.targets" />
```
vatsan-madhavan: Have you tested the build of DotNET-wpf-int with this change?

stevenbrix: Is this a general question or is it specific to this change (CodeAnalysis.props -> CodeAnalysis.targets)?

vatsan-madhavan: It's re: changes to `Microsoft.DotNet.Arcade.Wpf.Sdk`

stevenbrix: There's no reason to believe that this simple change will cause any issues. None of the internal projects are building with `<EnableAnalzers>true</EnableAnalyzers>`, so the build won't start failing due to analyzers being integrated where they weren't before

stevenbrix: but i'll go ahead and make sure and report back if it's clean

stevenbrix: it builds fine

PR #519, opened 2019-04-07
Make build.sh executable

hughbe:


hughbe:
Lets us run `./build.sh` from the prompt

stevenbrix:
@hughbe i don't see any changes to the file, what exactly was changed?

hughbe:
All that’s changed is the flags of the file. Previously it was not executable but I ran `chmod +x` allowing it to be executable. There’s no change in the content of the file

stevenbrix:
Awesome, thanks @hughbe!

PR #525, opened 2019-04-08
Clean up XmlnsDictionary

thomasclaudiushuber:
Use nameof, var where type is obvious from assignment, and expression bodies properties for one-line getters


thomasclaudiushuber:
Sorry, I just saw in the [readme](https://github.com/dotnet/wpf/blob/master/src/Microsoft.DotNet.Wpf/src/PresentationFramework/README.md) at PresentationFramework this after creating this PR:

> Until PresentationFramework is open-sourced, the files under this folder should be modified only with careful coordination with the WPF team

If a clean up like this is ok, I'll go this evening through the other files too

stevenbrix:
No worries @thomasclaudiushuber! We appreciate the good faith, I know I miss things like that all the time :)

LGTM, @ryalanms might know more about whether this is OK too merge at this point

thomasclaudiushuber:
@stevenbrix I close this one and we do stuff like this after 3.0 with roslyn analyzers.

PR #534, opened 2019-04-08
Use RuntimeFrameworkVersion in builds

vatsan-madhavan:
Adds support for using RuntimeFrameworkVersion in our builds


ericstj:
Tests will fail unless you acquire the shared framework for them to run against.  See https://github.com/dotnet/winforms/pull/708#issuecomment-481039436

vatsan-madhavan:
@ericstj - The tests aren't yet running in a testhost built from the sdk used to build the product bits. We will get this right along with the testhost work.

PR #543, opened 2019-04-09
Add helper script and update developer guide

stevenbrix:
This PR adds a helper script that I've been using locally that helps me test my local changes. It also updates the developer-guide.md on how to use the script effectively, as well as some other helpful information about testing locally built assemblies. I'm not sure how to test PBT yet, and hopefully @ryalanms can share more information on how that can be done.

Note: This script will currently only copy System.Xaml, but as soon as we start bringing more assemblies it will just work


eng/copy-wpf.ps1
```diff
@@ -0,0 +1,126 @@
...
+    $locallyBuiltBinaryLocationBase = Join-Path $RepoRoot "artifacts\bin"
+    CopyNativeBinariesToPublishLocation $publishLocation $locallyBuiltBinaryLocationBase 
+    CopyManagedBinariesToPublishLocation $publishLocation  $locallyBuiltBinaryLocationBase
+    CopyThemeBinariesToPublishLocation $publishLocation  $locallyBuiltBinaryLocationBase
+    CopyUIAutomationBinariesToPublishLocation $publishLocation  $locallyBuiltBinaryLocationBase
+}
+
+function CopyNativeBinariesToPublishLocation($publishLocation, $localBinLocation)
+{
+    $NativeBinaries = "D3DCompiler", "PenImc", "PresentationNative", "wpfgfx"
```
vatsan-madhavan: We also depend on vcruntime140_cor3. This dependency will light-up as soon as the managed portion of /dllrename functionality in the linker starts working.

vatsan-madhavan: All the relevant binaries are usually gathered under `artifacts\packaging\Debug\Microsoft.DotNet.Wpf.GitHub\lib\netcoreapp3.0` for easy packaging. You can probably gather them from there instead of having to maintain a hardcoded list here.

eng/copy-wpf.ps1
```diff
@@ -0,0 +1,126 @@
...
+$Config = if ($release) { "Release" } else { "Debug" }
+
+if ($help -or [string]::IsNullOrEmpty($destination)) 
+{
+    Print-Usage
+}
+elseif($shared)
+{
+    Write-Host "Copying binaries to shared installation"
+    $sharedLocation = Join-Path (Join-Path $env:ProgramFiles "dotnet\shared\Microsoft.WindowsDesktop.App") $destination
```
vatsan-madhavan: A shared installed can be somewhere besides %programfiles%.

Documentation/developer-guide.md
```diff
@@ -23,6 +23,58 @@ We use the following workflow for building and testing features and fixes.
...
+
+#### Testing API changes 
+The above instructions imply that you are testing assemblies that don't have any changes to the
+public API surface that the test application needs to use. If you need to test some API changes
+you are making, the C# team has done some great work to make this relatively straightforward.
+When the C# compiler detects a collision with assembly references, the assembly with the
+higher version number is chosen. Assuming our locally built binaries are newer than what is
+installed, we can then simply reference those local binaries directly from the project file, like this:
+
+*Note: you should build locally with the `-pack` param to ensure the binaries are put in the correct location.*
```
vatsan-madhavan: At this time, the binaries under artifacts\packaging aren't updated during rebuilds. They exist to help with CI builds primarily. If they are used for local dev builds, one should `rmdir` the `artifacts\packaging` folder before rebuilds to get a fresh copy each time. We will optimize the packaging targets to ensure that local dev workflows can leverage these as well (i.e., ensure that these files are updated after each rebuild) in near-future.

eng/copy-wpf.ps1
stevenbrix: Yeah that's what i had originally done, but i noticed that build -pack doesn't always update them. and it was one of those things that at the time i didn't want to go down the rabbit hole of figuring out was going on. i totally could be doing something wrong if no one else has seen this?

eng/copy-wpf.ps1
stevenbrix: yeah you are right, although i'm not sure how we could figure out where? regardless, i think after the dicussion we had yesterday i'm going to remove this in favor of doing a test-host installation now that i have a better understanding of how that is done.

Documentation/developer-guide.md
stevenbrix: ahh yeah i think this is what i was talking about in my other comment. 

i'll add that `rmdir` is needed for now and open an issue that i reference here

eng/copy-wpf.ps1
stevenbrix: I'm going to leave as is for now, I'd prefer to not have to rmdir, i do have a local version of this script that uses that directory that we can update once build -pack works as expected

stevenbrix: i see vcruntime140d.dll in my artifacts\bin directory right now, is this something that's going to come later?

stevenbrix:
Hmm, after seeing dotnet/arcade#2343 i'm thinking that maybe i should hold off and not add the install-testhost script if that does indeed allow us to install a local testhost to the .dotnet folder

eng/copy-wpf.ps1
stevenbrix: now that we are copying by package, i think this is no longer relevant

stevenbrix: now that you have fixed the packaging issue i've gone and updated this :) thanks!

eng/copy-wpf.ps1
```diff
@@ -0,0 +1,93 @@
...
+
+function Print-Usage()
+{
+    Write-Host "Usage: copy-wpf.ps1 -destination <value> [-arch <value>] [-release] [-testhost]"
+    Write-Host "    This script helps developers deploy wpf assemblies to the proper location for easy testing. See "
+    Write-Host "    developer-guide.md for more information on how to use this script."
+    Write-Host ""
+    Write-Host "Common parameters:"
+    Write-Host "  -destination <value>    Location of .csproj or .vbproj of application to test against. Ignored"
+    Write-Host "                          if the -testhost parameter is used."
```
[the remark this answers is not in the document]
stevenbrix: you are right, i had staged the fixed script instead of commiting. it's fixed now

Issue #350, opened 2019-02-13
"The Name 'XXX' has already been defined" build error occurs if ControlTemplate has visual element and VisualState with the same names 


pavelovcharov:
* .NET Core Version: 3.0.100-preview-010184
* Windows version: 1809 (17763.253)
* Does the bug reproduce also in WPF for .NET Framework 4.8?: No

 
 
 **Problem description:**
Build error occurs if ControlTemplate contains a visual element and VisualState with the same names

 **Actual behavior:** 
Build error "MC4101: The Name 'Normal' has already been defined. Names must be unique." occurs

 **Expected behavior:**
No build errors
 
 **Minimal repro:**
Create new .Net Core 3 wpf project, add the following ControlTemplate
```
    <Window.Resources>
        <ControlTemplate TargetType="Button" x:Key="buttonStyle">
            <Border x:Name="Normal">
                <VisualStateManager.VisualStateGroups>
                    <VisualStateGroup x:Name="TestGroup">
                        <VisualState x:Name="Normal" />
                    </VisualStateGroup>
                </VisualStateManager.VisualStateGroups>
            </Border>
        </ControlTemplate>
    </Window.Resources>

```
and try to build the project.

stevenbrix:
Thanks @pavelovcharov for filing this issue! This is interesting, in a .NET Framework WPF app, you see an intellisense error, and an error shows up in the ErrorsList, but doesn't cause your build to error. The UWP Xaml Compiler also treats this as an error condition as well, and your build will fail. However, I don't imagine there is any technical reason for this, as these names should never cause a collision.

@ryalanms i'm assigning this to you. i'd imagine we should make sure this doesn't cause errors so that porting from .NET Framework -> .NET Core is as easy as possible. however, we should continue having discussions so that these behaviors are consistent and both UWP and WPF behave the same

ryalanms:
This difference was due to a bug in BamlMapTable.

weltkante:
@ryalanms This doesn't look like a bugfix, if the same code works in Desktop Framework it should not be necessary to reinitialize static properties on every instance creation.

Note that your "fix" broke thread safety (see comment on PR)

Either make it threadsafe (probably will be a performance loss because you need a lot more locking now) or look for the real difference between .NET Core and Desktop Framework. The static initialization wasn't the difference.

PR #570, opened 2019-04-12
BamlMapTable should always reinitialize KnownAssemblyInfoRecord

### LINKED ISSUE #378: NETFX Regression: No longer can set name property to MarkupExtension in certain places
* .NET Core Version 3.0.100-preview-010184
* Windows version: Server 2019, 17763.253

I am porting a WPF control library to .NET Core. In this library, they define many `const` fields that they use in attributes, element names, etc. For example:
```cs
[TemplateVisualState(Name = MyState, GroupName = MyStateGroup)]
class MyControl : Control
{
   public const String MyState = "MyState";
   public const String MyStateGroup = "MyStateGroup";
}
```
```xml
<!-- somewhere inside of a control template -->
<VisualStateGroup Name="{x:Static local:MyControl.MyState}">...</VisualStateGroup>
```

The .NET Framework tooling works fine here. However, in .NET Core, trying to compile code using this pattern produces this error:

> error MC3079: MarkupExtensions are not allowed for Uid or Name property values, so '{x:Static local:MyControl.MyState}' is not valid.

[Here's a small repro that demonstrates this issue.](https://github.com/dotnet/wpf/files/2897472/Repro.zip) The application will compile fine if you set the target framework to `net472`, but it will not compile with the above error if you set it to `netcoreapp3.0`.

Note that this pattern doesn't work in all cases; I don't know the specifics on where it does and doesn't work, but I think it only works for files evaluated at runtime.

ryalanms:
NETFX Regression: No longer can set name property to MarkupExtension in certain places #378.

Move KnownAssemblyInfoRecord initialization out of BamlMapTable's static constructor to its constructor. (Fixes #378)

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Markup/BamlMapTable.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Markup/BamlMapTable.cs
@@ -72,20 +72,17 @@ namespace System.Windows.Markup
 
 #region Constructor
 
-        static BamlMapTable()
-        {
-            // Setup the assembly record for the known types of controls
-            KnownAssemblyInfoRecord = new BamlAssemblyInfoRecord();
-            KnownAssemblyInfoRecord.AssemblyId = -1;
-            KnownAssemblyInfoRecord.Assembly = ReflectionHelper.LoadAssembly(_frameworkAssembly, string.Empty);
-            KnownAssemblyInfoRecord.AssemblyFullName = KnownAssemblyInfoRecord.Assembly.FullName;
-        }
-
         internal BamlMapTable(XamlTypeMapper xamlTypeMapper)
         {
             Debug.Assert(null != xamlTypeMapper);
 
             _xamlTypeMapper = xamlTypeMapper;
+
+            // Setup the assembly record for the known types of controls
+            KnownAssemblyInfoRecord = new BamlAssemblyInfoRecord();
+            KnownAssemblyInfoRecord.AssemblyId = -1;
+            KnownAssemblyInfoRecord.Assembly = ReflectionHelper.LoadAssembly(_frameworkAssembly, string.Empty);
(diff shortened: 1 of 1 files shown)


ryalanms:
This change also brings namescoping rules in line with NETFX.  (Fixes issue #350)

vatsan-madhavan:
Looks ok to me. 

@nguerrera, would you mind taking a quick peek?

ryalanms:
Thanks.

weltkante:
@ryalanms @vatsan-madhavan This change doesn't look threadsafe to me. Originally, the static type constructor was threadsafe by the framework rules (because you can't access static properties before the static constructor has run), now you're setting static properties from an instance constructor. If msbuild or VisualStudio is doing anything multithreaded this will turn bad.

nguerrera:
Looks like KnownAssemblyInfoRecord can be made to be an instance field. Sorry that I missed that in the review. I would do that and rename it to _knownAssemblyInfoRecord for clarity.

As far as threading, build tasks are executed in parallel using different processes, not different threads by msbuild. I would generally recommend against relying on this. We discussed a long term plan to move everything to instance data off of the markup compiler, but that hasn't been done yet as the PBT code is rather tangled with statics inherited straight from  .NET Framework implementation, and untangling it is non-trivial. But perhaps that should be tracked by a GitHub issue?

For anything that needs to remain static in the short term pending that refactoring, make sure it is cleared like the others:

https://github.com/dotnet/wpf/blob/master/src/Microsoft.DotNet.Wpf/src/PresentationBuildTasks/MS/Internal/MarkupCompiler/MarkupCompiler.cs#L222

weltkante:
@ryalanms @nguerrera should I create a new issue for that since the PR is already merged?

ryalanms:
Thanks, @weltkante. I'll update the KnownAssemblyInfoRecord property to be an instance variable today.  The refactoring of statics is being tracked internally, but a adding public GitHub issue sounds reasonable, now that PresentationBuildTasks is public.

PR #589, opened 2019-04-15
BamlMapTable's KnownAssemblyInfoRecord should be an instance field

ryalanms:
System.Reflection.Assembly.ReflectionOnlyLoad is not supported on .NET core and has been replaced with MetadataLoadContext.  MetadataLoadContext is cleared between markup compile passes, invalidating all System.Relection.Assembly instances loaded by MetadataLoadContext.  

Various markup compilation errors were occuring because of failed property lookups due to comparisons against an invalid System.Reflection.Assembly instance (PresentationFramework) in BamlMapTable.KnownAssemblyInfoRecord.  PresentationFramework is now loaded each time BamlMapTable is created by XamlTypeMapper, per markup compile pass.

This change includes the recommendation from @weltkante's review to update BamlAssemblyInfoRecord on BamlMapTable to be an instance field.


ryalanms:
@weltkante: Thank you for reviewing.  I'll include you on future build task refactor threads.  Thanks.

weltkante:
As requested in the other PR I've created an issue

Fixes #591

Issue #587, opened 2019-04-15
Add attribute to C++/CLI assemblies


fadimounir:
As discussed over email, we need to add some attribute to the IJW assemblies (DirectWriteForwarder.dll and System.Printing.dll to help with this bug: https://github.com/dotnet/coreclr/issues/23872

For Preview 5, the coreclr bug might not get fixed in time, so we'll use the assembly attribute as a way to exclude these 2 problematic assemblies from being compiled into ReadyToRun images.

Proposed attribute to add:
`[assembly: AssemblyMetadata("Language", "C++")]`

fadimounir:
cc @vatsan-madhavan

vatsan-madhavan:
The latest core-sdk builds (3.0.100-preview5-011473) now have these attributes for WPF assemblies.

PR #596, opened 2019-04-16
Add AssemblyMetadata[Language] attribute

vatsan-madhavan:
This PR addresses #587 

/cc @fadimounir

### DIFF
--- a/eng/WpfArcadeSdk/tools/ExtendedAssemblyInfo.targets
+++ b/eng/WpfArcadeSdk/tools/ExtendedAssemblyInfo.targets
@@ -40,6 +40,24 @@
       <AssemblyMetadata Include="Commit" Condition="'$(SourceRevisionId)'!=''">
         <Value>$(SourceRevisionId)</Value>
       </AssemblyMetadata>
+
+      <!-- 
+        Crossgen wants to detect when an assembly is built from C++/CLI sources, 
+        since it is unable to function correctly today (ostensibly due to bugs). Since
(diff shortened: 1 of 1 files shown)


vatsan-madhavan:
@fadimounir, Please note that this attribute will not be added to the satellite assemblies (resource assemblies like WindowsBase.resources.dll etc.).

eng/WpfArcadeSdk/tools/ExtendedAssemblyInfo.targets
```diff
@@ -40,6 +40,26 @@
       <AssemblyMetadata Include="Commit" Condition="'$(SourceRevisionId)'!=''">
         <Value>$(SourceRevisionId)</Value>
       </AssemblyMetadata>
+
+      <!-- 
+        Crossgen wants to detect when an assembly is built from C++/CLI sources, 
+        since it is unable to function correctly today (ostensibly due to bugs). Since
+        /clr:pure is deprecated and WPF will eventually switch over to /clr (i.e., IJW),
+        Crossgen will eventually (i.e., in the near-term) fail to work on C++/CLI
```
fadimounir: Note: Crossgen does not support IJW today. There are no plans to support them so far

fadimounir:
LGTM.

Thanks @vatsan-madhavan for adding this metadata. We've identified the root cause of the bug and are investigating a fix, but it's useful anyways to have this attribute.

fadimounir:
> Please note that this attribute will not be added to the satellite assemblies (resource assemblies like WindowsBase.resources.dll etc.).

Yes. We don't crossgen those anyways

PR #598, opened 2019-04-16
convert gentracesources to tt

stevenbrix:
This PR updates the WpfArcadeSdk to move our gentracesources.pl and gentracestrings.pl to using design time t4 generation. Instead of these files being generated into `$(IntermediateOutputPath)` they are now put under `$(WpfSourceDir)`, and are easily discoverable in the project files. 

There are a few caveats in order to get this to work:
1. We have to ensure BuildDependsOn variable that T4 targets update doesn't get overwritten. So with SDK style projects, we have to manually import the Sdk.targets and then include the T4 targets after.
2. I wanted to decouple the GenTraceSources and GenAvMessages (similar to how we had 2 separate perl scripts before). In order to do this, those targets can't include the T4 targets, since then you get MSBuild authoring errors about the same target being imported twice.

Below is an example of how WindowsBase uses these targets. PresentationFramework and PresentationCore are similar, only without the `GenTraceSources.targets` import.

```
  <Import Project="Sdk.targets" Sdk="Microsoft.NET.Sdk" />
  <Import Project="$(WpfCodeGenDir)AvTrace\GenTraceSources.targets" />
  <Import Project="$(WpfCodeGenDir)AvTrace\GenAvMessages.targets" />
  <Import Project="$(WpfCodeGenDir)DesignTimeTextTemplating.targets" />
```

There will be an internal PR that shows how these changes are consumed. The goal here is that we will continue to move more and more of our codegen scripts to T4 templates, where we should be able to use this as a starting point for completing that work.


eng/WpfArcadeSdk/tools/CodeGen/DesignTimeTextTemplating.targets
```diff
@@ -0,0 +1,98 @@
...
+      already exists and is read-only. -->
+      <Output ItemName="NonGeneratedFiles" TaskParameter="NonGeneratedFiles"/>
+
+      <!-- List of assemblies that are required to run the processed code. -->
+      <Output ItemName="T4RequiredAssemblies" TaskParameter="RequiredAssemblies"/>
+    </PreprocessTemplates>
+  </Target>
+  <!-- End workaround -->
+  <ItemGroup>
+    <T4ParameterValues Include="NewtonsoftLocation">
```
vatsan-madhavan: Are you just trying to get to the full path of `Newtonsoft.Json.dll` within the Nuget Package? 

If so, you will likely want to add a dependency to [`NewtonSoft.Json`](https://github.com/dotnet/arcade/blob/f42c6d1de75067f8c610f160006ce02187da683d/Documentation/ArcadeConventions.md), then set [`GeneratePathProperty=true`](https://github.com/NuGet/Home/issues/7851) for the corresponding `PackageReference` (which you could likely add right here). 

Nuget would generate an MSBuild property for you like `$(PkgNewtonSoft_Json)`, which will point you to the root of the package on the file-system.

eng/WpfArcadeSdk/tools/CodeGen/AvTrace/GenTraceSources.targets
```diff
@@ -0,0 +1,26 @@
+<Project>
+  <PropertyGroup>
+    <WindowsBaseAvTraceMessageFile>$(WpfSourceDir)\WindowsBase\AvTraceMessages.json</WindowsBaseAvTraceMessageFile>
```
vatsan-madhavan: Could this be an opt-in? For e.g., if a project sets `$(AvTraceMessagesFile)`, then this target picks up the value and uses it? 

Alternatively, could this fit into `GenTraceSources.props` using a pattern somewhat similar to what is in use within `ShippingProjects.props`, for e.g., like this: 

```XML
<AvTraceEnabledProjects>
  WindowsBase;
  PresentationCore;
  PresentationFramework
  </AvTraceEnabledProjects>
```

In GenTraceSources.targets, 




```XML
<AvTraceMessageFile Condition="$(AvTraceEnabledProjects.Contains('$(NormalizedMSBuildProjectName)'))>$(MSBuildProjectDirectory)AVTraceMessages.json</AvTraceMessageFile>
```

eng/WpfArcadeSdk/tools/CodeGen/AvTrace/GenTraceSources.targets
```diff
@@ -0,0 +1,26 @@
...
+    <PresentationCoreAvTraceMessageFile>$(WpfSourceDir)\PresentationCore\AvTraceMessages.json</PresentationCoreAvTraceMessageFile>
+    <PresentationFrameworkAvTraceMessageFile>$(WpfSourceDir)\PresentationFramework\AvTraceMessages.json</PresentationFrameworkAvTraceMessageFile>    
+  </PropertyGroup>
+
+  <ItemGroup>
+    <None Include="$(MSBuildThisFileDirectory)PresentationTraceSources.tt">
+      <Generator>TextTemplatingFileGenerator</Generator>
+      <LastGenOutput>PresentationTraceSources.cs</LastGenOutput>
+      <OutputFilePath>$(WpfSourceDir)\WindowsBase\Generated\</OutputFilePath>
```
vatsan-madhavan: This is a design change - instead of generating ephemeral sources into `$(IntermediateOutputPath)`, now you are generating permanent sources under `$(WpfSourceDir)`. This seems consistent with the way TT is supposed to be used. This should be called out in the change-summary in both this PR and the internal one that corresponds.

eng/WpfArcadeSdk/tools/CodeGen/AvTrace/GenTraceSources.targets
```diff
@@ -0,0 +1,26 @@
...
+    </None>
+    <None Update="PresentationTraceSources.cs">
+      <DesignTime>True</DesignTime>
+      <AutoGen>True</AutoGen>
+      <DependentUpon>$(MSBuildThisFileDirectory)PresentationTraceSources.tt</DependentUpon>
+    </None>
+  </ItemGroup>
+  <ItemGroup>
+    <T4ParameterValues Include="AvTraceMessageFiles">
+      <Value>$(WindowsBaseAvTraceMessageFile);$(PresentationCoreAvTraceMessageFile);$(PresentationFrameworkAvTraceMessageFile)</Value>
```
vatsan-madhavan: Does this target run all at once for three projects, or does this run three times, once each per project? I'm trying to understand why all three of these values have to be embedded here permanently, rather than having these picked up dynamically one at a time.

eng/WpfArcadeSdk/tools/CodeGen/DesignTimeTextTemplating.targets
```diff
@@ -0,0 +1,98 @@
...
+       doesn’t run before the Build target. -->
+  <!-- <Import Project="Sdk.targets" Sdk="Microsoft.NET.Sdk" /> -->
+  <PropertyGroup>
+    <!-- Get the Visual Studio version: -->
+    <VisualStudioVersion Condition="'$(VisualStudioVersion)' == ''">16.0</VisualStudioVersion>
+    <!-- Keep the next element all on one line: -->
+    <VSToolsPath Condition="'$(VSToolsPath)' == ''">$(MSBuildExtensionsPath32)\Microsoft\VisualStudio\v$(VisualStudioVersion)</VSToolsPath>
+    <!-- This is what will cause the templates to be transformed when the project is built (default is false) -->
+    <TransformOnBuild>true</TransformOnBuild>
+    <!-- Set to true to force overwriting of read-only output files, e.g. if they're not checked out (default is false) -->
```
vatsan-madhavan: re "_they're not checked out_"

Does this matter in Git repos?

eng/WpfArcadeSdk/tools/CodeGen/DesignTimeTextTemplating.targets
```diff
@@ -0,0 +1,98 @@
...
+    <OverwriteReadOnlyOutputFiles>true</OverwriteReadOnlyOutputFiles>
+    <!-- Set to false to transform files even if the output appears to be up-to-date (default is true)  -->
```
vatsan-madhavan: I like the choice of _TransformOutOfDateOnly=false_

eng/WpfArcadeSdk/tools/CodeGen/DesignTimeTextTemplating.targets
```diff
@@ -0,0 +1,98 @@
...
+    <TransformOutOfDateOnly>false</TransformOutOfDateOnly>
+    <NewtonsoftVersion>12.0.2-beta1</NewtonsoftVersion>
```
vatsan-madhavan: `<NewtonsoftVersion>12.0.2-beta1</NewtonsoftVersion>` should be rewritten as something like `<NewtonsoftVersion>$(NewtonsoftJsonVersion)</NewtonsoftVersion>`. where `$(NewtonsoftJsonVersion)` is defined in `eng\Version.props` because of a DARC-created dependency.

eng/WpfArcadeSdk/tools/CodeGen/DesignTimeTextTemplating.targets
```diff
@@ -0,0 +1,98 @@
...
+  </PropertyGroup>
+  <Import Project="$(VSToolsPath)\TextTemplating\Microsoft.TextTemplating.targets" />
```
vatsan-madhavan: I would imagine that  `VSToolsPath` is defined outside of VS IDE, but within msbuild (i.e., when `'$(MSBuildRuntimeType)'!='Core'`)

What happens when `'$(MSBuildRuntimeType)'=='Core'` (for e.g., when invoking build using `dotnet.exe`)? 

Today, our build is not entirely compatible with `dotnet.exe` because of C++ involvement. That said, our C# projects are, in general, compatible with building with `dotnet.exe`. For e.g. `System.Xaml`, `WindowsBase`, `PresentationBuildTasks` etc. should be buildable with `dotnet.exe` (possibly with minor fixes to props/targets) beause they do not rely on C++/CLI or pure-C++ for builds. Will this continue to be true with the inclusion of `Microsoft.TextTemplating.targets` and related MSBuild Tasks?

eng/WpfArcadeSdk/tools/CodeGen/AvTrace/GenTraceSources.targets
stevenbrix: Yeah i had tried that initially (or some variation of it), there is such a headache with ordering of targets that i decided to have the projects manually include the targets that are needed

stevenbrix: Yeah i had tried that initially (or some variation of it), there is such a headache with ordering of targets that i decided to have the projects manually include the targets that are needed

stevenbrix: but maybe i can revisit it now that i have everything working and find a way to make it work

eng/WpfArcadeSdk/tools/CodeGen/AvTrace/GenTraceSources.targets
stevenbrix: Yeah this is one of the things i especially like about this :)

eng/WpfArcadeSdk/tools/CodeGen/AvTrace/GenTraceSources.targets
stevenbrix: This target only runs once for WindowsBase (it's the only project that includes it).  The GenTraceSources worked in the past was that it built the `PresentationTraceSources` class into WindowsBase only, and not the other assemblies. This is the class, and since it's public i felt it safest to not change this aspect of the design:
https://docs.microsoft.com/en-us/dotnet/api/system.diagnostics.presentationtracesources?view=netframework-4.7.2

I agree that this wasn't ideal, but i also felt this was simple and straightforward enough to not worry about it too much. I'm open to suggestions for improving it if you see something.

eng/WpfArcadeSdk/tools/CodeGen/DesignTimeTextTemplating.targets
stevenbrix: does darc work for non-dotnet built assemblies? i had purposefully isolated this from the rest of the project because i didn't want to start introducing this into the product unnecessarily

eng/WpfArcadeSdk/tools/CodeGen/AvTrace/GenTraceSources.targets
stevenbrix: i do want to write up some documentation relating to this. i'll add some to this PR

eng/WpfArcadeSdk/tools/CodeGen/DesignTimeTextTemplating.targets
stevenbrix: I'll plead guilty that i had gotten this bit from a blog post about integrating design time tt in an sdk style project :)

I think you are right that this is probably not necessary, or at least the comment is definitely not relevant for us.

eng/WpfArcadeSdk/tools/CodeGen/DesignTimeTextTemplating.targets
stevenbrix: Yeah i'll give this a go, although looks like `dotnet` builds are broken since System.Xaml.csproj is missing.  i notice we get warnings when you build in dotnet-wpf-int about this, but turns out it breaks core builds. i'll fix this up and make sure that it works

stevenbrix: the only other thing i saw was some targets defined in `ExtendedAssemblyInfo.targets` that are incompatible when building using `dotnet.exe`. I've updated them so that they don't run in that case and the build works

eng/WpfArcadeSdk/tools/CodeGen/AvTrace/GenTraceSources.targets
stevenbrix: i'm going to leave as is, i think it's cleaner that these targets aren't magically imported. couple that with the dependency on the ordering of including the Sdk.targets from Microsoft.NET.Sdk, this will easier to maintain

eng/WpfArcadeSdk/tools/CodeGen/AvTrace/GenTraceSources.targets
vatsan-madhavan: Another option is that each project opts-in by contributing to an `<ItemGroup>` of `@(AvTraceMessageFiles)`, but only one project _runs_ the target (which, as you say, is `WindowsBase`). At the time the target runs, the accumulated set of config files in `@(AvTraceMessageFiles)` gets processed.

eng/Version.Details.xml
```diff
@@ -75,5 +75,10 @@
       <Uri>https://github.com/dotnet/corefx</Uri>
       <Sha>071e6986b2a71e29cab3df4e2dd2cdf3fe3ce574</Sha>
     </Dependency>
+    <Dependency Name="Newtonsoft.Json" Version="12.0.1">
```
vatsan-madhavan: Can we use `System.Text.Json` instead? 

https://github.com/dotnet/corefx/blob/aa9b7c2632ae07b68c109b051faf35ea38491c63/src/System.Text.Json/porting_guide/README.md

stevenbrix: i had started looking into it and tried converting over. now i don't remember why i decided it wasn't worth it. i think i looked at arcade and they use newtonsoft as well. is there a particular reason you'd prefer to use System.Text.Json?

stevenbrix: I think what i found was there is no System.Text.Json package produced, there is the Microsoft.Bcl.Json.Sources package, but it specifcally says that shouldn't be included for netcore3 projects. anyways, json.net is the de-facto json parsing library for .net and is owned by a guy who works at microsoft, so i think it's fine to use

stevenbrix: i looked at the Microsoft.Bcl.Json.Sources package, and it just contains source code. This package seems special compared to regular corefx pacakges, and i'm not sure what it's intent is. I believe it looks like the JSon parsing code gets directly compiled into your project, meaning it is most likely something we can't use here. @ahsonkhan looks to be the owner/maintainer of this and might be able to shed some more insight here.

ahsonkhan: Here's the source package guide: https://github.com/dotnet/corefx/tree/aa9b7c2632ae07b68c109b051faf35ea38491c63/src/System.Text.Json/source_package

That is primarily for libraries that multi-target or target non-netcoreapp3.0 TFMs (say netfx472, or netstandard2.0).

https://github.com/dotnet/wpf/blob/71d36bacfd89c53e240f1ae542ab1c67bcdce780/packaging/Microsoft.DotNet.Arcade.Wpf.Sdk/Microsoft.DotNet.Arcade.Wpf.Sdk.ArchNeutral.csproj#L4

It looks to me that only netcoreapp3.0 is being targeted here. The System.Text.Json library APIs are built-in as part of .NET Core 3.0. Yo shouldn't need to reference any additional package.

eng/WpfArcadeSdk/tools/CodeGen/AvTrace/PresentationTraceSources.tt
```diff
@@ -0,0 +1,54 @@
...
+{
+    /// <summary>Access point for TraceSources</summary>
+    public static partial class PresentationTraceSources
+    {
+<# 
+   foreach (var avTracMessageFile in AvTraceMessageFiles.Split(';'))
+   {
+		string fileName = this.Host.ResolvePath(avTracMessageFile);
+		string jsonText = File.ReadAllText(fileName);
+		dynamic jsonObj = JsonConvert.DeserializeObject(jsonText);
```
ahsonkhan: If all you are doing is fetching the sources property and getting it's name, and do not need the `dynamic` functionality, then the `JsonDocument` (within System.Text.Json namespace) works for your use case here.

ahsonkhan: Otherwise, if you need `dynamic`, then you would have to keep using `Newtonsoft.Json` since we don't support that functionality yet.

eng/Version.Details.xml
stevenbrix: the thing to consider here is that this is running in the msbuild process to generate code that then is compiled into our netcore3 assemblies. so we need .net fx support as well

vatsan-madhavan: `System.Text.Json` is buit-in , and we get security updates etc. for free. 

Existing code (like Arcade) using `Newtonsoft.Json` is not suprising considering `System.Text.Json` is new to the ecosystem.  New code choosing `Newtonsoft.json` over `System.Text.Json` -  esp. code that is part of netcoreapp3.0  - suggests that `System.Text.Json` may not be meeting the needs of its intended audience. Can we articulate where it fails our needs and file bugs (unless it is already covered by @ahsonkhan below)?

vatsan-madhavan: The discussion about Json parsers got me thinking about this a bit more and I realized that we haven't quite discussed the use of Json as a design-point. 

We have complicated codegen data in XML (along with XSD for validations) that I don't anticipate being converted to another format. Given we expect to retain at least some codegen driven by XML data, it might be worht consisering whether a consistent design across the board that standardizes on XML would be more maintainable in the long term. Besides,  inbox XML readers/parsers are full-featured and you woldn't have to pull in an external package .

eng/WpfArcadeSdk/tools/CodeGen/AvTrace/PresentationTraceSources.tt
stevenbrix: i don't think there is a hard requirement, in fact i didn't like that i had used it, and i'm going to remove them entirely :)

eng/Version.Details.xml
stevenbrix: Sorry, I must not be explaning how this works very clearly.

This should not be considered a netcoreapp3.0 project, the text templating engine runs inside the msbuild process, which is a .net fx process. Therefore, the `System.Text.Json` library is not available. Even if it was a .NET Core process, there is also no `System.Text.Json` package or library that we can reference like we do here with `Newtonsoft.Json`, or at least I have not found one. Please correct me if I'm wrong here, @ahsonkhan, but I have not been able to find one at all when looking at the https://dotnetfeed.blob.core.windows.net/dotnet-core/index.json feed. This is confirmed by the fact that if I try to add a reference to `System.Text.Json` to WPF like we do with every other CoreFx pacakge, `darc update-dependencies -c ".NET Core 3 Dev"` fails. **Even** if we had a package, we still can't use it because the built-in VisualStudio text templating engine doesn't support .NET Core, and so we can't load the assembly anyways.

We could switch over to using  https://github.com/mono/t4, which does support .NET Core, but at that point if the argument is that we aren't willing to use Newtonsoft because it isn't built-in, then we probably won't be willing to use https://github.com/mono/t4 for the same reason, and then we are stuck and can't really do anything.

As for using JSON over XML, converting between the two is incredibly trivial, so i don't have a real issue with moving towards XML. IMO JSON is less verbose and an overall easier/cleaner format to read, but I don't feel super strongly about it.

stevenbrix: I found https://github.com/dotnet/corefx/issues/33115 which mentions that `System.Text.Json` doesn't support JSON schemas, so if that is something we want, then we can't use it either. Although I don't think that would be a legitimate reason, I think having the .xsd is overkill for our usage since nothing is escaping the bounds of this repository.

stevenbrix: I opened this issue regarding lack of `System.Text.Json` package: https://github.com/dotnet/corefx/issues/37043

ahsonkhan: Please leverage the source package to see if it works for your needs. Aspnet is consuming it today and the guide I listed above can provide some help. Feel free to see how they are using it or ping me if you need some help.

https://github.com/aspnet/AspNetCore/search?q=Microsoft.Bcl.Json.Sources&unscoped_q=Microsoft.Bcl.Json.Sources

It should version fine with darc.

https://github.com/aspnet/AspNetCore/blob/9212f4a3df81ed28707d56e1b339f334fa48eb36/eng/Version.Details.xml#L288-L291

ahsonkhan: Regarding finding a "package", particularly on .NET Core 3.0, there isn't one because the API is **built-in**. The System.Text.Json.dll (assembly) is available in the box to automatically reference. You won't find a specific NuGet package for it since it isn't required on 3.0. Think of it like string or stream (or XmlReader). You don't go install/reference a NuGet package to get those APIs, they just exist. All you need to do is add a using directive (i.e. `System.Text.Json`).

Now, as you mentioned, if this runs on netfx (which confused me a bit since the TFM here is netcoreapp3.0), then yes, you need to reference the netstandard2.0 compatible NuGet package which contains the same API surface as what's built-in on core.

Does that help clarify things?

vatsan-madhavan: > Even if we had a package, we still can't use it because the built-in VisualStudio text templating engine doesn't support .NET Core, and so we can't load the assembly anyways.

>  yes, you need to reference the netstandard2.0 compatible NuGet package

@stevenbrix, excellent clarification! An analogous problem arises for the compilation of WPF applications themselves, which is why `Microsoft.NET.Sdk.WindowsDesktop` carries `PresentationBuildTasks.dll` for both `netcoreapp2.1` and `net472` tfm's. The former is used for `dotnet` builds and the latter for VisualStudio builds.

eng/WpfArcadeSdk/tools/CodeGen/DesignTimeTextTemplating.targets
vatsan-madhavan: Even if DARC doesn't work for dependency updates (you'll have to test and identify whether a dependency is meaningful for this package), please don't hard-code the version here unless you have a dependency on a specific version. Instead, add the version into eng\Versions.props and consume an MSBuild variable here. You just won't have a DARC dependency, but that's ok.

eng/Version.Details.xml
vatsan-madhavan: Re XSD - You are qute right in observing that XSD (or even JSON schema checks) would be an overkill for this PR. 

We have at least one codegnerator that builds the shared protocol between the UI and the render thread and also helps generate a debugger extensions, along with the render-thread commanding infrastructure.  It already uses XML + XSD and will likely be our most-significant generator (asides from the ones that are involved in the Theme assemblies).  

I'm not suggesting that we use XSD etc. everywhere. I'm suggesting that we consider XML as a standard language for codegen-data so that our codegenerators remain uniform, allow for XSD's where we think they make sense, and relieve us of the need to search for compatible parsers etc.

vatsan-madhavan: @ahsonkhan - the TFM in the project you are looking at is not related to the runtime environment where  `System.Text.Json` would be used.  The reference is supplied to the text-templating engine, and not to the project. The Text-templating engine runs in the context of either MSBuild or `dotnet`. 

What @stevenbrix  needs is the ability to reference an adaptive version of `System.Text.Json` - i.e., being able to reference `netstandard2.0` or `net4x` when invoked by MSBuild, and the ability to reference `netcoreapp2.x` or `netcoreapp3.0` when used from within `dotnet`. 

See how we do this in [`Microsoft.NET.Sdk.WindowsDesktop` ](https://github.com/dotnet/wpf/blob/b25e77aac80430a01deac92b48a90c0ab27d35f1/src/Microsoft.DotNet.Wpf/src/PresentationBuildTasks/Microsoft.WinFx.targets#L7):

```XML
    <_PresentationBuildTasksTfm Condition="'$(MSBuildRuntimeType)' == 'Core'">netcoreapp2.1</_PresentationBuildTasksTfm>
    <_PresentationBuildTasksTfm Condition="'$(MSBuildRuntimeType)' != 'Core'">net472</_PresentationBuildTasksTfm>
    <_PresentationBuildTasksAssembly Condition="'$(_PresentationBuildTasksAssembly)'==''">$(MSBuildThisFileDirectory)..\tools\$(_PresentationBuildTasksTfm)\PresentationBuildTasks.dll</_PresentationBuildTasksAssembly>

```

eng/Version.Details.xml
```diff
@@ -75,5 +75,10 @@
...
+      <Uri>https://github.com/JamesNK/Newtonsoft.Json</Uri>
```
vatsan-madhavan: There may be no real dependency arc possible from `JamesNK`

eng/Versions.props
```diff
@@ -4,6 +4,7 @@
     <VersionPrefix>4.8.0</VersionPrefix>
     <PreReleaseVersionLabel>preview5</PreReleaseVersionLabel>
     <AssemblyVersion>4.0.0.0</AssemblyVersion>
+    <NewtonsoftJsonVersion>12.0.1</NewtonsoftJsonVersion>
```
vatsan-madhavan: consider moving this to a separate section and annotating it with a comment that this is a static version. Also consider renaming as `NewtonsoftJsonVersionConstant` or some such name .

eng/Version.Details.xml
ahsonkhan: Presumably, a netstandard2.0 compatible NuGet package for `System.Text.Json` will help with that. @stevenbrix, let me know if that does not work for your use case here (once its available).

https://github.com/dotnet/corefx/issues/37126
https://github.com/dotnet/corefx/pull/37129

eng/WpfArcadeSdk/tools/CodeGen/DesignTimeTextTemplating.targets
stevenbrix: i converted to using xml

eng/Version.Details.xml
stevenbrix: So per other comments about which data format we should use (JSON or XML). I switched over to using XML so that we have a standard data format across all our code gen

eng/Versions.props
stevenbrix: newtonsoft has been removed

eng/WpfArcadeSdk/tools/CodeGen/DesignTimeTextTemplating.targets
stevenbrix: yeah i hear ya. i moved over to xml though so i'm just removing this

eng/WpfArcadeSdk/tools/CodeGen/DesignTimeTextTemplating.targets
```diff
@@ -0,0 +1,89 @@
+<Project>
+  <ItemGroup>
+    <Service Include="{508349b6-6b84-4df5-91f0-309beebad82d}" />
+  </ItemGroup>
+  <!-- Including the Microsoft.TextTemplating.targets requires us to manually import Sdk.Targets because
```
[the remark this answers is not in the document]
stevenbrix: yeah i meant to remove this entirely and just put it in the documentation

Documentation/codegen.md
```diff
@@ -0,0 +1,104 @@
+# CodeGen in the dotnet/wpf repo
+
+The following document describes how code generation in this repo works. The goal is to have all our code generation done through the use of T4 text generation. See the offical [Visual Studio T4 documentation](https://docs.microsoft.com/en-us/visualstudio/modeling/design-time-code-generation-by-using-t4-text-templates?view=vs-2019) for more information.
+
+## Design-Time vs Run-Time T4 templates
+Currently, we are evaluating the use of design-time text templates. This gives us the ability to simply add the templates and associated targets to the build, without the need of maintaining a separate tool to do run-time generation. Including the `Microsoft.TextTemplating.targets` requires us to manually import Sdk.Targets because it needs to be imported after Sdk.targets. This causes the `BuildDependsOn` variable, which is modified by the T4 targets, to be overwritten, so the `TransformAll` target doesn’t run before the Build target. The boilerplait for including design-time templates has been encapsulated in the `$(WpfCodeGenDir)DesignTimeTextTemplating.targets` file, so the pattern for enabling these in a project looks like this:
```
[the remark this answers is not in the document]
stevenbrix: Rewording to:

Including the `Microsoft.TextTemplating.targets` requires us to manually import `Sdk.Targets` because the `BuildDependsOn` variable, which is modified by the T4 targets, would otherwise be overwritten by the automatic inclusion of `Sdk.targets` when using the SDK-style project format

PR #637, opened 2019-05-03
Adding initial helix support 

stevenbrix:
### Testing in Helix
This PR enables running our DRT tests on Helix. I'd recommend seeing the official Helix [readme](https://github.com/dotnet/arcade/blob/master/src/Microsoft.DotNet.Helix/Sdk/Readme.md) if you are interested in some of the general Helix concepts. I'll briefly outline what we are doing that is a bit unique:

**NOTE**: The DRTs aren't currently running, as we are still waiting on the package to be published to a feed. But I have validated that locally I am able to reference a locally hand-crafted `Microsoft.DotNet.Wpf.DncEng.Test` package and run tests out of it. 

1. Helix has built-in support for running xUnit tests. Since we are not using xUnit, we have to manually setup our machines so that they work with QualityVault and STI.
2. We manually restore the `helixpublish.proj` project during CI builds. This allows us to package reference the `Microsoft.DotNet.Wpf.DncEng.Test` package to pull down the binaries that we need. We can extract them into a local folder using the MSBuild property `RestorePackagesPath`, so that when the project is restored, we know the exact relative location of the packages.
3. We need to create a "TestHost" installation of dotnet on the Helix machine. The default Helix support for installing dotnet only installs version 2.x, and only installs the `Microsoft.NETCore.App` runtime, and not the `Microsoft.WindowsDesktop.App` or `Microsoft.AspNetCore.App` runtimes. I opted to doing this manually and copy it out of the `.dotnet` folder that gets created as part of ci builds, rather than running an installer on the Helix machine. I figured reducing the number of dependencies on making external network calls and downloading items from the internet the better.
4. We copy both the nuget packages we care care about and the dotnet install in the `prepare-helix-payload.ps1` script. This gets invoked before we actually run tests. Helix allows you to specify a `HelixCorrelationPayload` directory, where this directory gets deployed to the Helix machine, and is made available in your various helix commands with the `HELIX_CORRELATION_PAYLOAD` environment variable.
5. Helix and Azure Pipelines can report xUnit logs, so we will be updating QualityVault to produce an xUnit compatible log. We will then need to copy that log to a known location for it to be picked up.

### How we are running tests
1. Currently, the `HelixQueues` that we selected are Windows.10.Amd64.Open;Windows.7.Amd64.Open;Windows.10.Amd64.ServerRS5.Open. Essentially, this translates to: "Latest Windows 10", "Windows 7", and "Windows 10 RS5 Server" addition. I felt this enables tests to run on some of the most interesting and important SKUs, including Windows Server, without overloading the Helix servers and/or making CI runs take an unnecessarily long time to run. I decided to peg the Windows Server to a particular OS build, because I felt it added a little more variety than always testing latest Windows Client and latest Windows Server. These decisions are of course up for discussion and aren't set in stone.
2. In similar vain, we only run test passes for Release builds, mainly to save time and resources running duplicate tests in a near-identical environment. If we'd rather run Debug tests so that we can catch asserts in our CI gates, I'm open to that as well.


eng/helix/configure-helix-machine.ps1
```diff
@@ -0,0 +1,18 @@
+# This file prepares the helix machine for our tests runs
+
+$dotnetLocation = Join-Path (Split-Path -Parent $script:MyInvocation.MyCommand.Path) "dotnet"
+
+Write-Output "Adding dotnet location to path: $dotnetLocation"
+
+# Prepend the path to ensure that dotnet.exe is called from there.
+$env:path="$dotnetLocation;$env:path"
+
+# Don't look outside the TestHost to resolve best match SDKs.
```
vatsan-madhavan: Why do we still use the concept of a TestHost -  shouldn't this instead be relying upon Arcade's multi-framework support and running off of `Microsoft.NetCore.App` version corresponding to `RuntimeFrameworkVersion` + project references to WPF assemblies, and impliict support provided by Arcade for `DOTNET_MULTILEVEL_LOOKUP=0` etc. when `ContinousIntegrationBuild==true`?

eng/helix/configure-helix-machine.ps1
```diff
@@ -0,0 +1,18 @@
...
+$env:DOTNET_MULTILEVEL_LOOKUP = 0
+
+# Disable first run since we do not need all ASP.NET packages restored.
```
vatsan-madhavan: Why disable this? 

By design, all 3 framework packages are always restored. We should retain whatever customer experiences we plan on shipping.

eng/helix/configure-helix-machine.ps1
```diff
@@ -0,0 +1,18 @@
...
+$env:DOTNET_SKIP_FIRST_TIME_EXPERIENCE=1
+
+# Ensure that the dotnet installed at our dotnetLocation is used
+$env:DOTNET_ROOT=$dotnetLocation
+Set-Variable -name "DOTNET_ROOT(x86)" -value $dotnetLocation
```
vatsan-madhavan: Will this ever use the `x64` version?

src/Microsoft.DotNet.Wpf/test/BranchCommon/data/CommonData.csproj
```diff
@@ -0,0 +1,19 @@
+﻿<Project Sdk="Microsoft.Build.NoTargets/1.0.40">
```
vatsan-madhavan: This is not how we use NoTargets Sdk. Please use the pattern already in place.

src/Microsoft.DotNet.Wpf/test/DRT/DrtXaml/XamlTestClasses/XamlTestClasses.csproj
```diff
@@ -4,10 +4,13 @@
...
     <Platforms>AnyCPU;x64</Platforms>
     <UseWPF>true</UseWPF>
+    <OutputType>Library</OutputType>
+    <DefineConstants>$(DefineConstants);OLDRESOURCES</DefineConstants>
+    <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
   </PropertyGroup>
   
   <PropertyGroup>
-    <SignAssembly>true</SignAssembly>
+    <SignAssembly>false</SignAssembly>
```
vatsan-madhavan: Are the test assemblies getting signed? If so, what is the signing scheme being employed?

src/Microsoft.DotNet.Wpf/test/DRT/DrtXaml/FriendWithoutKey/XamlTestClasses.FriendWithoutKey.csproj
```diff
@@ -0,0 +1,12 @@
+﻿<Project Sdk="Microsoft.NET.Sdk.WindowsDesktop">
+    <PropertyGroup>
+    <AssemblyName>XamlTestClasses.FriendWithoutKey</AssemblyName>
+    <OutputType>Library</OutputType>
+    <DefineConstants>$(DefineConstants);OLDRESOURCES</DefineConstants>
+    <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
+    <SignAssembly>false</SignAssembly>
```
vatsan-madhavan: SignAssembly - configure policy centrally.

src/Microsoft.DotNet.Wpf/test/Directory.Build.props
```diff
@@ -1,20 +1,56 @@
...
-  </PropertyGroup>
-  <Import Condition="Exists('../Directory.Build.props')" Project="../Directory.Build.props" />
-  <PropertyGroup>
-    <WpfTestsDir>$(MsBuildThisFileDirectory)</WpfTestsDir>
-    <SignAssembly>false</SignAssembly>
-    <IsTestProject>true</IsTestProject>
-    <IsShipping>false</IsShipping>
-    <EnableXlfLocalization>false</EnableXlfLocalization>
-    <CLSCompliant>false</CLSCompliant>
+    <WpfTestBasePath>$(MsBuildThisFileDirectory)</WpfTestBasePath>
```
vatsan-madhavan: paths should be configured centrally, and not in projects ad-hoc.

src/Microsoft.DotNet.Wpf/test/Directory.Build.props
```diff
@@ -1,20 +1,56 @@
...
+    <DrtCommonRoot>$(WpfTestBasePath)\Common\DRT</DrtCommonRoot>
+    <WpfSharedRoot>$(WpfTestBasePath)\Shared</WpfSharedRoot>
+    <PublishRoot>$(RepoRoot)artifacts\test\</PublishRoot>
+    <TargetFramework>netcoreapp3.0</TargetFramework>
+    <CopyLocalLockFileAssemblies>true</CopyLocalLockFileAssemblies>
```
vatsan-madhavan: This like `CopyLocalLockfileAssemblies` should be configured centrally for test projects (or for wpfTest==true) and not per-project/ad-hoc.

src/Microsoft.DotNet.Wpf/test/Directory.Build.props
```diff
@@ -1,20 +1,56 @@
...
+    <RuntimeIdentifiers>win-x64;win-x86</RuntimeIdentifiers>
+    <PlatformFolder>x86</PlatformFolder>
```
vatsan-madhavan: What does PlatformFolder do?

src/Microsoft.DotNet.Wpf/test/PublishAfterBuild.targets
```diff
@@ -0,0 +1,18 @@
+<?xml version="1.0" encoding="utf-8"?>
```
vatsan-madhavan: What are the pros/cons of this target being left here, vs. being moved under eng\WpfArcadeSdk\tools/

vatsan-madhavan: Likewise, discuss each individualized props/targets file that appears under test\. Each of them breaks the general model that build-related intelligence is refactored out of src\ into a separate location.

eng/helix/configure-helix-machine.ps1
stevenbrix: that's fine, the arcade script does this but it probably isn't necessary

eng/helix/configure-helix-machine.ps1
stevenbrix: i'm taking a walk, crawl, run approach here. We already know how to run DRTs as they are, and we're now trying to move that into Helix. I'd rather not try to develop the new method of how to build and run the DRTs at the same time that we're also doing something else. that is just going to slow down the entire process.

vatsan-madhavan: We should use the right approach in production codebases. 

The right approach is is in fact the easier one, and I don't believe it fits the mode of walk-crawl-run. In my estimation, the changes required to build these projects using `Microsoft.Net.Sdk` should take a couple of hours of work at most (for System.Xaml DRTs') - I know because I tried doing it once as a trial until I had to stop due to the lack of `multi-framework` and `InternalMarkupCompilation` support. They are both available to you readily now.

PR #663, opened 2019-05-07
Fix ResX Code Generation

### LINKED ISSUE #662: Fix ResX code generation due to changes in Arcade
Latest Arcade changes in https://github.com/dotnet/wpf/tree/darc-master-154bfb93-0564-405e-88b9-df2489fd6e99 caused ResX code generation to break.

This results in missing SR code files and is blocking DARC PRs.

rladuca:
Due to a change in Arcade in GenerateResxSource.targets, we need to set GenerateResxSource to true so that codegen runs on our EmbeddedResources.  This is blocking some DARC PRs with the latest Arcade changes.

fixes #662

### DIFF
--- a/eng/WpfArcadeSdk/tools/SystemResources.props
+++ b/eng/WpfArcadeSdk/tools/SystemResources.props
@@ -1,5 +1,16 @@
 <?xml version="1.0" encoding="utf-8"?>
 <Project>
+
+  <PropertyGroup>
+    <!-- 
+         Arcade requires us to set this property in order for it to appropriately update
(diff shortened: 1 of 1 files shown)


rladuca:
@vatsan-madhavan There is only one knob, the GenerateResxSource property.  I left GenerateSource set EmbeddedResource since, otherwise, we'd have to wait for DARC flow to get things working.  With both in there, we get current builds succeeding + builds with the new Arcade stuff.  Once that is pushed everywhere I can remove GenerateSource if we want.

PR #680, opened 2019-05-13
Fix bug where x:Static argument is treated as member of 'SystemParameters' by mistake

### LINKED ISSUE #679: XAML {x:Static FooClass.Border} will be compiled unexpectedly to {x:Static SystemParameters.Border}
* .NET Core Version: 3.0 Preview5
* Windows version: 10.0.18895.1000

 
 
 **Problem description:**

When assigning a WPF dependency property using `x:Static` markup extension and pass a CLR property named `Border` into it, the XAML compiler will compile this static property to `SystemParameters.Border` which is not expected.

 **Actual behavior:**

This first one will be compiled to the second.

```xml
<!-- The original code -->
<Button Content="{x:Static local:FooClass.Border}" />
```

```xml
<!-- The compiled code -->
<Button Content="{x:Static SystemParameters.Border}" />
```

 **Expected behavior:**

XAML compiler should not change the original property assigning of `x:Static` markup extension.
 
 **Minimal repro:**
 
1. Create a WPF (.NET Core / .NET Framework) project;
1. Write a class with a static property named `Border`;
1. Write a `x:Static` usage in a XAML file.

You can view the minimal reproduction on my GitHub project here:

- [walterlv/XamlBugs.StaticProperties](https://github.com/walterlv/XamlBugs.StaticProperties)

The key codes are here:

```csharp
// This namespace can be changed to any other and the bug still reproduces.
namespace XamlBugs.StaticProperties
{
    /// <summary>
    /// The class that provide a `Border` property.
    /// You can change this class name to any other and the bug still reproduces.
    /// </summary>
    public static class FooClass
    {
        /// <summary>
        /// The correct value for this property is `string.Empty`, but you'll see `1` in actual.
        /// </summary>
        public static string Border => "";
    }
}
```

```xml
<Window x:Class="XamlBugs.StaticProperties.MainWindow"
        xmlns:local="clr-namespace:XamlBugs.StaticProperties"
        mc:Ignorable="d"
    <Grid>

        <!-- When you run this application, and you'll see an incorrect content value of this `Button`.
             If you try to decompile this assembly and view the mainwindow.baml file,
             you'll find that the static value has been changed to `SystemParameters.Border` -->
        <!-- The XAML namespace prefix can be changed to any other and the bug still reproduces. -->
        <Button Content="{x:Static local:FooClass.Border}" />

        <!-- Don't uncomment this line, because it will cause a runtime error.
             The `FooClass.Border` property is a string,
             but it will be compiled to `SystemParameters.Border` which is not a string. -->
        <!--<TextBlock Text="{x:Static local:FooClass.Border}" />-->
        
    </Grid>
</Window>
```

KodamaSakuno:
Fixes #679

The issue is that the original code forgets to check `memberInfo`'s declaring type. When there's a property which shares a same member name of `SystemParameters` or `SystemResourceKeyID`, it'll be treated as member of `SystemParameters`.

The following content is a part of compiled baml of issue demo. Generated by dnSpy.

Before:
```
ElementStart [TypeId=0xff02, Flags=0x00]
    ContentProperty [AttributeId=0xff2b]
    LineNumberAndPosition [LineNumber=0x0000000f, LinePosition=0x0000000a]
    ElementStart [TypeId=0xffc9, Flags=0x00]
        PropertyWithExtension [AttributeId=0xfff2, Flags=0x025a, ValueId=0xfe52]
        LinePosition [LinePosition=0x00000011]
    ElementEnd
ElementEnd
```
After:
```
ElementStart [TypeId=0xff02, Flags=0x00]
    ContentProperty [AttributeId=0xff2b]
    LineNumberAndPosition [LineNumber=0x0000000f, LinePosition=0x0000000a]
    ElementStart [TypeId=0xffc9, Flags=0x00]
        TypeInfo [TypeId=0x0001, AssemblyId=0x0000, TypeFullName="XamlBugs.StaticProperties.FooClass"]
        AttributeInfo [AttributeId=0x0001, OwnerTypeId=0x0001, AttributeUsage=0x00, Name="Border"]
        PropertyWithExtension [AttributeId=0xfff2, Flags=0x025a, ValueId=0x0001]
        LinePosition [LinePosition=0x00000011]
    ElementEnd
ElementEnd
```

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Markup/BamlMapTable.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Markup/BamlMapTable.cs
@@ -1509,9 +1509,7 @@ namespace System.Windows.Markup
                     Debug.Assert(targetType != null && memberName != null);
 
                     MemberInfo memberInfo = XamlTypeMapper.GetStaticMemberInfo(targetType, memberName, false);
-                    if (memberInfo is PropertyInfo &&
-                        targetType.Assembly == XamlTypeMapper.AssemblyPF &&
-                        targetType.FullName == "System.Windows.SystemParameters")
+                    if (memberInfo is PropertyInfo)
                     {
                         // see if the value is a static SystemResourceKey property from the themes
                         memberId = SystemResourceKey.GetBamlIdBasedOnSystemResourceKeyId(targetType, memberName);
--- a/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Markup/BamlMapTable.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Markup/BamlMapTable.cs
@@ -1509,7 +1509,9 @@ namespace System.Windows.Markup
                     Debug.Assert(targetType != null && memberName != null);
 
                     MemberInfo memberInfo = XamlTypeMapper.GetStaticMemberInfo(targetType, memberName, false);
-                    if (memberInfo is PropertyInfo)
+                    if (memberInfo is PropertyInfo &&
+                        targetType.Assembly == XamlTypeMapper.AssemblyPF &&
+                        targetType.FullName == "System.Windows.SystemParameters")
                     {
                         // see if the value is a static SystemResourceKey property from the themes
                         memberId = SystemResourceKey.GetBamlIdBasedOnSystemResourceKeyId(targetType, memberName);
--- a/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/SystemResourceKey.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/SystemResourceKey.cs
@@ -364,6 +364,7 @@ namespace System.Windows
             short memberId = 0;
             string srkField = null;
             bool isKey = false;
+            bool found = true;
 
             if (memberName.EndsWith("Key", false, TypeConverterHelper.InvariantEnglishUS))
             {
@@ -382,9 +383,14 @@ namespace System.Windows
             {
                 srkField = memberName;
             }
+#if PBTCOMPILER
+            
+            found &= targetType.Assembly == XamlTypeMapper.AssemblyPF &&
+                targetType.FullName == "System.Windows.SystemParameters";
+                
+#endif
 
-            if (targetType.Assembly == XamlTypeMapper.AssemblyPF &&
-                targetType.FullName == "System.Windows.SystemParameters" &&
+            if (found &&
                 Enum.TryParse(srkField, out SystemResourceKeyID srkId))
             {
                 bool isExtended = ((short)srkId > SystemResourceKeyIDExtendedStart &&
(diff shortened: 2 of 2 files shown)


src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Markup/BamlMapTable.cs
```diff
@@ -1509,7 +1509,9 @@ internal void LoadTypeInfoRecord(BamlTypeInfoRecord record)
                     Debug.Assert(targetType != null && memberName != null);
 
                     MemberInfo memberInfo = XamlTypeMapper.GetStaticMemberInfo(targetType, memberName, false);
-                    if (memberInfo is PropertyInfo)
+                    if (memberInfo is PropertyInfo &&
```
walterlv: I find that most WPF code files put && at the head of the code line instead of the end of it.

weltkante:
Are you sure you want to fix this in the caller and don't want to put a fix in `SystemResourceKey.GetBamlIdBasedOnSystemResourceKeyId`? It receives the `targetType` as argument and then just ignores it when selecting the memberId from the enum.

walterlv:
I really don't like this naming `XamlTypeMapper.AssemblyPF`, I spend too much time to find that it indicates the PresentationFramework assembly.

src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Markup/BamlMapTable.cs
KodamaSakuno: I follow the style like line 450 in the same file.

KodamaSakuno:
@weltkante Ok, I moved it.

src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/SystemResourceKey.cs
```diff
@@ -387,16 +383,9 @@ internal static short GetBamlIdBasedOnSystemResourceKeyId(Type targetType, strin
...
-            }
-            catch (ArgumentException)
-            {
-                found = false;
-            }
-
-            if (found)
+            if (targetType.Assembly == XamlTypeMapper.AssemblyPF &&
+                targetType.FullName == "System.Windows.SystemParameters" &&
+                Enum.TryParse(srkField, out SystemResourceKeyID srkId))
```
miloush: Enum.TryParse is not available on .NET 3.0 and 3.5 which support WPF and out variables require C# 7.0. Might be OK, just pointing out this is changing runtime and compilation requirements.

weltkante: This repository is for .NET Core 3+ only, so you can assume thats available.

KodamaSakuno: It's SDK code, not runtime. And then, as far as I'm concerned, .Net Framework 3.0 and 3.5 are both out of support at present, so they're impossible to get this change, especially runtime. What's more, it's unnecessary to build WPF app with SDK source code referenced. So it doesn't matter to use C# 7.0+. :)

rchauhan18 (approved):

rchauhan18:
Thanks @KodamaSakuno for your contribution.

PR #703, opened 2019-05-17
Fix packaging to place RID specific binaries under runtimes\<rid>\native

vatsan-madhavan:
WPF's internal transport packages currently place RID specific binaries under lib\<RID> folder. 

This change moves them under runtimes\<RID>native\ folder, per spec at [Supporting Multiple .NET Versions](https://docs.microsoft.com/en-us/nuget/create-packages/supporting-multiple-target-frameworks). 

When referencing one of these packages from a test, we noticed that the native assemblies were not getting included in `deps.json`, which lead to the identification of this problem. Placing the binaries under the correct location remedies the `deps.json` creation errors.


vatsan-madhavan:
FYI @dagood

PR #711, opened 2019-05-19
RuntimeFrameworkVersion enablement - final step (test projects)

vatsan-madhavan:
This is the final change that brings tests into the fold of building using `RuntimeFrameworkVersion`.

This completes the work that we started at the end of Preview 4 to to use `RuntimeFrameworkVersion` to build the product bits (https://github.com/dotnet/wpf/pull/534, https://github.com/dotnet/wpf/pull/604) in order to improve the coherent-build process for core-sdk. Tests could not be enabled at that time because Arcade SDK lacked the capability to install multiple side-by-side SDK's (an extra version corresponding to `$(MicrosoftNETCoreAppVersion)` (defined in `eng\Versions.props`), needed to run tests since they would build against that version of `Microsoft.NetCore.App`). 

This missing functionality has since been enabled by https://github.com/dotnet/arcade/pull/2343 and https://github.com/dotnet/arcade/pull/2815; along with these, https://github.com/dotnet/wpf/pull/703 helps fix a prerequisite problem. Taken together, we are now able to build the entire repo using `RuntimeFrameworkVersion` against `$(MicrosoftNETCoreAppVersion)` version of `Microsoft.NETCore.App`


eng/WpfArcadeSdk/tools/TestProjects.targets
```diff
@@ -0,0 +1,26 @@
+<Project>
+  <PropertyGroup Condition="'$(WpfTest)'=='true'">
+    <RuntimeIdentifier Condition="('$(Platform)'!='AnyCPU' and '$(Platform)'!='Win32') Or $(Platform.EndsWith('64'))">win-$(Platform)</RuntimeIdentifier>
+    <RuntimeIdentifier Condition="'$(Platform)'=='AnyCPU' or '$(Platform)'=='Win32' or '$(Platform)'=='x86'">win-x86</RuntimeIdentifier>
+  </PropertyGroup>
+
+  <PropertyGroup Condition="'$(IsTestProject)' == 'true' or '$(WpfTest)'=='true'">
+    <TestArchitectures Condition="'$(PlatformTarget)' == '' or '$(PlatformTarget)' == 'AnyCpu'">x86</TestArchitectures>
```
rladuca: Might be worth a comment here that Arcade handles x64 correct, but doesn't correlate AnyCpu to x86.  I saw you have one in configure-toolset.ps1.

eng/WpfArcadeSdk/tools/TestProjects.targets
```diff
@@ -0,0 +1,26 @@
...
+  
+  <ItemGroup Condition="'$(WpfTest)'=='true'">
+    <PackageReference Include="runtime.$(RuntimeIdentifier).$(MicrosoftDotNetWpfDncEngPackage)" 
+                      Version="$(MicrosoftDotNetWpfDncEngVersion)">
+      <CopyLocal>true</CopyLocal>
+      <GeneratePathProperty>true</GeneratePathProperty>
+    </PackageReference>
+
+    <None Condition="'$(RuntimeIdentifier)'=='win-x64'" 
+          Include="$(Pkgruntime_win-x64_Microsoft_DotNet_Wpf_DncEng)\lib\$(RuntimeIdentifier)\*.dll"
```
rladuca: Might be worth a comment that these exist to ensure native dependencies are binplaced correctly with the test.

PR #726, opened 2019-05-22
Change DWriteForwarder to use __uuidof

### LINKED ISSUE #725: Change DWriteForwarder to use __uuidof
See: https://github.com/dotnet/coreclr/issues/24705

A workaround to this JIT issue is to sidestep our use of the GUID struct and just call __uuidof.

rladuca:
Removing use of GUID struct in initialization of the DWrite factory.  This is a workaround for a JIT bug that marshals incorrect GUID data, causing the cast to fail with E_NOINTERFACE.

fixes #725


ryalanms:
Why wasn't the macro used initially? Looks good.

rladuca:
@ryalanms There was an older bug in .NET, probably hadn't been relevant for a long time, see the comment in Factory.h.

PR #734, opened 2019-05-23
Update OSVersionHelper.cs

meziantou:
Change `OperatingSystemVersion.Windows10RS3` to `OperatingSystemVersion.Windows10RS5`

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/Shared/System/Windows/InterOp/OSVersionHelper.cs
+++ b/src/Microsoft.DotNet.Wpf/src/Shared/System/Windows/InterOp/OSVersionHelper.cs
@@ -278,7 +278,7 @@ namespace Microsoft.Internal.Interop
         {
             if (IsOsWindows10RS5OrGreater)
             {
-                return OperatingSystemVersion.Windows10RS3;
+                return OperatingSystemVersion.Windows10RS5;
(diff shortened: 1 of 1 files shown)


rladuca:
Yep, this is in both places.  GetOsVersion() is only used in error text for CompositeFont parsing, which basically never occurs, so that's why it never affected anything.

ryalanms:
Good catch, @meziantou.

vatsan-madhavan:
Historical note: 

When we added these checks a few years ago, `RtlVerifyVersionInfo` was only exposed through the DDK headers - it was not documented as an entrypoint in `ntdll.dll`. So we consumed it in our native layer (`PresentationNative`, I think), and P/Invoked into our native DLL's from managed code. This is the current design.

Over the years, `RtlVerifyVersionInfo` has since been documented as being available as a user-mode entrypoint  [in `ntdll.dll`](https://docs.microsoft.com/en-us/windows-hardware/drivers/ddi/content/wdm/nf-wdm-rtlverifyversioninfo), which means that we can P/Invoke directly into `ntdll` now. 

[We only consume Win32 API's that are public and documented in WPF]

vatsan-madhavan:
@meziantou, thanks for the catch and the patch!

PR #753, opened 2019-05-24
Type equivalent performance

facebamm:
## Output 
|  | sourceByte | sourceInt |
| -- | -- | -- |
| TypecastingPtrDesc | 350ms  | 600ms  | 
| PtrDesc | 290ms | 276ms |

## Type equivalent performance test case 

```cs
        {
            Array sourceByte = new Byte[] { 5, 6, 7 };
            Array sourceInt = new int[] { 1, 2, 3 };
            Stopwatch watch = new Stopwatch();

            Console.WriteLine("## sourceByte ##");
            watch.Restart();
            for (int i = 0; i < 10_000_000; i++)
            {
                TypecastingPtrDesc(sourceByte);
            }
            watch.Stop();
            Console.WriteLine($"TypecastingPtrDesc: {watch.Elapsed.Milliseconds}ms");
            watch.Restart();
            for (int i = 0; i < 10_000_000; i++)
            {
                PtrDesc((byte[])sourceByte);
            }
            watch.Stop();
            Console.WriteLine($"PtrDesc: {watch.Elapsed.Milliseconds}ms");

            Console.WriteLine("## sourceInt ##");
            watch.Restart();
            for (int i = 0; i < 10_000_000; i++)
            {
                TypecastingPtrDesc(sourceInt);
            }
            watch.Stop();
            Console.WriteLine($"TypecastingPtrDesc: {watch.Elapsed.Milliseconds}ms");
            watch.Restart();
            for (int i = 0; i < 10_000_000; i++)
            {
                PtrDesc((int[])sourceInt);
            }
            watch.Stop();
            Console.WriteLine($"PtrDesc: {watch.Elapsed.Milliseconds}ms");
        }


        private unsafe static void TypecastingPtrDesc(System.Array value)
        {
            switch (value)
            {
                case byte[] byteArray:
                    fixed (void* valuePtr = byteArray)
                    {
                        for (int iElement = 0, nElement = byteArray.Length; iElement < nElement; iElement++)
                        {
                            *((int*)valuePtr + iElement) = iElement;
                        }
                    }
                    break;
                case int[] intArray:
                    fixed (void* valuePtr = intArray)
                    {
                        for (int iElement = 0, nElement = intArray.Length; iElement < nElement; iElement++)
                        {
                            *((int*)valuePtr + iElement) = iElement;
                        }
                    }
                    break;
            }
        }
        private unsafe static void PtrDesc<T>(T[] pixels) where T : unmanaged
        {
            fixed (void* pixelsPtr = pixels)
            {
                for (int iElement = 0, nElement = pixels.Length; iElement < nElement; iElement++)
                {
                    *((byte*) pixelsPtr + iElement) = (byte) iElement;
                }
            }
        }
```

### DIFF
--- a/eng/Version.Details.xml
+++ b/eng/Version.Details.xml
@@ -7,61 +7,61 @@
     </Dependency>
   </ProductDependencies>
   <ToolsetDependencies>
-    <Dependency Name="Microsoft.Win32.Registry" Version="4.6.0-preview6.19270.12" CoherentParentDependency="Microsoft.NETCore.App">
+    <Dependency Name="Microsoft.Win32.Registry" Version="4.6.0-preview6.19270.15" CoherentParentDependency="Microsoft.NETCore.App">
       <Uri>https://github.com/dotnet/corefx</Uri>
-      <Sha>ef1a5aa730098b6c3350977a991232c1ff11cfe3</Sha>
+      <Sha>11eba7947e9736109c588a1cfcb77987cb80f570</Sha>
     </Dependency>
-    <Dependency Name="System.CodeDom" Version="4.6.0-preview6.19270.12" CoherentParentDependency="Microsoft.NETCore.App">
+    <Dependency Name="System.CodeDom" Version="4.6.0-preview6.19270.15" CoherentParentDependency="Microsoft.NETCore.App">
       <Uri>https://github.com/dotnet/corefx</Uri>
-      <Sha>ef1a5aa730098b6c3350977a991232c1ff11cfe3</Sha>
+      <Sha>11eba7947e9736109c588a1cfcb77987cb80f570</Sha>
     </Dependency>
-    <Dependency Name="System.Configuration.ConfigurationManager" Version="4.6.0-preview6.19270.12" CoherentParentDependency="Microsoft.NETCore.App">
+    <Dependency Name="System.Configuration.ConfigurationManager" Version="4.6.0-preview6.19270.15" CoherentParentDependency="Microsoft.NETCore.App">
       <Uri>https://github.com/dotnet/corefx</Uri>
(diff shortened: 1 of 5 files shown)


weltkante:
Leaving aside that the PR history is messed up, I think you are just removing a single typecast. The performance overhead of that single typecast is irrelevant compared to copying the rest of the array content, transferring it to the rendering backend and loading it onto the GPU.

In other words, your testcase is testing the wrong thing. You are loading 10 million images each containing just 3 pixels. You are also skipping the whole rendering backend overhead of 10 million images, so even if your images were tiny, saving a single cast is probably not even observable in a real application.

facebamm:
@weltkante 
ok, it might be possible that my example is not the best But it shows what I mean.

By using the ``System.Array`` I don't know which type is inside the Array and if I have a look inside the ``CachedBitmap`` method then I can see, that it only supports basic .net types at the moment.

For more dynamically work it might be better to use the ``unmanaged`` keyword for supporting an ``Pixel[] : struct`` and getting a pointer without many typecasting

vatsan-madhavan:
- This PR is unlikely to be merged as-is - please [open an issue](https://github.com/dotnet/wpf/blob/master/Documentation/acceptance_criteria.md) to discuss
  - This also affords the WPF team an opportunity to decide whether we'd like to invite a PR. 
- State the problem that is being solved and the solution approach being recommended. 
  - It's not clear whether _one specific instance_ of runtime overhead involving typecasting is being solved (and why), or whether the goal to solve _a class_ of such problems (for e.g., searching and finding instances where `unmanaged` would prove to be useful) across the whole codebase, etc. 
- Please consider benchmarking improvements _as applied_ to WPF. 
  - For e.g., tests/benchmarks on a prototype that shows what specific WPF scenarios are improved, and share repeatable methodology and measurements. 
 
Please continue this conversation in a newly opened issue.

Issue #763, opened 2019-05-25
System.Reflection.MetadataLoadContext has an incoherent dependency on System.Memory -> causes build breaks in PresentationBuildTasks


vatsan-madhavan:
See https://github.com/dotnet/corefx/issues/37943 for context. 

- [x] Implement a workaround
  - #764 
- [x] Wait for corefx to fix  https://github.com/dotnet/corefx/issues/37943
- [x] Undo the workaround

rladuca:
CoreFX has been updated and we're seeing that in the DARC flow [here](https://github.com/dotnet/wpf/pull/826).

As such, we need to make the final change to remove the pinned packages and close the loop.

PR #764, opened 2019-05-25
Fix PBT build - workaround for corefx/37943

vatsan-madhavan:
Workaround for https://github.com/dotnet/corefx/issues/37943

### DIFF
--- a/eng/Version.Details.xml
+++ b/eng/Version.Details.xml
@@ -1,87 +1,87 @@
 <?xml version="1.0" encoding="utf-8"?>
 <Dependencies>
   <ProductDependencies>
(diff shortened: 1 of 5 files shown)


ryalanms:
The PBT output by users/vatsan/pbt-srmdlc-fixup successfully compiled some test XAML.

vatsan-madhavan:
> The PBT output by users/vatsan/pbt-srmdlc-fixup successfully compiled some test XAML.

Thanks - a second test is always great to have!

vatsan-madhavan:
> Seems fine to do for the short term. Should we have a tracking issue to remove the workaround and simply reference MetadataLoadContext once the dependencies of that are fixed up?

I don't think we should close https://github.com/dotnet/wpf/issues/763 until the right fix is in place.

PR #794, opened 2019-05-30
Remove PresentationCore-CommonResources and move composite fonts to PresentationCore

### LINKED ISSUE #793: Remove PresentationCore-CommonResources
`PresentationCore-CommonResources` is currently used to house the composite font files as a resource due to an issue with getting them to be included when we were building `PresentationCore` as a mixed DLL from a C++/CLI object file and a C# netmodule.

We have since removed this (via creating `DirectWriteForwarder`) so the main reason for keeping the resource out of `PresentationCore` directly no longer exists.

As such, we should remove this DLL and move the composite font resources into `PresentationCore` directly.

rladuca:
fixes #793


rladuca:
@miguep, I already fixed the DRT, see my internal branch of wpf_test.

vatsan-madhavan:
This requires core-setup changes for `FrameworkListProfile` item information. I think it's best to move this over to Preview 7.

src/Microsoft.DotNet.Wpf/src/PresentationCore/PresentationCore.csproj
```diff
@@ -1421,6 +1420,21 @@
     <ProjectReference Include="$(WpfSourceDir)UIAutomation\UIAutomationProvider\UIAutomationProvider.csproj" />
   </ItemGroup>
 
+  <!-- 
+    Need Microsoft.WinFx.targets to embed *.compositefont
```
vatsan-madhavan: Can PBT function before PresentationCore is built and available?

rladuca:
> This requires `core-setup` changes to `FrameworkListProfile` item information, I think. Probably best to do this in Preview 7.

Yeah, I'm perfectly happy to hold onto this PR until post-Preview 6.  I just wanted to get the bulk of the change up and out of the way ASAP.

eng/WpfArcadeSdk/tools/ShippingProjects.props
```diff
@@ -5,7 +5,6 @@
       PresentationBuildTasks;
       PresentationCore;
       DirectWriteForwarder;
-      PresentationCore-CommonResources;
```
vatsan-madhavan: I don't see a removal from `<UseMicrosoftSharedKeyId>` section. Looks like we have always been missing `PresentationCore-CommonResources` from that section. 

Becasue`PresentationCore-CommonResources`  was a _breakaway_ from `PresentationCore`, I we should have reasoned our way into assigning it the same _key_ as `PresentationCore`...

src/Microsoft.DotNet.Wpf/src/PresentationCore/PresentationCore.csproj
rladuca: It seems to have been fine.  I never added this change in the first place and I am not 100% sure exactly what PBT has to do with including XML files as a resource, but I wanted to preserve what was done in PresentationCore-CommonResources.csproj.  Apart from that, I definitely see the correct assemblies coming out, I can check the build logs to see what PBT is up to during PCore builds.

rladuca: Talked offline about this, and PBT doesn't need PCore just to do resource generation.

vatsan-madhavan:
> > This requires `core-setup` changes to `FrameworkListProfile` item information, I think. Probably best to do this in Preview 7.
> 
> Yeah, I'm perfectly happy to hold onto this PR until post-Preview 6. I just wanted to get the bulk of the change up and out of the way ASAP.

@dagood What's the best way to propagate this type of information to core-setup? Just make the change and let the build in core-setup fail?

eng/WpfArcadeSdk/tools/ShippingProjects.props
rladuca: Yeah, I just removed what reference we had.  Like we talked about, I think it's fine to leave the repo alone until next week when we can check this in without messing with Preview6 and the concern will go away.

rladuca:
Marking No Merge until post-preview 6.

dagood:
> > > This requires `core-setup` changes to `FrameworkListProfile` item information, I think. Probably best to do this in Preview 7.
> > 
> > 
> > Yeah, I'm perfectly happy to hold onto this PR until post-Preview 6. I just wanted to get the bulk of the change up and out of the way ASAP.
> 
> @dagood What's the best way to propagate this type of information to core-setup? Just make the change and let the build in core-setup fail?

In this specific case I wouldn't expect a build break--`Profile` classification is loose. Extra entries don't affect anything, and missing ones are treated as having no `Profile`. (But this is probably a bad thing and we should require a 1:1 mapping so things don't slip by. Well, more ideally we would programmatically determine the list, or at least maintain it in the WPF and WinForms repos rather than Core-Setup.)

Generally speaking, going ahead with breaking changes and keeping an eye on the resulting update PR to ping someone if needed seems to me like the most effective way to do things. The downside is that updates are blocked until it's fixed, and we don't want to be in a blocked state when Preview 6 snaps because then the break needs to be fixed in two places. So, I agree that waiting is a good move.

PR #795, opened 2019-05-30
Add PresentationFramework extensions to WPF solution

ryalanms:
Adds extensions to the WPF solution file, and includes a minor update to .editorconfig for .sln line endings.


Microsoft.Dotnet.Wpf.sln
```diff
@@ -1,562 +1,632 @@
-﻿
-Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio Version 16
```
vatsan-madhavan: Can this be re-done to limit changes to what is needed and no more ?

rladuca: I think it's probably best to do the config changes as a quick follow-up PR so we don't get the noisy diff.

ryalanms: It should be good now.

PR #796, opened 2019-05-30
Revert PresentationFramework to .NET Framework translations for DE

ryalanms:
A previous loc check-in translated locked strings, only in DE.   I'm reviewing other loc updates to verify this didn't happen elsewhere.

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/PresentationFramework/Resources/xlf/Strings.de.xlf
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationFramework/Resources/xlf/Strings.de.xlf
@@ -1,5 +1,5 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
-<xliff xmlns="urn:oasis:names:tc:xliff:document:1.2" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" version="1.2" xsi:schemaLocation="urn:oasis:names:tc:xliff:document:1.2 xliff-core-1.2-transitional.xsd">
+<xliff xmlns="urn:oasis:names:tc:xliff:document:1.2" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" version="1.2" schemaLocation="urn:oasis:names:tc:xliff:document:1.2 xliff-core-1.2-transitional.xsd">
   <file datatype="xml" source-language="en" target-language="de" original="../Strings.resx">
     <body>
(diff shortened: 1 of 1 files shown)


ryalanms:
There are two strings for each key combination, one for the CommandHandler, and one for the description of the command handler.  Only the description should be translated.  For some reason, the CommandHandler string was also translated, but only for DE.

      </trans-unit>
      <trans-unit id="KeyToggleInsert">
        <source>Insert</source>
        <target state="translated">Einfg</target>
        <target state="translated">Insert</target>
        <note />

      <trans-unit id="KeyToggleInsertDisplayString">
        <source>Insert</source>
        <target state="translated">Einfg</target>
        <note />
      </trans-unit>

vatsan-madhavan:
Wasn't the plan to _eliminate_ locked strings (and move them to sources as either _const,_ or _static_ strings), rather than having them continue to exist in these xlf files?

ryalanms:
Yes, PresentationFramework has a set of locked strings that need to be moved in to static strings.  This fixes the crash for Preview 6 until that can be done.

vatsan-madhavan:
> Yes, PresentationFramework has a set of locked strings that need to be moved in to static strings. This fixes the crash for Preview 6 until that can be done.

Do we have issues opened for each of the 'locked strings need to be moved' work ? Are there other assemblies this category? if not, let's open some some of these issues so we can prioritize them.

ryalanms:
I'm working on getting PresentationFramework updated today.

Lakritzator:
I would not have guessed this solution, the issue here is that the translation is actually not a translation. 😕 

Thanks for fixing it, the fix seems to be available in 3.0.100-preview7-012267.

onovotny:
I'm getting this for many cultures, not just German, here:
 https://github.com/NuGetPackageExplorer/NuGetPackageExplorer/issues/716

Spanish, Russian, Japanese, and German. None in English though.

brux88:
i have the same error for italian language

weltkante:
@brux88 The fix is in preview7, so getting the error in preview6 is normal. Preview7 is currently getting ready to be released (github master is already on preview8), if you don't want to wait you can use [nightly builds](https://github.com/dotnet/core/blob/master/daily-builds.md) or as a workaround set the current culture to english during startup (`CultureInfo.CurrentUICulture = CultureInfo.GetCultureInfo("en-US");`)

PR #802, opened 2019-05-31
Adding initial Helix support

### LINKED ISSUE #782: Get CI builds running in Helix

stevenbrix:
### Testing in Helix
This PR enables running our DRT tests on Helix. I'd recommend seeing the official Helix [readme](https://github.com/dotnet/arcade/blob/master/src/Microsoft.DotNet.Helix/Sdk/Readme.md) if you are interested in some of the general Helix concepts. I'll briefly outline what we are doing that is a bit unique:

1. Helix has built-in support for running xUnit tests. Since we are not using xUnit, we have to manually setup our machines so that they work with QualityVault and STI. During the build, we create a payload directory that contains the infrastructure we need. A single project (in our case `DrtXaml`) is responsible for creating this directory (see instances of the MSBuild property `CreateTestPayload`).
2. After the build is done, we utilize Arcade's `AfterSolutionBuild.targets` extension point to finish creating the rest of the payload if the `-test` parameter is passed to the build. Here we add the just built DRTs and if `-ci` was **not** passed into to the build, run the tests. 
3. Helix allows you to specify a `HelixCorrelationPayload` directory, where this directory gets deployed to the Helix machine, and is made available in your various helix commands with the `HELIX_CORRELATION_PAYLOAD` environment variable. We use the payload directory created described above.
4. Helix and Azure Pipelines can report xUnit logs, so QualityVault was updated to produce an xUnit compatible log. We then copy that log to a known location for it to be picked up. This location can be in any subfolder of the `HelixWorkItem` working directory. 

## How we are running tests
1. Currently, the `HelixQueues` that we selected are Windows.10.Amd64.Open;Windows.7.Amd64.Open;Windows.10.Amd64.ServerRS5.Open. Essentially, this translates to: "Latest Windows 10", "Windows 7", and "Windows 10 RS5 Server" addition. This enables tests to run on some of the most interesting and important SKUs, without overloading the Helix servers and/or making CI runs take an unnecessarily long time to run.
2. In a similar fashion, we only run test passes for Release builds, mainly to save time and resources running duplicate tests in a near-identical environment.

Fixes #782


Documentation/developer-guide.md
```diff
@@ -23,6 +23,28 @@ We use the following workflow for building and testing features and fixes.
...
+
+```
+  A total of 1 test Infos were processed, with the following results.
+   Passed: 1
+   Failed (need to analyze): 0
+   Failed (with BugIDs): 0
+   Ignore: 0
+
+```
+If there were any failures, you can cd into $(RepoRoot)\artifacts\test\$(Configuration)\$(Platform)\Test and run the tests manually with the `/debugtests` flag. This will open a cmd window, where you can open the test executable in Visual Studio and debug it. For example, to debug DrtXaml.exe you would enter this into the command window:
```
rladuca: This doesn't really tell anyone how to run the tests.  You should talk about the cmd scripts and what they do (e.g. RunDrts, etc.)  For instance, it can be confusing to run `RunDrtsDebug` and not realize you are debugging QV and not the tests.

Documentation/developer-guide.md
```diff
@@ -23,6 +23,28 @@ We use the following workflow for building and testing features and fixes.
...
+
+`"%ProgramFiles%\Microsoft Visual Studio\2019\Preview\Common7\IDE\devenv.exe" DrtXaml.exe`
```
rladuca: There are actually cmd scripts that come with QV for this.  When the debug prompt is up, you just type "vs" or "windbg" and it will load the appropriate debugger.  The instructions for this are in the debug window itself, but might make sense to mention it here.

eng/WpfArcadeSdk/tools/CreateTestPayload.targets
```diff
@@ -0,0 +1,49 @@
...
+    <MicrosoftDotNetWpfTestPath>$(Pkgruntime_win-x86_Microsoft_DotNet_Wpf_Test)\tools\$(LocalRuntimeIdentifier)\Test\</MicrosoftDotNetWpfTestPath>
+  </PropertyGroup>
+
+  <PropertyGroup Condition="'$(LocalRuntimeIdentifier)'=='win-x64'">
+    <MicrosoftDotNetWpfTestPath>$(Pkgruntime_win-x64_Microsoft_DotNet_Wpf_Test)\tools\$(LocalRuntimeIdentifier)\Test\</MicrosoftDotNetWpfTestPath>
+  </PropertyGroup>
+
+  <Target Name="CreateTestPayload" BeforeTargets="Test" AfterTargets="Build">
+    <ItemGroup>
+      <MicrosoftDotNetWpfTestContents Include="$(MicrosoftDotNetWpfTestPath)\**\*.*" />
```
rladuca: Extra backslash here after the path var.  Generally, path vars will always include a trailing backslash so you shouldn't add one during use.

rladuca: I'd scrub this file for cases where this occurs

rladuca: Just spot checked some other targets and we generally need a scrub of that, oops.

eng/WpfArcadeSdk/tools/RunDrtsLocal.targets
```diff
@@ -0,0 +1,7 @@
+<Project>
+  <!-- The internal and external repos each define their own drts to add that need running. -->
```
rladuca: I generally like this idea, let me potentially suggest an improvement.

Instead of a complicated target file being the dictator of what gets added, can't we have a simple ItemGroup in a props file?  Then the messier parts of the target (like platform detection, etc) just live in the WPF Arcade SDK.  The individual repos just change up the ItemGroup includes.

eng/AddDrtsToPayload.targets
```diff
@@ -0,0 +1,26 @@
...
+    <PlatformFolder Condition="'$(Platform)'=='x64'">x64</PlatformFolder>
+
+    <!-- For the test payload, we default to x86 for AnyCPU platform --> 
+    <PayloadPlatformFolder>x86</PayloadPlatformFolder>
+    <PayloadPlatformFolder Condition="'$(Platform)'=='x64'">x64</PayloadPlatformFolder>
+    <PayloadBaseLocation>$(RepoRoot)artifacts\test\$(Configuration)\$(PayloadPlatformFolder)</PayloadBaseLocation>
+    <PayloadDrtLocation>$(PayloadBaseLocation)\Test\DRT</PayloadDrtLocation>
+  </PropertyGroup>
+
+  <ItemGroup>
```
rladuca: As I said below, it feels like this ItemGroup should live in a props file under eng, where the more targetish bits can go down into the SDK.

eng/AfterSolutionBuild.targets
```diff
@@ -0,0 +1,3 @@
+<Project>
+    <Import Project="$(MSBuildThisFileDirectory)WpfArcadeSdk\tools\RunDrtsLocal.targets" Condition="'$(Test)'=='True'" />
```
rladuca: The WPF Arcade SDK defines path vars that can be used to plumb down to the target you want.  It seems to me that if you used `WpfArcadeSdkToolsDir` you could move this targets file down into the SDK itself without an issue.  Then, if you do the props stuff I was talking about elsewhere, only the props file requires syncing across repos.  Well, apart from DARC deps.

rladuca: Bah!  Misunderstood this entirely!  Nevermind about moving things down.

eng/WpfArcadeSdk/tools/CreateTestPayload.targets
```diff
@@ -0,0 +1,49 @@
+<Project>
+    <!-- We use a local RID here since we don't want to change the build layout by setting the actual RID. -->
+  <PropertyGroup>
+    <LocalRuntimeIdentifier Condition="'$(Platform)'=='x64' or '$(Platform)'=='x86' or '$(Platform)'=='arm' or '$(Platform)'=='arm64'">win-$(Platform)</LocalRuntimeIdentifier>
```
rladuca: This is now at least 3 places (counting InjectModuleInitializer.targets) where we do this.  We should think about re-factoring this out into something common.

eng/WpfArcadeSdk/tools/configure-helix-machine.ps1
```diff
@@ -0,0 +1,30 @@
...
+$runtimes = dotnet --list-runtimes
+foreach ($rt in $runtimes)
+{
+    if ($rt.StartsWith("Microsoft.WindowsDesktop.App"))
+    {
+        $version = $rt.Split(" ")[1]
+    }
+}
+
+# Rewrite the *.runtimeconfig.json files to match the version of the runtime on the machine
```
rladuca: Maybe I am misunderstanding this, but don't we dictate the version of the runtime on the Helix machine?  Why would we need to re-write it?  Multi-framework should mean that our deps.json is built targeting the chosen framework version from global,.json, right?  So we just need to ensure we have that framework on the machine.  What am I missing?

eng/helixpublish.proj
```diff
@@ -0,0 +1,122 @@
+<Project Sdk="Microsoft.DotNet.Helix.Sdk" DefaultTargets="Test">
+  <PropertyGroup>
+    <TargetFramework>netcoreapp3.0</TargetFramework>
+    <IncludeDotNetCli>true</IncludeDotNetCli>
+    <DotNetCliPackageType>runtime</DotNetCliPackageType>
+    <!-- Set to match global.json.tools.dotnet by pre-build.ps1 -->
+    <DotNetCliVersion>$(MicrosoftNetCoreAppVersion)</DotNetCliVersion>
+    <DotNetCliRuntime>win-x86</DotNetCliRuntime>
```
rladuca: Is `RuntimeIdentifier` set at this point, can you just use that?

eng/helixpublish.proj
```diff
@@ -0,0 +1,122 @@
...
+  </PropertyGroup>
+
+  <ItemGroup>
+    <HelixCorrelationPayload Include="$(ArtifactsDir)\test\$(Configuration)\$(Platform)" /> 
+    <HelixWorkItem Include="Xaml">
+      <Timeout>00:20:00</Timeout>
+      <Command>call %HELIX_CORRELATION_PAYLOAD%\runtests.cmd /Area=Xaml</Command>
+    </HelixWorkItem>
+   
+   <!--
```
rladuca: You have the ItemGroup that defines the DRT set.  Can you use that ItemGroup here to build your work items by batching?

rladuca:
Just marking NO MERGE until Preview 6 clears.

eng/AddDrtsToPayload.targets
vatsan-madhavan: Reuse `$(Architecture)` instead of computing yet another property derived from `$(Platform)`

stevenbrix: Are we anticipating packaging the DRTs that built out in the open along with the infrastructure? I was thinking we'd just package the infrastructure and run a different suite of tests internal.

Documentation/developer-guide.md
stevenbrix: those don't actually seem to work. that only launches vs in a way that debugs sti

eng/WpfArcadeSdk/tools/configure-helix-machine.ps1
stevenbrix: the infrastructure isn't built out in the open, so i was getting errors about not being able to find the proper version of WindowsDesktop.App. This should only be a temporary thing until we move it out into the open.

eng/AddDrtsToPayload.targets
stevenbrix: This one is a bit confusing, and maybe i should add some more documentation on it, but Arcade doesn't provide the same set of properties to the `AfterSolutionBuild.proj` build that they do to the restore or build projects:

https://github.com/dotnet/arcade/blob/b5016f5688dc8ca9f3e4811ee7e2e86ad8907a40/src/Microsoft.DotNet.Arcade.Sdk/tools/Build.proj#L225

eng/helixpublish.proj
stevenbrix: i only see `RuntimeIdentifiers` set at this point. although i can refactor some of the common logic out so we should be able to use that here

eng/helixpublish.proj
stevenbrix: yeah that's a good suggestion. i was thinking about it, but I really want to codegen that DRT set based off DrtList.xml, so I figured it could wait until then. although there's no real reason for waiting.

eng/WpfArcadeSdk/tools/configure-helix-machine.ps1
stevenbrix: actually i'm now wondering if this only worked because the helix machine had some old install of the WindowsDesktop runtime on it. now that i've updated the DotNetCliVersion to use the version of NetCoreApp that we compile WPF against, this doesn't work. We're going to publish the test infrastructure package as self-contained, so we can minimize these dependencies. once it's open-sourced we can compile it like we do today and things will just work

rladuca: The infrastructure isn't WindowsDesktop, afaik.  @miguep, nothing in QV or STI has anything to do with WindowsDesktop, correct?

rladuca: Was the infra package built during the time period where Arcade had the version bug?

rladuca: Publishing as self contained doesn't test what we want to test, unless you also copy over the entirety of the dotnet called out in multi-framework.  Also how does SCA infra work with a build -test run, what are we testing in that instance?

Documentation/developer-guide.md
rladuca: STI runs the test, that's the point.

rladuca: I see, that's probably the issue.  You can use windbg with attach to child-process to pick it up, but makes sense.

eng/WpfArcadeSdk/tools/configure-helix-machine.ps1
stevenbrix: QV has a dependency on WinForms for displaying the progress in the bottom right. We should be able to depend on the private WinForms package once we move it out into the open. Otherwise, I'd vote for removing the dependency entirely so it can just build against the `Microsoft.NET.Sdk`. What does SCA mean?

Which version bug are you referring to? I don't think we technically care about what QV or STI run against, as long as the actual tests themselves don't run against it. Unfortunately, building the infra as `SelfContained` breaks the `build -test` scenario (which seems weird to me). The test executables are now no longer able to find the version of `Microsoft.NETCore.App` that gets installed by Arcade in the `.dotnet` folder. I'm not exactly sure why this is, I would think the self contained QV, STI, and DrtXaml would still inherit the proper environment variables that point to the local install. There must be something here about a self-contained process calling `Process.Start` on a non-self-contained app that breaks this scenario. Maybe @ericstj can share more info about this.

stevenbrix:
I removed the no merge tag since the preview6 branch has snapped.

For more detailed information on some of the workarounds, see this comment:
https://github.com/dotnet/wpf/issues/816#issuecomment-498834686

PR #808, opened 2019-06-01
Add back ValueSerializerAttribute TypeForward to System.Xaml

### LINKED ISSUE #807: System.Xaml missing TypeForward for System.Windows.Markup.ValueSerializerAttribute
The type forward for ValueSerializerAttribute is [here](https://github.com/dotnet/wpf/blob/master/src/Microsoft.DotNet.Wpf/src/System.Xaml/System/Windows/Markup/ValueSerializerAttribute.cs), but this file is excluded from [System.Xaml.csproj](https://github.com/dotnet/wpf/blob/master/src/Microsoft.DotNet.Wpf/src/System.Xaml/System.Xaml.csproj).  This is due to historical reasons during porting,  We were attempting to support some scenarios (netstandard 2.1) that would not have this type.

This is an ApiCompat break and should be fixed.

rladuca:
Fixes #807

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/System.Xaml/System.Xaml.csproj
+++ b/src/Microsoft.DotNet.Wpf/src/System.Xaml/System.Xaml.csproj
@@ -60,7 +60,6 @@
 
   <ItemGroup>
     <Compile Remove="System\Xaml\XamlProperty.cs" />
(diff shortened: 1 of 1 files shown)


rladuca:
NO MERGE until Preview 6 clears.

PR #831, opened 2019-06-04
Remove pinned version of System.Memory

vatsan-madhavan:
https://github.com/dotnet/wpf/issues/763

corefx has fixed the underlying bug - removing pinned version of System.Memory. We need to cherry pick this into release/3.0 as well.


vatsan-madhavan:
I was able to build all the samples (well, almost all - we always get 3 failures as a baseline) under https;//github.com/microsoft/wpf-samples using both dotnet.exe and msbuild.exe using the new bits.

PR #862, opened 2019-06-04
Crash in DataGridItemAutomationPeer

### LINKED ISSUE #860: Crash in VS opening 'Naming'
Port from .Net 4.8 bug.

SamBent:
VS 2017 and 2019 crash when
a. Automation is enabled (e.g. Narrator or Inspect are running)
b. User clicks on the Text Editor\C#\Code Style\Naming page in Tools\Options
(also other similar pages)

This is a regression in 4.8.  The bad pages contain a DataGrid whose cells contain a ComboBox, and whose virtualizing mode is Recycling

Fixes #860


rladuca:
> Thanks Sam.
> Are there any other changes we nee to port from the 4.8 branch? I remember we had a task open to refresh the code.

There are going to be some more changes both in test code and potentially some product code ports that were missed.  We'll have to do a diff against 4.8 again soon, just to be sure.  @miguep when was the last sync between the two?

Issue #241, opened 2019-01-07
Stub CAS/Security in wpf


hughbe:
From https://github.com/dotnet/wpf/pull/214#discussion_r244214372

I'm guessing cause the source is shared we can't but opening for discussion

vatsan-madhavan:
@ojhad, since related PR have all been merged, I recommend closing this issue. 

For cleanup of remaining items, like various `using System.Security.Permissions;` and miscellaneous comments (which implicate System.Windows.Extensions), please open a new tracking issue. 

/cc @grubioe

PR #869, opened 2019-06-04
Hollowed out public types extending from System.Security.Permissions types

ojhad:
**Currently this PR does not include Type Forwarding of these types**

Part of Issue #241 
Hollowing out public types that should be moved out of WPF to System.Security.Permissions.dll


vatsan-madhavan:
Your build is passing, but tests are failing (so, well, the build is not passing, really). Do you have a sense for what's going on? 

```
   at DRT.DrtBase.ReportException(Object o) in /_/src/Microsoft.DotNet.Wpf/test/Common/TestServices/DrtBase.cs:line 2102
   at DRT.DrtBase.Run(String[] args) in /_/src/Microsoft.DotNet.Wpf/test/Common/TestServices/DrtBase.cs:line 296
   at DrtXaml.XamlDrt.RunTests(String[] args) in /_/src/Microsoft.DotNet.Wpf/test/DRT/DrtXaml/DrtXaml/DrtXamlBase.cs:line 99
   at DrtXaml.TestFrontEnd.<>c__DisplayClass1_0.<RunSuite>b__0() in /_/src/Microsoft.DotNet.Wpf/test/DRT/DrtXaml/DrtXaml/DrtXamlBase.cs:line 70
   at DrtXaml.TestFrontEnd.<>c__DisplayClass2_0.<StartSTATask>b__0() in /_/src/Microsoft.DotNet.Wpf/test/DRT/DrtXaml/DrtXaml/DrtXamlBase.cs:line 80
--- End of stack trace from previous location where exception was thrown ---
   at DrtXaml.TestFrontEnd.RunSuite(String[] data) in /_/src/Microsoft.DotNet.Wpf/test/DRT/DrtXaml/DrtXaml/DrtXamlBase.cs:line 70
--- End of stack trace from previous location where exception was thrown ---
```

src/Microsoft.DotNet.Wpf/src/System.Xaml/System/Xaml/Permissions/XamlAccessLevel.cs
```diff
@@ -7,144 +7,15 @@
...
-        // Stored as string: we need to store the assembly and type names, rather than Assembly or
-        // Type references, because permissions can be serialized, and we don't want to force an
-        // assembly load on deserialization in a different AppDomain.
-        
-        public AssemblyName AssemblyAccessToAssemblyName
-        {
-            get { return new AssemblyName(AssemblyNameString); }
-        }
-
+        private XamlAccessLevel(string assemblyName, string typeName) { }
```
vatsan-madhavan: Do we need to retain the private constructor?

src/Microsoft.DotNet.Wpf/src/System.Xaml/System/Xaml/Permissions/XamlAccessLevel.cs
```diff
@@ -7,144 +7,15 @@
...
+        public static XamlAccessLevel AssemblyAccessTo(Assembly assembly) { return default(XamlAccessLevel); }
+        public static XamlAccessLevel AssemblyAccessTo(AssemblyName assemblyName) { return default(XamlAccessLevel); }
+        public static XamlAccessLevel PrivateAccessTo(Type type) { return default(XamlAccessLevel); }
+        public static XamlAccessLevel PrivateAccessTo(string assemblyQualifiedTypeName) { return default(XamlAccessLevel); }
+        public AssemblyName AssemblyAccessToAssemblyName { get; set; }
```
vatsan-madhavan: Initialize this to default value perhaps?

src/Microsoft.DotNet.Wpf/src/System.Xaml/System/Xaml/Permissions/XamlLoadPermission.cs
```diff
@@ -14,327 +12,23 @@ namespace System.Xaml.Permissions
...
-                AllowedAccess = s_emptyAccessLevel;
-            }
-            else
-            {
-                Debug.Assert(!isUnrestricted);
-                AllowedAccess = new ReadOnlyCollection<XamlAccessLevel>(allowedAccess);
-            }
-        }
-
         public IList<XamlAccessLevel> AllowedAccess { get; private set; } // ReadOnlyCollection
```
vatsan-madhavan: Remove comment

src/Microsoft.DotNet.Wpf/src/System.Xaml/System/Xaml/Permissions/XamlLoadPermission.cs
```diff
@@ -14,327 +12,23 @@ namespace System.Xaml.Permissions
...
-            public const string Unrestricted = "Unrestricted";
-        }
+        public override IPermission Copy() { return default(IPermission); }
+        public override void FromXml(SecurityElement elem) { }
+        public bool Includes(XamlAccessLevel requestedAccess) { return false;  }
+        public override IPermission Intersect(IPermission target) { return default(IPermission); }
+        public override bool IsSubsetOf(IPermission target) { return false; }
+        public override SecurityElement ToXml() { return default(SecurityElement); }
+        public override IPermission Union(IPermission other) { return default(IPermission); }
+        public bool IsUnrestricted() { return false; }
```
vatsan-madhavan: Make this true

src/Microsoft.DotNet.Wpf/src/WindowsBase/System/Security/Permissions/MediaPermission.cs
```diff
@@ -1,687 +1,58 @@
...
-        private MediaPermissionImage _mediaPermissionImage ; 
-}
-
-
-    ///<summary> 
-    /// Imperative attribute to create a MediaPermission. 
-    ///</summary> 
+                               MediaPermissionVideo permissionVideo, 
+                               MediaPermissionImage permissionImage ) { }
+        public bool IsUnrestricted() { return false; }
```
vatsan-madhavan: return true

src/Microsoft.DotNet.Wpf/src/WindowsBase/System/Security/Permissions/WebBrowserPermission.cs
```diff
@@ -1,427 +1,36 @@
...
-                return;
-            }
-
-            this._webBrowserPermissionLevel = WebBrowserPermissionLevel.None;
-
-            String level = securityElement.Attribute("Level");
+        public WebBrowserPermission() { }       
+        public WebBrowserPermission(PermissionState state) { }
+        public WebBrowserPermission(WebBrowserPermissionLevel webBrowserPermissionLevel) { }
+        public bool IsUnrestricted() { return false;  }
```
vatsan-madhavan: return true

vatsan-madhavan:
@ericstj , We'd like to move these types to `System.Security.Permissions`.

ericstj:
> We'd like to move these types to System.Security.Permissions, and type-forward to it

Excellent, please open a PR.

> make these dangling-references from WindowsDesktop. App

What kind of dangling-references?  Dangling type-forwards: sure.  We do the same for our mscorlib.dll facade.  Dangling type-refs from attribute applications / actual usage: no-go : JIT would throw.

src/Microsoft.DotNet.Wpf/src/System.Xaml/System/Xaml/Permissions/XamlAccessLevel.cs
ericstj: Private constructor ensures that you don't get a default public parameter-less one.

src/Microsoft.DotNet.Wpf/src/System.Xaml/System/Xaml/Permissions/XamlLoadPermission.cs
ericstj: You may still want to back this with an actual empty readonly list so that you provide an instance as before instead of null.

vatsan-madhavan:
> What kind of dangling-references? Dangling type-forwards: sure. We do the same for our mscorlib.dll facade. Dangling type-refs from attribute applications / actual usage: no-go : JIT would throw.

Dangling type-forwards only. 

The goal is to eliminate all actual type-refs before RTM and end up with ref assemblies that don't require S.S.Permissions to build (module type-forwards). You'll start seeing a related PR's to get to this shortly (I'll cc you)

vatsan-madhavan:
@ojhad, once the changes are made and the test failure is figured out, let's merge the changes into WPF and _then_ move the dummied-out impl. to corefx `s.s.permissions`.

ericstj:
Be aware that we're trying to lock down all API changes in CoreFx by Preview7 which means this would need to make it into corefx in the next couple weeks.  For a move like this, I'd recommend adding types in CoreFx, letting them flow upstack, then swap from type-def to typeforward when the packages reach WPF.  It does mean that we'll have a couple SDK builds with the types in both places, but that only breaks apps trying to compile against them which we don't expect a lot of folks to need to do.

src/Microsoft.DotNet.Wpf/src/System.Xaml/System/Xaml/Permissions/XamlAccessLevel.cs
```diff
@@ -58,7 +58,7 @@ public static XamlAccessLevel PrivateAccessTo(string assemblyQualifiedTypeName)
             int nameBoundary = assemblyQualifiedTypeName.IndexOf(',');
             if (nameBoundary < 0)
             {
-                throw new ArgumentException(SR.Get(SRID.ExpectedQualifiedTypeName, assemblyQualifiedTypeName), nameof(assemblyQualifiedTypeName));
+                throw new ArgumentException("", nameof(assemblyQualifiedTypeName));
```
vatsan-madhavan: nit: consider string.Emtpy

src/Microsoft.DotNet.Wpf/src/System.Xaml/System/Xaml/Permissions/XamlLoadPermission.cs
```diff
@@ -16,6 +16,10 @@ public sealed class XamlLoadPermission : CodeAccessPermission, IUnrestrictedPerm
         public XamlLoadPermission(PermissionState state) { }
         public XamlLoadPermission(XamlAccessLevel allowedAccess) { }
         public XamlLoadPermission(IEnumerable<XamlAccessLevel> allowedAccess) { }
+        [ComVisible(false)]
+        public override bool Equals(object obj) { return false; }
```
weltkante: This is not a valid implementation because `x.Equals(x)` returns false. You should use `ReferenceEquals(this, obj)` or throw an exception if you don't want to support equality.

vatsan-madhavan: This impl. will get deleted from dotnet/wpf shortly, and we'll consume the one from https://github.com/dotnet/corefx/pull/38405. Let's make sure we get the corefx PR right.

PR #880, opened 2019-06-05
Enable APICompat in WPF Builds

### LINKED ISSUE #724: Enable APICompat in WPF Builds
Enable API Compat against net48 ref assemblies on WPF Builds. 

This will prevent future breaks like #515

rladuca:
Fixes #724 

See: https://github.com/rladuca/ApiCompatNotes/blob/master/ApiCompat%20Notes.txt

This sets up and enables API Compat against .NET 4.8 reference assemblies and also ensures compat between hand-crafted ref assemblies and their lib counterparts.


eng/WpfArcadeSdk/tools/ApiCompat.targets
```diff
@@ -0,0 +1,67 @@
...
+    <RunApiCompat Condition="'$(RunNetFrameworkApiCompat)'=='true' or '$(RunReferenceApiCompat)'=='true'">true</RunApiCompat>
+
+    <RunApiCompatForSrc>false</RunApiCompatForSrc>
+    <RunApiCompatForSrc Condition="'$(RunNetFrameworkApiCompat)'=='true'">true</RunApiCompatForSrc>
+
+    <RunMatchingRefApiCompat>false</RunMatchingRefApiCompat>
+    <RunMatchingRefApiCompat Condition="'$(RunReferenceApiCompat)'=='true'">true</RunMatchingRefApiCompat>
+  </PropertyGroup>
+
+  <PropertyGroup>
```
stevenbrix: Is this needed anymore?

stevenbrix: that does not at all look like what i thought i was commenting on. one sec

eng/WpfArcadeSdk/tools/ApiCompat.targets
```diff
@@ -0,0 +1,67 @@
...
+    <TaskFactory>RoslynCodeTaskFactory</TaskFactory>
```
stevenbrix: is this `TaskFactory` needed now that you don't have an inline task?

rladuca: Yeah I have another change already that will address it, I'll resolve when I push it.

rladuca:
Having some issues getting ref compat to run, working them out now.

eng/WpfArcadeSdk/tools/ApiCompat.targets
```diff
@@ -0,0 +1,55 @@
+<Project>
+  <ItemGroup>
+    <PackageReference Include="Microsoft.DotNet.ApiCompat" Version="$(MicrosoftDotNetApiCompatVersion)"/>
```
vatsan-madhavan: Can we condition this `PackageReference` on `"'$(RunNetFrameworkApiCompat)' == 'true'"` ?

eng/WpfArcadeSdk/tools/ApiCompat.targets
```diff
@@ -0,0 +1,55 @@
...
+  </ItemGroup>
+
+  <!-- Only enable API Compat when needed for some building assembly -->
+  <PropertyGroup Condition="'$(RunNetFrameworkApiCompat)'==''">
+    <RunNetFrameworkApiCompat>false</RunNetFrameworkApiCompat>
+    <RunNetFrameworkApiCompat Condition="'$(Net48CompatNeededProjects)'!='' 
+                                          and $(Net48CompatNeededProjects.Contains('$(MSBuildProjectName)'))">true</RunNetFrameworkApiCompat>
```
vatsan-madhavan: This kind of pattern usually fails when dealing with generated projects, for e.g., `System.Windows.Controls.Ribbon_dklsjfl_wpftmp.csproj`

eng/WpfArcadeSdk/tools/ApiCompat.targets
```diff
@@ -0,0 +1,55 @@
...
+  </PropertyGroup>
+
+  <PropertyGroup Condition="'$(RunReferenceApiCompat)'==''">
+    <RunReferenceApiCompat>false</RunReferenceApiCompat>
+
+    <RefBin>$(ArtifactsDir)bin\$(AssemblyName)\$(Configuration)\$(Platform)\$(AssemblyName).dll</RefBin>
```
vatsan-madhavan: Is there no other property that can give us the value needed for `RefBin`? This seems fragile. `$(OutputAssembly)`, or `$(IntermediateAssembly)`, perhaps?

eng/WpfArcadeSdk/tools/ApiCompat.targets
rladuca: Good call, I can add a scan for wpftmp as well.

eng/WpfArcadeSdk/tools/ApiCompat.targets
rladuca: Sure, makes sense.

eng/WpfArcadeSdk/tools/ApiCompat.targets
rladuca: I have to rework this whole section.  The issue is there is no good way to ensure that a hand-crafted ref assembly (or cycle-breaker) is being built after the implementation assembly.  This means the generic reverse API compat from Arcade isn't useful.

CoreFX has a separate project that does this and we may have to do that + dynamically project reference everything we need to build from that project, then call the APICompat task directly via batching.

Currently I think we should split that work off from the .NET 4.8 compat because it doesn't directly influence RC API compat and we can do it later on.  Thoughts?

rladuca:
Talked with @vatsan-madhavan and we're going to go ahead with a change to this to ensure that the right API Compat can be run against any binary combination.

rladuca:
We can now run API compat both against .NET 4.8 and lib vs hand-crafted ref.  Still don't have cycle-breakers working, but that's not much different from hand-crafted ref.

eng/WpfArcadeSdk/tools/ApiCompat.targets
```diff
@@ -57,14 +77,12 @@
...
+  <PropertyGroup Condition="'$(RunNetFrameworkApiCompat)' == 'true' or '$(RunRefApiCompat)'=='true'">
     <_apiCompatTargetSuffix>$(TargetGroup)</_apiCompatTargetSuffix>
     <_apiCompatTargetSuffix Condition="'$(_apiCompatTargetSuffix)' == ''">$(TargetFramework)</_apiCompatTargetSuffix>
 
-    <ApiCompatBaseline Condition="!Exists('$(ApiCompatBaseline)')">$(MSBuildProjectDirectory)\ApiCompatBaseline.$(_apiCompatTargetSuffix).txt</ApiCompatBaseline>
-    <ApiCompatBaseline Condition="!Exists('$(ApiCompatBaseline)')">$(MSBuildProjectDirectory)\ApiCompatBaseline.txt</ApiCompatBaseline>
-
-    <TargetsTriggeredByCompilation Condition="'$(RunWpfApiCompat)' == 'true'">$(TargetsTriggeredByCompilation);WpfValidateApiCompatForSrc</TargetsTriggeredByCompilation>
+    <TargetsTriggeredByCompilation Condition="'$(RunNetFrameworkApiCompat)' == 'true'">$(TargetsTriggeredByCompilation);WpfValidateApiCompatForNetFramework</TargetsTriggeredByCompilation>
+    <AfterLinkTargets Condition="'$(RunRefApiCompat)' == 'true'">$(AfterLibTargets);WpfValidateApiCompatForRef</AfterLinkTargets>
```
rladuca: Add comment on why we need AfterLinkTargets here instead of TargetsTiggeredByCompilation.

We probably should also condition these on ManagedCxx so that we use AfterLinkTargets when dealing with a lib project that is C++\CLI and TargetsTriggeredByCompilation when dealing with C#.  This will allow this to continue working if we pivot to ref assemblies generated by GenAPI.

eng/WpfArcadeSdk/tools/ApiCompat.targets
```diff
@@ -0,0 +1,225 @@
...
+      <!-- Work backwards to find the output for a hand-crafted ref assembly. -->
+      <ResolvedImplementationAssembly Include="$(BaseOutputPath)..\$(AssemblyName)-ref\**\$(AssemblyName).dll" />
+    </ItemGroup>
+  </Target>
+
+  <!-- 
+       The below is a modified version of Arcade's ApiCompat targets file.
+       https://github.com/dotnet/arcade/blob/master/src/Microsoft.DotNet.ApiCompat/build/Microsoft.DotNet.ApiCompat.targets
+       
+       We use a custom version to provide greater control over which assembly pairs are being examined.
```
stevenbrix: What is it about Arcade's implementation that doesn't work for us?

src/Microsoft.DotNet.Wpf/ApiCompat/Baselines/PresentationBuildTasks-Net48.baseline.txt
```diff
@@ -0,0 +1,3 @@
+Compat issues with assembly PresentationBuildTasks:
```
stevenbrix: so we shouldn't expect any further changes to these files between now and RTM, correct?

eng/WpfArcadeSdk/tools/ApiCompat.targets
```diff
@@ -0,0 +1,225 @@
...
+    </Exec>
+
+    <!--
+      To force incremental builds to show failures again we are invalidating
+      one compile input.
+    -->
+    <Touch Condition="'$(ApiCompatExitCode)' != '0'" Files="$(IntermediateOutputPath)$(_ApiCompatSemaphoreFile)" AlwaysCreate="true">
+      <Output TaskParameter="TouchedFiles" ItemName="FileWrites" />
+    </Touch>
+    <Error Condition="'$(ApiCompatExitCode)' != '0'" Text="ApiCompat failed for '$(TargetPath)'" />
```
stevenbrix: i think it would be really beneficial to a blurb in the developer-guide describing what this error means, and then tell the user to see the developer-guide for more information on what this error means

eng/WpfArcadeSdk/tools/ApiCompat.targets
rladuca: That's what the sentence describes, we wanted greater control over what files we pass in for both the contract and implementation.  That's pretty much it.

src/Microsoft.DotNet.Wpf/ApiCompat/Baselines/PresentationBuildTasks-Net48.baseline.txt
rladuca: These baselines are always updated manually (usually by building with a specific Arcade flag set).  We don't expect there to be any ApiCompat breaks that will require us to do that though.

eng/WpfArcadeSdk/tools/ApiCompat.targets
rladuca: Yeah I think I'll add some ApiCompat docs generally that point to Arcade and also describe what we did.  Should have done that from the start.

eng/WpfArcadeSdk/tools/ApiCompat.targets
rladuca: This will resolve as part of adding a doc for this.  I'll point to the doc and it will have a better discussion.

rladuca:
> See comments about documentation and improving the error message when api compat fails

I resolved your comments against the documentation.  Let me know if you have any specific comments relating to that.

PR #899, opened 2019-06-08
port 2 changes from .NET 4.8: crash in PropertyChangeEventManager and GDI handle leak

### LINKED ISSUE #896: WPF application crashes with System.InvalidOperationException: Collection was modified after the enumerator was instantiated. at System.Collections.Specialized.ListDictionary.NodeEnumerator.MoveNext()
(Port of bugfix in .NET 4.8)

The exception occurs as follows:

1. A property is changed
2. The weak-event system starts delivery of the PropertyChanged event
3. One of the handlers calls (indirectly) a method that
4. Pushes a dispatcher frame, starting a nested message pump
5. A cleanup task reaches the front of the dispatcher queue (before the frame stops pumping)
6. PropertyChangedEventManager.Purge iterates over its dictionary.  When it reaches the ListenerList corresponding to the property from (1), it's marked "in-use" (from (2)), so the list is cloned and the clone is inserted into the dictionary.
7. The next MoveNext() from (6) throws the exception.

### LINKED ISSUE #897: GDI handle leak when application is PerMonV2 aware
(Port of bugfix in .NET 4.8)

The root cause for this is a handle leak in WindowStartupTopLeftPointHelper.IdentifyScreenTopLeft when the application is PerMonV2 aware.

The screen HDC used in this function is never released.  This causes a GC when the window being created paints as DoPaint's HDC creation will always push the handle count in HandleCollector.HandleType above the threshold value.  So a new window, in most scenarios under PerMonV2, will cause a GC.

SamBent:
Fixes #896, fixes #897.

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Window.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Window.cs
@@ -399,7 +399,7 @@ namespace System.Windows
 
             Debug.Assert(_threadWindowHandles == null, "_threadWindowHandles must be null before enumerating the thread windows");
 
-            // NOTE: 
+            // NOTE:
             // _threadWindowHandles is created here.  This reference is nulled out in EnableThreadWindows
             // when it is called with a true parameter.  Please do not null it out anywhere else.
             // EnableThreadWindow(true) is called when dialog is going away.  Once dialog is closed and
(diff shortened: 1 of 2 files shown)


src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Window.cs
```diff
@@ -455,7 +455,7 @@ public Nullable<bool> ShowDialog()
             }
             catch
             {
-                // NOTE: 
+                // NOTE:
                 // See BUG 929740.
```
vatsan-madhavan: Please remove this bug ID.

PR #912, opened 2019-06-11
Fix a incorrect spelled word.

walterlv:
Actually, the correct word is `ReferencePath`.

```xml
<ReferencePath Include="xxx" />
```

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/PresentationBuildTasks/Microsoft/Build/Tasks/Windows/GenerateTemporaryTargetAssembly.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationBuildTasks/Microsoft/Build/Tasks/Windows/GenerateTemporaryTargetAssembly.cs
@@ -11,7 +11,7 @@
 //              to compile it.
 //              
 //              The new project file will replace all the Reference Items with the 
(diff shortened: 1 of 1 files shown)


stevenbrix:
@walterlv looks like something is up with our pipeline, although the setting that the error message is telling us to enable is set, so i'm hoping that a simple retry will fix things. There obviously isn't anything wrong with your change here.

stevenbrix:
it looks like other prs are succeeding, so i'm restarting these checks and hopefully this will succeed

vatsan-madhavan:
@stevenbrix I had to restart it a 3rd time. Perhaps open a core-eng issue?

stevenbrix:
@vatsan-madhavan yeah I'll do that in the morning

vatsan-madhavan:
> yeah I'll do that in the morning

Just so you know, _Make Secrets Available to Builds of Forks_ setting is turned on in our pipeline, so if  a suggestion is made that we turn this on in an attempt to solve this particular build failure, you know that's probably not the incantation we are looking for ;-) 

image

lindexi:
@walterlv You may need to merge master

vatsan-madhavan:
long overdue - closing and reopening the PR did the trick with AzDo.

PR #919, opened 2019-06-11
Fix incorrect localization causing crashes in RegisterCommandHandler and KeyGestures

### LINKED ISSUE #913: Culture de-DE issue with DataGrid
* .NET Core Version: 3.0.100-preview7-012343

* Windows version: Windows 10 1903 & Windows 7 (6.1 build 7601 Service Pack 1)

 **Problem description:**
 When the UI Culture is de-DE, any interaction (wheel, click) with a DataGrid will throw an exception (see below). This is related to https://github.com/dotnet/wpf/issues/684

 **Actual behavior:**
```
Unhandled exception. System.Reflection.TargetInvocationException: Exception has been thrown by the target of an invocation. ---> System.ArgumentException: Requested value 'BILDAUF' was not found.
   at System.Enum.TryParseByName(RuntimeType enumType, String originalValueString, ReadOnlySpan`1 value, Boolean ignoreCase, Boolean throwOnFailure, UInt64& result)
   at System.Enum.TryParseInt32Enum(RuntimeType enumType, String originalValueString, ReadOnlySpan`1 value, Int32 minInclusive, Int32 maxInclusive, Boolean ignoreCase, Boolean throwOnFailure, TypeCode type, Int32& result)
   at System.Enum.TryParse(Type enumType, String value, Boolean ignoreCase, Boolean throwOnFailure, Object& result)
   at System.Enum.Parse(Type enumType, String value, Boolean ignoreCase)
   at System.Windows.Input.KeyConverter.GetKey(String keyToken, CultureInfo culture)
   at System.Windows.Input.KeyConverter.ConvertFrom(ITypeDescriptorContext context, CultureInfo culture, Object source)
   at System.Windows.Input.KeyGestureConverter.ConvertFrom(ITypeDescriptorContext context, CultureInfo culture, Object source)
   at System.ComponentModel.TypeConverter.ConvertFromInvariantString(String text)
   at System.Windows.Input.KeyGesture.CreateFromResourceStrings(String keyGestureToken, String keyDisplayString)
   at System.Windows.Input.KeyGesture.AddGesturesFromResourceStrings(String keyGestures, String displayStrings, InputGestureCollection gestures)
   at System.Windows.Input.ComponentCommands.LoadDefaultGestureFromResource(Byte commandId)
   at System.Windows.Input.RoutedCommand.GetInputGestures()
   at System.Windows.Input.RoutedCommand.get_InputGesturesInternal()
   at System.Windows.Input.CommandBindingCollection.FindMatch(Object targetElement, InputEventArgs inputEventArgs)
   at System.Windows.Input.CommandManager.TranslateInput(IInputElement targetElement, InputEventArgs inputEventArgs)
   at System.Windows.UIElement.OnMouseWheelThunk(Object sender, MouseWheelEventArgs e)
   at System.Windows.Input.MouseWheelEventArgs.InvokeEventHandler(Delegate genericHandler, Object genericTarget)
   at System.Windows.RoutedEventArgs.InvokeHandler(Delegate handler, Object target)
   at System.Windows.RoutedEventHandlerInfo.InvokeHandler(Object target, RoutedEventArgs routedEventArgs)
   at System.Windows.RouteItem.InvokeHandler(RoutedEventArgs routedEventArgs)
   at System.Windows.EventRoute.InvokeHandlersImpl(Object source, RoutedEventArgs args, Boolean reRaised)
   at System.Windows.EventRoute.InvokeHandlers(Object source, RoutedEventArgs args)
   at System.Windows.UIElement.RaiseTrustedEvent(RoutedEventArgs args)
   at System.Windows.UIElement.RaiseEvent(RoutedEventArgs args, Boolean trusted)
   at System.Windows.Input.InputManager.ProcessStagingArea()
   at System.Windows.Input.InputManager.ProcessInput(InputEventArgs input)
   at System.Windows.Input.InputProviderSite.ReportInput(InputReport inputReport)
   at System.Windows.Interop.HwndMouseInputProvider.ReportInput(IntPtr hwnd, InputMode mode, Int32 timestamp, RawMouseActions actions, Int32 x, Int32 y, Int32 wheel)
   at System.Windows.Interop.HwndMouseInputProvider.FilterMessage(IntPtr hwnd, WindowMessage msg, IntPtr wParam, IntPtr lParam, Boolean& handled)
   at System.Windows.Interop.HwndSource.InputFilterMessage(IntPtr hwnd, Int32 msg, IntPtr wParam, IntPtr lParam, Boolean& handled)
   at System.Windows.Threading.Dispatcher.PushFrame(DispatcherFrame frame)
   at System.Windows.Window.Show()
   at System.Windows.Window.ShowDialog()
   at Caliburn.Micro.WindowManager.ShowDialog(Object rootModel, Object context, IDictionary`2 settings)
```

 **Expected behavior:**
 No exception

 **Minimal repro:**
 https://github.com/Lakritzator/WpfDataGridCultureBug
In fact this is just a "dotnet new wpf", making sure the Thread.CurrentThread.CurrentUICulture is set to the "de-DE" culture, and change the grid in MainWindow.xaml to DataGrid.
Run the example, and use the scrollwheel (or click) on the DataGrid.

I was going to supply a PR, but I'm confused! In https://github.com/dotnet/wpf/pull/796 the translations are reverted and seem to be okay.

But in the final file https://github.com/dotnet/wpf/blob/master/src/Microsoft.DotNet.Wpf/src/PresentationFramework/Resources/xlf/Strings.de.xlf#L2952 there are still translations which probably cause the issue.

And somehow I don't know where the changes come from...

Maybe @ryalanms knows what happened here.

ryalanms:
Reviewed key strings, updated strings.resx files, and updated xlf files.  

Confirmed updates fix NuGet Package Explorer and loc test application provided by @Lakritzator.  (Fixes #913.)


vatsan-madhavan:
Removing the "LOC CHECKIN" prefix from the title. This is not a conventional loc-checkin.

src/Microsoft.DotNet.Wpf/src/PresentationBuildTasks/Microsoft/Build/Tasks/Windows/FileClassifier.cs
```diff
@@ -82,7 +82,7 @@ public override bool Execute()
 
             try
             {
-                TaskHelper.DisplayLogo(Log, SR.Get(SRID.FileClassifierTask));
+                TaskHelper.DisplayLogo(Log, "FileClassifier");
```
vatsan-madhavan: Why are these (and similar) lookups changed from `SR.Get(SRID.SomeName)` to `"SomeName"` ?

src/Microsoft.DotNet.Wpf/src/PresentationBuildTasks/Resources/Strings.resx
```diff
@@ -132,30 +132,6 @@
   <data name="FileResolved" xml:space="preserve">
     <value>Input file '{0}' is resolved to new relative path '{1}' at directory '{2}'.</value>
   </data>
-  <data name="MarkupCompilePass1Task" xml:space="preserve">
```
vatsan-madhavan: Was there a hint in the resx xml previously that indicated that this item ("MarkupCompilePass1Task") should not have been localized?

vatsan-madhavan:
Can you summarize (in a table, perhaps?) what strings are being removed from resx files?

src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Controls/FlowDocumentReader.cs
```diff
@@ -35,6 +35,9 @@ namespace System.Windows.Controls
     [ContentProperty("Document")]
     public class FlowDocumentReader : Control, IAddChild, IJournalState
     {
+        internal const string KeySwitchViewingMode = "Ctrl+M";
```
vatsan-madhavan: why `internal` (as against `private`)?

rladuca:
It was hard to keep it all in my head to determine, so I'll ask, are there any duplicate strings being defined in multiple files that could be hoisted up into a common file?  If so, please do so.

rladuca:
Removed string resources could have been used by others in the past.  Are we worried about those no longer being accessible?

vatsan-madhavan:
> Removed string resources could have been used by others in the past. Are we worried about those no longer being accessible?

What's a concrete scenario that might break?

rladuca:
> > Removed string resources could have been used by others in the past. Are we worried about those no longer being accessible?
> 
> What's a concrete scenario that might break?

Someone could be looking directly at the string resources and either displaying them or parsing them out for their own needs.  I don't have a specific example for sure, but resources are publicly available at runtime and now these are not a part of that set.

vatsan-madhavan:
> Someone could be looking directly at the string resources and either displaying them or parsing them out for their own needs. 

Retaining the old strings seems like a costly (i.e, confusion and bug-prone) effort on our side. We'd end up with two sets of strings (resources and const strings) - which would require careful documentation to ensure that we only use the const-strings in code and never use the resource strings. Localization will continue happening (and we'd continue having to ensure that these strings remain _locked_, despite the fact that we don't them, nor are we sure anyone else uses them). 

On balance, I think we should just take this change now as a clean-break. Anyone depending on these resources will just have to adapt a bit.

rladuca:
> > Someone could be looking directly at the string resources and either displaying them or parsing them out for their own needs.
> 
> Retaining the old strings seems like a costly (i.e, confusion and bug-prone) effort on our side. We'd end up with two sets of strings (resources and const strings) - which would require careful documentation to ensure that we only use the const-strings in code and never use the resource strings. Localization will continue happening (and we'd continue having to ensure that these strings remain _locked_, despite the fact that we don't them, nor are we sure anyone else uses them).
> 
> On balance, I think we should just take this change now as a clean-break. Anyone depending on these resources will just have to adapt a bit.

That's fair enough, anyone accessing these should really be going through the various RoutedUICommands anyway.

src/Microsoft.DotNet.Wpf/src/PresentationBuildTasks/Microsoft/Build/Tasks/Windows/FileClassifier.cs
ryalanms: The generated SRIDs are string variables containing the string of the resource ID (e.g., string FileClassifierTask == "FileClassifierTask").  In this case, the actual resource lookup is the string value "FileClassifierTask" which resolves to the untranslated string "FileClassifier".  For the untranslated case with only a single instance, I've removed the SRID and the resource lookup and replaced it with a hardcoded string.

src/Microsoft.DotNet.Wpf/src/PresentationBuildTasks/Resources/Strings.resx
ryalanms: This is no way to represent that data in the resx file.

ryalanms:
## Strings moved from strings.resx files ##

| PresentationBuildTasks  |
| --------------------------- |
| FileClassifierTask |
| GetWinFXPathTask |
| MarkupCompilePass1Task |
| MarkupCompilePass2Task |
| MergeLocalizationDirectivesTask |
| ResourcesGeneratorTask |
| UidManagerTask |
| UpdateManifestForBrowserApplicationTask |

| PresentationFramework |
| --------------------------- |
| InkCanvasDeselectKey   |
| KeyAlignCenter |
| KeyAlignLeft |
| KeyAlignRight |
| KeyAltUndo |
| KeyApplyDoubleSpace |
| KeyApplyOneAndAHalfSpace |
| KeyApplySingleSpace |
| KeyBackspace |
| KeyCopy |
| KeyCopyFormat |
| KeyCtrlInsert |
| KeyCut |
| KeyDecreaseFontSize |
| KeyDecreaseIndentation |
| KeyDelete |
| KeyDeleteColumns |
| KeyDeleteNextWord |
| KeyDeletePreviousWord |
| KeyEnterLineBreak |
| KeyEnterParagraphBreak |
| KeyIncreaseFontSize |
| KeyInsertColumns |
| KeyInsertRows |
| KeyInsertTable |
| KeyMergeCells |
| KeyMoveDownByLine |
| KeyMoveDownByPage |
| KeyMoveDownByParagraph |
| KeyMoveLeftByCharacter |
| KeyMoveLeftByWord |
| KeyMoveRightByCharacter |
| KeyMoveRightByWord |
| KeyMoveToColumnEnd |
| KeyMoveToColumnStart |
| KeyMoveToDocumentEnd |
| KeyMoveToDocumentStart |
| KeyMoveToLineEnd |
| KeyMoveToLineStart |
| KeyMoveToWindowBottom |
| KeyMoveToWindowTop |
| KeyMoveUpByLine |
| KeyMoveUpByPage |
| KeyMoveUpByParagraph |
| KeyPasteFormat |
| KeyRedo |
| KeyRemoveListMarkers |
| KeyResetFormat |
| KeySelectDownByLine |
| KeySelectDownByPage |
| KeySelectDownByParagraph |
| KeySelectLeftByCharacter |
| KeySelectLeftByWord |
| KeySelectRightByCharacter |
| KeySelectRightByWord |
| KeySelectToColumnEnd |
| KeySelectToColumnStart |
| KeySelectToDocumentEnd |
| KeySelectToDocumentStart |
| KeySelectToLineEnd |
| KeySelectToLineStart |
| KeySelectToWindowBottom |
| KeySelectToWindowTop |
| KeySelectUpByLine |
| KeySelectUpByPage |
| KeySelectUpByParagraph |
| KeyShiftBackspace |
| KeyShiftDelete |
| KeyShiftInsert |
| KeyShiftInsert |
| KeyShiftSpace |
| KeySpace |
| KeySplitCell |
| KeySwitchViewingMode |
| KeyTabBackward |
| KeyTabForward |
| KeyToggleBullets |
| KeyToggleInsert |
| KeyToggleNumbering |
| KeyToggleSubscript |
| KeyToggleSuperscript |
| KeyToggleUnderline |
| KeyUndo |
| ListBoxSelectAllKey |
| Switch_ViewingMode |
| ZoomPercentageConverterStringFormat |

| PresentationCore |
| --------------------------- |
| BrowseBackKey |
| BrowseForwardKey |
| BrowseHomeKey |
| BrowseStopKey |
| ContextMenuKey |
| CopyKey |
| CutKey |
| DeleteKey |
| ExtendSelectionDownKey |
| ExtendSelectionLeftKey |
| ExtendSelectionRightKey |
| ExtendSelectionUpKey |
| FindKey |
| HelpKey |
| KeyCopy |
| KeyCopyFormat |
| KeyCtrlInsert |
| KeyShiftDelete |
| KeyShiftDelete |
| KeyShiftInsert |
| MoveDownKey |
| MoveFocusBackKey |
| MoveFocusDownKey |
| MoveFocusForwardKey |
| MoveFocusPageDownKey |
| MoveFocusPageUpKey |
| MoveFocusUpKey |
| MoveLeftKey |
| MoveRightKey |
| MoveToEndKey |
| MoveToHomeKey |
| MoveToPageDownKey |
| MoveToPageUpKey |
| MoveUpKey |
| NewKey |
| OpenKey |
| PasteKey |
| PrintKey |
| PrintPreviewKey |
| PropertiesKey |
| RedoKey |
| ReplaceKey |
| SaveKey |
| ScrollPageDownKey |
| ScrollPageUpKey |
| SelectAllKey |
| SelectToEndKey |
| SelectToHomeKey |
| SelectToPageDownKey |
| SelectToPageUpKey |
| StopKey |
| UndoKey |

src/Microsoft.DotNet.Wpf/src/PresentationBuildTasks/Microsoft/Build/Tasks/Windows/UidManager.cs
```diff
@@ -74,7 +74,7 @@ public UidManager() : base(SR.SharedResourceManager)
         /// </summary>
         public override bool Execute()
         {
-            TaskHelper.DisplayLogo(Log, SR.Get(SRID.UidManagerTask));
+            TaskHelper.DisplayLogo(Log, "UidManagerTask");
```
rladuca: The string value here was just "UidManager".

src/Microsoft.DotNet.Wpf/src/PresentationBuildTasks/Microsoft/Build/Tasks/Windows/FileClassifier.cs
vatsan-madhavan: If `"FileClassifier"` continues to reside in `strings.resx`, we should continue using `SR.Get()`. 
If `"FileClassifier"` has been converted into a `const string`, then just use the corresponding variable. 

There is no need to create a shortcut here even for the scenarios that you've described - it is likely to confuse future maintainers of this code (_The generated SRIDs are string variables containing the string of the resource ID (e.g., string FileClassifierTask == "FileClassifierTask")_

The real problem is that we continue to have string resources that are not useful (like `FileClassifierTask`). That should be refactored in a fundamentally different manner (perhaps open an Issue to track it).

src/Microsoft.DotNet.Wpf/src/PresentationBuildTasks/Microsoft/Build/Tasks/Windows/UidManager.cs
vatsan-madhavan: I don't understand this resolution.

src/Microsoft.DotNet.Wpf/src/PresentationBuildTasks/Microsoft/Build/Tasks/Windows/FileClassifier.cs
vatsan-madhavan: These can be removed from strings.resx, added to `const string` list, and accessed as before (i.e., `SRID.FileClassifierTask` is replaced by ` const string FileClassifierTask`)

vatsan-madhavan:
In your tables, I can't tell whether the list consists of strings, or resource names. It would be best if the table were structured as a {resource name, string value} pair.

ryalanms:
Removing a string from .resx removes that string's corresponding generated SRID.  The table contains resource IDs, which were used as const string names where possible.

src/Microsoft.DotNet.Wpf/src/PresentationBuildTasks/Microsoft/Build/Tasks/Windows/UidManager.cs
rladuca: Unless you were pre-resolving, this isn't correct still.

rladuca: My bad, I was in some sort of half refreshed state, this is good now.

src/Microsoft.DotNet.Wpf/src/PresentationBuildTasks/Microsoft/Build/Tasks/Windows/FileClassifier.cs
```diff
@@ -82,7 +82,7 @@ public override bool Execute()
 
             try
             {
-                TaskHelper.DisplayLogo(Log, SR.Get(SRID.FileClassifierTask));
+                TaskHelper.DisplayLogo(Log, FileClassifierTaskString);
```
vatsan-madhavan: Based on our conversation, you are going to change these to `nameof(FileClassifier)` etc., right?

PR #934, opened 2019-06-12
Bindings to System.Dynamic types fail to resolve

### LINKED ISSUE #695: Binding to ExpandoObject Properties throws Binding Exception
* .NET Core Version: 3.0.100-preview5-011568, Commit: b487ff10aa
 OS Name:     Windows
 OS Version:  10.0.15063
 OS Platform: Windows
 RID:         win10-x64
 Base Path:   C:\Program Files\dotnet\sdk\3.0.100-preview5-011568\

Host (useful for support):
  Version: 3.0.0-preview5-27626-15
  Commit:  61f30f5a23

* Windows version: Windows 10 Version 1703 (Build: 15063.1746)
* Visual Studio version: 16.1.0 Preview 3
* Does the bug reproduce also in WPF for .NET Framework 4.8?: No, Binding works as expected
 
 **Problem description:**
Binding an ExpandoObject in WPF Control (ex. Textbox.Text) throws BindingExpression Error in .NET Core WPF, but Binding works in .NET Framework (4.8 and 4.7.2)
 
 **Actual behavior:**
Throwing BindingExpression Error
System.Windows.Data Error: 40
BindingExpression path error: 'Test' property not found on 'object' ''ExpandoObject' (HashCode=58204539)'. BindingExpression:Path=TestProperty.Test; DataItem='MainViewModel' (HashCode=16745860); target element is 'TextBlock' (Name=''); target property is 'Text' (type 'String')

 **Expected behavior:**
 Binding Property as in .NET Framework

 **Minimal repro:**
```csharp
using System.ComponentModel;
using System.Dynamic;

namespace WpfApp1.ViewModels
{
	public class MainViewModel : INotifyPropertyChanged
	{
		public MainViewModel()
		{
			TestProperty = new ExpandoObject();
			TestProperty.Test = "abc";
			TestProperty.ABC = "xyz";
		}

		private dynamic TestPropertyValue = default;
		public dynamic TestProperty { get { return TestPropertyValue; } set { TestPropertyValue = value; NotifyPropertyChanged(nameof(TestProperty)); } }

		private void NotifyPropertyChanged([CallerMemberName] string propertyName = "")
		{
			PropertyChanged?.Invoke(this, new PropertyChangedEventArgs(propertyName));
		}
	}
}
```
```xaml
<Window x:Class="WpfApp1.MainWindow" xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"	xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"	xmlns:d="http://schemas.microsoft.com/expression/blend/2008"
				xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006" xmlns:viewmodels="clr-namespace:WpfApp1.ViewModels" 	mc:Ignorable="d" DataContext="{DynamicResource ResourceKey=MainVM}">
	<Window.Resources>
		<viewmodels:MainViewModel x:Key="MainVM" />
	</Window.Resources>
	<StackPanel Orientation="Vertical">
		<TextBlock Text="{Binding Path=TestProperty.Test}"/>
		<TextBlock Text="{Binding Path=TestProperty.ABC}"/>
	</StackPanel>
</Window>
```

ryalanms:
`MS.Internal.SystemCoreHelper.IsIDynamicMetaObjectProvider` was returning false for `System.Dynamic.ExpandoObject` causing the PropertyPath resolution to fail. 

On .NET Core, System.Core is now a facade assembly that will never be loaded by the CLR.  **All System.Core types are now forwarded to other assemblies.**  This change updates the required assemblies to those containing the forwarded types (System.Runtime, loaded by default, and System.Linq.Expressions).  AssemblyHelper.cs will now check for System.Linq.Expressions instead of System.Core before loading the extension assembly. 

 Fixes #695.  A larger architectural fix for the extension assemblies may happen after RTM.

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/WindowsBase/MS/Internal/AssemblyHelper.cs
+++ b/src/Microsoft.DotNet.Wpf/src/WindowsBase/MS/Internal/AssemblyHelper.cs
@@ -42,7 +42,7 @@ namespace MS.Internal
         System_Xml_Linq,
         System_Private_Xml_Linq,
         System_Data,
-        System_Core,
+        System_Linq_Expressions,
     }
 
     [FriendAccessAllowed]
@@ -180,8 +180,13 @@ namespace MS.Internal
         // load the extension class for System.Core
         internal static SystemCoreExtensionMethods ExtensionsForSystemCore(bool force=false)
         {
+            // System.Core is always loaded by default on .NET Framework. On .NET Core, 
+            // System.Core is a facade assembly never loaded by the CLR, and System.Core types 
+            // have been forwarded to System.Runtime (always loaded by default) and 
+            // System.Linq.Expressions.  Only load the extension assembly if System.Linq.Expressions 
+            // has already been loaded (e.g., a System.Dynamic type has been created).  
             if (_systemCoreExtensionMethods == null &&
-                (force || IsLoaded(UncommonAssembly.System_Core)))
+                (force || IsLoaded(UncommonAssembly.System_Linq_Expressions)))
             {
                 _systemCoreExtensionMethods = (SystemCoreExtensionMethods)LoadExtensionFor("SystemCore");
             }


rladuca:
The last build here succeeded, this looks like another instance of https://github.com/dotnet/core-eng/issues/6416.

There is no reason not to merge this.

Issue #935, opened 2019-06-13
Throw IndexOutOfRangeException in WispLogic.CoalesceAndQueueStylusEvent


lindexi:
* .NET Core Version: 3.0.100-preview5-011568
* Windows version: 1903 18362.145
* Does the bug reproduce also in WPF for .NET Framework 4.8?: Find in 4.7

 
 
 **Problem description:**

The application crash.

We find many customs will break the application when he touch the application. I catch the `AppDomain.CurrentDomain.UnhandledException` and find the WispLogic.CoalesceAndQueueStylusEvent may be break the application. 

See: https://stackoverflow.com/questions/46049673/wpf-touch-application-partially-freezes-on-net-framework-4-7

And [Bruno V](https://stackoverflow.com/users/3357566/bruno-v) tell me that it fix in .NET 4.7.1, but it seems not fix in dotnet core.

I can not debug it, because 200 users found the crash among 300,000 users.
 
 **Actual behavior:** 

exceptions  

ExceptionType: System.IndexOutOfRangeException

callstack for crashes

```
   at System.Collections.Generic.Dictionary`2.Insert(TKey key, TValue value, Boolean add)
   at System.Windows.Input.StylusWisp.WispLogic.CoalesceAndQueueStylusEvent(RawStylusInputReport inputReport)
   at System.Windows.Input.StylusWisp.WispLogic.ProcessInputReport(RawStylusInputReport inputReport)
   at System.Windows.Input.PenContext.FirePackets(Int32 stylusPointerId, Int32[] data, Int32 timestamp)
   at System.Windows.Input.PenThreadWorker.FlushCache(Boolean goingOutOfRange)
   at System.Windows.Input.PenThreadWorker.FireEvent(PenContext penContext, Int32 evt, Int32 stylusPointerId, Int32 cPackets, Int32 cbPacket, IntPtr pPackets)
   at System.Windows.Input.PenThreadWorker.ThreadProc()
   at System.Threading.ThreadHelper.ThreadStart_Context(Object state)
   at System.Threading.ExecutionContext.RunInternal(ExecutionContext executionContext, ContextCallback callback, Object state, Boolean preserveSyncCtx)
   at System.Threading.ExecutionContext.Run(ExecutionContext executionContext, ContextCallback callback, Object state, Boolean preserveSyncCtx)
   at System.Threading.ExecutionContext.Run(ExecutionContext executionContext, ContextCallback callback, Object state)
[...]

PR #936, opened 2019-06-13
Catch all the exception in touch process This MR do not expect to merge

lindexi:
Because we can not use any way to repair it when the code throw the exception and the Stylus Input thread will exit

Maybe it can fix https://github.com/dotnet/wpf/issues/935

It is an evil code

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Input/Stylus/Wisp/PenThreadWorker.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Input/Stylus/Wisp/PenThreadWorker.cs
@@ -1477,45 +1477,55 @@ namespace System.Windows.Input
                                 break;
                             }
                         }
-                        if (evt != PenEventTimeout)
+
+                        try
                         {
-                            // dispatch the event
+                            if (evt != PenEventTimeout)
+                            {
+                                // dispatch the event
 #if TRACEPTW
-                            Debug.WriteLine (String.Format("PenThreadWorker::ThreadProc - FireEvent [evt={0}, stylusId={1}]", evt, stylusPointerId));
+                                Debug.WriteLine (String.Format("PenThreadWorker::ThreadProc - FireEvent [evt={0}, stylusId={1}]", evt, stylusPointerId));
 #endif
                             
-                            // This comment addresses and IndexOutOfRangeException in PenThreadWorker which is related and likely caused by the above.
-                            // This index is safe as long as there are no corruption issues within PenIMC.  There have been
-                            // instances of IndexOutOfRangeExceptions from this code but this should not occur in practice.
-                            // If this throws, check that the handles list generated in CPimcContext::GetPenEventMultiple
-                            // is not corrupted (it has appropriate wait handles and does not point to invalid memory).
-                            PenContext penContext = _penContexts[iHandleEvt].Target as PenContext;
-                            // If we get an event from a GC'd PenContext then just ignore.
-                            if (penContext != null)
-                            {
-                                FireEvent(penContext, evt, stylusPointerId, cPackets, cbPacket, pPackets);
-                                penContext = null;
-                            }
-                        }
-                        else
-                        {
-#if TRACEPTW
-                            Debug.WriteLine (String.Format("PenThreadWorker::ThreadProc - FlushInput"));
-#endif
-                            FlushCache(true);
-
-                            // we hit the timeout, make sure that all our devices are in the correct out-of-range state
-                            // we are doing this to compinsate for drivers that send a move after they send a outofrange
-                            for (int i = 0; i < _penContexts.Length; i++)
-                            {
-                                PenContext penContext = _penContexts[i].Target as PenContext;
+                                // This comment addresses and IndexOutOfRangeException in PenThreadWorker which is related and likely caused by the above.
+                                // This index is safe as long as there are no corruption issues within PenIMC.  There have been
+                                // instances of IndexOutOfRangeExceptions from this code but this should not occur in practice.
+                                // If this throws, check that the handles list generated in CPimcContext::GetPenEventMultiple
+                                // is not corrupted (it has appropriate wait handles and does not point to invalid memory).
+                                PenContext penContext = _penContexts[iHandleEvt].Target as PenContext;
+                                // If we get an event from a GC'd PenContext then just ignore.
                                 if (penContext != null)
                                 {
-                                    // we send 0 as the stulyspointerId to trigger code in PenContext::FirePenOutOfRange
-                                    penContext.FirePenOutOfRange(0, Environment.TickCount);
+                                    FireEvent(penContext, evt, stylusPointerId, cPackets, cbPacket, pPackets);
                                     penContext = null;
                                 }
                             }
+                            else
+                            {
+#if TRACEPTW
+                                Debug.WriteLine (String.Format("PenThreadWorker::ThreadProc - FlushInput"));
+#endif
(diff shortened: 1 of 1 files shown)


weltkante:
I'd prefer bugs being fixed instead of being ignored. If you ignore the exceptions they will never get fixed. It just leads to problems of the type "my touch input stops responding" because the thread terminated and can't recover.

If the WPF team really wants a catch-all handler it should have associated diagnostics and not just ignore the exception.

Personally I have my own top level exception handlers on AppDomain etc. to report (and potentially ignore) unhandled exceptions. I believe the option to ignore exceptions should be left to the application programmer, not be made at the framework level.

src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Input/Stylus/Wisp/PenThreadWorker.cs
```diff
@@ -1477,46 +1477,56 @@ internal void ThreadProc()
...
+                                        // we send 0 as the stulyspointerId to trigger code in PenContext::FirePenOutOfRange
+                                        penContext.FirePenOutOfRange(0, Environment.TickCount);
+                                        penContext = null;
+                                    }
                                 }
                             }
                         }
+                        catch (Exception)
+                        {
+                            // If the code above throw an exception that the Stylus Input thread will be break. And we can not use any way to repair it.
```
walterlv: Some grammar issues.

```csharp
// If the code above throws an exception the Stylus Input thread will break and there is nothing we can do to repair it.
```

src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Input/Stylus/Wisp/PenThreadWorker.cs
```diff
@@ -1477,46 +1477,56 @@ internal void ThreadProc()
...
+                            // Maybe only some touch device be crash but other touch device can run still.
```
walterlv: Only some of the touch devices may crash and the others can still run normally.

src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Input/Stylus/Wisp/PenThreadWorker.cs
```diff
@@ -1477,46 +1477,56 @@ internal void ThreadProc()
...
+                            // To catch all the exception to make the application can run when the code throw an exception.
```
walterlv: We catch all the exceptions so that the application can still be alive.

lindexi:
@weltkante I think the better way is to offer some way to restart the Stylus Input thread or some way to reset the status

walterlv:
@weltkante You mean that we'd better unload the touch module and then initialize it again during the `catch` statement?

We have some other similar situations that reinitializing can fix stylus issues such as these below:

- [c# - WPF applications stop responding to touches after adding or removing tablet devices - Stack Overflow](https://stackoverflow.com/q/51854593/6233938)
- [c# - WPF touch application (partially) freezes on .NET Framework 4.7 - Stack Overflow](https://stackoverflow.com/q/46049673/6233938)

lindexi:
> It just leads to problems of the type "my touch input stops responding"

@weltkante The user can only fix it by reboot the device when the fullscreen application running in the device which only supports touch input. The application stops responding to touch input is an important crash for the touch application. 

But I must admit that this is not a good way and it is an evil code. We should not ignore the exception.

weltkante:
In my opinion WPF needs to deal gracefully with hardware devices being (un)plugged or becoming unusable. If it can't do that I prefer an application crash (usually surfaced via unhandled exceptions) over stopping to respond, so I can bring up a diagnostic interface to report the problem to our support. Our customers usually have support contracts and if they have faulty devices preventing proper usage of our software we may be able to root cause the problem this way. If the input just stops working this is much harder to diagnose on our side and the customer isn't any more happy either.

I undestand not everyone wants that, thats why I brought up the option of configurable top level exception handling, like WinForms allows in [Application.SetUnhandledExceptionMode](https://docs.microsoft.com/de-de/dotnet/api/system.windows.forms.application.setunhandledexceptionmode?view=netframework-4.8). I don't know if WPF already has a similar setting. If you want to ignore exceptions you can't recover from, you'll have to give the application the chance to abort (or restart) the whole program.

> You mean that we'd better unload the touch module and then initialize it again during the catch statement?

If this is guaranteed to fix the problem you can do that. But what if the device is broken and you end up an infinite loop of ignoring exceptions? I don't want to burn a CPU core at 100% just repeatedly ignoring the same exception.

lindexi:
This MR may fix some friend throw exception in StylusPlugIn.

    {
        public MainWindow()
        {
            InitializeComponent();
            StylusPlugIns.Add(new Foo());
        }
    }

    public class Foo : StylusPlugIn
    {
        /// <inheritdoc />
        {
            throw new Exception();
            base.OnStylusDown(rawStylusInput);
        }
    }

The OnStylusDown running in Stylus Input thread and if some friend throws any exceptions that will break the thread. And the application will stop responding touch.

But I do not think we can throw any unintended exceptions in StylusPlugIn.

See [StylusPlugIn Class (System.Windows.Input.StylusPlugIns)](https://docs.microsoft.com/en-us/dotnet/api/system.windows.input.stylusplugins.stylusplugin?view=netframework-4.8 )

> If you use a StylusPlugIn inside a control, you should test the plug-in and control extensively to make sure they do not throw any unintended exceptions.

See [Intercepting Input from the Stylus](https://msdn.microsoft.com/en-us/data/ms749105(v=vs.80) )

> If a StylusPlugIn throws or causes an exception, the application will close. You should thoroughly test controls that consume a StylusPlugIn and only use a control if you are certain the StylusPlugIn will not throw an exception.

stevenbrix:
@weltkante @walterlv @lindexi thank you for the great discussion on this PR!

My personal feeling on this issue, and I think there is consensus on this thread, is that catching all exceptions and ignoring them is not the right fix for this issue.

As for something analogous to the `Application.SetUnhandledExceptionMode` in WinForms, does using `Application.Current.DispatcherUnhandledException` solve your issue, or is there something fundamentally lacking?

lindexi:
Thank you @stevenbrix . The core issue is how to fix the touch break when some core throw an exception in the stylus plugin method. Maybe it is the WPF StylusPlugIn API lack a little design. But I do not know how to fix the design.

It will break the stylus input thread when we do not capture the StylusPlugIn exception, so I do not think we can use `Application.Current.DispatcherUnhandledExceptio` can solve it.

Do you mean to use ExceptionDispatchInfo to capture the exception and rethrow? See [Catch and rethrow the exception in StylusPlugIn by lindexi · Pull Request #945 · dotnet/wpf](https://github.com/dotnet/wpf/pull/945/files )

stevenbrix:
> It will break the stylus input thread when we do not capture the StylusPlugIn exception, so I do not think we can use `Application.Current.DispatcherUnhandledException` can solve it.

Makes sense since that only catches exceptions on the UI thread. 

> Do you mean to use ExceptionDispatchInfo to capture the exception and rethrow? See [Catch and rethrow the exception in StylusPlugIn by lindexi · Pull Request #945 · dotnet/wpf](https://github.com/dotnet/wpf/pull/945/files)

@lindexi, that is definitely a better overall solution. Let's continue the discussion on that PR. Do you mind if close out this PR?

stevenbrix:
@lindexi I'm just going to close this so it doesn't linger. Discussion has hopped over to the original issue you opened #1037

PR #945, opened 2019-06-13
Catch and rethrow the exception in StylusPlugIn

lindexi:
This code running in Stylus Input thread and this thread is a backgroud thread that will be break when anyone throw an exception in this code

This MR may fix the problem that some friend throws an exception in StylusPlugIn.

Just like this code. The code will break the Stylus Input thread.

    {
        public MainWindow()
        {
            InitializeComponent();
            StylusPlugIns.Add(new Foo());
        }
    }

    public class Foo : StylusPlugIn
    {
        /// <inheritdoc />
        {
            throw new Exception();
        }
    }

The OnStylusDown running in Stylus Input thread and if some friend throws any exceptions that will break the thread. And the application will stop responding touch. Of course, the actual situation may be more complicated. Maybe some of my friends just didn't realize that the code he wrote was unstable. But add `try catch` to all the method is an evil code.

But I do not think we can throw any unintended exceptions in StylusPlugIn.

See [StylusPlugIn Class (System.Windows.Input.StylusPlugIns)](https://docs.microsoft.com/en-us/dotnet/api/system.windows.input.stylusplugins.stylusplugin?view=netframework-4.8 )

> If you use a StylusPlugIn inside a control, you should test the plug-in and control extensively to make sure they do not throw any unintended exceptions.

See [Intercepting Input from the Stylus](https://msdn.microsoft.com/en-us/data/ms749105(v=vs.80) )

> If a StylusPlugIn throws or causes an exception, the application will close. You should thoroughly test controls that consume a StylusPlugIn and only use a control if you are certain the StylusPlugIn will not throw an exception.

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Input/Stylus/Common/StylusPlugin.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Input/Stylus/Common/StylusPlugin.cs
@@ -92,17 +92,34 @@ namespace System.Windows.Input.StylusPlugIns
             // Only fire if plugin is enabled and hooked up to plugincollection.
             if (__enabled && _pic != null)
             {
-                switch (rawStylusInput.Report.Actions)
+                try
                 {
-                    case RawStylusActions.Down:
-                        OnStylusDown(rawStylusInput);
-                        break;
-                    case RawStylusActions.Move:
-                        OnStylusMove(rawStylusInput);
-                        break;
-                    case RawStylusActions.Up:
-                        OnStylusUp(rawStylusInput);
-                        break;
+                    switch (rawStylusInput.Report.Actions)
+                    {
+                        case RawStylusActions.Down:
+                            OnStylusDown(rawStylusInput);
+                            break;
+                        case RawStylusActions.Move:
+                            OnStylusMove(rawStylusInput);
+                            break;
+                        case RawStylusActions.Up:
+                            OnStylusUp(rawStylusInput);
+                            break;
+                    }
+                }
+                catch (Exception e)
+                {
+                    // this code running in Stylus Input thread
+                    // and the developer will re-write the method with their code
+                    // If they throw any exceptions careless, that the thread will break and the application crash and the application will stop responding touch
+                    // But we should not ignore the exception, and we can re-throw it in the main thread
+                    // Why we should use InvokeAsync to replace Invoke? Because maybe some friend will make the main thread wait for the code above
+                    _pic.Element.Dispatcher.InvokeAsync(() =>
+                    {
+                        System.Runtime.ExceptionServices.ExceptionDispatchInfo.Capture(e).Throw();
+                    },
+                        // Why we should use `Send` priority? Maybe the main thread is busy that the developer can not find the exception timely
+                        DispatcherPriority.Send);
                 }
             }
         }


stevenbrix:
@lindexi is there an issue related to this? i'm trying to understand what this change is doing. From my limited understanding, this sounds like it can break apps that currently depend on how component is handling exceptions.

src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Input/Stylus/Common/StylusPlugin.cs
```diff
@@ -92,17 +92,34 @@ internal void RawStylusInput(RawStylusInput rawStylusInput)
...
+                            OnStylusMove(rawStylusInput);
+                            break;
+                        case RawStylusActions.Up:
+                            OnStylusUp(rawStylusInput);
+                            break;
+                    }
+                }
+                catch (Exception e)
+                {
+                    // this code running in Stylus Input thread
```
stevenbrix: i think this can be worded a little more clearly:


>This code is running on the Stylus Input thread and has the chance to call out into app code. If the app code throws an exception that is not caught, then the app will crash. We don't want to ignore all exceptions, so we catch the exception and dispatch it to the main thread, allowing developers to handle it inside the handler for the `Application.DispatcherUnhandledException` event.

Does that sound right?

stevenbrix:
It also sounds like if we make this change in post 3.0 timeframe that the documentation will need to be updated, correct?

src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Input/Stylus/Common/StylusPlugin.cs
lindexi: Thanks @stevenbrix . I will change it.

lindexi:
> @lindexi is there an issue related to this? i'm trying to understand what this change is doing. From my limited understanding, this sounds like it can break apps that currently depend on how component is handling exceptions.

Thanks @stevenbrix . And I will new a issus about it.

lindexi:
> It also sounds like if we make this change in post 3.0 timeframe that the documentation will need to be updated, correct?

Yes, we should change the [StylusPlugIn Class (System.Windows.Input.StylusPlugIns)](https://docs.microsoft.com/en-us/dotnet/api/system.windows.input.stylusplugins.stylusplugin?view=netframework-4.8 ) document and remove this word `If you use a StylusPlugIn inside a control, you should test the plug-in and control extensively to make sure they do not throw any unintended exceptions.` , because now we can still run when the app code throws an exception in StylusPlugIn method

stevenbrix:
> because now we can still run when the app code throws an exception in StylusPlugIn method

Well you still would need to handle the exception in your `UnhandledException` handler, right? This won't magically make things resilient to the `StylusPlugin` throwing exceptions will it?

stevenbrix:
@lindexi, this is a better proposed solution than #936 

Can you provide some screenshots of what the caught exception looks like? Does it contain the inner exception with the original callstack?

stevenbrix:
I'm also curious to know if this is something that other community members think is a good idea.  It seems reasonable to me, but I'm wondering if there are other, more full featured solutions that could possibly cover other scenarios as well.

On #936, @weltkante brought up how WinForms has `Application.SetUnhandledExceptionMode`. Would something like that apply here?

I'm certainly not trying to discount this effort done here, just want to make sure we are thorough and design the right thing :)

/cc @dotMorten

dotMorten:
Thanks for the mention. If I understand this issue correctly, I'm concerned there's even an exception being thrown here, whether it's in the background or moved to the main-thread. IMHO the problem is still there, because I can't handle the unhandled exception (unless I put in an unhandled exception handler which IMHO is nothing but a band-aid). It's basically just an exception waiting for kill your app regardless.

Instead, I think first the proper fix would be to ensure this code doesn't throw, so there's nothing to catch. I'm hoping that's possible. In the event that there isn't, it means it's an expected exception, and it should IMHO just be dealth with the right way and swallowed. At no point should an exception be thrown that the user can't directly guard against in a try/catch around a method call (since it's an event, there is nothing to try/catch).

Looking at the example in the linked issue, IMHO the proper fix here is to fix the stylus plug-in. That code shouldn't throw, but just catching and swallowing it would mean the developer would never know the plugin was bad.

stevenbrix:
> IMHO the problem is still there, because I can't handle the unhandled exception (unless I put in an unhandled exception handler which IMHO is nothing but a band-aid). It's basically just an exception waiting to kill your app regardless.

Yeah this is sort of my thought, but I don't think you can always catch all `StylusPlugIn` exceptions. I'm sure there are scenarios where it's fundamentally not possible if you are using a 3rd party library and maybe they contain a StylusPlugin and do some stuff with it. And in your app you never use it directly? You could test as much as you want, but a component could always  throw an exception in some case you aren't aware of. 

> I'm hoping that's possible. In the event that there isn't, it means it's an expected exception, and it should IMHO just be dealth with the right way and swallowed. At no point should an exception be thrown that the user can't directly guard against in a try/catch around a method call (since it's an event, there is nothing to try/catch).

If it's in the `UnhandledException` handler, you can set it to handled so that the app doesn't crash, correct?

> Looking at the example in the linked issue, IMHO the proper fix here is to fix the stylus plug-in. That code shouldn't throw, but just catching and swallowing it would mean the developer would never know the plugin was bad.

This is a very trivial case, it would be great to see a more complicated scenario where this makes more sense (if one exists).


I also want to think of what the end-user experience is here. Do we want apps to crash just because a stylus plugin threw an exception? What if I'm writing a document with a pen and oops, the stylus stopped working, ok i'll just type. I'd prefer every WPF app be robust by default in this scenario, rather than leave this responsibility to the app developer. When apps crash like that users think it's an issue with Windows, and they are partially correct because we should be preventing scenarios like that if it's possible the app can continue functioning in a usable way. Let the user manually close the app if something becomes so broken.  But I also don't know what swallowing exceptions here does, is it possible to recover the Stylus and get it to continue working?

rladuca:
If an exception in a StylusPlugin didn't kill the Stylus thread the rest of the application would, generally, function well.  There could be some erroneous behavior that is application specific, but not framework specific.  For instance, if the application was using a plugin to limit stylus input to a specific area, then a failure in a plugin could then expand the input area unintentionally.

This sort of application specific consequence can be handled if we maintain the ability to get the errors that are arising and react to them.  Why not provide a (ThreadStatic) event in StylusPlugin for error trapping and let the application developer decide what to do with them?  If the error isn't marked as handled after the event handlers are called, allow the exception to crash the Stylus thread (for compat reasons).

It feels like we should be having this sort of design discussion in the issue, and not in a PR.  I suggest continued discussion move over to there.

stevenbrix:
> It feels like we should be having this sort of design discussion in the issue, and not in a PR. I suggest continued discussion move over to there.

Fair point, I'll continue the discussion there

PR #946, opened 2019-06-13
Ensure the identity name is the same for C# and VB templates

jmarolf:
Having mismatched identities means that shim templates cannot be re-used across languages in VS


jmarolf:
fixes [AB#902637](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/902637)

PR #949, opened 2019-06-13
Include GAC in changed ref assembly check during WPF build on .NET Framework

ryalanms:
**PresentationBuildTasks** is the only multi-targeted WPF component.  Both .NET Core and .NET Framework PBT assemblies are produced by the WPF build.

PresentationBuildTasks contains Global Assembly Cache code that was #ifdef'd out that should be included when compiling PresentationBuildTasks for .NET Framework.  **The Global Assembly Cache does not exist in .NET Core.**


src/Microsoft.DotNet.Wpf/src/PresentationBuildTasks/MS/Internal/Tasks/TaskHelper.cs
```diff
@@ -212,17 +212,17 @@ internal static bool IsValidCultureName(string name)
             return bValid;
         }
 
-#if false
+#if NETFX
```
vatsan-madhavan: Please add a comment near `#if NETFX` indicating that it continues to be useful in the .NET Core 3.0+ codebase, and exists intentionally. 

We are in the process of ripping out `#if NETFX` from the codebase and most of those are _TODO's_ or "remove this section" type conditionals. Having a nearby comment would reduce confusion during refactoring/cleanup.

src/Microsoft.DotNet.Wpf/src/PresentationBuildTasks/MS/Internal/Tasks/TaskHelper.cs
```diff
@@ -318,6 +316,8 @@ internal static bool CouldReferenceAssemblyBeChanged(string assemblyPath, string
                         }
                     }
                 }
+#if NETFX
```
vatsan-madhavan: Can this be refactored into a separate method, so that `#if NETFX` is not necessary inside the body of `CouldReferenceAssemblyBeChanged` ? 

Something like this: 

```C#

internal static bool CouldReferenceAssemblyBeChanged(...)
{
   ...
    // Check the Global Assembly Cache (GAC) on .NET Framework only.
    TestGacPaths(ref bCouldBeChanged); 
   ...
}


private static void TestGacPaths(ref bool bCouldBeChanged)
{
#if NETFX
    IEnumerable<string> gacRoots = GetGacRoots();
    if (gacRoots != null)
    {
        foreach(var gacRoot in gacRoots)
        {
            if (.... )
            {
                bCouldBeChanged = false;
            }
        }
    }
#endif 
}

```

src/Microsoft.DotNet.Wpf/src/PresentationBuildTasks/MS/Internal/Tasks/TaskHelper.cs
vatsan-madhavan: At this point this comment is probably irrelevant. All the _todo_ style `#if NETFX` entries have been removed from the codebase. rebase or merge with master and you should be good to go. Commenting is probably still a good idea for documentation.

ryalanms:
MetadataLoader is an 'all or nothing' mechanism for retaining loading assemblies.  The 'CouldReferenceAssmeblyBeChanged' logic no longer makes sense when reflection-only assemblies are only retained per compilation step, then released with the MetadataLoader is disposed.

The logic with the old ReflectionOnlyLoad logic was to test if an assembly had been changed, and if so, remove it from the cache (releasing it) and set it to load from memory.  Now, we do that between tasks.  Any assembly that was in the GAC was 'unchangeable' and never removed from the cache.  

This whole mechanism no longer makes sense with MetadataLoader, the ReflectionOnly load replacement for .NET Core, so I'm removing CouldReferenceAssemblyBeChanged and dependent methods.  Again, all the assemblies will be released between compile passes.

weltkante:
@ryalanms At the top of the PR you were saying

> PresentationBuildTasks is the only multi-targeted WPF component. Both .NET Core and .NET Framework PBT assemblies are produced by the WPF build.

I don't think .NET Framework has `MetadataLoadContext` so could you clarify whether this component is going to be used in Desktop .NET Framework or not? Because if it is then you just removed a code path which was dedicated to Desktop Framework based on assumptions only present in .NET Core ...

PR #975, opened 2019-06-15
Change _TestHelixAgentPool from 19H1 to RS5

### LINKED ISSUE #952: WPF Builds fail intermittently due to timeouts
WPF builds fail intermittently due to timeouts when _Run Developer Regression Tests on Helix Machine_ stage fails to start within its timeout period (60 mins). 

Example of a failed build: https://dev.azure.com/dnceng/internal/_build/results?buildId=224164 

= 

`_TestHelixAgentPool` in `eng\pipeline.yml` is currently set to `Windows.10.Amd64.Open%3bWindows.7.Amd64.Open%3bWindows.10.Amd64.ServerRS5.Open` 

It should be changed to `Windows.10.Amd64.Open%3bWindows.7.Amd64.Open%3bWindows.10.Amd64.Client19H1.Open`. 

In short, stop running tests on _ServerRS5_. It's a "server core" machine machine, and seems to have problems running tests right now. Change it to _Client19H1_.

vatsan-madhavan:
Changing `_TestHelixAgentPool` from 19H1 to RS5 seems to bring down execution time in Helix to approx 15 mins. 

https://dev.azure.com/dnceng/public/_build/results?buildId=226613&view=logs 

Looks like the 19H1 pool may have been the reason our builds have been suffering from chronic timeouts, much like the ServerRS5 pool prior to that. 

Fixes #952

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -36,3 +36,5 @@ jobs:
         queue: buildpool.windows.10.amd64.vs2019.pre
       # runAsPublic is used in expressions, which can't read from user-defined variables
       runAsPublic: false
+    timeoutInMinutes: 120 # how long to run the job before automatically cancelling; see https://github.com/dotnet/wpf/issues/952
+
--- a/eng/pipeline.yml
+++ b/eng/pipeline.yml
@@ -50,7 +50,7 @@ jobs:
         - name: _PlatformArgs
           value: /p:Platform=$(_Platform)
         - name: _TestHelixAgentPool
-          value: 'Windows.10.Amd64.Open%3bWindows.7.Amd64.Open%3bWindows.10.Amd64.Client19H1.Open'
+          value: 'Windows.10.Amd64.Open%3bWindows.7.Amd64.Open%3bWindows.10.Amd64.ClientRS5.Open'
         - name: _HelixStagingDir
           value: $(BUILD.STAGINGDIRECTORY)\helix\functests
         - name: _HelixSource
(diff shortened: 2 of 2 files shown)


vatsan-madhavan:
@msftbot, hold this PR for 0 minutes; do not require any signoffs.

vatsan-madhavan:
@msft-github-bot, hold this PR for 0 minutes; do not require any signoffs.

vatsan-madhavan:
This time RS5 seemed to complete quickly, but generic Win10 and Win7 took a lot of time. 

I’m not sure what the right solution should be. @mattgal any suggestions?

MattGal:
@vatsan-madhavan I'm observing that certain teams are doing work items that take 1.5+ hours EACH due to setting special environment variables for JIT stress and such;  when these teams (i.e. coreCLR) run their tests on any given queue, that queue will be backed up a bit because of the slow running work items sent.

I'll try to identify the folks doing this and see if we can maybe not do this every PR, but for now the best solution is to have longer build timeouts; if you don't the work still gets done, but the build gets retried; this is only making the problem worse and not improving your speeds at all.

vatsan-madhavan:
> I'll try to identify the folks doing this and see if we can maybe not do this every PR, but for now the best solution is to have longer build timeouts; 

Alright, we'll try to up our timeout to 120 mins for now. Is there a tracking issue we can follow to keep up? I'd really like to keep our build times not become arbitrarily long. Can we chat about the possibility of getting separate Helix queues for our team (perhaps shared with WinForms) maybe? 

/cc @AdamYoblick

AdamYoblick:
Definitely share with us, we’re all friendly over here 😊

vatsan-madhavan:
We will add back Win7 after Preview 7. For now, we will test RS5 only.

PR #978, opened 2019-06-15
corrected line breaks within the file dialog resource strings.

### LINKED ISSUE #977: File Dialog messages that contain \r\n render the \r\n as text
* .NET Core Version: 3.0.0-preview6-27804-01
* Windows version: Windows 10 Build 18917.rs_prerelease.190607-1942 and Windows 7 sp1

 **Problem description:**
When you use an Open File Dialog and provide an invalid file OR if you use a Save File Dialog and try to overwrite an existing file the messagebox renders the \r\n as text instead of  a line break.

This happens if you use the Microsoft.Win32 base implementations or the System.Windows.Forms varients

 **Actual behavior:** 
image
 
 **Expected behavior:**
image

 **Minimal repro:**
```
var saveDialog = new Microsoft.Win32.SaveFileDialog
{
	OverwritePrompt = true
};
var success = saveDialog.ShowDialog();
var saveDialogForms = new System.Windows.Forms.SaveFileDialog
{
	OverwritePrompt = true
};
var successForms = saveDialog.ShowDialog();
```

mjjames:
Fixes #977 

I noticed in all of the other resource strings only \n is used not \r\n. Updated all of the file dialog resources to use \n instead of the \r\n.

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/PresentationFramework/Resources/ExceptionStringTable.txt
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationFramework/Resources/ExceptionStringTable.txt
@@ -1672,10 +1672,10 @@ FlowDocumentReader_MultipleViewProvider_ScrollViewName=Scroll
 
 ; File and Save Dialog error messages
 FileDialogBufferTooSmall=Too many files selected. Select fewer files and try again.
-FileDialogCreatePrompt='{0}' does not exist.\r\nDo you want to create it?
-FileDialogFileNotFound='{0}' does not exist.\r\nVerify that the file name is correct.
+FileDialogCreatePrompt='{0}' does not exist.\nDo you want to create it?
+FileDialogFileNotFound='{0}' does not exist.\nVerify that the file name is correct.
 FileDialogInvalidFileName='{0}' is not a valid file name.
-FileDialogOverwritePrompt='{0}' already exists.\r\nDo you want to replace it?
+FileDialogOverwritePrompt='{0}' already exists.\nDo you want to replace it?
 
 ; Navigation Window
 NavWindowMenuCurrentPage=Current Page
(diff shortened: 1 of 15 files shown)


Gillibald:
Isn't the issue that `\r\n` isn't recognized as a line break? This is just hiding it. `\r\n` is totally valid on Windows.

mjjames:
As part of this effort I did check that messagebox.show has no issues with the \r\n . I have made an assumption that the other resource entries are intentionally using just \n and they are doing so for a reason that I couldn’t immediately see.

My initial investigation seemed to show everything looking correct for the other resource entries which is why the assumption was made that these had to be updated.

One thing I hadn’t tried was the combination of string.format with messagebox.show nor using the get resource method and seeing if the issue laid in there instead.

More than happy to be corrected if I have missed something else.

weltkante:
I agree with @Gillibald this doesn't seem to be the right fix, see note on issue. If `\n` would work then you'd see a line break after the (literal) `\r`

Can you make sure you can reproduce the original bug without your fix?

mjjames:
more than happy to abandon this PR then. The issue 100% does exist as is, my repo in the original bug filing shows the issue as does a bug raised on the Nuget Package Explorer repository https://github.com/NuGetPackageExplorer/NuGetPackageExplorer/issues/705

weltkante:
@mjjames I'm not saying the issue doesn't exist, I just would like to get to the root cause of the problem. So you are saying you can reproduce this bug with a custom WPF build using the original strings (to make sure the bug isn't already fixed in master) and the bug goes away after removing `\r` in the resource xml files and rebuilding?

stevenbrix:
@mjjames don't abandon this pr just yet. Do we know when this started happening? I'd be curious if this has been in WPF since we first ported it to core, or if something regressed along the way

weltkante:
@stevenbrix its a general problem and also present in WinForms (see dotnet/winforms#1076)

As said before I don't think this PR fixes the issue, the problem is (probably) that nobody is translating C# escape sequences in XLF files, and that problem should be independent of `\r\n` vs `\n` issues

weltkante:
#977 should have been fixed by #1285 so this PR probably can be dropped now, there should be no escape sequences left to replace

Issue #585, opened 2019-04-15
Behavior change regarding seekability of streams returned for ZipArchive


ojhad:
* .NET Core Version: 3.0.100-preview4

 
 
 **Problem description:**

There is a behavior change regarding the seekability of returned ZipArchiveEntry streams.

 **.NET Framework behavior:** 

The WindowsBase implementation of Zip had the necessary functionality that operated on the returned deflate streams which provided support for seeking (However, in doing so, it incurred a lot of extra CPU and file system usage)

 **.NET Core behavior:**
The current implementation in CoreFX (System.IO.Packaging) no longer provides this functionality, resulting in [streams that are being returned](https://github.com/dotnet/corefx/blob/6322a406f88e854c5b827ab838f988f8992048f8/src/System.IO.Packaging/src/System/IO/Packaging/ZipStreamManager.cs#L86) to be non-seekable. 

**Related Issue:**
https://github.com/dotnet/corefx/issues/34219

jbartlau:
I'm also having an issue with this exception if I call

```cs
var document = new XpsDocument(@"repro.xps", FileAccess.Read);
var paginatorSource = (IDocumentPaginatorSource)document.GetFixedDocumentSequence();
```

on the attached XPS file (which was created with the XPS Document Writer driver). See my comment at #575. Exception details:

```
System.NotSupportedException
  HResult=0x80131515
  Message=This operation is not supported.
  Source=System.IO.Compression
  StackTrace:
   at System.IO.Compression.DeflateStream.get_Position()
   at System.IO.Packaging.ZipWrappingStream.get_Position()
   at MS.Internal.IO.Packaging.SynchronizingStream.get_Position() in /_/src/Microsoft.DotNet.Wpf/src/shared/MS/Internal/IO/SynchronizingStream.cs:line 207
   at MS.Internal.IO.Packaging.DeobfuscatingStream.Read(Byte[] buffer, Int32 offset, Int32 count) in /_/src/Microsoft.DotNet.Wpf/src/PresentationCore/MS/Internal/IO/Packaging/DeobfuscatingStream.cs:line 56
   at MS.Internal.FontCache.FontSource.StreamToByteArray(Stream fontStream) in /_/src/Microsoft.DotNet.Wpf/src/PresentationCore/MS/Internal/FontCache/FontSource.cs:line 419
[...]

PR #995, opened 2019-06-18
Wrap non-seekable stream in XamlToRtfWriter

### LINKED ISSUE #597: Printing fails in WPF
* .NET Core Version: 3.0.0-preview4-27613-07
* Windows version: Win7, Win8.1, Win10
 
 **Problem description:**

attempt to print using the following code snippet

```csharp
PrintQueue.CreateXpsDocumentWriter(LocalPrintServer.GetDefaultPrintQueue()).Write(visual);
```
 
 **Actual behavior:** 
Exception Info: System.IO.IOException: Cannot retrieve parts of writeonly container
   at System.IO.Packaging.Package.ThrowIfWriteOnly()
   at System.IO.Packaging.Package.GetRelationshipsByType(String relationshipType)
   at System.Windows.Xps.Packaging.XpsManager.GetXpsDocumentStartingPart(Package package)
   at System.Windows.Xps.Packaging.XpsDocument.AddFixedDocumentSequence()
   at System.Windows.Xps.Serialization.XpsPackagingPolicy.AcquireXmlWriterForFixedDocumentSequence()
   at System.Windows.Xps.Serialization.XpsSerializationManager.AcquireXmlWriter(Type writerType)
   at System.Windows.Xps.Serialization.ReachHierarchySimulator.SimulateBeginFixedDocumentSequence()
   at System.Windows.Xps.Serialization.ReachHierarchySimulator.BeginConfirmToXPSStructure(Boolean mode)
   at System.Windows.Xps.Serialization.XpsSerializationManager.SaveAsXaml(Object serializedObject)
   at System.Windows.Xps.XpsDocumentWriter.SaveAsXaml(Object serializedObject, Boolean isSync)
 
 **Expected behavior:**
Printing succeeds 

In Win7 this may or may not be an issue depending on the printer capabilities (Xps capable printer). In RS2+ this is also not an issue

 
Adding additional detail  from VSO 847076

Print using the below line of code:

            PrintQueue.CreateXpsDocumentWriter(LocalPrintServer.GetDefaultPrintQueue()).Write(visual);

Observe a System.IO.IOException in System.IO.Packaging.

In general it seems like corefx's version of System.IO.Packaging does not provide the support needed by WPF's XPS APIs to produce a correct XPS package to either send to an XPS compatible printer or save to disk.

This was found through failures in DrtInterleavedPackage which essentially simulates the way our XPS APIs work by directly calling into the System.IO.Packaging APIs.

System.IO.Packaging in Core dropped streaming creation support, which in the original implementation was the only way to create a Write-only package that would allow interleaving. Seems like there have been some changes to enable Write-only package creation without streaming, but there is no interleaving support, and some of the actions that the XPS code is trying to perform are no longer allowed under the new Write-only scenario.

We should investigate whether our XPS code needs to change to accommodate for the support available in System.IO.Packaging, or if we need to re-introduced some of the dropped support.

The reason this is broken only in the mentioned OS is because of the new XPS Object Model API support we added in 4.6, essentially the old method of printing required us to write an XPS document package directly to a stream given to us by the printer. the XPS Object Model APIs instead expose the entire XPS protocol through interfaces, so we just call methods on said interfaces and the XPS document that is sent to the printer is created for us.

Also Win7 will still print if the printer is GDI.

### LINKED ISSUE #858: [netcore3] XamlToRtfWriter.WriteShapeImage fails on unseekable streams
In PresentationFramework (XamlToRtfWriter.cs), the 'WriteShapeImage' function fails to due change in seekability of the zip streams returned in Read mode. The function reads from the same stream twice. This becomes a problem because after the first read, the stream becomes disposed, which causes an exception to be thrown when trying to set the position of the stream back to 0 in order to read the stream again. 

Github issue #585  is related to the root cause 

**Steps to reproduce:**

1. Run DigitalDocuments/...../DrtSequence
2. Observe following exception thrown when trying to update the stream position in function 'ConvertToImageHexDataString'

System.ObjectDisposedException: 'A stream from ZipArchiveEntry has been disposed.

Object name: 'System.IO.Compression.SubReadStream'.'

stevenbrix:
When we started using `System.IO.Packaging` that came in corefx as part of this commit [71c9337d](https://dev.azure.com/dnceng/internal/_git/dotnet-wpf-int/commit/71c9337d68a8c7825388b3387840ea90acfde860?refName=refs%2Fheads%2Fmaster), the behavior of the streams that were returned from `PackagePart.GetStream` are now non-seekable. The behavior change of this issue is described in #585, with links to the corresponding corefx issue that discusses the behavior change. The workaround for the behavior change is to wrap the stream in a `MemoryStream` to allow seek-ability.

Due to the refactor, the zip implementation inside of `WindowsBase` is no longer needed. I've removed it as part of this change since this should have been removed with the original refactor, but was missed.

The `PackUriHelper.GetPartUri` method was also missing in CoreFx. This was added with https://github.com/dotnet/corefx/pull/38699. I've updated the code to remove more of our custom duplicated logic in favor for what is in corefx. No code that is in our internal branch depends on these APIs, they already use the `PackUriHelper` APIs that come from `System.IO.Packaging`.

Fixes #858 
Fixes #597


src/Microsoft.DotNet.Wpf/src/WindowsBase/WindowsBase.csproj
```diff
@@ -331,35 +328,6 @@
     <Compile Include="System\Windows\Markup\XmlPartReader.cs" />
   </ItemGroup>
   <!-- ZIP sources -->
```
vatsan-madhavan: remove comment

stevenbrix:
> lgtm.
> 
> the previously failing tests pass now, do they?

@vatsan-madhavan, yeah the tests pass. Some of theme are no longer needed (like DrtZipIO) and quite a few failures were also just due to test code needing to be updated. Most of the tests use reflection to discover and invoke methods and that is the biggest reason they've started failing (rather than breaking at compile time). They were either trying to reflect and invoke internal methods that were removed, or the parameters were slightly different (removal of the `bool streaming` parameter from `Package.Open`). Most tests don't even rely on the streaming functionality that existed. This is at least true for the DRTs, I'm not sure what subset of feature tests that might be broken, but fixing those should be fairly straightforward as well.

PR #996, opened 2019-06-18
Add FlagsToEnable property to JumpTask to allow JumpList to set the SHELL_LINK_DATA_FLAGS enumeration on the shell link

### LINKED ISSUE #950: New feature: Add property to JumpTask to allow for elevated jump list entry
Background: PowerShell Core has a lot of custom code [here](https://github.com/PowerShell/PowerShell/tree/master/src/Microsoft.PowerShell.ConsoleHost/WindowsTaskbarJumpList) that makes COM calls to the Windows API for creating a jumplist entry to run PowerShell as an Administrator. However, in both .Net Core 2.1 and .Net Core 3.0-preview5, it has been observed that the CLR fails sporadically with a fatal error: https://github.com/PowerShell/PowerShell/issues/9295

WPF has APIs for creating a taskbar jump list but not for launching the application as elevated. PowerShell would therefore like to rather use WPF to minimize the amount of (possibly buggy code).

Proposal:

We add a boolean property to `JumpTask` so that when when `JumpList.Apply()` gets called the application will be elevated. The default will be to start unelevated.
The implementation will be quite simple, all we need to do is to use `SLDF_RUNAS_USER` enum on [SHELL_LINK_DATA_FLAGS](https://docs.microsoft.com/en-us/windows/desktop/api/shlobj_core/ne-shlobj_core-shell_link_data_flags). We will be calling [IShellLinkDataList::GetFlag](https://docs.microsoft.com/en-us/windows/desktop/api/shobjidl_core/nf-shobjidl_core-ishelllinkdatalist-getflags)`, adding the enum value using XOR and then call [IShellLinkDataList::SetFlags](https://docs.microsoft.com/en-us/windows/desktop/api/shobjidl_core/nf-shobjidl_core-ishelllinkdatalist-setflags) on the `shellLink`  object of type `IShellLinkW` in JumpList.

Are you OK with this proposal that includes no breaking change?
I would first code this from a theoretical perspective, open a draft PR and from then on I would be grateful for some assistance for being able to test it and will have to do some homework to get familiar with this repo (I have never committed to .Net Core before although I use it a lot)

bergmeister:
Closes #950

Add FlagsToEnable property to JumpTask to allow JumpList to set the [SHELL_LINK_DATA_FLAGS](https://docs.microsoft.com/en-us/windows/desktop/api/shlobj_core/ne-shlobj_core-shell_link_data_flags) enumeration on the shell link. This enables various scenarios such as e.g. being able to request elevation of a JumpTask or allowing the link to point to another shell. Because there are so many scenarios, I decided to expose this low level enumeration. Do we need to create an enum wrapper for possible [SHELL_LINK_DATA_FLAGS](https://docs.microsoft.com/en-us/windows/desktop/api/shlobj_core/ne-shlobj_core-shell_link_data_flags) values or shall we leave it up to people? This has the complexity that some enumerations were only introduced in newer versions of Windows (e.g. Win8) or can we disregard this since support for Windows 7 runs out in 6 months anyway?

Context: This is so that PowerShell does not have to maintain hundreds of lines of COM code just for the `Run As Administrator` jump task and also receive the more sophisticated handling by WPF as it has shown that PowerShell's custom code can sometimes cause fatal CLR errors.

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/PresentationFramework/MS/Internal/AppModel/ComGuids.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationFramework/MS/Internal/AppModel/ComGuids.cs
@@ -33,6 +33,8 @@ namespace MS.Internal.AppModel
         public const string ShellFolder = "000214E6-0000-0000-C000-000000000046";
         /// <summary>IID_IShellLink</summary>
         public const string ShellLink = "000214F9-0000-0000-C000-000000000046";
+        /// <summary>IID_IShellLinkDataList</summary>
+        public const string ShellLinkDataList = "45E2B4AE-B1C3-11D0-B92F-00A0C90312E1";
         /// <summary>IID_IShellItem</summary>
         public const string ShellItem = "43826d1e-e718-42ee-bc55-a1e261c37bfe";
         /// <summary>IID_IShellItem2</summary>
--- a/src/Microsoft.DotNet.Wpf/src/PresentationFramework/MS/Internal/AppModel/ShellProvider.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationFramework/MS/Internal/AppModel/ShellProvider.cs
@@ -434,6 +434,30 @@ namespace MS.Internal.AppModel
     [SecurityCritical(SecurityCriticalScope.Everything), SuppressUnmanagedCodeSecurity]
     [
         ComImport,
+        InterfaceTypeAttribute(ComInterfaceType.InterfaceIsIUnknown),
+        Guid(IID.ShellLinkDataList),
+    ]
+    internal interface IShellLinkDataListW
+    {
+        [PreserveSig]
+        Int32 AddDataBlock(IntPtr pDataBlock);
+
+        [PreserveSig]
+        Int32 CopyDataBlock(UInt32 dwSig, out IntPtr ppDataBlock);
+
+        [PreserveSig]
+        Int32 RemoveDataBlock(UInt32 dwSig);
+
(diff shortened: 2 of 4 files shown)


wjk:
My thoughts on this PR: There are a lot of flags available in the listed docs page, many/most of which I wouldn't use very often, if ever (`SLDF_RUN_IN_SEPARATE`, anyone?). Furthermore, I don't know if specifying/not specifying some of these flags would break anything in the Jump List (e.g. what would happen if `SLDF_HAS_NAME` was not set/cleared, and a name was specified regardless?). Moreover, having reviewed #950, I would think that a simple boolean property that maps to `SLDF_RUNAS_USER` would be sufficient for your and my use cases. Hope this helps!

bergmeister:
The idea was to keep it open to give developers more option but I am also aware that the .Net framework prefers simpler APIs and not expose 'naked' low level detail. I'd be happy with your approach as well but at this point I would like to get a review and formal agreement on the decision by the product team please

bergmeister:
@ojhad Any chance someone can review this PR or its associated issue as I'd like to get this feature into 3.0

ryalanms:
This is a public API change and must go through API review.

miloush:
There should be an enum with the corresponding flags instead of the int, and there is no reason for the property to be an array rather than just the enum. It probably should also be called just `Flags`, or `LinkFlags` or similar, like the other properties who have the same behaviour (i.e. it's not `DescriptionToSet`). And property documentation should start with accessors, i.e. "Gets or sets..."

ryalanms:
Adding @MikeHillberg for API review...

PR #1005, opened 2019-06-18
Add control library templates

jmarolf:
fixes https://github.com/dotnet/project-system/issues/4715

PR #1007, opened 2019-06-19
open xaml file first in editor

jmarolf:
fixes [AB#832526](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/832526)

### DIFF
--- a/packaging/Microsoft.Dotnet.Wpf.ProjectTemplates/content/WpfApplication-CSharp/.template.config/template.json
+++ b/packaging/Microsoft.Dotnet.Wpf.ProjectTemplates/content/WpfApplication-CSharp/.template.config/template.json
@@ -59,7 +59,8 @@
     {
       "condition": "(HostIdentifier != \"dotnetcli\" && HostIdentifier != \"dotnetcli-preview\")",
       "path": "MainWindow.xaml.cs"
-    }
+    },
+    { "path": "MainWindow.xaml" }
   ],
   "defaultName": "WpfApp1",
   "postActions": [
@@ -74,11 +75,11 @@
     },
     {
       "condition": "(HostIdentifier != \"dotnetcli\" && HostIdentifier != \"dotnetcli-preview\")",
-      "description": "Opens MainWindow.xaml.cs in the editor",
+      "description": "Opens MainWindow.xaml in the editor",
       "manualInstructions": [ ],
       "actionId": "84C0DA21-51C8-4541-9940-6CA19AF04EE6",
       "args": {
-        "files": "1"
+        "files": "1;2"
       },
       "continueOnError": true
     }
--- a/packaging/Microsoft.Dotnet.Wpf.ProjectTemplates/content/WpfApplication-CSharp/.template.config/template.json
+++ b/packaging/Microsoft.Dotnet.Wpf.ProjectTemplates/content/WpfApplication-CSharp/.template.config/template.json
@@ -60,7 +60,10 @@
       "condition": "(HostIdentifier != \"dotnetcli\" && HostIdentifier != \"dotnetcli-preview\")",
       "path": "MainWindow.xaml.cs"
     },
-    { "path": "MainWindow.xaml" }
+    { 
+      "condition": "(HostIdentifier != \"dotnetcli\" && HostIdentifier != \"dotnetcli-preview\")",
+      "path": "MainWindow.xaml"
+    }
   ],
   "defaultName": "WpfApp1",
   "postActions": [
--- a/packaging/Microsoft.Dotnet.Wpf.ProjectTemplates/content/WpfApplication-VisualBasic/.template.config/template.json
+++ b/packaging/Microsoft.Dotnet.Wpf.ProjectTemplates/content/WpfApplication-VisualBasic/.template.config/template.json
@@ -59,7 +59,8 @@
     {
       "condition": "(HostIdentifier != \"dotnetcli\" && HostIdentifier != \"dotnetcli-preview\")",
       "path": "MainWindow.xaml.vb"
-    }
+    },
+    { "path": "MainWindow.xaml" }
   ],
   "defaultName": "WpfApp1",
   "postActions": [
@@ -74,11 +75,11 @@
     },
     {
       "condition": "(HostIdentifier != \"dotnetcli\" && HostIdentifier != \"dotnetcli-preview\")",
-      "description": "Opens MainWindow.xaml.vb in the editor",
+      "description": "Opens MainWindow.xaml in the editor",
       "manualInstructions": [ ],
       "actionId": "84C0DA21-51C8-4541-9940-6CA19AF04EE6",
       "args": {
-        "files": "1"
+        "files": "1;2"
       },
       "continueOnError": true
     }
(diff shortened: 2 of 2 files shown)


vatsan-madhavan:
@jmarolf - Please start by opening an issue per [guidelines](https://github.com/dotnet/wpf/blob/master/Documentation/contributing.md). Let's discuss what you are trying to achieve  and we can identify the appropriate solution, timelines etc. fromt there. Same goes for #1005

jmarolf:
@debonte can you file an issue on this repo and talk to @vatsan-madhavan 	to justigy this change?

debonte:
@jmarolf, I opened #1017

debonte:
@jmarolf, two thoughts on this change:
1. It looks like this will result in only opening MainWindow.xaml. While this is the expected behavior for Blend, in VS the WPF Framework experience is that both MainWindow.xaml and MainWindow.xaml.cs/vb are open (and the XAML file is active).
2. I wonder if the choice of which documents to open should be made in the .vstemplate (or at least be overridable there) instead of in the dotnet template?

mgoertz-msft:
@vatsan-madhavan We have to do something about the Blend experience. It is not acceptable for Blend to open a C# file instead of a XAML file when creating a new project.

Perhaps there is something we can do with the HostIdentifier condition to get one experience in Blend and another in VS.

stevenbrix:
This change seems righteous, we should be striving to have the right user experience in both VS and Blend.

> I wonder if the choice of which documents to open should be made in the .vstemplate (or at least be overridable there) instead of in the dotnet template?

@debonte i don't see a VSTemplate, but from the json schema it looks like you can have multiple `primaryOutputs` and specify which one is the default. Does that fix the scenario?

debonte:
>  it looks like you can have multiple primaryOutputs and specify which one is the default. Does that fix the scenario?

@stevenbrix I'm not super-concerned where it's done. I just thought the .NET Core SDK team might not want to encode VS/Blend-specific behavior in their templates.

I think HostIdentifier may currently be a fixed value for both VS and Blend (VSConfigurationHelpers.DefaultHostId), so I think that would need to be changed first.

Btw, from @jmarolf's changes it's not clear to me if its the primaryOutputs or the postActions that actually determines which files are opened.

> i don't see a .vstemplate

Look for Wpf.vstemplate either under c:\Program Files (x86)\Microsoft Visual Studio\<vs-instance>\Common7\IDE\Extensions or %LOCALAPPDATA%\Microsoft\VisualStudio\<vs-instance-id>

stevenbrix:
Ah ok i see, i thought this was something in our repo, but those ship with VS.

> I just thought the .NET Core SDK team might not want to encode VS/Blend-specific behavior in their templates.

I think this is a good general goal to strive for, we should strive to be unaware of the tools. How was this done in framework?

debonte:
> I think this is a good general goal to strive for. How was this done in framework?

The .NET Framework .vstemplates are self-contained. They don't call into the dotnet template engine. Those .vstemplates have metadata for each file that is generated, some of which specifies which files should be opened and in which order. See #1017 for a snippet that shows how it controls the MainWindow.xaml/MainWindow.xaml.cs behavior.

jmarolf:
> The .NET Framework .vstemplates are self-contained. They don't call into the dotnet template engine. Those .vstemplates have metadata for each file that is generated, some of which specifies which files should be opened and in which order. See #1017 for a snippet that shows how it controls the MainWindow.xaml/MainWindow.xaml.cs behavior.

I'll look and see if there is a way to get this behavior in the .NET Core templates.  Today [this](https://devdiv.visualstudio.com/DevDiv/_git/WebTools?path=%2Fsrc%2FTemplateEngine%2FMicrosoft.VisualStudio.TemplateEngine.UI%2FPostCreationActions%2FOpenFileInEditorPostCreationAction.cs&version=GBmaster) is all the action is doing (essentially DTE.ItemOperations.OpenFile for a given list of files).

vatsan-madhavan:
I was able to spend some time looking over the C# template changes - I'll assume that the discussion extends to the VB template symmetrically. 

"open xaml file first in editor" is exactly what it seems to be doing - good so far. This is what I'm noticing overall relative to the _spec_ #1017: 


|       | Spec                                     | Actual                                   |
|-------|------------------------------------------|------------------------------------------|
| **VS**    | - Open MainWindow.xaml & MainWindow.xaml.cs<br>- Focus on MainWindow.xaml | Per spec                                 |
| **Blend** | - Open MainWindow.xaml only<br>- Focus on MainWindow.xaml | - Opens MainWindow.xaml & MainWindow.xaml.cs<br>- Focus on MainWindow.xaml<br>(Behavior identical to VS) |


- Unless there is a way to differentiate Blend and VS today - for e.g., a difference in `HostIdentifier` - it's not clear to me that achieving parity with spec will be possible on Blend.  Perhaps the tentative approach should be to accept the current limitation in Blend (i.e., the fact that MainWindow.xaml.cs is also opened in the background), until its `HostIdentifier` can be changed to something unique? 
-  It's not clear to me either whether `primaryOutputs` or `postActions` that determines which file is opened. I think it's a combination of both. 
  - Empirically, `postActions.args.files` seems to control the order in which the files appear. If we had 3 files in `primaryOutputs` listed in the order `{MainWindow.xaml.cs, MainWindow.xaml, App.xaml}`, and `postActions.args.file="1;3;2"`, this is what you'd see: 
    - image


What kind of testing has this proposed change undergone?

packaging/Microsoft.Dotnet.Wpf.ProjectTemplates/content/WpfApplication-CSharp/.template.config/template.json
```diff
@@ -59,7 +59,8 @@
     {
       "condition": "(HostIdentifier != \"dotnetcli\" && HostIdentifier != \"dotnetcli-preview\")",
       "path": "MainWindow.xaml.cs"
-    }
+    },
+    { "path": "MainWindow.xaml" }
```
vatsan-madhavan: There is no condition here similar to `MainWindow.xaml.cs`. Can you help me understand whether this was intentional, and if so, why we'd want `MainWindow.xaml` included in `primaryOutputs` when `HostIdentifer=="dotnetcli*"`, for e.g? 

I notice that `postActions` is set up not to react to this addition even when `Hostidentifer=="dotnetcli*"`, so perhaps its not a big problem. We should probably still maintain the right conditions if its easy to do so.

mgoertz-msft:
OK, I guess I misunderstood. It looked like this change is needed to open the XAML file but you weren't going to take it because of the NO MERGE label.

As long as the XAML file is active in Blend on File->New Project then it is just a minor issue that there is also a C# file in the background. We could address that in a separate issue.

vatsan-madhavan:
> OK, I guess I misunderstood. It looked like this change is needed to open the XAML file but you weren't going to take it because of the NO MERGE label.

To clarify, the NO MERGE label was added because the PR didn't have issue description, what was being fixed, testing details etc, and we didn't want an accidental merge before Preview 7 completed. 

I also reached out @jmarolf by email asking for additional details, and explaining that we just had a regression in project templates and wanted to be careful in Preview 7. 

Once P7 is done, we'll remove the NO MERGE label.

mgoertz-msft:
Sounds good, thank you.

vatsan-madhavan:
Removing *NO MERGE* since #1100 is completed and master has switched to _Preview 8_. This PR can continue on with usual due-diligence.

mgoertz-msft:
I just chatted with DoRon and as long as the .xaml file ends up being the active one then that would be acceptable for Blend.

vatsan-madhavan:
> I just chatted with DoRon and as long as the .xaml file ends up being the active one then that would be acceptable for Blend.

Thanks, that's good to know! 

@jmarolf, do you have a sense for when we can expect an update from you re: other open questions?

jmarolf:
@vatsan-madhavan 	ready for review

vatsan-madhavan:
@jmarolf  - lgtm!

PR #1009, opened 2019-06-19
Cleanup #if NETFX guarded code and remove XBAP related types

vatsan-madhavan:
Fixes [AB#897927](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/897927) and [AB#652415](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/652415)

- Cleans up `#if NETFX` guarded code that was ported from .NET Framework
- Removes and/or fixes the impl/behavior of XBAP related types


src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Input/Win32KeyboardDevice.cs
```diff
@@ -68,32 +68,9 @@ protected override KeyStates GetKeyStatesFromSystem(Key key)
...
-                    case Key.LeftAlt:
-                    case Key.RightAlt:
-                    case Key.LeftCtrl:
-                    case Key.RightCtrl:
-                    case Key.LeftShift:
-                    case Key.RightShift:
-                        getKeyStatesFromSystem = true;
-                        break;
-                }
+                getKeyStatesFromSystem = true;
```
rladuca: There doesn't seem to be a need for `getKeyStatesFromSystem` anymore.

vatsan-madhavan: removed.

PR #1024, opened 2019-06-19
Fix globbing in ResolveRefApiCompatItems

### LINKED ISSUE #1023: Reference API Compat can crash with multiple configuration/release combinations.
We added API Compat in #724 .

The ResolveRefApiCompatItems target was matching the reference assembly file for use as ResolvedImplementationAssembly via a non-specific glob.  This can pick up more than one item when there are multiple combinations of configuration and platform sitting in the artifacts directory.

rladuca:
Fixes #1023 

Makes the globbing in ResolveRefApiCompatItems more specific so that it always picks exactly one reference assembly.


rladuca:
@msftbot require at least 1 sign-off

Issue #1022, opened 2019-06-19
TextBox's AcceptsReturn does not work


miguep:
* .NET Core Version: 3.0.100-preview7-012511
* Windows version: Win10
* Does the bug reproduce also in WPF for .NET Framework 4.8?: No

 **Problem description:**
A simple TextBox with AcceptsReturn="True" does not behave as expected
 
 **Actual behavior:** 
Upon clicking the TextBox and hitting Enter key, nothing happens
 
 **Expected behavior:**
Upon clicking the TextBox and hitting Enter key, a new line should be added to the TextBox's text.
 
 **Minimal repro:**
 ``` xaml
<TextBox AcceptsReturn="True"></TextBox>
```

PR #1025, opened 2019-06-19
Fix CommandHandler registration issue identified by DRT failure

ryalanms:
Fixes DRT failures and #1022.


stevenbrix:
I looked back over #919 and it's not immediately obvious how this regression started happening, and why these exact changes are the fix. Can you provide more information on what happened? That way we can all learn and maybe be lucky enough to avoid these issues in the future.

ryalanms:
Translations in .NET Framework were in different file formats and split across multiple files.  .resx does not support the metadata contained in these files.  .NET Framework contained strings in the satellite assemblies that were not translated for any language that were used to register key commands.  Some of these key commands were translated when we moved to .NET Core causing crashes at runtime due to lookup failures.  The original change moves the static strings out of the satellite assemblies to prevent accidental translation.

vatsan-madhavan:
Please hold for review from  @rladuca or me.

PR #1027, opened 2019-06-20
Add warnings to Microsoft.NET.Sdk.WindowsDesktop to improve UseWpf/UseWinforms and multi-targeting experience

### LINKED ISSUE #327: Cannot multi-target to netcoreapp < 3.0  and use WindowsDesktop SDK unconditionally
* .NET Core Version: 3.0  Preview 2
* Windows version: (`winver`): N/A, build time error
* Does the bug reproduce also in WPF for .NET Framework 4.8?: N/A
 
 **Problem description:**

If a project targets netcoreapp1.x or netcoreapp2.x in addition to netcoreapp3.0, and uses WindowsDesktop SDK in order to use WPF conditionally on 3.0, it fails to build. (Importing SDKs conditionally is possible, but difficult to get right and hard to maintain.)

The implicit framework reference is conditioned only on TargetFrameworkIdentifier being .NETCoreApp. It ignores TargetFrameworkVersion.

 **Minimal repro:**
`dotnet build` this:

``` xml
  <PropertyGroup>
   <TargetFrameworks>netcoreapp1.0;netcoreapp3.0</TargetFrameworks>
   <UseWPF Condition="'$(TargetFramework)' == 'netcoreapp3.0'">true</UseWPF>
  </PropertyGroup>
</Project>
```
 
 **Actual behavior:** 
```
error NETSDK1073: The FrameworkReference 'Microsoft.WindowsDesktop.App' was not recognized
```
 
 **Expected behavior:**

Successful build
 
cc @AArnott @dsplaisted @vatsan-madhavan

### LINKED ISSUE #685: Add Page automatically, remove from None automatically
When the SDK is "Microsoft.NET.Sdk.WindowsDesktop", it would be nice not to have to add `<UseWPF>true</UseWPF>`. Especially when migrating projects from old-style .csproj files, this is easy to forget and seems otherwise unnecessary since I already claimed what project type I'm going to build. And wouldn't this be true of both WPF and UWP projects anyway - to include *.xaml files as `<Page>` items?

Also, why must the IDE - or, when using VSCode, we (manually) - add `<None Remove="..."/>` when the targets could do this for any `<Page>` items. Could this also be done automatically in the targets for "Microsoft.NET.Sdk.WindowsDesktop"?

This could make WPF SDK-style .csproj files as slim as non-WPF SDK-style .csproj files.

### LINKED ISSUE #746: Support WPF and WinForms specific FrameworkReferences via Profiles
See https://github.com/dotnet/sdk/pull/3259 

- [x] Incorporate Wpf and WinForms specific profiles into `Microsoft.NET.Sdk.WindowsDesktop`
- [ ] Throw error when neither `UseWpf` nor `UseWindowsForms` is set to `true` when `Microsoft.NET.Sdk.WindowsDesktop` is in use

### LINKED ISSUE #867: Show error when neither UseWpf nor UseWindowsForms is set to true
When neither `UseWpf` nor `UseWindowsForms` is set to `true`, `Microsoft.NET.Sdk.WindowsDesktop` should show a (suppressible) error

vatsan-madhavan:
Fixes the following:

- Fixes #685: Remove **/*.xaml from None
- Fixes #327: Cannot multi-target netcoreap < 3.0 and use WindowsDesktop SDK unconditionally.
- Fixes #746: [Epic] Support WPF and WinForms specific FrameworkReferences a Profiles
  - Fixes #867: Show error when neither UseWpf nor UseWindowsForms is set to true

Also cleans up the way in which we import Microsoft.WinFX.targets - UseLegacyPresentationBuildTasks has been broken for some time now and unusable.

### DIFF
--- a/global.json
+++ b/global.json
@@ -1,6 +1,6 @@
 {
   "tools": {
-    "dotnet": "3.0.100-preview6-012264",
+    "dotnet": "3.0.100-preview7-012512",
     "runtimes": {
       "dotnet": [
         "2.1.7",
--- a/packaging/Microsoft.NET.Sdk.WindowsDesktop/targets/Microsoft.NET.Sdk.WindowsDesktop.props
+++ b/packaging/Microsoft.NET.Sdk.WindowsDesktop/targets/Microsoft.NET.Sdk.WindowsDesktop.props
@@ -23,7 +23,8 @@
     <_WindowsDesktopSdkTargetFrameworkVersionFloor>3.0</_WindowsDesktopSdkTargetFrameworkVersionFloor>
   </PropertyGroup>
 
-  <ItemGroup Condition=" '$(EnableDefaultItems)' == 'true' And '$(UseWPF)' == 'true' And ('$(_TargetFrameworkVersionWithoutV)' != '') And ('$(_TargetFrameworkVersionWithoutV)' >= '$(_WindowsDesktopSdkTargetFrameworkVersionFloor)')">
+  <ItemGroup Condition=" ('$(EnableDefaultItems)' == 'true') And ('$(UseWPF)' == 'true') And 
+                         ('$(_TargetFrameworkVersionWithoutV)' != '') And ('$(_TargetFrameworkVersionWithoutV)' >= '$(_WindowsDesktopSdkTargetFrameworkVersionFloor)')">
     <ApplicationDefinition Include="App.xaml"
                            Condition="'$(EnableDefaultApplicationDefinition)' != 'false' And Exists('$(MSBuildProjectDirectory)/App.xaml') And '$(MSBuildProjectExtension)' == '.csproj'">
       <Generator>MSBuild:Compile</Generator>
@@ -54,17 +55,22 @@
           Condition="'$(EnableDefaultApplicationDefinition)' != 'false' And '$(EnableDefaultPageItems)' != 'false'" />
   </ItemGroup>
 
-  <ItemGroup Condition="('$(DisableImplicitFrameworkReferences)' != 'true') And ('$(UseWPF)' == 'true') And ('$(UseWindowsForms)' == 'true') And ('$(TargetFrameworkIdentifier)' == '.NETCoreApp') And ('$(_TargetFrameworkVersionWithoutV)' != '') And ('$(_TargetFrameworkVersionWithoutV)' >= '$(_WindowsDesktopSdkTargetFrameworkVersionFloor)')">
-    <FrameworkReference Include="Microsoft.WindowsDesktop.App" IsImplicitlyDefined="true" />
-  </ItemGroup>
+  
+  <ItemGroup Condition="('$(DisableImplicitFrameworkReferences)' != 'true') And
+                      ('$(TargetFrameworkIdentifier)' == '.NETCoreApp') And
+                      ('$(_TargetFrameworkVersionWithoutV)' != '') And
+                      ('$(_TargetFrameworkVersionWithoutV)' >= '$(_WindowsDesktopSdkTargetFrameworkVersionFloor)')">
 
-  <ItemGroup Condition="('$(DisableImplicitFrameworkReferences)' != 'true') And ('$(UseWPF)' == 'true') And ('$(UseWindowsForms)' != 'true') And ('$(TargetFrameworkIdentifier)' == '.NETCoreApp') And ('$(_TargetFrameworkVersionWithoutV)' != '') And ('$(_TargetFrameworkVersionWithoutV)' >= '$(_WindowsDesktopSdkTargetFrameworkVersionFloor)')">
-    <FrameworkReference Include="Microsoft.WindowsDesktop.App.WPF" IsImplicitlyDefined="true" />
-  </ItemGroup>
+    <FrameworkReference Include="Microsoft.WindowsDesktop.App" IsImplicitlyDefined="true"
+                        Condition="('$(UseWPF)' == 'true') And ('$(UseWindowsForms)' == 'true')"/>
 
-  <ItemGroup Condition="('$(DisableImplicitFrameworkReferences)' != 'true') And ('$(UseWPF)' != 'true') And ('$(UseWindowsForms)' == 'true') And ('$(TargetFrameworkIdentifier)' == '.NETCoreApp') And ('$(_TargetFrameworkVersionWithoutV)' != '') And ('$(_TargetFrameworkVersionWithoutV)' >= '$(_WindowsDesktopSdkTargetFrameworkVersionFloor)')">
-    <FrameworkReference Include="Microsoft.WindowsDesktop.App.WindowsForms" IsImplicitlyDefined="true" />
+    <FrameworkReference Include="Microsoft.WindowsDesktop.App.WPF" IsImplicitlyDefined="true"
+                        Condition="('$(UseWPF)' == 'true') And ('$(UseWindowsForms)' != 'true')"/>
+
+    <FrameworkReference Include="Microsoft.WindowsDesktop.App.WindowsForms" IsImplicitlyDefined="true"
+                        Condition="('$(UseWPF)' != 'true') And ('$(UseWindowsForms)' == 'true')"/>
   </ItemGroup>
+  
 
   <!--
     Traditionally, Visual Studio has supplied these references for .NET Framework based
@@ -89,7 +95,8 @@
      Note:
        These are WPF specific version lower-bounds. Do not use $(_WindowsDesktopSdkTargetFrameworkVersionFloor) in place of '3.0' below.
   -->
-  <ItemGroup Condition="('$(TargetFrameworkIdentifier)' == '.NETFramework') And ('$(_TargetFrameworkVersionWithoutV)' != '') And ('$(_TargetFrameworkVersionWithoutV)' >= '$(_WindowsDesktopSdkTargetFrameworkVersionFloor)')">
+  <ItemGroup Condition="('$(TargetFrameworkIdentifier)' == '.NETFramework') And ('$(_TargetFrameworkVersionWithoutV)' != '') And 
+                        ('$(_TargetFrameworkVersionWithoutV)' >= '$(_WindowsDesktopSdkTargetFrameworkVersionFloor)')">
     <_WpfCommonNetFxReference Include="WindowsBase" Condition="'$(_TargetFrameworkVersionWithoutV)' >= '3.0'" />
     <_WpfCommonNetFxReference Include="PresentationCore" Condition="'$(_TargetFrameworkVersionWithoutV)' >= '3.0'" />
     <_WpfCommonNetFxReference Include="PresentationFramework" Condition="'$(_TargetFrameworkVersionWithoutV)' >= '3.0'" />
@@ -105,15 +112,16 @@
     <_WpfCommonNetFxReference Include="System.Windows.Controls.Ribbon" Condition="'$(_TargetFrameworkVersionWithoutV)' >= '4.5'" />
   </ItemGroup>
 
-  <ItemGroup Condition=" ('$(DisableImplicitFrameworkReferences)' != 'true') And ('$(TargetFrameworkIdentifier)' == '.NETFramework') And ('$(_TargetFrameworkVersionWithoutV)' != '') And ('$(_TargetFrameworkVersionWithoutV)' >= '$(_WindowsDesktopSdkTargetFrameworkVersionFloor)') And ('$(UseWPF)' == 'true') ">
-    <_SDKImplicitReference Include="@(_WpfCommonNetFxReference)" />
-  </ItemGroup>
-
-  <ItemGroup Condition=" ('$(DisableImplicitFrameworkReferences)' != 'true') And ('$(TargetFrameworkIdentifier)' == '.NETFramework') And ('$(_TargetFrameworkVersionWithoutV)' != '') And ('$(_TargetFrameworkVersionWithoutV)' >= '$(_WindowsDesktopSdkTargetFrameworkVersionFloor)') And ('$(UseWindowsForms)' == 'true') ">
-    <_SDKImplicitReference Include="System.Windows.Forms"/>
-  </ItemGroup>
-
-  <ItemGroup Condition=" ('$(DisableImplicitFrameworkReferences)' != 'true') And ('$(TargetFrameworkIdentifier)' == '.NETFramework') And ('$(_TargetFrameworkVersionWithoutV)' != '') And ('$(_TargetFrameworkVersionWithoutV)' >= '$(_WindowsDesktopSdkTargetFrameworkVersionFloor)') And ('$(UseWindowsForms)' == 'true') And ('$(UseWPF)' == 'true') ">
-    <_SDKImplicitReference Include="WindowsFormsIntegration"/>
+  <ItemGroup Condition="('$(DisableImplicitFrameworkReferences)' != 'true') And ('$(TargetFrameworkIdentifier)' == '.NETFramework') And 
+                        ('$(_TargetFrameworkVersionWithoutV)' != '') And ('$(_TargetFrameworkVersionWithoutV)' >= '$(_WindowsDesktopSdkTargetFrameworkVersionFloor)')">
+    
+    <_SDKImplicitReference Include="@(_WpfCommonNetFxReference)"
+                       Condition="'$(UseWPF)' == 'true'"/>
+    
+    <_SDKImplicitReference Include="System.Windows.Forms"
+                        Condition="('$(UseWindowsForms)' == 'true') " />
+    
+    <_SDKImplicitReference Include="WindowsFormsIntegration"
+                       Condition=" ('$(UseWindowsForms)' == 'true') And ('$(UseWPF)' == 'true') "/>
   </ItemGroup>
 </Project>
--- a/packaging/Microsoft.NET.Sdk.WindowsDesktop/targets/Microsoft.NET.Sdk.WindowsDesktop.props
+++ b/packaging/Microsoft.NET.Sdk.WindowsDesktop/targets/Microsoft.NET.Sdk.WindowsDesktop.props
@@ -91,15 +91,21 @@
     .NET 4.5+:  PresentationCore, PresentationFramework, WindowsBase, System.Xaml, 
                 UIAutomationClient, UIAutomationClientSideProviders, UIAutomationProvider, UIAutomationTypes
                 System.Windows.Controls.Ribbon
-                
-     Note:
-       These are WPF specific version lower-bounds. Do not use $(_WindowsDesktopSdkTargetFrameworkVersionFloor) in place of '3.0' below.
+
   -->
   <ItemGroup Condition="('$(TargetFrameworkIdentifier)' == '.NETFramework') And ('$(_TargetFrameworkVersionWithoutV)' != '') And 
                         ('$(_TargetFrameworkVersionWithoutV)' >= '$(_WindowsDesktopSdkTargetFrameworkVersionFloor)')">
-    <_WpfCommonNetFxReference Include="WindowsBase" Condition="'$(_TargetFrameworkVersionWithoutV)' >= '3.0'" />
-    <_WpfCommonNetFxReference Include="PresentationCore" Condition="'$(_TargetFrameworkVersionWithoutV)' >= '3.0'" />
-    <_WpfCommonNetFxReference Include="PresentationFramework" Condition="'$(_TargetFrameworkVersionWithoutV)' >= '3.0'" />
+
+    <!--
+      The following 3 _WpfCommonNetFxReference items normally require Condition="'$(_TargetFrameworkVersionWithoutV)' >= '3.0'", since
+      they are supported on .NET Framework 3.0 and above. 
+      
+      This condition is implicitly satisfied by '$(_TargetFrameworkVersionWithoutV)' >= '$(_WindowsDesktopSdkTargetFrameworkVersionFloor)' 
+      in the outer ItemGroup
+    -->
+    <_WpfCommonNetFxReference Include="WindowsBase" /> 
+    <_WpfCommonNetFxReference Include="PresentationCore" />
+    <_WpfCommonNetFxReference Include="PresentationFramework" />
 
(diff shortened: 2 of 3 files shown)


packaging/Microsoft.NET.Sdk.WindowsDesktop/targets/Microsoft.NET.Sdk.WindowsDesktop.targets
```diff
@@ -47,4 +36,27 @@
 
   </Target>
 
+  <!-- 
+    This warning can be suppressed by setting $(MSBuildWarningsAsMessages) property, like this:
+    <PropertyGroup>
+      <MSBuildWarningsAsMessages>$(MSBuildWarningsAsMessages);NETSDK1106</MSBuildWarningsAsMessages>
+    </PropertyGroup>
+  --> 
+  <Target Name="_WindowsDesktopFrameworkRequiresUseWpfOrUseWindowsForms"
```
vatsan-madhavan: Should we condition this check on `"'$(_TargetFrameworkVersionWithoutV)' >= '3.0'"` ? That way, when a `Microsoft.NET.Sdk.WindowsDesktop` app targets `netcorapp2.2`, it will get `WindowsDesktopFrameworkRequiresVersion30` warning, but not `WindowsDesktopFrameworkRequiresUseWpfOrUseWindowsForms`

packaging/Microsoft.NET.Sdk.WindowsDesktop/targets/Microsoft.NET.Sdk.WindowsDesktop.targets
```diff
@@ -1,20 +1,8 @@
...
- <Import Project="$(MSBuildToolsPath)\Microsoft.WinFX.targets" 
-         Condition ="'$(MSBuildRuntimeType)' != 'Core' and '$(UseLegacyPresentationBuildTasks)' == 'true'"/>
-
- <Import Project="Microsoft.WinFX.targets" 
-         Condition="'$(MSBuildRuntimeType)' == 'Core' or '$(UseLegacyPresentationBuildTasks)' != 'true'" />
-
-  <ItemGroup Condition=" '$(EnableDefaultItems)' == 'true' And '$(UseWPF)' == 'true'">
+  <Import Project="Microsoft.WinFX.targets" />
+  
+  <!-- $(TargetFrameworkIdentifier) can be '.NETCoreApp' or '.NETFramework' -->
```
vatsan-madhavan: Elaborate. 

Explain that WPF is supported on WindowsDesktop SDK on .NET Core 3.0+, or .NET Framework 3.0+.

packaging/Microsoft.NET.Sdk.WindowsDesktop/targets/Microsoft.NET.Sdk.WindowsDesktop.props
```diff
@@ -79,15 +87,15 @@
     <_WpfCommonNetFxReference Include="System.Windows.Controls.Ribbon" Condition="'$(_TargetFrameworkVersionWithoutV)' >= '4.5'" />
   </ItemGroup>
 
-  <ItemGroup Condition=" '$(DisableImplicitFrameworkReferences)' != 'true' and '$(TargetFrameworkIdentifier)' == '.NETFramework' And '$(UseWPF)' == 'true' ">
+  <ItemGroup Condition=" ('$(DisableImplicitFrameworkReferences)' != 'true') And ('$(TargetFrameworkIdentifier)' == '.NETFramework') And ('$(_TargetFrameworkVersionWithoutV)' != '') And ('$(_TargetFrameworkVersionWithoutV)' >= '3.0') And ('$(UseWPF)' == 'true') ">
     <_SDKImplicitReference Include="@(_WpfCommonNetFxReference)" />
   </ItemGroup>
 
-  <ItemGroup Condition=" '$(DisableImplicitFrameworkReferences)' != 'true' and '$(TargetFrameworkIdentifier)' == '.NETFramework' And '$(UseWindowsForms)' == 'true' ">
+  <ItemGroup Condition=" ('$(DisableImplicitFrameworkReferences)' != 'true') And ('$(TargetFrameworkIdentifier)' == '.NETFramework') And ('$(_TargetFrameworkVersionWithoutV)' != '') And ('$(_TargetFrameworkVersionWithoutV)' >= '3.0') And ('$(UseWindowsForms)' == 'true') ">
```
vatsan-madhavan: WinForms is supported on .NET Framework 1.0 onwards. Should this check be converted into `('$(_TargetFrameworkVersionWithoutV)' >= '3.0')` instead? 

That would mean that SDK style projects, specifically WindowsDesktop SDK, is supported for building .NET Framework 1.0 projects. Is this something we want to support (or _enable_, even if it were _unsupported_)? 

/cc @merriemcgaw, @nguerrera

stevenbrix: I'm confused, do you mean `('$(_TargetFrameworkVersionWithoutV)' >= '1.0')`? It looks like you already have `('$(_TargetFrameworkVersionWithoutV)' >= '3.0') `

>That would mean that SDK style projects, specifically WindowsDesktop SDK, is supported for building .NET Framework 1.0 projects

Are we supporting any .NET framework projects building with the WindowsDesktop SDK?

global.json
```diff
@@ -1,6 +1,6 @@
 {
   "tools": {
-    "dotnet": "3.0.100-preview6-012264",
+    "dotnet": "3.0.100-preview7-012512",
```
stevenbrix: trail blazing!

packaging/Microsoft.NET.Sdk.WindowsDesktop/targets/Microsoft.NET.Sdk.WindowsDesktop.props
vatsan-madhavan: Yes netfx projects are built with WindowsDesktop SDK. It’s supported.

vatsan-madhavan: Good catch, I did mean to ask about >= 1.0

global.json
stevenbrix: do we need to update to preview7 due to the CAS changes?

vatsan-madhavan: no, to get the new NetSdkWarning strings.

packaging/Microsoft.NET.Sdk.WindowsDesktop/targets/Microsoft.NET.Sdk.WindowsDesktop.props
```diff
@@ -13,28 +33,36 @@
...
+              Condition="'$(EnableDefaultPageItems)' != 'false'" />
+      <None Remove="@(ApplicationDefinition)"
+            Condition="'$(EnableDefaultApplicationDefinition)' != 'false'" />
+    -->
+    <None Remove="**/*.xaml"
+          Condition="'$(EnableDefaultApplicationDefinition)' != 'false' And '$(EnableDefaultPageItems)' != 'false'" />
   </ItemGroup>
 
-  <ItemGroup Condition="'$(DisableImplicitFrameworkReferences)' != 'true' And '$(TargetFrameworkIdentifier)' == '.NETCoreApp' And '$(UseWPF)' == 'true' And '$(UseWindowsForms)' == 'true'">
+  <ItemGroup Condition="('$(DisableImplicitFrameworkReferences)' != 'true') And ('$(UseWPF)' == 'true') And ('$(UseWindowsForms)' == 'true') And ('$(TargetFrameworkIdentifier)' == '.NETCoreApp') And ('$(_TargetFrameworkVersionWithoutV)' != '') And ('$(_TargetFrameworkVersionWithoutV)' >= '$(_WindowsDesktopSdkTargetFrameworkVersionFloor)')">
```
dsplaisted: I'd suggest putting the common clauses of the Condition on an `ItemGroup` shared between all 3 implicit framework references, and then having the `UseXXX` conditions on each `FrameworkReference`.  Something like this:

```xml
<ItemGroup Condition="('$(DisableImplicitFrameworkReferences)' != 'true') And
                      ('$(TargetFrameworkIdentifier)' == '.NETCoreApp') And
                      ('$(_TargetFrameworkVersionWithoutV)' != '') And
                      ('$(_TargetFrameworkVersionWithoutV)' >= '$(_WindowsDesktopSdkTargetFrameworkVersionFloor)')">

    <FrameworkReference Include="Microsoft.WindowsDesktop.App" IsImplicitlyDefined="true"
                        Condition="('$(UseWPF)' == 'true') And ('$(UseWindowsForms)' == 'true')"/>

    <FrameworkReference Include="Microsoft.WindowsDesktop.App.WPF" IsImplicitlyDefined="true"
                        Condition="('$(UseWPF)' == 'true') And ('$(UseWindowsForms)' != 'true')"/>

    <FrameworkReference Include="Microsoft.WindowsDesktop.App.WindowsForms" IsImplicitlyDefined="true"
                        Condition="('$(UseWPF)' != 'true') And ('$(UseWindowsForms)' == 'true')"/>
</ItemGroup>
```

packaging/Microsoft.NET.Sdk.WindowsDesktop/targets/Microsoft.NET.Sdk.WindowsDesktop.props
```diff
@@ -2,8 +2,28 @@
...
+    
+    Irrespective of whether '$(TargetFrameworkIdentifier)' is '.NETCoreApp' or '.NETFramework', 
+    the minimum value of $(_TargetFrameworkVersionWithoutV) we will be testing for is '3.0'
+    
+    Note:
+      Please see https://github.com/microsoft/msbuild/issues/3212 for a discussion about the use of 
+      the private $(_TargetFrameworkVersionWithoutV) property - which will likely remain supported and
+      is safe to use here. 
+  -->
+    <_WindowsDesktopSdkTargetFrameworkVersionFloor>3.0</_WindowsDesktopSdkTargetFrameworkVersionFloor>
```
vatsan-madhavan: The experience in SDK style projects for .NET Framework 3.0, 3.5 is a bit clunky. It's certainly not great for .NET Framework 1.0, 2.0 etc. 

There is a ref-assembly nuget package (https://www.nuget.org/packages?q=Microsoft.NETFramework.ReferenceAssemblies), but `ResGen.exe` doesn't work under `dotnet.exe` when building for .NET 3.0, .NET 3.5. I couldn't find targeting pack/ref-assembly nuget for .NET Framework 1.0, 2.0 etc . 

Given this, I think it would be best to limit the capability of WindowsDesktop SDK to .NET Framework 3.0+. 

WinForms is supported from .NET Framework 1.0 onwards - so we'd be reducing a bit of the 'supported profile' (1.0, 2.0 etc. doesn't work today). This shouldn't pose a problem for WPF (which is .NET 3.5+). 


/cc @merriemcgaw, @nguerrera, @dsplaisted

packaging/Microsoft.NET.Sdk.WindowsDesktop/targets/Microsoft.NET.Sdk.WindowsDesktop.props
```diff
@@ -58,12 +86,10 @@
...
-    Note:
-      Please see https://github.com/microsoft/msbuild/issues/3212 for a discussion about the use of 
-      the private $(_TargetFrameworkVersionWithoutV) property - which will likely remain supported and
-      is safe to use here. 
+     Note:
+       These are WPF specific version lower-bounds. Do not use $(_WindowsDesktopSdkTargetFrameworkVersionFloor) in place of '3.0' below.
   -->
-  <ItemGroup Condition="'$(TargetFrameworkIdentifier)' == '.NETFramework' And '$(_TargetFrameworkVersionWithoutV)' != ''">
+  <ItemGroup Condition="('$(TargetFrameworkIdentifier)' == '.NETFramework') And ('$(_TargetFrameworkVersionWithoutV)' != '') And ('$(_TargetFrameworkVersionWithoutV)' >= '$(_WindowsDesktopSdkTargetFrameworkVersionFloor)')">
     <_WpfCommonNetFxReference Include="WindowsBase" Condition="'$(_TargetFrameworkVersionWithoutV)' >= '3.0'" />
```
dsplaisted: The comparisons against 3.0 seem redundant now with the ItemGroup condition clause against `_WindowsDesktopSdkTargetFrameworkVersionFloor`

packaging/Microsoft.NET.Sdk.WindowsDesktop/targets/Microsoft.NET.Sdk.WindowsDesktop.props
```diff
@@ -79,15 +105,15 @@
     <_WpfCommonNetFxReference Include="System.Windows.Controls.Ribbon" Condition="'$(_TargetFrameworkVersionWithoutV)' >= '4.5'" />
   </ItemGroup>
 
-  <ItemGroup Condition=" '$(DisableImplicitFrameworkReferences)' != 'true' and '$(TargetFrameworkIdentifier)' == '.NETFramework' And '$(UseWPF)' == 'true' ">
+  <ItemGroup Condition=" ('$(DisableImplicitFrameworkReferences)' != 'true') And ('$(TargetFrameworkIdentifier)' == '.NETFramework') And ('$(_TargetFrameworkVersionWithoutV)' != '') And ('$(_TargetFrameworkVersionWithoutV)' >= '$(_WindowsDesktopSdkTargetFrameworkVersionFloor)') And ('$(UseWPF)' == 'true') ">
```
dsplaisted: This can also be simplified by putting the common clauses on an `ItemGroup` condition, and the `UseWindowsForms` and `UseWPF` conditions on the `_SDKImplicitReference` elements.

vatsan-madhavan:
> It would be nice to have tests covering these fixes, for example that you can multitarget .NET Core 2.x and 3.0 using the WindowsDesktop SDK. Do you have a way of testing the WindowsDesktop SDK in this repo?

Not yet. I'll sync with you next week to figure out how to seed this repo with some tests based off of dotnet/sdk. 

For now, these are tested locally using a WPF app, WinForms app and a Console App, and using `dotnet`,  `msbuild` and VS. 

```xml
<!-- WPF project -->

  <PropertyGroup>
    <TargetFrameworks>netcoreapp3.0;net40;net45;net46;net452;net472;netcoreapp2.2;net35</TargetFrameworks>
    <UseWPF>true</UseWPF> <!-- change to false to test NETSDK1106 -->
    <UseWindowsForms>false</UseWindowsForms>
    <SuppressNETCoreSdkPreviewMessage>true</SuppressNETCoreSdkPreviewMessage>
  </PropertyGroup>

<PropertyGroup>
  <!--<MSBuildWarningsAsMessages>$(MSBuildWarningsAsMessages);NETSDK1096;NETSDK1105;NETSDK1107</MSBuildWarningsAsMessages>-->
  <!--<MSBuildWarningsAsMessages>$(MSBuildWarningsAsMessages);NETSDK1105</MSBuildWarningsAsMessages>-->
</PropertyGroup>

  <PropertyGroup>
    <FrameworkPathOverride Condition="$(TargetFramework.StartsWith('net3'))">$(MSBuildProgramFiles32)\Reference Assemblies\Microsoft\Framework\.NETFramework\v3.5\Profile\Client</FrameworkPathOverride>
  </PropertyGroup>
...
</Project>
```

```xml
<!-- WinForms project -->

  <PropertyGroup>
    <TargetFrameworks Condition="'$(MSBuildRuntimeType)'!='Core'">netcoreapp3.0;net30;net35;net40;net45;net472</TargetFrameworks>
	<TargetFrameworks Condition="'$(MSBuildRuntimeType)'=='Core'">netcoreapp3.0;net40;net45;net472</TargetFrameworks>
    <UseWindowsForms>true</UseWindowsForms>
  </PropertyGroup>
  
    <PropertyGroup>
    <FrameworkPathOverride Condition="$(TargetFramework.StartsWith('net3'))">$(MSBuildProgramFiles32)\Reference Assemblies\Microsoft\Framework\.NETFramework\v3.5\Profile\Client</FrameworkPathOverride>
  </PropertyGroup>
</Project>
```

```xml
<!-- Console project -->

  <PropertyGroup>
    <OutputType>Exe</OutputType>
    <TargetFramework>netcoreapp3.0</TargetFramework>
    <UseWpf>true</UseWpf>
    <!--<MSBuildWarningsAsMessages>$(MSBuildWarningsAsMessages);NETSDK1096;NETSDK1105;NETSDK1107</MSBuildWarningsAsMessages>-->
  </PropertyGroup>
</Project>
```

packaging/Microsoft.NET.Sdk.WindowsDesktop/targets/Microsoft.NET.Sdk.WindowsDesktop.props
merriemcgaw: I'm totally OK with that @vatsan-madhavan. Technically 1.0 and 1.1 are out of mainstream support.

drewnoakes:
Would it be in scope to warn if property `<UseWinForms>` is defined? I'm sure lots of people would get tripped up on that.

vatsan-madhavan:
> Would it be in scope to warn if property `<UseWinForms>` is defined? I'm sure lots of people would get tripped up on that.

Are you suggesting that we should warn if `<UseWinForms>` is defined - which is an incorrect variant of `UseWindowsForms`? If this is what you have in mind, would you please open an issue and we can address this another PR?

RussKie:
> if `<UseWinForms>` is defined - which is an incorrect variant of `UseWindowsForms`

I'd think `UseWinForms` would be _a better_ (if not correct) setting to use. I speculate here, but based on my experience "WinForms" is significantly more recognisable than "Windows Forms"

drewnoakes:
@vatsan-madhavan yes that's what I meant. Thanks for creating the issue.

@RussKie I agree (I filed issues against the new project dialog to make sure "winforms" actually returned any results) but in code the full name "WindowsForms" is used throughout. This is why I think an additional warning here would help a non-trivial number of people succeed here, especially since the current design of the project file around enable winforms/wpf is so difficult for us on the project system team to tool that customers are likely to need to write this XML by hand from memory.

RussKie:
Thank you for clarifications.

I'd spin it from a customer perspective - how we call it in our codebase is an internal detail and immaterial outside the codebase.
What would make the most sense to our consumers?

dsplaisted:
It's not about internal details of the code base.  The namespace and assembly name is System.Windows.Forms, not something like System.WinForms.  That is visible to customers.  Project templates in VS, documentation, etc. all say Windows Forms instead of WinForms.

stevenbrix:
> It's not about internal details of the code base. The namespace and assembly name is System.Windows.Forms, not something like System.WinForms. That is visible to customers. Project templates in VS, documentation, etc. all say Windows Forms instead of WinForms.

Right, but no one calls WPF by it's actual name - Windows Presentation Foundation. Developers are lazy and don't want to type more then they have to. They are also opinionated though, and no answer is really more correct than the other one.  As a broad generalization of our kind, I think brevity is a close second behind readability/understanding. Obviously something like `<UseWF>` would be a bad idea because it's not immediately clear what "WF" is. Likewise, it would be equally bad to have `<UseWindowsPresentationFoundation>` because it doesn't help the user to understand its purpose anymore than `<UseWPF>` does.

I do agree with your sentiment that everything should be consistent! The command to create a new WinForms app is `dotnet  new winforms`, but in VS you create a new "Windows Forms App (.NET Core)":
image

Personally, I think most people are comfortable and familiar with the term "WinForms" and we should go with this everywhere. This would also make the VS experience more like WPF and say "WinForms App (.NET Core)" and then in the description say the actual name. But this is also just my opinion

image

dsplaisted:
Good point about `dotnet new winforms`.  It looks like we weren't consistent there.

vatsan-madhavan:
> Good point about `dotnet new winforms`. It looks like we weren't consistent there.

please don't change this to `dotnet new windowsforms` /cc @merriemcgaw

OliaG:
I agree that the template `dotnet new winforms` should remain the same.

My personal preference would be <UseWinForms> in the project file as well since it is consistent with <UseWPF>, clear (everyone knows what is WinForms, WindowsForms doesn't add any value), brief (users don't need to type more) and I'd also guesstimate the name "WinForms" is used more often than "Windows Forms". I'd make <UseWinForms> the correct property and prompt users if they are using <UseWindowsForms>. That being said I realize this would break existing WinForms Core apps, but if we want to make this change - the preview is the right time to do so.
 
@merriemcgaw, what do you think?

vatsan-madhavan:
> ...and I'd also guesstimate the name "WinForms" is used more often than "Windows Forms". I'd make the correct property... 
>  the preview is the right time to do so.

We are locking down tomorrow noon for preview 7. 

A change like this requires coordinated changes in dotnet/sdk and dotnet/wpf. I recommend that we pursue an approach that _augments_ the existing design by recognizing `WinForms` as a synonym for `WindowsForms` (assuming we want to pursue a change at all - I'll defer to the WinForms team for this decision). We can pursue such a  change now or in P8 (because it would be a non-breaking change).

davkean:
Lots of "gut feelings" in this thread. "Windows Forms" has been consistently through documentation, training, Visual Studio UI, templates, etc since the beginning of its existence. `dotnet new winforms` is the odd one out here. We should not change usage from "Windows Forms" -> "WinForms" (which would have a fairly significant impact across the product) without data/interviews showing that users are confused by this and this PR is not the right place.

drewnoakes:
Regardless of what's used, I maintain customers who hand edit their project file and use the incorrect version will appreciate guidance that they're not quite doing what's needed. Currently there's no feedback. This is not the pit of success.

dsplaisted:
Strongly agree with David, but also agree that either a warning for using the wrong property name (or possibly just allowing either property name) is a good idea.

davkean:
Yep, sorry didn't mean to say that shouldn't happen RE warning. Accepting both definitions I think would be a mistake, as there are already downstream consumers of this (our capabilities are already driven off this)

RussKie:
Following your argument, we should use `UseWindowsPresentationFoundation` 😉

vatsan-madhavan:
Please move the conversation over to https://github.com/dotnet/winforms/issues/1226

drewnoakes:
If we add a warning for `UseWinForms` we should add one for `UseWindowsPresentationFoundation` too.

PR #1028, opened 2019-06-20
Dev guide fix

dotMorten:
PropertyGroup tag wasn't closing and FrameworkReference must be inside an ItemGroup

### DIFF
--- a/Documentation/developer-guide.md
+++ b/Documentation/developer-guide.md
@@ -112,8 +112,7 @@ installed, we can then simply reference those local binaries directly from the p
     <RuntimeIdentifier>win-x86</RuntimeIdentifier>
   </PropertyGroup>
   <ItemGroup>
-    <Reference Include="$(WpfRepoRoot)\artifacts\packaging\$(WpfConfig)\Microsoft.DotNet.Wpf.GitHub\ref\netcoreapp3.0\*.dll" Private="false" />
-    <ReferenceCopyLocalPaths Include="$(WpfRepoRoot)\artifacts\packaging\$(WpfConfig)\Microsoft.DotNet.Wpf.GitHub\lib\netcoreapp3.0\*.dll" />
+    <Reference Include="$(WpfRepoRoot)\artifacts\packaging\$(WpfConfig)\Microsoft.DotNet.Wpf.GitHub\lib\netcoreapp3.0\*.dll" />
     <ReferenceCopyLocalPaths Include="$(WpfRepoRoot)\artifacts\packaging\$(WpfConfig)\Microsoft.DotNet.Wpf.GitHub\lib\$(RuntimeIdentifier)\*.dll" />
   </ItemGroup>
 ```
--- a/Documentation/developer-guide.md
+++ b/Documentation/developer-guide.md
@@ -157,7 +157,7 @@ If you can build directly from source, and want to test your application against
 ```xml
  <PropertyGroup>
     <MicrosoftWindowsDesktopAppVersion>3.0.0-preview5-27619-18</MicrosoftWindowsDesktopAppVersion>
- <PropertyGroup>
+ </PropertyGroup>
  <FrameworkReference Update="Microsoft.WindowsDesktop.App">
     <TargetingPackVersion>$(MicrosoftWindowsDesktopAppVersion)</TargetingPackVersion>
  </FrameworkReference>
--- a/Documentation/developer-guide.md
+++ b/Documentation/developer-guide.md
@@ -158,9 +158,11 @@ If you can build directly from source, and want to test your application against
  <PropertyGroup>
     <MicrosoftWindowsDesktopAppVersion>3.0.0-preview5-27619-18</MicrosoftWindowsDesktopAppVersion>
  </PropertyGroup>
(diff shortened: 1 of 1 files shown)


stevenbrix:
Can you also add the fixes described here: https://github.com/dotnet/coreclr/issues/25279#issuecomment-504156727

I think that should fix all the trouble we've been having. I was about to do this myself but you beat me to getting a PR out :)

dotMorten:
@stevenbrix Sure I can do that, but it sounded like you weren't completely sure that was the issue?

stevenbrix:
Oh no, that's definitely the fix. I'm not sure if something changed that caused the old method to no longer work, but the issue was that the local assemblies weren't being added to the deps.json. I looked into some binlogs and figured out which target was responsible for writing them and it makes sense why it doesn't work now, just not why it did before :)

dotMorten:
Should we add a comment that you need preview 7? I cannot get this working on preview 6 due to the security access thing.

Here are my steps for using a local build:
Install nightly preview7 (preview 6 release won't work)
from commandline:
```
git clone https://github.com/dotnet/wpf
wpf\build.cmd -pack -c Release
dotnet new wpf -o TestApp
```
Open TestApp.csproj and add the following to the project:

```xml
  <PropertyGroup>
     <!-- Change this value based on where your local repo is located -->
     <WpfRepoRoot>..\wpf</WpfRepoRoot>
     <!-- Change based on which assemblies you build (Release/Debug) -->
     <WpfConfig>Debug</WpfConfig>
     <!-- Publishing a self-contained app ensures our binaries are used. -->
     <SelfContained>true</SelfContained>
    <!-- The runtime identifier needs to match the architecture you built WPF assemblies for. -->
    <RuntimeIdentifier>win-x86</RuntimeIdentifier>
  </PropertyGroup>
  <ItemGroup>
    <Reference Include="$(WpfRepoRoot)\artifacts\packaging\$(WpfConfig)\Microsoft.DotNet.Wpf.GitHub\lib\netcoreapp3.0\*.dll" />
    <ReferenceCopyLocalPaths Include="$(WpfRepoRoot)\artifacts\packaging\$(WpfConfig)\Microsoft.DotNet.Wpf.GitHub\lib\$(RuntimeIdentifier)\*.dll" />
  </ItemGroup>
```
Save project and more importantly .sln file and close VS.
Create a file named nuget.config next to the .sln file you made and add (won't be necessary once Preview 7 ships):
```xml
<configuration>
  <packageSources>
    <add key="dotnet-core" value="https://dotnetfeed.blob.core.windows.net/dotnet-core/index.json" />
    <add key="dotnet-windowsdesktop" value="https://dotnetfeed.blob.core.windows.net/dotnet-windowsdesktop/index.json" />
    <add key="aspnet-aspnetcore" value="https://dotnetfeed.blob.core.windows.net/aspnet-aspnetcore/index.json" />
    <add key="aspnet-aspnetcore-tooling" value="https://dotnetfeed.blob.core.windows.net/aspnet-aspnetcore-tooling/index.json" />
    <add key="aspnet-entityframeworkcore" value="https://dotnetfeed.blob.core.windows.net/aspnet-entityframeworkcore/index.json" />
    <add key="aspnet-extensions" value="https://dotnetfeed.blob.core.windows.net/aspnet-extensions/index.json" />
    <add key="gRPC repository" value="https://grpc.jfrog.io/grpc/api/nuget/v3/grpc-nuget-dev" />
  </packageSources>
</configuration>
```
Reopen your solution and run.

dotMorten:
One more thing I'm finding: PresentationCore.dll is not copied over (or it's overwritten by the installed version), so I had to add this silly post-build step to be able to debug that one:
```
COPY $(WpfRepoRoot)\artifacts\packaging\Debug\Microsoft.DotNet.Wpf.GitHub\lib\netcoreapp3.0\PresentationCore.dll bin\Debug\netcoreapp3.0\win-x86\
```

stevenbrix:
> Should we add a comment that you need preview 7? I cannot get this working on preview 6 due to the security access thing.

Yeah, preview 7 is definitely needed. Although I'm not sure a comment about a specific version number is the right approach, unless there was a sure way to keep it up-to-date. The general philosophy should probably be *"if something isn't working right, first check the sdk version you have installed and what version of `Microsoft.NETCore.App` is along with it. Make sure that is in sync, or close enough, to the version specified in `eng\Version.props`.  Generally, these stay fairly compatible throughout the development life-cycle, however that isn't always the case. If you install a new SDK and your problem consists, please file an issue so it can be investigated.*

stevenbrix:
> One more thing I'm finding: PresentationCore.dll is not copied over (or it's overwritten by the installed version), so I had to add this silly post-build step to be able to debug that one:

What is the assembly version of PresentationCore? It sounds like it isn't overriding the one contained in the `Microsoft.WindowsDesktop.App` you have installed. This should be a fairly simple fix if that is the case, it might be that a certain MSBuild property isn't being set properly.

stevenbrix:
Ok, so I opened up PresentationCore.dll in ILspy and the Assembly Version info looks correct (or at least the same as other WPF assemblies). I'm going to file an issue, since that goes beyond updating this dev-guide. thanks for the updates!

PR #1035, opened 2019-06-20
Removed declarative CAS permission attributes and comments (Part 2)

ojhad:
Part of issue #241

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/DirectWriteForwarder/CPP/DWriteWrapper/TextAnalyzer.cpp
+++ b/src/Microsoft.DotNet.Wpf/src/DirectWriteForwarder/CPP/DWriteWrapper/TextAnalyzer.cpp
@@ -19,7 +19,6 @@ namespace MS { namespace Internal { namespace Text { namespace TextInterface
     /// <SecurityNote>
     /// Critical    - Calls critical AnalyzeExtendedAndItemize overload
     /// </SecurityNote>
-    [SecurityCritical]
     IList<Span^>^ TextAnalyzer::Itemize(
         __in_ecount(length) const WCHAR* text,
         UINT32                           length,
@@ -124,8 +123,6 @@ namespace MS { namespace Internal { namespace Text { namespace TextInterface
     ///               the pointer passed to Itemize() is constructed internally so the call 
     ///               to TextItemizer->Itemize() is safe.
     /// </SecurityNote>
-    [SecurityCritical]
-    [SecurityPermission(SecurityAction::Assert, UnmanagedCode=true)]
     IList<Span^>^ TextAnalyzer::AnalyzeExtendedAndItemize(
         TextItemizer^ textItemizer, 
         __in_ecount(length) const WCHAR *text, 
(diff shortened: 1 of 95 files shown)


vatsan-madhavan:
lgtm. 

did you get a chance to run an IL-diff on the resultant assemblies and the pre-change assemblies?

stevenbrix:
looks good!

ojhad:
> lgtm.
> 
> did you get a chance to run an IL-diff on the resultant assemblies and the pre-change assemblies?

I didn't get a chance to do that yet, but I will run it before merging this PR

src/Microsoft.DotNet.Wpf/src/System.Xaml/GlobalSuppressions.cs
```diff
@@ -124,8 +124,6 @@
 #endregion
 
 #region Microsoft.Security Suppressions
-[module: SuppressMessage("Microsoft.Security", "CA2103:ReviewImperativeSecurity", Scope = "member", Target = "System.Xaml.Permissions.XamlLoadPermission.#Copy()", Justification = "Reviewed by senior CLR security developer.")]
 [module: SuppressMessage("Microsoft.Security", "CA2103:ReviewImperativeSecurity", Scope = "member", Target = "MS.Internal.Xaml.Runtime.DynamicMethodRuntime.#.ctor(MS.Internal.Xaml.Runtime.XamlRuntimeSettings,System.Xaml.XamlSchemaContext,System.Xaml.Permissions.XamlAccessLevel)", Justification = "Reviewed by Microsoft.")]
```
rladuca: Why the others and not the 2 leftover suppressions?

src/Microsoft.DotNet.Wpf/src/DirectWriteForwarder/CPP/DWriteWrapper/TextAnalyzer.cpp
```diff
@@ -297,7 +277,6 @@ namespace MS { namespace Internal { namespace Text { namespace TextInterface
     }
 
 // Warning 4714 (__forceinline function not inlined)
-// is expected here because TextAnalyzer::GetGlyphs is marked with [SecurityCritical]
```
rladuca: This comment is strange now without this line.

src/Microsoft.DotNet.Wpf/src/DirectWriteForwarder/CPP/DWriteWrapper/TextAnalyzer.cpp
```diff
@@ -276,19 +276,6 @@ namespace MS { namespace Internal { namespace Text { namespace TextInterface
         }
     }
 
-// Warning 4714 (__forceinline function not inlined)
```
rladuca: I'm not so sure about suddenly inlining a set of function that were not inlined previously.  I understand the original reasoning for them not being inlined has gone away, but we need to be 100% certain there are no adverse affects from doing this.

Can we hold off on this, file an issue and do a more thorough examination of these changes with a larger test pass later?

rladuca: I think I misunderstood this, it's just disabling a warning, not enabling inlining, resolving.

vatsan-madhavan:
I did a quick IL-diff scan and the changes look ok to me. There is some compiler generated code in DirectWriteForwarder that uses `SecurityCriticalAttribute`  (for e.g., global method `_atexit_m`). We need to figure out a way to resolve these dependencies.

ojhad:
> I did a quick IL-diff scan and the changes look ok to me. There is some compiler generated code in DirectWriteForwarder that uses `SecurityCriticalAttribute` (for e.g., global method `_atexit_m`). We need to figure out a way to resolve these dependencies.

Will take a look and include this in the last PR

PR #969, opened 2019-06-14
Remove internal permissions types and most of SecurityHelper

ojhad:
Opened an [internal PR](https://dev.azure.com/dnceng/internal/_git/dotnet-wpf-int/pullrequest/1759?_a=overview) which should be merged first in order to avoid build errors.

Part if issue #241 

- Removed dependence on System.Security.Permission.dll for internal types that extend from System.Security.Permission.dll types
- Deleted internal permission types
- Deleted most of `SecurityHelper`

PR #1038, opened 2019-06-21
Added missing null check to uri in MediaPlayerState

### LINKED ISSUE #722: MediaPlayer cannot play back audio files from secure HTTPS sources
* .NET Core Version: 3.0.0-preview6-27720-09
* Windows version: (`winver`) 1903
 
 **Problem description:**
When pointing to a https source on MediaPlayer, you'll get a NullReferenceException. This doesn't occur with http.
However this doesn't happen in click-once deployments, because that's what it uses to check if the secure connection is used or not, but since most scenarios aren't click-once this fails because
 null is returned. 

It's clear the code here never assumed this wouldn't be used in a click-once scenario (appDeploymentUri us null so crashes on line 987):
https://github.com/dotnet/wpf/blob/9c3bd0ec0976d11b1b19a2125a30431afdb3e53d/src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Media/MediaPlayerState.cs#L979-L992

 **Actual behavior:** 
image

<details> 
<summary>StackTrace:</summary>
```
   at System.Windows.Media.MediaPlayerState.OpenMedia(Uri source)
   at System.Windows.Media.MediaPlayerState.SetSource(Uri source)
   at System.Windows.Media.MediaPlayerState.Open(Uri source)
   at System.Windows.Media.MediaPlayer.Open(Uri source)
   at test.MainWindow.Button_Click(Object sender, RoutedEventArgs e) in e:\GitHub\dotnet\test\MainWindow.xaml.cs:line 33
   at System.Windows.EventRoute.InvokeHandlersImpl(Object source, RoutedEventArgs args, Boolean reRaised)
   at System.Windows.UIElement.RaiseEvent(RoutedEventArgs e)
   at System.Windows.Controls.Primitives.ButtonBase.OnClick()
   at System.Windows.Controls.Button.OnClick()
   at System.Windows.Controls.Primitives.ButtonBase.OnMouseLeftButtonUp(MouseButtonEventArgs e)
   at System.Windows.UIElement.OnMouseLeftButtonUpThunk(Object sender, MouseButtonEventArgs e)
   at System.Windows.Input.MouseButtonEventArgs.InvokeEventHandler(Delegate genericHandler, Object genericTarget)
   at System.Windows.RoutedEventArgs.InvokeHandler(Delegate handler, Object target)
   at System.Windows.EventRoute.InvokeHandlersImpl(Object source, RoutedEventArgs args, Boolean reRaised)
   at System.Windows.UIElement.ReRaiseEventAs(DependencyObject sender, RoutedEventArgs args, RoutedEvent newEvent)
   at System.Windows.UIElement.OnMouseUpThunk(Object sender, MouseButtonEventArgs e)
   at System.Windows.Input.MouseButtonEventArgs.InvokeEventHandler(Delegate genericHandler, Object genericTarget)
   at System.Windows.RoutedEventArgs.InvokeHandler(Delegate handler, Object target)
   at System.Windows.EventRoute.InvokeHandlersImpl(Object source, RoutedEventArgs args, Boolean reRaised)
   at System.Windows.UIElement.RaiseTrustedEvent(RoutedEventArgs args)
   at System.Windows.Input.InputManager.ProcessStagingArea()
   at System.Windows.Input.InputProviderSite.ReportInput(InputReport inputReport)
   at System.Windows.Interop.HwndMouseInputProvider.ReportInput(IntPtr hwnd, InputMode mode, Int32 timestamp, RawMouseActions actions, Int32 x, Int32 y, Int32 wheel)
   at System.Windows.Interop.HwndMouseInputProvider.FilterMessage(IntPtr hwnd, WindowMessage msg, IntPtr wParam, IntPtr lParam, Boolean& handled)
   at System.Windows.Interop.HwndSource.InputFilterMessage(IntPtr hwnd, Int32 msg, IntPtr wParam, IntPtr lParam, Boolean& handled)
   at System.Windows.Threading.Dispatcher.PushFrame(DispatcherFrame frame)
   at System.Windows.Threading.Dispatcher.Run()
   at System.Windows.Application.RunDispatcher(Object ignore)
   at System.Windows.Application.RunInternal(Window window)
   at System.Windows.Application.Run(Window window)
   at System.Windows.Application.Run()
   at test.App.Main()
```
</details> 

 **Expected behavior:**
Audio plays
 
 **Minimal repro:**
 
```cs
MediaPlayer player = new MediaPlayer();
player.Open(new Uri("https://www.thesoundarchive.com/email/youGotmail.wav"));
```
Remove "s" from https, and the audio will play fine.

dotMorten:
Adds null check for when playing streams from https (would throw NullReferenceException otherwise). 
Fixes #722

This check matches what is already in place here:
https://github.com/dotnet/wpf/blob/master/src/Microsoft.DotNet.Wpf/src/Shared/MS/Internal/SecurityHelper.cs#L302

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Media/MediaPlayerState.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Media/MediaPlayerState.cs
@@ -877,7 +877,7 @@ namespace System.Windows.Media
                 {
                     // target is HTTPS. Then, elevate ONLY if we are NOT coming from HTTPS (=XDomain HTTPS app to HTTPS media disallowed)
                     Uri appDeploymentUri = SecurityHelper.ExtractUriForClickOnceDeployedApp();
-                    if (!SecurityHelper.AreStringTypesEqual(appDeploymentUri.Scheme, Uri.UriSchemeHttps))
+                    if (appDeploymentUri != null && !SecurityHelper.AreStringTypesEqual(appDeploymentUri.Scheme, Uri.UriSchemeHttps))
                     {
                         new WebPermission(NetworkAccess.Connect, BindUriHelper.UriToString(uriToOpen)).Assert();
                         elevated = true;


src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Media/MediaPlayerState.cs
```diff
@@ -877,7 +877,7 @@ private void OpenMedia(Uri source)
                 {
                     // target is HTTPS. Then, elevate ONLY if we are NOT coming from HTTPS (=XDomain HTTPS app to HTTPS media disallowed)
                     Uri appDeploymentUri = SecurityHelper.ExtractUriForClickOnceDeployedApp();
-                    if (!SecurityHelper.AreStringTypesEqual(appDeploymentUri.Scheme, Uri.UriSchemeHttps))
+                    if (appDeploymentUri != null && !SecurityHelper.AreStringTypesEqual(appDeploymentUri.Scheme, Uri.UriSchemeHttps))
```
vatsan-madhavan: In .NET Core, it seems to me that the method can probably boil down to something like shown below, and `DemandPermissions(Uri)` just gets deleted. /cc @ojhad , @rladuca thoughts? 

I didn't walk through `MILMedia.Open` carefully to see whether `null` would be always ok, but I'd imagine that would be the case... 

```C#
        private void OpenMedia(Uri source)
        {
            if (source != null && source.IsAbsoluteUri && source.Scheme == PackUriHelper.UriSchemePack)
            {
                try
                {
                    source = BaseUriHelper.ConvertPackUriToAbsoluteExternallyVisibleUri(source);
                }
                catch (InvalidOperationException)
                {
                    source = null;
                    _mediaEventsHelper.RaiseMediaFailed(new System.NotSupportedException(SR.Get(SRID.Media_PackURIsAreNotSupported, null)));
                }
            }

            // Setting a null source effectively disconects the MediaElement.
            if (source != null)
            {

                // get the base directory of the application
                Uri appBase = new Uri(AppDomain.CurrentDomain.BaseDirectory);
                // this extracts the URI to open
                Uri uriToOpen = ResolveUri(source, appBase);				
            }

            HRESULT.Check(MILMedia.Open(_nativeMedia, null));
        }
```

vatsan-madhavan: For the change as proposed, why not simply this? It just needs to fall-through into evaluating `elevated=true` one way or another. 

```C#
!SecurityHelper.AreStringTypesEqual(appDeploymentUri?.Scheme, Uri?.UriSchemeHttps)
```

dotMorten: I find `Uri?.UriSchemeHttps` odd, as it would never be null. I like the above better as it clearly reads "only check this if there's an appDeploymentUri"

stevenbrix:
Does this also fix the other scenarios called out in the issue:

>It's not only audio files, it's true for video too. And it's the same with System.Windows.Controls.MediaElement control.
This crashes has been around for years in WPF...

dotMorten:
> Does this also fix the other scenarios called out in the issue:

While I didn't explicitly test that, I didn't' find anywhere else where this code was used, and I believe MediaElement is built on top of the same thing, and uses the same MediaPlayerState, so it's quite likely it fixes that too.
With everything going https only, it's hard to use these APIs without https support.

stevenbrix:
Oh I'm not trying to question the validity of the issue, this should definitely be fixed. I just want to make sure we fix both known issues. It does look like `MediaElement` uses `MediaPlayer` under the covers (thankfully 😊 ), so I'm sure this scenario is fixed.

src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Media/MediaPlayerState.cs
rladuca: @ojhad Is this still around after your security fixups?

ojhad: All this code is removed as part of the cleanup in this PR: https://github.com/dotnet/wpf/pull/969
So I don't think this should be an issue anymore.

dotMorten: Looks like it! I'll close this if 969 gets merged.

dotMorten:
Closing now that #969 removed the problematic code.

PR #1056, opened 2019-06-21
Restore PackageReferences in temporary assembly projects to import build extensions

ryalanms:
**The temporary project is not getting props and targets files (build extension files) from PackageReferences defined in their parent project.**  This is causing failures in referenced packages that rely on build extensions (e.g., Google Protocol Buffers) to perform pre-compile tasks, such as code generation.  

Several options were considered, including consuming the generated props and targets that include the PackageReferences props/targets from the outer parent project, as the parent project `Restore` has already run and includes identical PackageReferences. However, we opted to run `Restore` directly on the temporary project instead, which allowed a fix without changes to external build files (other than a minor change to the Arcade SDK).  There are now two `Restore` operations: one for the parent project and one for the temporary project.  Note that `Restore` will not re-download packages but will generate the assets file, targets file, and props file.  These files include imports for build extensions in the PackageReferences.

These tasks could easily be condensed in to a single parameterized task, but breaking the steps in to three made a logs easier to read and improved debugging.  There are now three tasks: 

1. `CreateTemporaryTargetAssemblyProject`
2. `RestoreTemporaryTargetAssemblyProject` (which calls `RunProjectBuildTarget` with `Restore`)
3. `GenerateTemporaryTargetAssembly`

The first task creates a temporary project with random component in the name from the outer parent project.  `RestoreTemporaryTargetAssemblyProject` consumes the temporary project and runs `Restore`, generating assets, .props, and .targets files that include imports to any build extensions inside PackageReferences.  These files are then consumed in `GenerateTemporaryTargetAssembly` as part of the
normal build process for the temporary project.

Note that there is an Arcade SDK targets file change that needs to be coordinated with this change.  Workaround.props needs to be modified to remove a temporary workaround for the lack of PackageReference support to prevent duplicate import errors.  https://github.com/dotnet/arcade/pull/3127

(APICompat was re-baselined to account for the public API surface changes in PresentationBuildTasks.)

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/PresentationBuildTasks/Microsoft.WinFx.targets
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationBuildTasks/Microsoft.WinFx.targets
@@ -20,6 +20,8 @@
   <UsingTask TaskName="Microsoft.Build.Tasks.Windows.FileClassifier" AssemblyFile="$(_PresentationBuildTasksAssembly)" />
   <UsingTask TaskName="Microsoft.Build.Tasks.Windows.MarkupCompilePass2" AssemblyFile="$(_PresentationBuildTasksAssembly)" />
   <UsingTask TaskName="Microsoft.Build.Tasks.Windows.GenerateTemporaryTargetAssembly" AssemblyFile="$(_PresentationBuildTasksAssembly)" />
+  <UsingTask TaskName="Microsoft.Build.Tasks.Windows.CreateTemporaryTargetAssemblyProject" AssemblyFile="$(_PresentationBuildTasksAssembly)" />
+  <UsingTask TaskName="Microsoft.Build.Tasks.Windows.RunProjectBuildTarget" AssemblyFile="$(_PresentationBuildTasksAssembly)" />
   <UsingTask TaskName="Microsoft.Build.Tasks.Windows.MergeLocalizationDirectives" AssemblyFile="$(_PresentationBuildTasksAssembly)" />
 
 
@@ -380,6 +382,8 @@
   <PropertyGroup>
 
        <MarkupCompilePass2ForMainAssemblyDependsOn>
+                     CreateTemporaryTargetAssemblyProject;
+                     RestoreTemporaryTargetAssemblyProject;
                      GenerateTemporaryTargetAssembly;
                      MarkupCompilePass2;
                      AfterMarkupCompilePass2;
@@ -400,6 +404,57 @@
 
   </Target>
 
+  <!--
+
+                ==========================================
+                  CreateTemporaryTargetAssemblyProject
+                ==========================================
+
+                Name : CreateTemporaryTargetAssemblyProject
+
+  -->
+
+   <Target Name="CreateTemporaryTargetAssemblyProject"
+               Condition="'$(_RequireMCPass2ForMainAssembly)' == 'true' " >
+
+       <CreateTemporaryTargetAssemblyProject 
+                CurrentProject="$(MSBuildProjectFullPath)"
+                GeneratedCodeFiles="@(_GeneratedCodeFiles)"
+                CompileTypeName="Compile"
+                ReferencePath="@(ReferencePath)"
+                ReferencePathTypeName="ReferencePath" >
+           <Output TaskParameter="TemporaryTargetAssemblyProjectName" PropertyName="_TemporaryTargetAssemblyProjectName"/>
+       </CreateTemporaryTargetAssemblyProject>
+
+      <Message Text="GenerateTemporaryTargetAssembyProject project is $(_TemporaryTargetAssemblyProjectName)" />
+
+  </Target>
+
+  <!--
+
+                ==========================================
+                   RestoreTemporaryTargetAssemblyProject
+                ==========================================
+
+                Name : RestoreTemporaryTargetAssemblyProject
+
+  -->
+
+   <Target Name="RestoreTemporaryTargetAssemblyProject"
+               Condition="'$(_RequireMCPass2ForMainAssembly)' == 'true' " >
+
+      <Message Text="RestoreTemporaryTargetAssemblyProject project is $(_TemporaryTargetAssemblyProjectName)" Condition="'$(MSBuildTargetsVerbose)' == 'true'" />
+
+      <RunProjectBuildTarget 
+          ProjectName="$(_TemporaryTargetAssemblyProjectName)"
+          BuildTarget="Restore"
+          />
+
+       <CreateItem Include="$(IntermediateOutputPath)$(TargetFileName)" >
+               <Output TaskParameter="Include" ItemName="AssemblyForLocalTypeReference" />
+       </CreateItem>
+
+  </Target>
 
   <!--
 
@@ -418,13 +473,9 @@
 
        <GenerateTemporaryTargetAssembly
                 CurrentProject="$(MSBuildProjectFullPath)"
-                MSBuildBinPath="$(MSBuildBinPath)"
-                ReferencePathTypeName="ReferencePath"
-                CompileTypeName="Compile"
-                GeneratedCodeFiles="@(_GeneratedCodeFiles)"
-                ReferencePath="@(ReferencePath)"
                 IntermediateOutputPath="$(IntermediateOutputPath)"
                 AssemblyName="$(AssemblyName)"
+                TemporaryTargetAssemblyProjectName="$(_TemporaryTargetAssemblyProjectName)"
                 CompileTargetName="$(_CompileTargetNameForLocalType)"
                 GenerateTemporaryTargetAssemblyDebuggingInformation="$(GenerateTemporaryTargetAssemblyDebuggingInformation)"
                  >
--- a/src/Microsoft.DotNet.Wpf/src/PresentationBuildTasks/Microsoft/Build/Tasks/Windows/CreateTemporaryTargetAssemblyProject.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationBuildTasks/Microsoft/Build/Tasks/Windows/CreateTemporaryTargetAssemblyProject.cs
@@ -73,8 +73,6 @@ namespace Microsoft.Build.Tasks.Windows
         [System.Diagnostics.CodeAnalysis.SuppressMessage("Microsoft.Design", "CA1031:DoNotCatchGeneralExceptionTypes")]
         public override bool Execute()
         {
-            System.Diagnostics.Debugger.Launch();
-
             bool retValue = true;
 
             // Verification
--- a/src/Microsoft.DotNet.Wpf/src/PresentationBuildTasks/Microsoft/Build/Tasks/Windows/CreateTemporaryTargetAssemblyProject.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationBuildTasks/Microsoft/Build/Tasks/Windows/CreateTemporaryTargetAssemblyProject.cs
@@ -1,0 +1,391 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System;
+using System.IO;
+using System.Collections;
+
+using System.Globalization;
+using System.Diagnostics;
+using System.Reflection;
+using System.Resources;
+using System.Xml;
+
+using Microsoft.Build.Framework;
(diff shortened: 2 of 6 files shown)


walterlv:
@ryalanms Thanks for fixing this issue.
I have some NuGet packages which extend the building so should I have to check target building SDK version to determine whether should I hack to fix or use your fix?

src/Microsoft.DotNet.Wpf/src/PresentationBuildTasks/Microsoft/Build/Tasks/Windows/CreateTemporaryTargetAssemblyProject.cs
```diff
@@ -0,0 +1,391 @@
...
+
+        /// <summary>
+        /// ITask Execute method
+        /// </summary>
+        /// <returns></returns>
+        /// <remarks>Catching all exceptions in this method is appropriate - it will allow the build process to resume if possible after logging errors</remarks>
+        [System.Diagnostics.CodeAnalysis.SuppressMessage("Microsoft.Design", "CA1031:DoNotCatchGeneralExceptionTypes")]
+        public override bool Execute()
+        {
+            System.Diagnostics.Debugger.Launch();
```
walterlv: You've written a debugging code here. Maybe you should remove it in the product code.

src/Microsoft.DotNet.Wpf/src/PresentationBuildTasks/Microsoft/Build/Tasks/Windows/GenerateTemporaryTargetAssembly.cs
```diff
@@ -89,83 +89,34 @@ public GenerateTemporaryTargetAssembly()
         [System.Diagnostics.CodeAnalysis.SuppressMessage("Microsoft.Design", "CA1031:DoNotCatchGeneralExceptionTypes")]
         public override bool Execute()
         {
+            System.Diagnostics.Debugger.Launch();
```
walterlv: Another debug-only code here.

src/Microsoft.DotNet.Wpf/src/PresentationBuildTasks/Microsoft/Build/Tasks/Windows/RunProjectBuildTarget.cs
```diff
@@ -0,0 +1,90 @@
...
+
+        /// <summary>
+        /// ITask Execute method
+        /// </summary>
+        /// <returns></returns>
+        /// <remarks>Catching all exceptions in this method is appropriate - it will allow the build process to resume if possible after logging errors</remarks>
+        [System.Diagnostics.CodeAnalysis.SuppressMessage("Microsoft.Design", "CA1031:DoNotCatchGeneralExceptionTypes")]
+        public override bool Execute()
+        {
+            System.Diagnostics.Debugger.Launch();
```
walterlv: Another debug-only code here.

ryalanms:
Thanks for reviewing, @walterlv.  After this fix is in, there shouldn't be any need to use custom code to get the build extensions in PackageReferences to work in user code.

ryalanms:
Note that this will fail to build with duplicate imports until the Arcade SDK change is merged and picked up.

rladuca:
> Note that this will fail to build with duplicate imports until the Arcade SDK change is merged and picked up.

Arcade just merged the change.

src/Microsoft.DotNet.Wpf/src/PresentationBuildTasks/Microsoft.WinFx.targets
```diff
@@ -400,6 +404,57 @@
...
+  -->
+
+   <Target Name="RestoreTemporaryTargetAssemblyProject"
+               Condition="'$(_RequireMCPass2ForMainAssembly)' == 'true' " >
+
+      <Message Text="RestoreTemporaryTargetAssemblyProject project is $(_TemporaryTargetAssemblyProjectName)" Condition="'$(MSBuildTargetsVerbose)' == 'true'" />
+
+      <RunProjectBuildTarget 
+          ProjectName="$(_TemporaryTargetAssemblyProjectName)"
+          BuildTarget="Restore"
```
vatsan-madhavan: Use a pattern similar to that in   `<Target Name="_CompileTemporaryAssembly"  DependsOnTargets="BuildOnlySettings;ResolveKeySource;CoreCompile" />` and `       <_CompileTargetNameForLocalType Condition="'$(_CompileTargetNameForLocalType)' == ''">_CompileTemporaryAssembly</_CompileTargetNameForLocalType>` used in `GenerateTemporaryTargetAssembly`

src/Microsoft.DotNet.Wpf/src/PresentationBuildTasks/Microsoft/Build/Tasks/Windows/CreateTemporaryTargetAssemblyProject.cs
```diff
@@ -0,0 +1,389 @@
...
+using System.Reflection;
+using System.Resources;
+using System.Xml;
+
+using Microsoft.Build.Framework;
+using Microsoft.Build.Utilities;
+using MS.Utility;
+using MS.Internal.Tasks;
+
+// Since we disable PreSharp warnings in this file, PreSharp warning is unknown to C# compiler.
```
vatsan-madhavan: I don't think we need this and related suppressions - we don't really use PreSharp any more

src/Microsoft.DotNet.Wpf/src/PresentationBuildTasks/Microsoft/Build/Tasks/Windows/CreateTemporaryTargetAssemblyProject.cs
```diff
@@ -0,0 +1,389 @@
...
+        
+        #endregion Constructors
+
+        //------------------------------------------------------
+        //
+        //  Public Methods
+        //
+        //------------------------------------------------------
+
+        #region Public Methods
```
vatsan-madhavan: Don't use regions in new code.

src/Microsoft.DotNet.Wpf/src/PresentationBuildTasks/Microsoft/Build/Tasks/Windows/CreateTemporaryTargetAssemblyProject.cs
```diff
@@ -0,0 +1,389 @@
...
+        private void RemoveItemsByName(XmlDocument xmlProjectDoc, string sItemName)
+        {
+            if (xmlProjectDoc == null || String.IsNullOrEmpty(sItemName))
+            {
+                // When the parameters are not valid, simply return it, instead of throwing exceptions.
+                return;
+            }
+
+            //
+            // The project file format is always like below:
```
vatsan-madhavan: This comment is no longer accurate. For e.g., the SDK style projects no longer have `xmlns=""` in the `Project` element.

src/Microsoft.DotNet.Wpf/src/PresentationBuildTasks/Microsoft/Build/Tasks/Windows/CreateTemporaryTargetAssemblyProject.cs
```diff
@@ -0,0 +1,389 @@
...
+            //
+            //
+            // The order of children nodes under Project Root element is random
+            //
+
+            XmlNode root = xmlProjectDoc.DocumentElement;
+
+            if (root.HasChildNodes == false)
+            {
+                // If there is no child element in this project file, just return immediatelly.
```
vatsan-madhavan: _immediately_ - type

src/Microsoft.DotNet.Wpf/src/PresentationBuildTasks/Microsoft/Build/Tasks/Windows/CreateTemporaryTargetAssemblyProject.cs
```diff
@@ -0,0 +1,389 @@
...
+                XmlElement nodeGroup = root.ChildNodes[i] as XmlElement;
+
+                if (nodeGroup != null && String.Compare(nodeGroup.Name, ITEMGROUP_NAME, StringComparison.OrdinalIgnoreCase) == 0)
+                {
+                    //
+                    // This is ItemGroup element.
+                    //
+                    if (nodeGroup.HasChildNodes)
+                    {
+                        ArrayList itemToRemove = new ArrayList();
```
vatsan-madhavan: Don't use `ArrayList` - instead use `List<object>`

src/Microsoft.DotNet.Wpf/src/PresentationBuildTasks/Microsoft/Build/Tasks/Windows/CreateTemporaryTargetAssemblyProject.cs
```diff
@@ -0,0 +1,389 @@
...
+                            }
+                        }
+                    }
+
+                    //
+                    // Removed all the items with given name from this item group.
+                    //
+                    // Continue the loop for the next ItemGroup.
+                }
+
```
vatsan-madhavan: excess blank line

src/Microsoft.DotNet.Wpf/src/PresentationBuildTasks/Microsoft/Build/Tasks/Windows/CreateTemporaryTargetAssemblyProject.cs
```diff
@@ -0,0 +1,389 @@
...
+            }   // end of "for i" statement.
+
```
vatsan-madhavan: excess blank line

src/Microsoft.DotNet.Wpf/src/PresentationBuildTasks/Microsoft/Build/Tasks/Windows/RunProjectBuildTarget.cs
```diff
@@ -0,0 +1,88 @@
...
+
+namespace Microsoft.Build.Tasks.Windows
+{
+    #region RunProjectBuildTarget Task class
+
+    /// <summary>
+    /// </summary>
+    public sealed class RunProjectBuildTarget : Task
+    {
+        #region Constructors
```
vatsan-madhavan: Don't use regions in new code

src/Microsoft.DotNet.Wpf/src/PresentationBuildTasks/Microsoft/Build/Tasks/Windows/RunProjectBuildTarget.cs
```diff
@@ -0,0 +1,88 @@
...
+using System.Resources;
+using System.Xml;
+
+using Microsoft.Build.Framework;
+using Microsoft.Build.Utilities;
+
+using MS.Utility;
+using MS.Internal.Tasks;
+
+// Since we disable PreSharp warnings in this file, PreSharp warning is unknown to C# compiler.
```
vatsan-madhavan: We don't use PreSharp - remove comment and related warning suppressions

vatsan-madhavan:
The approach and the fix looks good to me overall. Given the overall level of churn, I'm hesitant to merge this so close to _Preview 7_ (check-ins close at noon tomorrow for WPF). 

I don't see clear reasons why this change should cause regressions. The approach is sound and I think the change looks good to me overall. That said, the general complexity of the fix makes me think that it is likely to bring with it a slightly higher possibility of bugs. I'd rather fork off _master_ to _preview/3.0_ (which happens on in two days on 6/26), and and _then_ merge this into _master_ for Preview 8 (so we would get early feedback on P8 builds). 

I've spoken with @ryalanms and he agrees that waiting for 2 days is reasonable. 

Marking as *NO MERGE* for now.

vatsan-madhavan:
Removing *NO MERGE* since #1100 is completed.

nguerrera:
Are you initiating a restore during the build? That is not allowed. It breaks the promise of build --no-restore, and will also mess with VS that handles restore differently than restore target.

nguerrera:
cc @davkean

nguerrera:
> Note that Restore will not re-download packages but will generate the assets file, targets file, and props file.

This seems risky. If there's any evaluation change in generated project that ends up changing a PackageReference it could still hit the network.

ryalanms:
Closing this, since we ended up taking a different implementation.  Thanks.

PR #1057, opened 2019-06-22
Convert C# Reference Assembly Generation to Use GenAPI

### LINKED ISSUE #932: Convert C# Reference Assembly Generation to Use GenAPI
Currently, C# assemblies are building reference assemblies via [ProduceReferenceAssembly](https://docs.microsoft.com/en-us/visualstudio/msbuild/common-msbuild-project-properties?view=vs-2019) (corresponding to the CSC flag [-refout](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/compiler-options/refout-compiler-option)).

This causes several issues:

- Reference assemblies contain internal types due to WPF's use of `InternalsVisibleTo` in many of its assemblies.
- Due to this, several reference assemblies may have dangling type references as we move toward the same reference assembly surface as .NET Framework.  This was also the case in .NET Framework for some reference assemblies, but this wasn't as strictly enforced.

Currently, the known dangling references are in the `PresentationCore` and `PresentationFramework` reference assemblies, the later only when `PresentationUI`'s reference assembly is removed.

To fix these, we want to switch to using the [GenAPI tool](https://github.com/dotnet/arcade/tree/a1805103791e43031355e11c0d037bca803a9593/src/Microsoft.DotNet.GenAPI) so that we can tailor the content of the reference assemblies to only contain necessary API surface area.  

Our builds directly use `ProjectReference` for needed assemblies that are in the same repository and will alter these for those that are not available in the same repository (via [WpfProjectReference.targets](https://github.com/dotnet/wpf/blob/master/eng/WpfArcadeSdk/tools/WpfProjectReference.targets)).  We likely will have to update this mechanic if internal building assemblies require internals from the reference assemblies.

- [x] Call GenAPI to generate `PresentationCore`'s reference assembly
- [x] Trim output to only needed types
- [x] Do the same for the rest of the publicly building C# assemblies
- [ ] Test package ingestion against dotnet-wpf-int to validate [WpfProjectReference.targets](https://github.com/dotnet/wpf/blob/master/eng/WpfArcadeSdk/tools/WpfProjectReference.targets)
- [ ] Fix issues as needed
- [ ] Apply GenAPI changes to dotnet-wpf-int C# assemblies

rladuca:
Fixes #932 

What work is currently in this PR:

### GitHub
* For all shipping C# reference assemblies built on GitHub generate their code from GenAPI
* Fixup any issues that result from GenAPI code gen (see ManualFixups.txt in this PR)
* Baseline all ApiCompat issues (there is a feedback loop here with fixups for a few commits)

### dotnet-wpf-int
* Switch ProjectReference substitution for dotnet-wpf-int to use runtime assemblies
* For all shipping C# reference assemblies built on dotnet-wpf-int generate their code from GenAPI
* Fixup any issues that result from GenAPI code gen (see ManualFixups.txt in this PR)
* Baseline all ApiCompat issues (there is a feedback loop here with fixups for a few commits)

### Testing Work:
- [x] Build our internal test corpus against new ref assemblies
- [x] Build our samples repo against new ref assemblies
- [x] Build NuGet Package Explorer against new ref assemblies
- [x] Verify there are no dangling references to assemblies not part of the reference pack 

### What you should center your review around

* Engineering system changes (in WpfArcadeSdk, ignore Arcade DARC flow)
* ApiCompat results and baselines
* Exclusion lists for GenApi (see GlobalApiExclusions.txt and GlobalAttrExclusions.txt)

These are what eventually feed into the reference assemblies and dictate the public surface area.

### Links

dotnet-wpf-int PR: https://dev.azure.com/dnceng/internal/_git/dotnet-wpf-int/pullrequest/1772?_a=overview

Upstream notification issue: https://github.com/dotnet/core-setup/issues/6901


eng/WpfArcadeSdk/tools/GenApi/GlobalApiExclusions.txt
```diff
@@ -0,0 +1,23 @@
+// NOTE: 
+//   The following commented out removals are due to issues with generating derived classes.
```
ryalanms: Is APICompat able to use the same baselines for ref and runtime assemblies?

rladuca: Runtime assemblies were never baselined against .NET 4.8.  We picked up the compiler generated ref output and baselined that.  The filename just had the runtime assembly name.  In any case I need to clean those files and remove the runtime assemblies from the list of apicompat runs.

rladuca: I actually found some bugs in the ApiCompat implementation and I'm working them out right now.

src/Microsoft.DotNet.Wpf/ApiCompat/Baselines/System.Xaml-ref.baseline.txt
```diff
@@ -0,0 +1,27 @@
+Compat issues with assembly System.Xaml:
+CannotChangeAttribute : Attribute 'System.AttributeUsageAttribute' on 'System.Windows.Markup.AcceptedMarkupExtensionExpressionTypeAttribute' changed from '[AttributeUsageAttribute(4, AllowMultiple=true, Inherited=true)]' in the contract to '[AttributeUsageAttribute(AttributeTargets.Class, AllowMultiple=true, Inherited=true)]' in the implementation.
+CannotChangeAttribute : Attribute 'System.AttributeUsageAttribute' on 'System.Windows.Markup.AmbientAttribute' changed from '[AttributeUsageAttribute(196, Inherited=true)]' in the contract to '[AttributeUsageAttribute(AttributeTargets.Class | AttributeTargets.Method | AttributeTargets.Property, Inherited=true)]' in the implementation.
+CannotChangeAttribute : Attribute 'System.AttributeUsageAttribute' on 'System.Windows.Markup.ConstructorArgumentAttribute' changed from '[AttributeUsageAttribute(128, AllowMultiple=false, Inherited=false)]' in the contract to '[AttributeUsageAttribute(AttributeTargets.Property, AllowMultiple=false, Inherited=false)]' in the implementation.
```
rladuca: These sorts of errors are due to GenAPI converting the enum directly to the underlying type (usually int).

Should we be fixing these up to the actual enum type in the reference assembly instead of baselining them away?  The compiler doesn't seem to care much (apart from a few instances).

rladuca:
.NET Framework reference assemblies contain resources (BAML/strings resx, etc).  Should we also include these in .NET Core?

Documentation/gen-api.md
```diff
@@ -0,0 +1,26 @@
...
+GenAPI is run only on-demand.  In the event that a change to a runtime assembly creates new public surface area, a developer will see an [ApiCompat](api-compat.md) error between the reference assembly and the runtime assembly.  In order to address this, the developer must run GenAPI to generate new reference assembly code.
+### Running GenAPI
+GenAPI can be run by setting the following MSBuild property while building.
+```
+/p:GenerateReferenceAssemblySource=true
+```
+When a build is run with that property enabled, GenAPI will read the runtime assembly and generate a new `{AssemblyName}.cs` file under the ref directory in the assembly's source tree.
+
+This new file will contain the newly created surface area and will need to be checked in along with the runtime assembly change.  The next build without `GenerateReferenceAssemblySource` enabled will no longer display an ApiCompat error as the surface area will now match the baseline.
+### Issues with GenAPI
```
stevenbrix: is all of this extra work needed because we use the `InternalsVisibleToAttribute`?

src/Microsoft.DotNet.Wpf/src/Shared/ref/AttributesForReferenceAssemblies.cs
```diff
@@ -3,8 +3,9 @@
 // See the LICENSE file in the project root for more information.
 
 //
+// This file specifies various assembly level attributes.
 //
 
 using System.Runtime.CompilerServices;
 
-[assembly:TypeForwardedTo(typeof(System.Windows.Input.ICommand))]
```
stevenbrix: this is weird, and maybe it's just a bug in GitHub, but did this `ICommand.cs` file get removed?

rladuca:
> In your description you mention a "ManualFixups.txt" file in the PR, but i don't see that anywhere. was that removed or changed to a different file name?

I removed it from the docs and the repo.  Basically, it would confuse more than help in light of some internals needing to be kept for XAML compilation.

eng/WpfArcadeSdk/tools/GenApi/GlobalApiExclusions.txt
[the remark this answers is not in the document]
rladuca: I can add to the docs, but this is an exclusions file that GenAPI uses to explicitly remove types.

src/Microsoft.DotNet.Wpf/ApiCompat/Baselines/PresentationFramework-ref-Net48.baseline.txt
```diff
@@ -1,15 +1,23 @@
 Compat issues with assembly PresentationFramework:
+MembersMustExist : Member 'Microsoft.Win32.FileDialog.CreateVistaDialog()' does not exist in the implementation but it does exist in the contract.
+MembersMustExist : Member 'Microsoft.Win32.FileDialog.ProcessVistaFiles(MS.Internal.AppModel.IFileDialog)' does not exist in the implementation but it does exist in the contract.
```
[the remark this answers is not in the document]
rladuca: This is ApiCompat vs .NET 4.8 assemblies.  This is saying, "This exists in 4.8 but not in .NET Core".  That's why it's baselined because we are intentionally removing this from the ref surface area.

See: https://github.com/dotnet/wpf/blob/master/Documentation/api-compat.md

src/Microsoft.DotNet.Wpf/ApiCompat/Baselines/PresentationCore-ref-Net48.baseline.txt
```diff
@@ -0,0 +1,24 @@
+Compat issues with assembly PresentationCore:
+CannotRemoveBaseTypeOrInterface : Type 'System.Windows.LocalizabilityAttribute' does not implement interface 'System.Runtime.InteropServices._Attribute' in the implementation but it does in the contract.
+CannotRemoveAttribute : Attribute 'MS.Internal.PresentationCore.FriendAccessAllowedAttribute' exists on 'System.Windows.Input.InputEventArgs' in the contract but not the implementation.
+MembersMustExist : Member 'System.Windows.Media.Brush.AddRefOnChannelCore(System.Windows.Media.Composition.DUCE.Channel)' does not exist in the implementation but it does exist in the contract.
+MembersMustExist : Member 'System.Windows.Media.Brush.GetChannelCore(System.Int32)' does not exist in the implementation but it does exist in the contract.
```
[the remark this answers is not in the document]
rladuca: Same as above

Documentation/gen-api.md
rladuca: It's a combination of a few things, `InternalsVisibleTo` certainly complicates things between assemblies, but also XAML compilation will reflect and try to find internal API in order to take certain shortcuts and optimizations.  (Look at [IAddChildInternal](https://github.com/dotnet/wpf/search?q=IAddChildInternal&unscoped_q=IAddChildInternal) for example).

There are also issues where we have public API that has internal abstract members, which confuses GenAPI.  GenAPI itself just sometimes adds strange things, like private dummy int fields and properties.  No idea why, there are just bugs.  I intend to make small reproduction applications when I have time and file bugs against GenAPI, but that's future stuff.

src/Microsoft.DotNet.Wpf/src/Shared/ref/AttributesForReferenceAssemblies.cs
rladuca: I did remove it because it only contained a type forward.  I moved that type forward to the assembly attribute file so I could easily include it in the ref assembly.

I don't know why git decided it was renamed into my new assembly attribute file, but it did.

Documentation/gen-api.md
weltkante: Are those bugs reported back to the Roslyn team so they can be fixed?

rladuca: GenAPI doesn't use Roslyn it uses [Microsoft.CCI](https://github.com/Microsoft/cci)).  Originally, we were using Roslyn to generate our reference assemblies, but the use of InternalsVisibleTo was creating dangling references.  We filed https://github.com/dotnet/roslyn/issues/36409 to address these issues.  In the meantime, this was the fix.

I haven't created issues against GenAPI yet, I haven't had the time to create minimal reproductions of the various bugs.

src/Microsoft.DotNet.Wpf/ApiCompat/Baselines/System.Xaml-ref.baseline.txt
rladuca: Decompiling the new reference assemblies associates these appropriately, I don't think this is a problem, but I am willing to change this if someone thinks it makes sense.

eng/WpfArcadeSdk/tools/ShippingProjects.props
```diff
@@ -52,24 +52,55 @@
...
+      ReachFramework-PresentationFramework-api-cycle;
+      ReachFramework-System.Printing-api-cycle;
+      System.Printing-PresentationFramework-api-cycle
     </CycleBreakerProjects>
-    <HandCraftedReferenceProjects>
-        System.Printing-ref
-    </HandCraftedReferenceProjects>
+
+    <!-- 
+      Reference assembly projects are categorized by what repo they are shipping in.
```
vatsan-madhavan: Can these all be put in external repo under shared, then consumed from Microsoft.DotNet.Arcade.Wpf.Sdk (which brings along all sources under Shared\**/* ?

src/Microsoft.DotNet.Wpf/src/PresentationCore/otherassemblyattrs.cs
```diff
@@ -21,6 +21,7 @@
 
 [assembly: TypeForwardedTo(typeof(System.Windows.Markup.IUriContext))]
 [assembly: TypeForwardedTo(typeof(System.Windows.Media.TextFormattingMode))]
+[assembly: TypeForwardedTo(typeof(System.Windows.Input.ICommand))]
```
vatsan-madhavan: Was this missed previously?

eng/WpfArcadeSdk/tools/ShippingProjects.props
rladuca: We'd still need logic to disable picking up ref outputs in the internal packaging, so I am not sure we'd reduce complexity by doing that.

src/Microsoft.DotNet.Wpf/src/PresentationCore/otherassemblyattrs.cs
rladuca: It was, originally, inside ICommand.cs.  I removed that file and bumped the forward to the attributes file so it could be easily included in the ref assembly.

vatsan-madhavan:
> .NET Framework reference assemblies contain resources (BAML/strings resx, etc). Should we also include these in .NET Core?

No, let's not include these.

eng/WpfArcadeSdk/tools/ShippingProjects.props
rladuca: We could probably start moving most stuff out like that though, including cycle breakers.

Documentation/gen-api.md
stevenbrix: If i'm adding a new function to a class, what's the difference between adding the API to the reference assembly manually and using GenAPI? I might not fully understand the problem we are trying to solve here, but it seems like it would be much simpler to just manually modify the reference assembly. Is there some part of our workflow that makes things more complicated?

Documentation/gen-api.md
```diff
@@ -0,0 +1,46 @@
...
+Often, GenAPI will generate code output that will contain code that is either private, internal, or creates build errors.  For this reason a developer usually cannot just use the output of GenAPI directly.  Instead, the developer should do the following:
+* Build with GenAPI enabled
+* Diff the output file against the previous version
+* Extract just the new surface area (and related code) from the generated code
+* Revert the generated file
+* Add back the new surface area to the reference assembly code
+* Ensure that nothing in the new surface area is private or internal unless requried by XAML compilation or other reference assemblies
```
stevenbrix: It doesn't sound like a good idea to have internal/private APIs required by the XAML compiler. I understand that IAddChildInternal exists today, but is this something we want to put in as being an ok thing to do in the future?

rladuca: This is more if you make a change to an existing class that uses something like that, the entire class will get re-generated, sans the IAddChildInternal.  So you need to ensure that you're not messing with that piece.

rladuca: We can have a discussion about fixing the docs up later, at this point I don't want to spin another build just due to doc updates.  Helix queues are killing me today.

Documentation/gen-api.md
rladuca: I missed this.  The C# reference assemblies didn't exist in a manual fashion before this (back in the old internal repo maybe), so we could not just add things.  So GenAPI was used as an initial seed for the most part.  For some changes, it's probably easy enough just to pop the change into the reference assembly.  Other changes it might be more difficult and require basically mirrored code changes.  Also if you make a small change sans GenAPI then someone makes a very large change and uses it, it'll likely generate a noisier diff since your code location might not be the same.

It's probably the simplest and most thorough to GenAPI and use that output to manually add to the reference assembly.  It's pretty easy (mostly) to diff the generated output against the current surface and just pick and choose your new stuff.

Long term we'd rather just have Roslyn allow us to say, "All publics, but also these specific internals" in a nice configurable fashion and call it a day.  But we need something in the short term.

Issue #579, opened 2019-04-13
RichTextBox: Copy to Clipboard throws NotSupportedException


jbe2277:
* .NET Core SDK: 3.0.100-preview5-011256 \[e6f157e5a4\]
* Microsoft.WindowsDesktop.App 3.0.0-preview5-27610-56
* VS 2019: 16.1.0 Preview 1.0
* Windows version: Win10 1809 (Azure VM, connected via Remote Desktop)
* Does the bug reproduce also in WPF for .NET Framework 4.8?: No

Copy to Clipboard (e.g. press Ctrl+C) within the RichTextBox throws a NotSupportedException with the message: `No data is available for encoding 1252. For information on defining a custom encoding, see the documentation for the Encoding.RegisterProvider method.`

**Reproduce:**
* Clone [https://github.com/jbe2277/waf.git](https://github.com/jbe2277/waf.git)
* Checkout the **netcoreapp3.0** branch
* Open `System.Waf.sln` in VS 2019.
* Set `Samples/Writer/Writer.Presentation` as start-up project
* Start the application and click on New.
* Insert some text, select the text and press Ctrl + C.

**Expected:** The selected text should be copied to the Clipboard. You could compare this with the **master** branch (targets .NET 4.6.1)

**Actual:** The application crashes.

**Related issue:** I get the same exception when I try to save the RichTextBox content as RTF file. Just press Save in the Writer application.

**Details:** Copy to Clipboard throws the exception:
```
System.NotSupportedException
  HResult=0x80131515
  Message=No data is available for encoding 1252. For information on defining a custom encoding, see the documentation for the Encoding.RegisterProvider method.
  Source=System.Private.CoreLib
  StackTrace:
   at System.Text.Encoding.GetEncoding(Int32 codepage) in F:\workspace\_work\1\s\src\System.Private.CoreLib\shared\System\Text\Encoding.cs:line 287

System.Private.CoreLib.dll!System.Text.Encoding.GetEncoding(int codepage) Line 287
PresentationFramework.dll!System.Windows.Documents.FontTableEntry.ComputePreferredCodePage()
PresentationFramework.dll!System.Windows.Documents.XamlToRtfWriter.XamlToRtfWriter(string xaml)
PresentationFramework.dll!System.Windows.Documents.XamlRtfConverter.ConvertXamlToRtf(string xamlContent)
[...]

PR #1068, opened 2019-06-24
Fix crash when copying from RichTextBox

ojhad:
Fixes issue #664, #579 

The cause of the crash was that we were missing an encoding (Windows-1252) that is not available by default in core. In order to obtain it, we need to register an encoding provider. I added an internal encoding class to make sure we have this encoding provider registered.
I wrapped the uses of the static Encoding class that pass in parameters such as code page id or some `Encoding` parameters (as opposed to the places where we are just accessing properties)

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/Shared/MS/Internal/Text/InternalEncoding.cs
+++ b/src/Microsoft.DotNet.Wpf/src/Shared/MS/Internal/Text/InternalEncoding.cs
@@ -22,5 +22,10 @@ namespace MS.Internal.Text
         {
             return Encoding.GetEncoding(codepage);
         }
+
+        internal static byte[] Convert(System.Text.Encoding srcEncoding, System.Text.Encoding dstEncoding, byte[] bytes)
+        {
+            return Encoding.Convert(srcEncoding, dstEncoding, bytes);
+        }
     }
 }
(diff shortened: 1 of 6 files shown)


src/Microsoft.DotNet.Wpf/src/Shared/MS/Internal/Text/InternalEncoding.cs
```diff
@@ -0,0 +1,33 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System.Text;
+
+namespace MS.Internal.Text
+{
+    // We use encodings that are not provided by default in core.
+    // This class makes sure that we register extra providers that are required before use.
```
lindexi: Could you change it to document comment?

    /// <summary>
    /// The comment
    /// </summary>

src/Microsoft.DotNet.Wpf/src/Shared/MS/Internal/Text/InternalEncoding.cs
```diff
@@ -0,0 +1,33 @@
...
+
+        internal static Encoding Default
+        {
+            get
+            {
+                return Encoding.Default;
+            }
+        }
+
+    };
```
watfordgnf:
```suggestion
    }
```

vatsan-madhavan:
Removing *NO MERGE* since #1100 is completed.

PR #1077, opened 2019-06-24
Define _WindowsDesktopSdkTargetFrameworkVersionFloor before  Microsoft.NET.Sdk.WindowsDesktop.targets is imported

vatsan-madhavan:
Define `_WindowsDesktopSdkTargetFrameworkVersionFloor` before `Microsoft.NET.Sdk.WindowsDesktop.targets` is imported

This is needed since we changed the way props are defined in #1027

### DIFF
--- a/eng/WpfArcadeSdk/tools/Pbt.props
+++ b/eng/WpfArcadeSdk/tools/Pbt.props
@@ -13,6 +13,14 @@
   </PropertyGroup>
 
   <!-- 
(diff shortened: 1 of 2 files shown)


vatsan-madhavan:
Having trouble completing PR builds in < 60 mins due to Helix machine acquisition issues. The build otherwise passes cleanly. Forcing the PR to merge to unblock others in the team.

PR #1078, opened 2019-06-24
Update how timeout is defined in yml files.

vatsan-madhavan:
Update how timeout is defined in yml files.

We just noticed that our 120 min timeout is not working correctly, causing backlogs in build and preventing work from progressing. Trying to define this a different way.


vatsan-madhavan:
Related: https://github.com/dotnet/wpf/pull/1077

PR #1080, opened 2019-06-25
System.Printing does not consume lib assemblies from WPF GitHub package

rladuca:
Updating generated CppCliHelper project to reference and use RID-specific WPF GitHub package.

This is the WpfArcadeSdk piece of the change.  The latest commit in https://dev.azure.com/dnceng/internal/_git/dotnet-wpf-int/pullrequest/1772?_a=overview contains the internal piece of the fix.


vatsan-madhavan:
I recommend not waiting for Helix based tests to complete. It doesn't give you any value and likely going to block you for hours when the tests run for < 2 mins. The core build seems to have passed. Consider overriding policies and merging.

PR #1081, opened 2019-06-25
yml improvements and helix decoupling to separate pipeline

### LINKED ISSUE #1121: Improve CI Build Times
Since #782, CI build times have gone up > 1hr. We need to improve CI build times and still retain some mechanism for running tests.

vatsan-madhavan:
YAML improvements - various. 

Fixes #1121 

Intended to prevent build backlogs due to slow and overlapping builds that were observed to bottleneck PR'son 6/24. 

- Introduces a new AzDo build _public_ pipeline - `dotnet-wpf-citests`
  - This is a test-signed pipeline
  - CI builds are disabled on this pipeline. Only PR builds are allowed [This is an AzDo pipeline setting]
    - image
  - Will create and test _Release_ builds only on _Helix_
  - Helix test-runs can be disabled by disabling the pipeline on AzDo
    - image
  - Main AzDo pipeline `dotnet-wpf CI` will no longer run any tests (in _Helix_ or on build machines locally). 
    - It will only validate that the build succeeds. 
    - We will open a new issue to re-add the capability to run tests locally on build-machines (could be something we want to turn on selectively). 
- CI builds will be [batched](https://docs.microsoft.com/en-us/azure/devops/pipelines/yaml-schema?view=azure-devops&tabs=schema)
  -  PR builds will remain un-batched as before
- `Documentation/*` updates will not trigger builds. 

/cc @dotnet/wpf-developers

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -33,7 +33,7 @@ trigger:
     # - arcade
   paths:
     exclude:
-    -Documentation/*
+    - Documentation/*
 
 pr:
   autoCancel: true
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -14,6 +14,11 @@ variables:
   _PublishUsingPipelines: true
   _DotNetArtifactsCategory: WINDOWSDESKTOP
 
+# This is set in the pipeline directly 
+# When set to false, CI tests will not be enabled in builds. 
+#
+# _ContinuousIntegrationTestsEnabled: true
+
 # Setting batch to true, triggers one build at a time.
 # if there is a push while a build in progress, it will wait,
 # until the running build finishes, and produce a build with all the changes
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -17,7 +17,7 @@ variables:
 # This is set in the pipeline directly 
 # When set to false, CI tests will not be enabled in builds. 
 #
-# _ContinuousIntegrationTestsEnabled: true
+# _ContinuousIntegrationTestsEnabled: false
 
 # Setting batch to true, triggers one build at a time.
 # if there is a push while a build in progress, it will wait,
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -14,16 +14,31 @@ variables:
   _PublishUsingPipelines: true
   _DotNetArtifactsCategory: WINDOWSDESKTOP
 
+# Setting batch to true, triggers one build at a time.
+# if there is a push while a build in progress, it will wait,
+# until the running build finishes, and produce a build with all the changes
+#
 # only trigger ci builds for the master and release branches
 trigger:
-- master
-- release/3.0
-- arcade
+  batch: true 
+  branches:
+    include: 
+    - master
+    - release/3.0
+    # - arcade
+  paths:
+    exclude:
+    -Documentation/*
 
-# To be added in the future when VSTS supports this feature
-# only trigger pull request builds for the master branch
-# pr:
-# - master
+pr:
+  autoCancel: true
+  branches:
+    include:
+    - master
+    - release/3.0
+  paths:
+    exclude:
+    - Documentation/*
 
 # Call the pipeline.yml template, which does the real work
 jobs:
--- a/eng/pipeline.yml
+++ b/eng/pipeline.yml
@@ -157,4 +157,4 @@ jobs:
           HelixAccessToken: $(_HelixToken)              # only defined for internal CI
           Creator: $(_HelixCreator)
           SYSTEM_ACCESSTOKEN: $(System.AccessToken)
-        condition: and(succeeded(), eq(variables['_BuildConfig'], 'Release'), eq(variables['_HelixPublicBuildPipeline'], 'true')) # on helix machine (with real signing) when running on the public build pipeline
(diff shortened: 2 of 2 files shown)


vatsan-madhavan:
@stevenbrix This PR attempts to separate build-validation from test-validation and attempts to run them in  parallel in two separate build-pipelines. 

I'm not sure that we really get much out of it - maybe it gives us a small dial. For e.g., we can configure msftbot to check for `dotnet-wpf CI` and `dotnet-wpf-tests` routinely when `auto-merge` label is set, but when a different label is used (say, `fast-auto-merge`), perhaps then it will only check for `dotnet-wpf CI`). 

/cc @rladuca

stevenbrix:
@vatsan-madhavan does creating two build pipelines cause issues with packaging? Are the terms "pipeline" and "build definition" synonyms, or do they represent different things? If I remember correctly, there was an issue with some Arcade packages being produced from two different build definitions. I guess as long as the CI definition doesn't publish packages then we should be good?

vatsan-madhavan:
> does creating two build pipelines cause issues with packaging? Are the terms "pipeline" and "build definition" synonyms, or do they represent different things? 

I think it can. 

They are the same. _Pipeline_ is less ambiguous. _Build Definition_ may refer to the pipeline, or the YAML (although I suspect that it really refers to the pipeline only). 

Using multiple pipelines like corefx means we should ensure that only one pipeline ever runs for CI builds. That's easy to do - we can disable CI builds at pipeline definition level in AzDo and enable it just for PR builds. 

=

My goal here is to experiment and learn what capabilities exist so we might choose from a palette over the next few days. I'm not really recommending that we employ 2 or more pipelines. If it turns out that using 2 pipelines would be helpful, we can do that - otherwise just stick to 1.

vatsan-madhavan:
I spent looking around to see what others are doing. 

Here is what AspNet looks like (e.g. https://github.com/aspnet/AspNetCore/pull/11660). Only their main CI pipeline is marked as _required_:

image

CoreClr is likewise splitting their PR builds into multiple pipelines, with several dedicated to testing (e.g., https://github.com/dotnet/coreclr/pull/25350). None of the pipelines are marked as _required_ (maybe not the setting we want), and the PR-merge history suggests that merging PR's with less-than-perfect pass-rate is the norm. 

There are others that use only 1 pipeline (for e.g., MSBuild which is closer to us - averaging 34 mins, which includes testing. I suspect msbuild can run tests safely on the build-box...)

Overall, I like AspNet's approach.

vatsan-madhavan:
@rladuca Take a look at this version. <10 mins to complete the main (_required_) portion of the CI.

rladuca:
> @rladuca Rob LaDuca FTE Take a look at this version. <10 mins to complete the main (_required_) portion of the CI, and generally _holds_ only 2 machines while helix baesd tests are being run.
> 
> Disabling helix runs is as easy as disabling the dotnet-wpf-citests pipeline on AzDo (no yaml changes required).
> 
> What we _don't_ get are local test-runs on build-machines.

This seems fine, but I think we should get a quick PR out and just disable Helix in the current builds.  
@stevenbrix can we just quickly shut Helix off while we figure out the grand scheme of things w.r.t. multiple pipelines and possibly DRTs on build machines?  I was thinking we turn it off, then try out DRTs on build machines after your PR, see what shakes out failure wise.  Then we can start doing some of this more complicated work.  At least we won't wait an hour every build in the interim.

vatsan-madhavan:
@rladuca, @stevenbrix, Still leaving as `*NO` MERGE*`, but removing WIP label. I want to merge this if you are ok with it.

rladuca:
> @rladuca, @stevenbrix, Still leaving as `*NO` MERGE*`, but removing WIP label. I want to merge this if you are ok with it.

Can we turn off the test pipeline?  It's going to consume 2 build machines each PR build and it's not really that useful.

vatsan-madhavan:
> This seems fine, but I think we should get a quick PR out and just disable Helix in the current builds.

@rladuca This pretty much does exactly that - disables Helix in the current CI builds. It also spins up another pipeline for CI builds, which we don't have to keep enabled just yet (that's a policy conversation)

vatsan-madhavan:
> > @rladuca, @stevenbrix, Still leaving as `*NO` MERGE*`, but removing WIP label. I want to merge this if you are ok with it.
> 
> Can we turn off the test pipeline? It's going to consume 2 build machines each PR build and it's not really that useful.

easily done - yes.

rladuca:
> > > @rladuca, @stevenbrix, Still leaving as `*NO` MERGE*`, but removing WIP label. I want to merge this if you are ok with it.
> > 
> > 
> > Can we turn off the test pipeline? It's going to consume 2 build machines each PR build and it's not really that useful.
> 
> easily done - yes.

Yeah I'm fine with that then, we should iterate further on getting @stevenbrix 's DRT changes to run on build machine and see the fail rates, etc.  Then we can change things.

stevenbrix:
I like what you've done!

>This seems fine, but I think we should get a quick PR out and just disable Helix in the current builds.
@stevenbrix can we just quickly shut Helix

Simplest thing would be to just make the condition always evaluate to false. Maybe adding a pipeline variable that we can pass to turn them on/off from the pipeline itself (default should be on). If we want to disable tests, this is probably the PR to do it. I don't think there is anything weird going on with Helix for us, what we're seeing is about what everyone else is seeing as well. I do agree that is probably overkill for what we are trying to accomplish right now. Once we have more tests running it will make more sense.

The WinUI team has done some cool stuff around determining whether a PR requires tests (for example, not running tests for simple documentation updates) that we can probably piggy back on. You should go check that out.

vatsan-madhavan:
For now, I'm going to disable _citests_ pipeline and merge this. We can make improvements to yml even more if we want to. Let's regroup and chat more.

PR #1111, opened 2019-06-27
packaging test updates and updates to packaging tests

### LINKED ISSUE #816: Port test infrastructure to the open and remove dependency on running as an admin
I started digging into this a little bit, but there are a lot more dependencies on registry access than I had anticipated. This will most likely take a few days by someone more knowledgeable than I am and take some careful scanning of the code. Since we want to move this out in the open as well, I feel both these items can be done at the same time.

### LINKED ISSUE #1051: [DRT] Re-Enable GetPartUri code in DRTPackagingAPIs
Copied from AzDo: 843574

 Re-enable code that uses currently missing API 'GetPartUri' when the implementation is merged into CoreFX (System.IO.Packaging).
Code snippet is in DRTPackagingApis.cs in the DigitalDocuments Area under a TODO comment. Search for 'GetPartUri'.

### LINKED ISSUE #1052: [DRT] DrtInterleavedPackage fails after System.IO.Packaging refactoring
Copied from AzDO: 848772

Steps to reproduce:

1. Run DrtInterleavedPackage
2. Observe NullReferenceException from missing PAckage.Open method we reflect into.

The purpose of this test is to simulate the way XpsDocument calls into System.IO.Packaging to achieve package generation for printing and saving to disk. since System.IO.Packaging changed significantly in the support for certain features we need to first figure out how to achieve package generation correctly in product code, once we do that we can look into changing this test to reflect that.

stevenbrix:
I took advantage of this unique situation to give myself a clever little title ;)

This PR does a few things:
1. Updates our packaging to create a `Microsoft.DotNet.Wpf.DncEng.Test` package that contains our test infrastructure that compiles against `Microsoft.NETCore.App` instead of `Microsoft.WindowsDesktop.App`. This should be a fairly temporary solution until we open source the infrastructure entirely. There will be an internal PR that consumes this, and then once that is done we can update this repo to consume it.
2. Adds a few of the packaging tests that don't have any special test content files and fixes them from the packaging refactor. The rest of the tests will be updated in the internal repo until we have time to go through OSS cleanup on them.

Fixes #1051 
Fixes #1052 
Fixes #816


eng/WpfArcadeSdk/tools/SdkReferences.targets
```diff
@@ -194,14 +194,14 @@
...
       <ItemGroup>
-        <Reference Include="C:\Users\username\.nuget\packages\microsoft.private.winforms\4.8.0-preview4.19170.4\lib\netcoreapp3.0\Accessibility.dll" />
+        <Reference Include="C:\Users\username\.nuget\packages\microsoft.private.winforms\4.8.0-preview4.19170.4\lib\netcoreapp3.0\Accessibility.dll"/>
       </ItemGroup>
    -->
   <Target Name="ResolveWinFormsReferences"
           Condition="'@(MicrosoftPrivateWinFormsReference)'!='' And $(TargetFramework.StartsWith('netcoreapp3'))">
     <ItemGroup>
-      <Reference Include="$(PkgMicrosoft_Private_Winforms)\ref\$(TargetFramework)\%(MicrosoftPrivateWinFormsReference.Identity).dll"
-                 Condition="Exists('$(PkgMicrosoft_Private_Winforms)\ref\$(TargetFramework)\%(MicrosoftPrivateWinFormsReference.Identity).dll')">
```
vatsan-madhavan: Why?

eng/WpfArcadeSdk/tools/Packaging.targets
```diff
@@ -120,6 +120,16 @@ $(PreparePackageAssetsDependsOn):
                 Condition="'$(DestinationSubFolder)'!='' and '$(IsPackagingProject)'!='true' and '@(PackagingAssemblyContent)'==''">
       <Output ItemName="PackageAsset" TaskParameter="Include"/>
     </CreateItem>
+    <CreateItem Include="$(OutDir)*.deps.json"
+                AdditionalMetadata="RelativePath=$(ArtifactsPackagingDir)$(NormalizedPackageName)\$(DestinationSubFolder)"
+                Condition="'$(DestinationSubFolder)'!='' and '$(IsPackagingProject)'!='true' and '@(PackagingAssemblyContent)'==''">
```
vatsan-madhavan: I’d rather these weren’t added globally, and instead these were added by projects that need them as PackagingContent. 

I won’t have an objection to the introduction of an extensibility point for `PreparePackageAssets`, where about test specific target globally adds more items to `PackageAsset`.

stevenbrix: huh, this is odd, i don't remember adding this. i wonder if there was a bad merge or something on my end

vatsan-madhavan: :)

eng/WpfArcadeSdk/tools/SdkReferences.targets
stevenbrix: because the reference assembly can't be loaded at runtime so qualityvault (and any exe) would fail to run

eng/WpfArcadeSdk/tools/Packaging.targets
stevenbrix: ohhh nvm, why are these considered different then everything else?

eng/WpfArcadeSdk/tools/SdkReferences.targets
vatsan-madhavan: These are intended to be build time refs - not intended to supply runtime assemblies.

eng/WpfArcadeSdk/tools/Packaging.targets
vatsan-madhavan: Separation of concerns. These targets exist to serve the needs to shipping product. Needs of tests can either be served by them as-is, or using an extensibility.

stevenbrix:
> Sdk style projects should not have ToolsVersion. Something’s off with the solution configuration or your VS setup if those were auto generated...

Yeah VS is auto-generating this and won't load the project without it. I'll try updating my VS2019 install to see if that helps

eng/WpfArcadeSdk/tools/SdkReferences.targets
vatsan-madhavan: I don't think conflating ref-assemblies with lib-assemblies is a good practice in general. We do that for WPF assemblies for very special reasons, but we don't have good reasons to do that for WinForms assemblies. 

These are used to build the product - and we should continue building strictly against the ref-assemblies only.

vatsan-madhavan: As a general rule, any changes to build-orchestration that are needed by tests should not create side-effects upon product builds. 

We can think about better ways to achieve what you are trying to do here. For e.g., if the tests needs access to WinForms runtime-assemblies somehow, we'll need to solve that problem differently than to change what we build-against (ref-assemblies).

eng/WpfArcadeSdk/tools/Packaging.props
```diff
@@ -25,8 +25,8 @@
   <PropertyGroup>
     <GitHubTransportPackageName>Microsoft.DotNet.Wpf.GitHub</GitHubTransportPackageName>
     <DncEngTransportPackageName>Microsoft.DotNet.Wpf.DncEng</DncEngTransportPackageName>
+    <DncEngTestPackageName>Microsoft.DotNet.Wpf.DncEng.Test</DncEngTestPackageName>
```
rladuca: If you're building this in GitHub, it's not really a DncEng package.  I think we should name it based on where it is built, not where it ends up.

rladuca: Maybe I am misunderstanding the intent of the package is this being built solely in int or from GitHub?

eng/WpfArcadeSdk/tools/ShippingProjects.props
```diff
@@ -176,6 +185,10 @@
  
     <!-- WindowsDesktopSdk overrides Internal, External etc. -->
     <RepoLocation Condition="$(WindowsDesktopSdkProject.Contains('$(NormalizedMSBuildProjectName)'))">WindowsDesktopSdk</RepoLocation>
+
+    <!-- Test package overrides Internal, External etc. -->
+    <RepoLocation Condition="$(InternalTestProjects.Contains('$(NormalizedMSBuildProjectName)'))">InternalTest</RepoLocation>
```
rladuca: This may not be applicable to test, but I've seen bugs where you have something like:
`TestProject1`
`TestProject1ThatShips`
In these scenarios, just matching the project name would match both, which may be undesirable.  I've been adding the separator character to overcome this, so `'$(NormalizedMSBuildProjectName);'` to give an exact match.

src/Microsoft.DotNet.Wpf/test/XAML/DRT/DrtXaml/XamlTestClasses.FriendWithKey/XamlTestClasses.FriendWithKey.csproj
```diff
@@ -1,8 +1,11 @@
...
+﻿<Project Sdk="Microsoft.NET.Sdk" ToolsVersion="Current">
   <PropertyGroup>
     <AssemblyName>XamlTestClasses.FriendWithKey</AssemblyName>
     <RootNamespace>XamlTestClasses.FriendWithKey</RootNamespace>
     <Platforms>AnyCPU;x64</Platforms>
     <OutputType>Library</OutputType>
   </PropertyGroup>
-</Project>
+  <ItemGroup>
+    <InternalsVisibleTo Include="XamlTestClasses" />
```
rladuca: This is actually new to me (that Arcade supports this) and generally pretty cool.

eng/WpfArcadeSdk/tools/DrtsToRun.props
```diff
@@ -1,9 +1,32 @@
 <Project>
   <ItemGroup>
     <!--
-    List of Drts to run that are built out of dotnet/wpf. Note that this is incredibly
+    List of Drts (grouped by area) to run that are built out of dotnet/wpf. Note that this is incredibly
```
rladuca: Think about this mechanism being expanded to FeatureTests.  It could get pretty unwieldy.

Maybe we won't have a similar style of things going on there, but we should probably be sure that what mechanisms we use for DRTs are very similar to FeatureTests.  Maybe these should be automatically generated by a target parsing the QV files (like DrtList.xml and DiscoveryInfo.xml).

Maybe the best course of action would be using QV itself to generate the list that gets batched into Helix because then we could apply the discovery mechanism with specific variables that can trim/alter the test list.

Probably not a *now* thing, but might be something to think about going forward.

eng/WpfArcadeSdk/tools/SdkReferences.targets
stevenbrix: sounds good

eng/WpfArcadeSdk/tools/ShippingProjects.props
stevenbrix: it might not apply, but that's a good idea anyways, thanks!

src/Microsoft.DotNet.Wpf/src/WindowsBase/WindowsBase.csproj
```diff
@@ -439,6 +439,14 @@
     </ProjectReference>
   </ItemGroup>
   
+  <ItemGroup>
+     <!--
+     Allow test projects access to internals (but only those that require it). Arcade will generate a file like
+     "WindowsBase.InternalsVisibleTo.cs" with the proper assembly name and PublicKey for us
+     -->
+     <InternalsVisibleTo Include="DrtPackagingApis" />
```
vatsan-madhavan: In our codebase, we use `ExtendedAssemblyInfo` for this. See example [here](https://github.com/dotnet/wpf/blob/031b79d9aec73dc35d17792f1446a60423db564a/eng/WpfArcadeSdk/tools/ExtendedAssemblyInfo.props#L10)

src/Microsoft.DotNet.Wpf/test/XAML/DRT/DrtXaml/XamlTestClasses.FriendWithKey/XamlTestClasses.FriendWithKey.csproj
stevenbrix: yeah it is nice!! i totally found it out on accident haha

eng/WpfArcadeSdk/tools/DrtsToRun.props
stevenbrix: what do we do for feature tests, is there a FeatureTestList.xml like DRTs? My plan is to generate this file based on that list, which should be fairly straightforward to do

src/Microsoft.DotNet.Wpf/src/WindowsBase/WindowsBase.csproj
vatsan-madhavan: I would recommend adding this to `OtherAssemblyAttrs.cs` directly, and introducing the assembly name into `BuildInfo` first under `shared\RefAssemblyAttrs.cs`. 

This allows us to maintain a single uniform system for tracking `InternalsVisibleTo` relationships. Having this encoding inside a `csproj` makes this harder to maintain, and potentially troubleshoot problems in the future. 

Any assemblies requested internals access to `WindowsBase` etc. should have its key fingerprint declared clearly in `RefAssebmlyAttrs.cs`. Having a separate section in there for test assemblies would be preferable.

eng/WpfArcadeSdk/tools/DrtsToRun.props
rladuca: Unfortunately FeatureTests aren't that straightforward.  In the same directory with DrtList, is DiscoveryInfo,xml.  This doesn't list everything though, you need to use QV to get the full test discovery I believe, which is how I generated that excel file we had with the test lists.

I had to hack QV for that, but you can probably add an option to spit out the discovery list and we can run it as a one off.

@miguep None of this sounds completely incorrect does it?  : - )

eng/WpfArcadeSdk/tools/Packaging.props
vatsan-madhavan: The name of the package is being added to the props file - that's all. It will get consumed during build orchestration in dotnet-wpf-int where the package named m.d.w.d.test will be produced.

stevenbrix: solely in int

rladuca: Got it, my main concern at this point is the infra package being built internally, then flipping to GitHub at some point.  Maybe we can just open the infra as well?

Minor concern though, just give it some thought,

src/Microsoft.DotNet.Wpf/src/WindowsBase/WindowsBase.csproj
stevenbrix: That seems like a lot more maintenance and overhead. Arcade generates this all for us based on the key you are using and all you have to is add this in your .csproj. We could go about removing all of that and moving to this model, but that would be out of scope for this PR.

vatsan-madhavan: When we tried to remove this from `otherassemlyattrs.cs` early in .NET Core 3 and "clean up" this information, we broke API compat widely and spent several days chasing the mistakes and fixing problems in the build, then resurrecting this information from history.

stevenbrix: can you elaborate more on what happened? how does changing `InternalsVisibleTo` attributes break API compat?

rladuca:
@stevenbrix If you're making bug fixes to DRTs, please make sure there are separate issues for them (apart from the conglomerate issues we have) and that you immediately port these to the internal test repo.  If this PR takes more time to get merged, then you will be leaving us with test issues in our private runs in the interim period and possibly someone will try to fix them not realizing that you've already done so.

PR #1140, opened 2019-07-02
Fix for Relative Uri's in Flowdocument being broken

ojhad:
Issue #1141 

This test was failing because we were trying to read the scheme of a relative URI (which was throwing an exception). Removed this statement as the check should have been removed as part of the changes that were made here: https://github.com/dotnet/wpf/pull/969/files#diff-d6cd62f214719773a7a26ab93a0235b8

Ran a full DRT test pass on this change for verification.

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Documents/FixedPage.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Documents/FixedPage.cs
@@ -615,8 +615,6 @@ namespace System.Windows.Documents
             Debug.Assert(dpo != null, "GetLinkUri shouldn't be called for non-DependencyObjects.");
 
             if (inputUri != null)
-            // We want to allow navigation to pack:// Uris as we may get these from automation, but we 
-            // wouldn't support this if an actual Uri were given as a pack:// Uri
             {
                 //
                 // First remove the fragment, this is to prevent escape in file URI case, for example,
--- a/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Documents/FixedPage.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Documents/FixedPage.cs
@@ -614,8 +614,7 @@ namespace System.Windows.Documents
             DependencyObject dpo = element as DependencyObject;
             Debug.Assert(dpo != null, "GetLinkUri shouldn't be called for non-DependencyObjects.");
 
-            if (inputUri != null &&
-                (inputUri.Scheme == PackUriHelper.UriSchemePack && !String.IsNullOrEmpty(inputUri.Fragment)))
+            if (inputUri != null)
             // We want to allow navigation to pack:// Uris as we may get these from automation, but we 
             // wouldn't support this if an actual Uri were given as a pack:// Uri
             {


miguep:
Is there an issue for this?

This goes a bit beyond just the test, I think this affects at least all relative hyperlinks in FlowDocument.

src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Documents/FixedPage.cs
```diff
@@ -615,8 +615,6 @@ internal static Uri GetLinkUri(IInputElement element, Uri inputUri)
             Debug.Assert(dpo != null, "GetLinkUri shouldn't be called for non-DependencyObjects.");
 
             if (inputUri != null)
-            // We want to allow navigation to pack:// Uris as we may get these from automation, but we
```
miguep: I think this comment is in regards to the logic below, not the check you removed

rladuca:
> Is there an issue for this?
> 
> This goes a bit beyond just the test, I think this affects at least all relative hyperlinks in FlowDocument.

@ojhad Please make sure there is a detailed issue for this PR.

ojhad:
> Is there an issue for this?
> 
> This goes a bit beyond just the test, I think this affects at least all relative hyperlinks in FlowDocument.

Created issue and updated description.

In this specific method, there is already code accounting for relative URIs:
https://github.com/dotnet/wpf/blob/3e2293178da75df58a1f4ac878b4c5563511dea7/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Documents/FixedPage.cs#L634

In the removed `CanNavigateToUri` method, it would return true if the uri was ever relative due to a
`!uri.IsAbsoluteUri` check: [link to where the method was deleted](https://github.com/dotnet/wpf/pull/969/files#diff-d6cd62f214719773a7a26ab93a0235b8)

Is there somewhere else you see this change could be an issue @miguep ?

src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Documents/FixedPage.cs
ojhad: Yeah, I was on the fence on whether to remove this because I think its both (it was placed right under `(inputUri.Scheme == PackUriHelper.UriSchemePack && !String.IsNullOrEmpty(inputUri.Fragment))` in order to explain why this extra check was there).

However, I guess its better to leave this in as it does describe other functionality. Re-added comment.

stevenbrix: this comment does seem out-of-date now, it sounds like it was referring only to the check. before the removal, the if statement read like this:

```
if (inputUri != null &&
   (CanNavigateToUri(inputUri) ||
   (inputUri.Scheme == PackUriHelper.UriSchemePack && !String.IsNullOrEmpty(inputUri.Fragment))))
```

so i think the comment was about the fact that `CanNavigateToUrl` would return false and there were scenarios where we wanted to navigate even when that returned false. but now it seems we always navigate, unless i'm misunderstanding something?

miguep: I think you're right @stevenbrix , upon first reading it I assumed it was an overview of the logic below, but looking at it again it makes more sense, we should remove it.

rladuca:
Can you make sure you run all the DRTs against this change and that this failure is fixed and other failures are at baseline?

ojhad:
> Can you make sure you run all the DRTs against this change and that this failure is fixed and other failures are at baseline?

Re-ran DRTs to verify the failure is fixed and that failures are at baseline.

PR #1157, opened 2019-07-03
Use nameof instead of magic string

Youssef1313:


src/Microsoft.DotNet.Wpf/src/PresentationFramework/MS/Internal/Annotations/Anchoring/TreeNodeSelectionProcessor.cs
```diff
@@ -101,12 +101,12 @@ public override IList<DependencyObject> GetSelectedNodes(Object selection)
...
         {
             if (selection == null)
-                throw new ArgumentNullException("selection");
+                throw new ArgumentNullException(nameof(selection));
 
             UIElement element = selection as UIElement;
             if (element == null)
             {
-                throw new ArgumentException(SR.Get(SRID.WrongSelectionType), "selection");
+                throw new ArgumentException(SR.Get(SRID.WrongSelectionType), nameof(element));
```
stevenbrix: i think this makes sense to be`nameof(selection)` since that the `ArgumentException` is expressing that the `selection` paramter is the wrong type.

src/Microsoft.DotNet.Wpf/src/PresentationFramework/MS/Internal/Annotations/Anchoring/TreeNodeSelectionProcessor.cs
```diff
@@ -122,12 +122,12 @@ public override UIElement GetParent(Object selection)
...
         {
             if (selection == null)
-                throw new ArgumentNullException("selection");
+                throw new ArgumentNullException(nameof(selection));
 
             Visual element = selection as Visual;
 
             if (element == null)
-                throw new ArgumentException(SR.Get(SRID.WrongSelectionType), "selection");
+                throw new ArgumentException(SR.Get(SRID.WrongSelectionType), nameof(element));
```
stevenbrix:
```suggestion
                throw new ArgumentException(SR.Get(SRID.WrongSelectionType), nameof(selection));
```

src/Microsoft.DotNet.Wpf/src/PresentationFramework/MS/Internal/Annotations/Anchoring/TreeNodeSelectionProcessor.cs
stevenbrix:
```suggestion
                throw new ArgumentException(SR.Get(SRID.WrongSelectionType), nameof(selection));
```

stevenbrix: i think this makes sense to benameof(selection) since that the ArgumentException is expressing that the selection paramter is the wrong type.

Youssef1313:
@stevenbrix

stevenbrix:
Thanks @Youssef1313! Sorry for the delay, I appreciate you sticking with it 👍

Issue #1160, opened 2019-07-03
TextBoxView causes memory leak for 2 seconds after unloading host control


batzen:
* .NET Core Version: All versions
* Windows version: 1909
* Does the bug reproduce also in WPF for .NET Framework 4.8?: Yes

 **Problem description:**
TextBoxView causes memory leak for 2 seconds after unloading host control.
That's caused by the `DispatcherTimer` `_throttleBackgroundTimer`.
This might not be a problem during usage by a human user, but when i use my custom UI test framework this is a real problem.

 **Actual behavior:**
TextBoxView causes memory leak for 2 seconds after unloading host control.
 
 **Expected behavior:**
TextBoxView does not leak memory after unloading host control
 
 **Minimal repro:**
- Create a window containing a `TextBox`
- Create a new instance of the window open/show the window, focus the `TextBox` and close that window multiple times during a 2 second timespan in a loop

grubioe:
This issue will be considered once all tests are ported.

batzen:
@grubioe any news on this?

oklemencic:
We are also waiting for this issue. Any news when this will be integrated into 3.x or .NET 5 / .NET 6?

fooubar:
Waiting on this one as well.

PR #1161, opened 2019-07-03
Fixing TextBoxView memory leak for 2 seconds after unloading host control

batzen:
This is a fix for #1160

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/PresentationFramework/MS/Internal/documents/TextBoxView.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationFramework/MS/Internal/documents/TextBoxView.cs
@@ -44,8 +44,6 @@ namespace System.Windows.Controls
         {
             Invariant.Assert(host is Control);
             _host = host;
-
-            ((Control)_host).Unloaded += OnHostUnloaded;
         }
 
         #endregion Constructors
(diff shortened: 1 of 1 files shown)


src/Microsoft.DotNet.Wpf/src/PresentationFramework/MS/Internal/documents/TextBoxView.cs
```diff
@@ -44,6 +44,8 @@ internal TextBoxView(ITextBoxViewHost host)
         {
             Invariant.Assert(host is Control);
             _host = host;
+
+            ((Control)_host).Unloaded += OnHostUnloaded;
```
0xd4d: This event is never unregistered after it's been raised.

weltkante: I think this code path can be reached multiple times in the lifetime of a TextBox (via AttachFromVisualTree/DetachFromVisualTree pairs) so you need to unsubscribe when TextBoxView is detached.

Generally I'd prefer if the owner would notify the TextBoxView instead of the TextBoxView subscribing with the owner.

batzen:
Will have a look and change it to the way you suggested. Just give me a few days.

src/Microsoft.DotNet.Wpf/src/PresentationFramework/MS/Internal/documents/TextBoxView.cs
batzen: Will have a look and change it to the way you suggested. Just give me a few days.

batzen: The best way to achieve this would be to add the event subscription code to `EnsureTextContainerListeners` and to `RemoveTextContainerListeners` as those are already called from the host to ensure proper initialization and disposal.
What do you think about that?

weltkante: Yeah that should work and is probably intended for that purpose, didn't see it yesterday when I did a quick check of who calls the constructor.

batzen:

Issue #1166, opened 2019-07-03
Validation adorner does show up for controls with initial errors inside an Expander


batzen:
* .NET Core Version: All
* Windows version: 1903
* Does the bug reproduce also in WPF for .NET Framework 4.8?: Yes

 **Problem description:**
This issue occurs on controls that "hide" their child controls. These controls are, for example, `Expander` and `TabControl`.

 **Actual behavior:**
Validation adorner does not show up for not yet visible controls with validation errors.
 
 **Expected behavior:**
Validation adorner should show up as soon as controls get visible if they were not visible before but have validation errors.
 
 **Minimal repro:**
- Create an object that implements `INotifyDataErrorInfo`
- Always return `true` in `HasErrors`
- Always return some text in `GetErrors`
- Create a property on that object
- Create a `TextBox` and bind to the property on the object
- Place the `TextBox` inside a collapsed (`IsExpanded=False`) `Expander`
- Expand the `Expander`, no validation adorner is shown

grubioe:
To be considered when all tests are ported

batzen:
@grubioe any news on this?

grubioe:
@fabiant3  can provide the latest status.

pchaurasia14:
@dipeshmsft - Here is the repro for this issue. 
https://github.com/pchaurasia14/validationadornerissue

h3xds1nz:
This was fixed in #1167, hence closing.

PR #1167, opened 2019-07-03
Fixing validation adorner for not yet visible controls

batzen:
This is a fix for #1166

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Controls/Validation.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Controls/Validation.cs
@@ -420,9 +420,7 @@ namespace System.Windows.Controls
 
         private static void ShowValidationAdornerWhenAdornerSiteGetsVisible(object sender, DependencyPropertyChangedEventArgs e)
         {
-            var adornerSite = sender as UIElement;
-
-            if (adornerSite == null)
+            if (!(sender is UIElement adornerSite))
             {
                 return;
             }
--- a/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Controls/Validation.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Controls/Validation.cs
@@ -345,7 +345,6 @@ namespace System.Windows.Controls
             }
         }
 
-
         internal static void ShowValidationAdorner(DependencyObject targetElement, bool show)
         {
             // If the element has a VisualStateGroup for validation, then dont show the Adorner
@@ -362,7 +361,6 @@ namespace System.Windows.Controls
                 ShowValidationAdornerHelper(targetElement, adornerSite, show);
             }
         }
-
 
         private static bool HasValidationGroup(FrameworkElement fe)
         {
@@ -420,6 +418,26 @@ namespace System.Windows.Controls
             return null;
         }
 
+        private static void ShowValidationAdornerWhenAdornerSiteGetsVisible(object sender, DependencyPropertyChangedEventArgs e)
+        {
+            var adornerSite = sender as UIElement;
+
+            if (adornerSite == null)
+            {
+                return;
+            }
+
+            adornerSite.IsVisibleChanged -= ShowValidationAdornerOnVisible;
+
+            DependencyObject targetElement = GetValidationAdornerSiteFor(adornerSite);
+            if (targetElement == null)
+            {
+                targetElement = adornerSite;
+            }
+
+            ShowValidationAdornerHelper(targetElement, adornerSite, (bool)e.NewValue && GetHasError(targetElement), false);
+        }
+
         private static void ShowValidationAdornerHelper(DependencyObject targetElement, DependencyObject adornerSite, bool show, bool tryAgain)
         {
             UIElement siteUIElement = adornerSite as UIElement;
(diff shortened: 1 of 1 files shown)


src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Controls/validation.cs
```diff
@@ -420,6 +418,26 @@ private static object ShowValidationAdornerOperation(object arg)
...
+        private static void ShowValidationAdornerWhenAdornerSiteGetsVisible(object sender, DependencyPropertyChangedEventArgs e)
+        {
+            var adornerSite = sender as UIElement;
+
+            if (adornerSite == null)
+            {
+                return;
+            }
+
+            adornerSite.IsVisibleChanged -= ShowValidationAdornerOnVisible;
```
weltkante: Was this intended to unsubscribe the handler added from ShowValidationAdornerHelper?
If so, shouldn't it unsubscribe `ShowValidationAdornerWhenAdornerSiteGetsVisible` instead of `ShowValidationAdornerOnVisible`?

batzen: It should. Somehow my latest edits are lost. Will fix this tomorrow.

src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Controls/validation.cs
```diff
@@ -420,6 +418,26 @@ private static object ShowValidationAdornerOperation(object arg)
             return null;
         }
 
+        private static void ShowValidationAdornerWhenAdornerSiteGetsVisible(object sender, DependencyPropertyChangedEventArgs e)
+        {
+            var adornerSite = sender as UIElement;
```
lindexi: Suggestion: using `is` pattern

```

if (! sender is UIElement adornerSite)
{
    return;
}
```

batzen: Done

src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Controls/validation.cs
batzen: Done.

batzen:

dipeshmsft:
Hey @batzen, we ran the test pass on this PR and are ready to take this PR. Can you please resolve the conflicts in this PR? Thanks a lot for this contribution.

batzen:
@dipeshmsft There were no conflicts.
I rebased just to be sure.

After pulling in all the changes from main i can't compile anymore as i get the error
```
C:\DEV\OSS_Own\dotnet\wpf\eng\WpfArcadeSdk\tools\Wpf.Cpp.targets(552,5): error : 'WriteOnlyWhenDifferent' attribute on
"C:\DEV\OSS_Own\dotnet\wpf\artifacts\obj\uce\Win32\Debug\CppSupport\CppSupport_uce.csproj" takes effect only when 'Over
write="true"'. [C:\DEV\OSS_Own\dotnet\wpf\src\Microsoft.DotNet.Wpf\src\WpfGfx\core\uce\uce.vcxproj]
    0 Warning(s)
    179 Error(s)
```
Any hints on how to get it to compile?

src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Controls/Validation.cs
```diff
@@ -432,10 +448,19 @@ private static void ShowValidationAdornerHelper(DependencyObject targetElement,
...
                     {
-                        // try again later, perhaps giving layout a chance to create the adorner layer
-                        adornerSite.Dispatcher.BeginInvoke(DispatcherPriority.Loaded,
-                                    new DispatcherOperationCallback(ShowValidationAdornerOperation),
-                                    new object[]{targetElement, adornerSite, show});
+                        // Check if the element is visible, if not try to show the adorner again once it gets visible.
+                        // This is needed because controls hosted in Expander or TabControl don't have a parent/AdornerLayer till the Expander is expanded or the TabItem is selected.
+                        if (siteUIElement.IsVisible == false)
+                        {
+                            siteUIElement.IsVisibleChanged += ShowValidationAdornerWhenAdornerSiteGetsVisible;
```
miloush, line 455: Can the element not become IsVisible = true before the loading finishes? Because then the repeated attempt would fail and you would not get the Loaded attempt which you get now otherwise.

batzen: I don't know how a not loaded control should be visible.

dipeshmsft:
> @dipeshmsft There were no conflicts. I rebased just to be sure.
> 
> After pulling in all the changes from main i can't compile anymore as i get the error
> 
> ```
> C:\DEV\OSS_Own\dotnet\wpf\eng\WpfArcadeSdk\tools\Wpf.Cpp.targets(552,5): error : 'WriteOnlyWhenDifferent' attribute on
> "C:\DEV\OSS_Own\dotnet\wpf\artifacts\obj\uce\Win32\Debug\CppSupport\CppSupport_uce.csproj" takes effect only when 'Over
> write="true"'. [C:\DEV\OSS_Own\dotnet\wpf\src\Microsoft.DotNet.Wpf\src\WpfGfx\core\uce\uce.vcxproj]
>     0 Warning(s)
>     179 Error(s)
> ```
> 
> Any hints on how to get it to compile?

@batzen I see, I will put up a workaround that I had for this up after which you will be able to build it locally.

dipeshmsft:
@batzen I have made the necessary changes in this PR #7678, you can try it once now.

src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Controls/Validation.cs
dipeshmsft: @miloush were you able to get any instances where the above condition will hold true? Or if you can suggest a way to check that?

miloush: Also what if the control never gets visible? Aren't we leaking references by subscribing to IsVisibleChanged?

batzen: We won't leak as the handler is static and not an instance member, but i will also test that.

Will have a look at the sample you posted.

miloush: RenderTargetBitmap might be another scenario that could be affected.

Since the code already protects against re-entrancy, why don't we just keep the current codepath _and_ add your visibility check after?

miloush: The basic premise is that `IsVisible` is set to true for all controls before the `Loaded` events are fired (and the controls first displayed) and that is still before the dispatcher gets to the Loaded priority operations. That is a lot of opportunities in between for people to mess up with the tree.

batzen: @dipeshmsft @miloush I made changes that should prevent the code from introducing any regressions.
I also confirmed that the changes don't lead to leaks if an element never gets visible.

dipeshmsft (approved):

dipeshmsft:
@batzen, thanks for your contribution. Once, the right area of the test is open-sourced, I think we should add the tests for this fix.

PR #1169, opened 2019-07-03
Preventing focus steal if parent window/application is not active

### LINKED ISSUE #1168: Hovering over MenuItem causes the window to get activated
* .NET Core Version: All
* Windows version: 1903

 **Problem description:**
 
 **Actual behavior:**
Hovering over a `MenuItem` causes the window to get activated.
 
 **Expected behavior:**
Hovering over a `MenuItem` does not activate the window.
 
 **Minimal repro:**
- Create a new window containing a `MenuItem`
- Focus a different window
- Hover over the `MenuItem`
- Window gets activated

batzen:
This fixes #1168

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Controls/MenuItem.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Controls/MenuItem.cs
@@ -2498,7 +2498,10 @@ namespace System.Windows.Controls
             // but if we fail to focus we should still select.
             // (This is to help enable focusless menus).
             // Check IsKeyboardFocusWithin to allow rich content within the menuitem.
-            if (!IsKeyboardFocusWithin)
+            if (!IsKeyboardFocusWithin
+                // but only aquire focus the window we are inside is currently active
+                // otherwise we would potentially steal focus from other applications.
+                && Window.GetWindow(this)?.IsActive == true)
             {
                 Focus();
             }
--- a/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Controls/MenuItem.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Controls/MenuItem.cs
@@ -2501,7 +2501,7 @@ namespace System.Windows.Controls
             if (!IsKeyboardFocusWithin
                 // But only acquire focus if the window we are inside of is currently active or there is no window.
                 // Otherwise we would potentially steal focus from other applications.
-                && Window.GetWindow(this)?.IsActive ?? true)
+                && (Window.GetWindow(this)?.IsActive ?? true))
             {
                 Focus();
             }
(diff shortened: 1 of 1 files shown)


batzen:

dipeshmsft:
@batzen, we included this PR in our June 23 CTP and found that DrtControls fails after including the fix. The fix itself seems logical and we are currently investigating the test and the reason for the failure. It may be possible that we will need to modify the test. We will get back to you with more details regarding this.

In the meanwhile, if you would like, you can run the test yourself. The test is present in the [wpf-test](https://github.com/dotnet/wpf-test) repository here https://github.com/dotnet/wpf-test/tree/main/src/Test/Controls/DRT/Controls
The steps for building and running the tests are mentioned in the repo.

batzen:
The cause for the failure is more likely because the test does not use a window.
So my code should check `Window.GetWindow(this)?.IsActive != false` instead of `Window.GetWindow(this)?.IsActive == true`.
What do you think?

dipeshmsft:
This suggestion seem to work when tried. However, I feel that this `Window.GetWindow(this)?.IsActive ?? true` makes the intent more clear. What are you thoughts on this ?

Moreover @batzen can you make the necessary changes to this PR before we take it up for the next test pass.

rchauhan18:
Hey @batzen  , your suggestion to change ```Window.GetWindow(this)?.IsActive == true```  to 
``` Window.GetWindow(this)?.IsActive != false``` is correct. We have tested this PR with this change and all the test are passing and we are planning to merge it. 
Can you update this PR ?

**Comment by batzen** [comment]
@rchauhan18 @dipeshmsft I took the `??` approach as that really felt more clear here.
Thanks for the suggestion.

File: src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Controls/MenuItem.cs [hunk 1]
```diff
@@ -2498,7 +2498,10 @@ private void FocusOrSelect()
             // but if we fail to focus we should still select.
             // (This is to help enable focusless menus).
             // Check IsKeyboardFocusWithin to allow rich content within the menuitem.
-            if (!IsKeyboardFocusWithin)
+            if (!IsKeyboardFocusWithin
+                // But only aquire focus if the window we are inside of is currently active or there is no window.
```
ThomasGoulet73, line 2502: Nit:
```suggestion
                // But only acquire focus if the window we are inside of is currently active or there is no window.
```

Suggested rewording (Makes it more clear to me but feel free to dismiss this suggestion):
```suggestion
                // But only acquire focus if the window containing this menu item is currently active or if this menu item is not in a window.
```

batzen: Will fix the typo, but the comment above mine also talks about "we".
I don't think it helps if we write out in which class we currently are as it's a source comment, that seems redundant.

ThomasGoulet73:
```
I don't think it helps if we write out in which class we currently are as it's a source comment, that seems redundant.
```
Fair enough.

dipeshmsft (approved):

src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Controls/MenuItem.cs
dipeshmsft: This conversation seems resolved.

dipeshmsft:
Thanks @batzen for the fix and making the necessary updates. We will merge this PR by the start of next week.

rchauhan18:
Thanks @batzen for the contribution.

PR #1171, opened 2019-07-03
Using a stopwatch and timespan instead of DateTime and uint

### LINKED ISSUE #1170: TextBoxView uses DateTime.Now for time measurements
* .NET Core Version: All
* Windows version: 1903

 **Problem description:**
` TextBoxView` uses `DateTime.Now` inside `FullMeasureTick` to measure a maximum time frame of 200 milliseconds for line measurements.
As `DateTime.Now` is "slow" and inaccurate a `Stopwatch` should be used and the time frame should be expressed as a `TimeSpan` instead of int or uint.
The same `TimeSpan` change should apply to the `_throttleBackgroundTimer`.

batzen:
This fixes #1170

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/PresentationFramework/MS/Internal/documents/TextBoxView.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationFramework/MS/Internal/documents/TextBoxView.cs
@@ -1139,7 +1139,7 @@ namespace System.Windows.Controls
                 // all background layout.  This leaves the TextBox responsive
                 // to user input.
                 _throttleBackgroundTimer = new DispatcherTimer(DispatcherPriority.Background);
-                _throttleBackgroundTimer.Interval = new TimeSpan(0, 0, _throttleBackgroundSeconds);
+                _throttleBackgroundTimer.Interval = _throttleBackgroundTimeSpan;
                 _throttleBackgroundTimer.Tick += new EventHandler(OnThrottleBackgroundTimeout);
             }
             else
@@ -2242,7 +2242,7 @@ namespace System.Windows.Controls
             // Calculate a stop time.
             // We limit work to just a few milliseconds per iteration
             // to avoid blocking the thread.
-            DateTime stopTime;
+            TimeSpan stopTimeout;
 
             if ((ScrollBarVisibility)((Control)_host).GetValue(ScrollViewer.VerticalScrollBarVisibilityProperty) == ScrollBarVisibility.Auto)
             {
@@ -2251,15 +2251,15 @@ namespace System.Windows.Controls
                 // our interaction with ScrollViewer.  Disable background layout to
                 // mitigate the problem until we can take a real fix in v.next.
 
-                stopTime = DateTime.MaxValue;
+                stopTimeout = TimeSpan.MaxValue;
             }
             else
             {
-                stopTime = DateTime.Now.AddMilliseconds(_maxMeasureTimeMs);
+                stopTimeout = _maxMeasureTime;
             }
 
             // Format lines until we hit the end of document or run out of time.
-
+            var measureStopwatch = System.Diagnostics.Stopwatch.StartNew();
             do
             {
                 using (line)
@@ -2281,7 +2281,7 @@ namespace System.Windows.Controls
                     endOfParagraph = line.EndOfParagraph;
                 }
             }
-            while (!endOfParagraph && DateTime.Now < stopTime);
+            while (!endOfParagraph && measureStopwatch.Elapsed < stopTimeout);
 
             if (!endOfParagraph)
             {
@@ -3263,11 +3263,11 @@ namespace System.Windows.Controls
         private EventHandler UpdatedEvent;
 
         // Max time slice to run FullMeasureTick.
-        private const uint _maxMeasureTimeMs = 200;
+        private static TimeSpan _maxMeasureTime = TimeSpan.FromMilliseconds(200);
 
         // Number of seconds to disable background layout after receiving
         // user input.
-        private const int _throttleBackgroundSeconds = 2;
+        private static TimeSpan _throttleBackgroundTimeSpan = TimeSpan.FromSeconds(2);
 
         #endregion Private Fields
     }


src/Microsoft.DotNet.Wpf/src/PresentationFramework/MS/Internal/documents/TextBoxView.cs
```diff
@@ -3263,11 +3263,11 @@ internal int Offset
         private EventHandler UpdatedEvent;
 
         // Max time slice to run FullMeasureTick.
-        private const uint _maxMeasureTimeMs = 200;
+        private static TimeSpan _maxMeasureTime = TimeSpan.FromMilliseconds(200);
```
lindexi: Could you make the readonly field as `static readonly` key word

src/Microsoft.DotNet.Wpf/src/PresentationFramework/MS/Internal/documents/TextBoxView.cs
```diff
@@ -3263,11 +3263,11 @@ internal int Offset
...
 
         // Number of seconds to disable background layout after receiving
         // user input.
-        private const int _throttleBackgroundSeconds = 2;
+        private static TimeSpan _throttleBackgroundTimeSpan = TimeSpan.FromSeconds(2);
```
lindexi: Could you add the `readonly` key word?

src/Microsoft.DotNet.Wpf/src/PresentationFramework/MS/Internal/documents/TextBoxView.cs
batzen: Done

src/Microsoft.DotNet.Wpf/src/PresentationFramework/MS/Internal/documents/TextBoxView.cs
batzen: Done

src/Microsoft.DotNet.Wpf/src/PresentationFramework/MS/Internal/documents/TextBoxView.cs
```diff
@@ -2281,7 +2281,7 @@ private Size FullMeasureTick(double constraintWidth, LineProperties lineProperti
                     endOfParagraph = line.EndOfParagraph;
                 }
             }
-            while (!endOfParagraph && DateTime.Now < stopTime);
+            while (!endOfParagraph && measureStopwatch.Elapsed < stopTimeout);
```
stevenbrix: do you know how sensitive this code-path is to the accuracy of this measurement? the differences pointed out by SO make me wonder if we could possibly alter behavior that was dependent on the inaccuracy of this: https://stackoverflow.com/questions/2923283/stopwatch-vs-using-system-datetime-now-for-timing-events

stevenbrix:
Hey @batzen, thanks for filing this issue and sending out a PR! Sorry for the lack of communication on this PR, we're all pretty new to working in OSS and haven't been the best at following through with things.

I'd like to help get this PR moving along, since it seems like this is the right thing to do, I just want to make sure we aren't regressing anything.

src/Microsoft.DotNet.Wpf/src/PresentationFramework/MS/Internal/documents/TextBoxView.cs
batzen: I don't know how sensitive the code is to the accuracy. I proposed this change not only for increased accuracy, but primarily for increased performance.
Accessing DateTime.Now is much slower than using a stopwatch. And every bit of performance gain we can achieve, especially if it's easy, should be taken, i guess.
I could write a benchmark to show the real difference if you want me to.

batzen:
No worries, i didn't expect fast reactions as i already imagined that the whole team is busy open sourcing the code and getting it ready for the public.
I am very happy that WPF is finally open source.

At the time i wrote the PR there were no tests and i could run, or i was unable to find those.
Is there anything i can do to help to prevent regressions?

src/Microsoft.DotNet.Wpf/src/PresentationFramework/MS/Internal/documents/TextBoxView.cs
weltkante: It'll probably never matter, but `DateTime.Now` has correctness issues when being used for measuring timeouts or implementing timers, as the clock can be changed both manually and automatically (time zone changes esp. scheduled ones like summer/winter time changes). `DateTime.UtcNow` avoids the timezone problem and is faster than `DateTime.Now` but is still suspect to manual clock changes.

batzen:
@stevenbrix any news on this?

stevenbrix:
@batzen thanks for the ping. I haven't been working on WPF, and I actually can't help you out here anymore unfortunately 😞 

@fabiant3 or @SamBent might be able to help provide more clarity here, and what the plan is for getting WPF into a state where we can comfortably accept community PRs

stevenbrix:
To further explain my reasoning/hesitation on this change:

I just want to make sure we've done our due diligence, because a 15-20% difference in a measurement seems like it could have unintended consequences.  For example, if we're 20% slower and *think* that we're giving 200ms to read the paragraph, but in reality this completes in 220-230ms in most cases, then we have a change in behavior. Before we'd be exiting the loop because `endOfParagraph` was true, but now we're exiting because `measureStopwatch.Elapsed < stopTimeout` is true. It's unclear to me if this means we have a bug or not. Do you have any idea on the differences in these code paths:

```charp
if (!endOfParagraph)
{
       // Ran out of time.  Defer to background layout.
       SetFlags(true, Flags.BackgroundLayoutPending);
       this.Dispatcher.BeginInvoke(DispatcherPriority.Background, new DispatcherOperationCallback(OnBackgroundMeasure), null);
}
else
{
        // Finished the entire document.  Stop background layout.
        SetFlags(false, Flags.BackgroundLayoutPending);
}
```

batzen:
Well, currently, by using DateTime the code suffers from these differences. It might change it's behavior on every invocation. Using a Stopwatch would make the behavior stable and predictable.
I don't understand why making the code work correctly would be an issue here.

batzen:

singhashish-wpf:
This PR has passed smoke tests (DRTs) and is under review.

batzen:
@singhashish-wpf any news? The PR seems to have passed DRTs over two years ago. Has it been reviewed?

Issue #316, opened 2019-02-01
CommandParameter is null when call CanExecute method at first time


chenyj796:
* .NET Core Version: (3.0.0-preview-27122-01)
* Windows version: (`windows 10`)
* Does the bug reproduce also in WPF for .NET Framework 4.6.1?: Yes

 
 
 **Problem description:**

I have a TestCommand as below:
```csharp
public class TestCommand : ICommand
{
    public event EventHandler CanExecuteChanged;

    public bool CanExecute(object parameter)
    {
        return true;
    }

    public void Execute(object parameter)
    {
    }
}
```
and then I use it in a xaml file:
```xml
<Button Content="Test" Command="{StaticResource testCommand}" CommandParameter="abcdef"/>
```
 
 **Actual behavior:** 
 At first time the CanExecute method is called, the parameter's value is null。
If I change the XAML property's order as below, the parameter's value is abcdef (that's ok):
```xml
<Button Content="Test" CommandParameter="abcdef" Command="{StaticResource testCommand}"/>
```
 **Expected behavior:**
 The parameter's value is not null at each time, regardless of the XAML order

 **Minimal repro:**

brian-reichle:
@brandonhood, or the button could just defer the call between `BeginInit` and `EndInit`?

chenyj796:
In my opinion, the XAML equals to the codes:
```csharp
var button = new Button
{
    Content = "Test",
    Command = ...
    CommandParameter = ...
};
```
and the CanExecute method will be called by some methods such as Load ...

brian-reichle:
The .NET framework version of the xaml parser honours the [ISupportInitialize](https://docs.microsoft.com/en-us/dotnet/api/system.componentmodel.isupportinitialize?view=netframework-4.7.2) interface (which is implemented by FrameworkElement), so its actually more like this:

```csharp
var button = new Button();
((ISupportInitialize)button).BeginInit();
button.Content = "Test";
button.Command = ...
button.CommandParameter = ...
((ISupportInitialize)button).EndInit();
```

stevenbrix:
I agree that this behavior is incorrect. I think something like this should be pretty straightforward to do.  `FrameworkElement` has an `OnInitialized` override that `Button` should be able to invoke the command from. However, we don't know who is currently depending on the current behavior, so this is something that we can look into fixing after the initial release of .NET Core 3.

vflame:
[...]

PR #1179, opened 2019-07-05
Delay first execution of ICommand.CanExecute(object parameter) until …

MichaelVach:
…after initialization. Issue #316

Revert changes to Hyperlink.cs


vatsan-madhavan:
@MichaelVach, Thanks for the contribution here - we appreciate it! 

We are not accepting [behavior changing PR's at this time](https://github.com/dotnet/wpf/blob/master/Documentation/contributing.md) with very rare exceptions (for e.g., something critical to the release of .NET Core 3.0) - so I'm closing this PR. Let's continue discussing #316 and agreeing on a fix approach that factors in app-compat concerns. Once you have a fix approach that is accepted by the product team and .NET 5 contributions have been opened up, we will be happy to reconsider PR's for this if it makes sense.

Issue #1183, opened 2019-07-06
WPF program crashes with the nightly build sdk


Berrysoft:
* .NET Core Version: 3.0.100-preview8-012893
* Windows version: 10.0.18392.1000
* Does the bug reproduce also in WPF for .NET Framework 4.8?: No

 **Problem description:**
The program just crashes.
 **Actual behavior:** 
 ```
System.TypeLoadException
  HResult=0x80131522
  Message=Could not load type 'System.Xaml.Permissions.XamlAccessLevel' from assembly 'System.Security.Permissions, Version=4.0.2.0, Culture=neutral, PublicKeyToken=cc7b13ffcd2ddd51'.
  Source=PresentationFramework
  StackTrace:
   at System.Windows.Markup.XamlReader.LoadBaml(Stream stream, ParserContext parserContext, Object parent, Boolean closeStream)
   at System.Windows.Application.LoadBamlStreamWithSyncInfo(Stream stream, ParserContext pc)
   at System.Windows.Application.LoadComponent(Uri resourceLocator, Boolean bSkipJournaledProperties)
   at System.Windows.Application.DoStartup()
   at System.Windows.Application.<.ctor>b__1_0(Object unused)
   at System.Windows.Threading.ExceptionWrapper.InternalRealCall(Delegate callback, Object args, Int32 numArgs)
   at System.Windows.Threading.ExceptionWrapper.TryCatchWhen(Object source, Delegate callback, Object args, Int32 numArgs, Delegate catchHandler)
   at System.Windows.Threading.DispatcherOperation.InvokeImpl()
   at System.Windows.Threading.DispatcherOperation.InvokeInSecurityContext(Object state)
   at MS.Internal.CulturePreservingExecutionContext.CallbackWrapper(Object obj)
   at System.Threading.ExecutionContext.RunInternal(ExecutionContext executionContext, ContextCallback callback, Object state)
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Threading.ExecutionContext.RunInternal(ExecutionContext executionContext, ContextCallback callback, Object state)
   at System.Threading.ExecutionContext.Run(ExecutionContext executionContext, ContextCallback callback, Object state)
   at MS.Internal.CulturePreservingExecutionContext.Run(CulturePreservingExecutionContext executionContext, ContextCallback callback, Object state)
   at System.Windows.Threading.DispatcherOperation.Invoke()
[...]

PR #1192, opened 2019-07-08
Add PackageReference to account for XamlAccessLevel move

ojhad:
Related to issue #1183 

`XamlAccessLevel` was moved to System.Windows.Extensions.dll as part of the following PR: https://github.com/dotnet/corefx/pull/38908
Updating System.Xaml.csproj to have this package reference so that when dependencies are updated, the type can be found in its new location


vatsan-madhavan:
This was missing a similar `<PackageReference />` for `System.Xaml-ref.csproj`. I added it directly to the branch as a separate commit; completing this PR since we need this change immediately & @ojhad unavailable today.

Issue #1015, opened 2019-06-19
Expander control is not narrated


grubioe:
Copied from ADO ID: 850630

**Repro Steps:**
Run the attached sample program with Narrator running and focus on the Expander button. (For example, focus a ListItem and use the arrow keys to move focus to the Expander header.)

[WpfApp1.zip](https://github.com/dotnet/wpf/files/3307148/WpfApp1.zip)

**Actual Result:**
Narrator focus does not move to the Expander and it is not read.

**Expected Result:**
Narrator focus should move to the Expander and it should read its label.

**User Impact:​**
Users with visual disabilities will get confused while working with the application if screen reader is not reading information about control​ and its statuses

arpitmathur:
MAS rule we're not compliant with is MAS 1.3.1 – Info and Relationships

arpitmathur:
Fixed in #1278

PR #1278, opened 2019-07-16
Fixing MAS Compliance Issues

### LINKED ISSUE #791: Accessibility: (combobox control) An element of the given type should not support the Scroll pattern.
Copied from VSO ID: 691457

Does the bug reproduce also in WPF for .NET Framework 4.8?: Yes

 **Problem description:**
MAS-4.2.8: (combobox control) An element of the given type should not support the Scroll pattern.
 
  **Minimal repro:**

Application: WPF Application project in C# with combo box control

Issue Details
An element of the given type should not support the Scroll pattern.

Element path: combo box "test"

To Reproduce and How to Fix guidance: Please open the attached Keros Test File (*.k4dtest) to reproduce the error in Keros for Desktop.

[691457.zip](https://github.com/dotnet/wpf/files/3237825/691457.zip)

This accessibility issue was found using Keros for Desktop, a tool that helps debug and find accessibility issues earlier. Get more information & download this tool at https://aka.ms/K4D.

Screen

arpitmathur:
Fixes #791, #1015 

This pull request contains two changes.
1. Setting ScrollPattern on ComboBox to only be enabled when its dropdown is opened.
2. Ensuring that expander is added to GroupItemDataAutomationPeer if it is the only child. This ensures that the expander is accessible to screen readers.

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Automation/Peers/ComboBoxAutomationPeer.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Automation/Peers/ComboBoxAutomationPeer.cs
@@ -61,9 +61,9 @@ namespace System.Windows.Automation.Peers
             {
                 iface = this;
             }
-            else if (pattern == PatternInterface.Scroll)
+            else if (pattern == PatternInterface.Scroll && !owner.IsDropDownOpen)
             {
-                if (owner.IsDropDownOpen) iface = this;
+                iface = this;
             }
             else
             {
--- a/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Automation/Peers/ComboBoxAutomationPeer.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Automation/Peers/ComboBoxAutomationPeer.cs
@@ -51,15 +51,19 @@ namespace System.Windows.Automation.Peers
         override public object GetPattern(PatternInterface pattern)
         {
             object iface = null;
+            ComboBox owner = (ComboBox)Owner;
 
             if (pattern == PatternInterface.Value)
             {
-                ComboBox owner = (ComboBox)Owner;
                 if (owner.IsEditable) iface = this;
             }
             else if(pattern == PatternInterface.ExpandCollapse)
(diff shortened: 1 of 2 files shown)


src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Automation/Peers/ComboBoxAutomationPeer.cs
```diff
@@ -51,16 +51,20 @@ override protected string GetClassNameCore()
...
-                ComboBox owner = (ComboBox)Owner;
                 if (owner.IsEditable) iface = this;
             }
             else if(pattern == PatternInterface.ExpandCollapse)
             {
                 iface = this;
             }
+            else if (pattern == PatternInterface.Scroll)
+            {
+                if (owner.IsDropDownOpen) iface = this;
```
miguep: You shouldn't just return "this" here, you should be calling the base method, because it does not return "this" it returns the AutomationPeer for the ScrollHost. Short-circuit when the drop-down is closed, but still call the base method when it is open.

https://github.com/dotnet/wpf/blob/8c7ab1592b6c70e732106aeabf58f215d4b71c22/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Automation/Peers/ItemsControlAutomationPeer.cs#L36-L59

src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Automation/Peers/GroupItemAutomationPeer.cs
```diff
@@ -152,6 +145,22 @@ protected override List<AutomationPeer> GetChildrenCore()
                             _expanderPeer.GetChildren();
                         }
                     }
+                    Panel itemsHost = owner.ItemsHost;
+
+                    if (itemsHost == null)
+                    {
+                        if (_expanderPeer == null)
```
rladuca: Don't use unbracketed if statements

rladuca:
Branch is all fixed up now, going to sign off.

rladuca:
@msftbot require 1 sign off before merging

rladuca:
@msftbot wait 5 minutes before merging

PR #1285, opened 2019-07-16
Update localized strings that render \r\n as text

### LINKED ISSUE #977: File Dialog messages that contain \r\n render the \r\n as text

ryalanms:
**Problem description:**
When you use an Open File Dialog and provide an invalid file OR if you use a Save File Dialog and try to overwrite an existing file the messagebox renders the \r\n as text instead of  a line break.

See issue #977 for a full description. 

Confirmed these updates fix File Dialog messages in English and German.  (Fixes #977.)

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/PresentationFramework/Resources/Strings.resx
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationFramework/Resources/Strings.resx
@@ -1027,10 +1027,12 @@
     <value>Too many files selected. Select fewer files and try again.</value>
   </data>
   <data name="FileDialogCreatePrompt" xml:space="preserve">
-    <value>'{0}' does not exist.\r\nDo you want to create it?</value>
+    <value>'{0}' does not exist.
+Do you want to create it?</value>
   </data>
   <data name="FileDialogFileNotFound" xml:space="preserve">
-    <value>'{0}' does not exist.\r\nVerify that the file name is correct.</value>
+    <value>'{0}' does not exist.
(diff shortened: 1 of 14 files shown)


ryalanms:
> Did the translated files automatically adapt to this change while preserving the translated text?

No.  The xlf generation doesn't touch translated strings (target), it only adds untranslated strings to the xlfs (translation unit, source, and target) and/or removes strings (translation units) from the xlfs.  The translated strings (targets) in the xlfs need to be updated manually.

ryalanms:
> Did we do a global search for `"\\r|\\n|\\r\\n"` in `.resx` files and make sure that these are the only instances of this particular problem?

Yes.

PR #1286, opened 2019-07-16
Remove support for `UseLegacyDangerousClipboardDeserializationMode`

### LINKED ISSUE #1132: Removing support for `UseLegacyDangerousClipboardDeserializationMode and `restrictDeserialization`
Related:
 - https://github.com/dotnet/wpf/issues/1131
 - https://github.com/dotnet/corefx/issues/38760

Harden the codebase by removing support for `UseLegacyDangerousClipboardDeserializationMode and `restrictDeserialization` near https://github.com/dotnet/wpf/search?l=C%23&q=binaryformatter&type=

vatsan-madhavan:
Remove support for `UseLegacyDangerousClipboardDeserializationMode` and permanently disallow deserialization of dangerous types.

- Removes support for AppContext flag `UseLegacyDangerousClipboardDeserializationMode`
- Permanently limits clipboard-deserialziation to primitive non-text types only.

Fixes #1132

PS: Creating a WIP PR since this hasn't gone through enough testing yet.

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/dataobject.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/dataobject.cs
@@ -2882,6 +2882,10 @@ namespace System.Windows
                     // * Out of these, we will disallow deserialization of 
                     // DataFormats.StringFormat to prevent potentially malicious objects from
                     // being deserialized as part of a "text" copy-paste or drag-drop.
+                    // TypeRestrictingSerializationBinder will throw when it encounters 
+                    // anything other than strings and primitives - this ensures that we will
+                    // continue successfully deserializing basic strings while rejecting other 
+                    // data types that advertise themselves as DataFormats.StringFormat.
                     // 
                     // The rest of the following formats are pre-defined in the OS, 
                     // they are not managed objects - an so we will not attempt to deserialize them. 
--- a/src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/dataobject.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/dataobject.cs
@@ -2863,7 +2863,7 @@ namespace System.Windows
                     {
                         data = ReadBitmapSourceFromHandle(hglobal);
                     }
-                    // Limit deserialization to only primitive types, which consist of the following:
+                    // Limit deserialization to DataFormats that correspond to primitives, which are:
                     //
                     // DataFormats.CommaSeparatedValue
                     // DataFormats.FileDrop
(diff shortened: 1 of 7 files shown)


src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/dataobject.cs
```diff
@@ -2863,13 +2863,29 @@ private object GetDataFromHGLOBAL(string format, IntPtr hglobal)
...
+                    // DataFormats.Rtf
+                    // DataFormats.Serializable
+                    // DataFormats.Text
+                    // DataFormats.UnicodeText
+                    // DataFormats.WaveAudio
+                    // DataFormats.Xaml
+                    // DataFormats.XamlPackage 
+                    // DataFormats.StringFormat
+                    // 
+                    // Out of these primitive types, we will disallow deserialization of
```
weltkante: Do you actually *disallow* StringFormat or are you just restricting it to safe types? The PR doesn't seem to contain any "disallow" logic and the existing code just seems to use `formatter.Binder = new TypeRestrictingSerializationBinder();` so I don't quite see how that disallows `DataFormats.StringFormat`.

Might want to fix comment to be more clear what actually happens.

src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/dataobject.cs
```diff
@@ -2863,13 +2863,29 @@ private object GetDataFromHGLOBAL(string format, IntPtr hglobal)
...
+                    // DataFormats.Text
+                    // DataFormats.UnicodeText
+                    // DataFormats.WaveAudio
+                    // DataFormats.Xaml
+                    // DataFormats.XamlPackage 
+                    // DataFormats.StringFormat
+                    // 
+                    // Out of these primitive types, we will disallow deserialization of 
+                    // DataFormats.StringFormat to prevent potentially malicious objects from
                     // being deserialized as part of a "text" copy-paste or drag-drop.
```
weltkante: Do you actually *disallow* StringFormat or are you just restricting it to safe types? The PR doesn't seem to contain any "disallow" logic and the existing code just seems to use `formatter.Binder = new TypeRestrictingSerializationBinder();` so I don't quite see how that disallows `DataFormats.StringFormat`.

Might want to fix comment to be more clear what actually happens.

miguep: I agree that the comment is not accurate. Here's some context on the scenario:

DataFormats is just the "bucket" in which the Clipboard data is set. Nothing enforces the type of data that is put in each bucket, so an application can choose to put a serialized object (of any type, not just a string) in the StringFormat bucket. That doesn't mean WPF should deserialize such objects, a malicious object could be serialized and put under the StringFormat object, for example, and WPF would then deserialize it. The way TypeRestrictingSerializationBinder works is, it throws when attempting to deserialize anything other than strings and primitives. This functionality was behind a compat flag, which is being removed in this PR.


The comment is mixing up the concepts of DataFormats and types.

miguep:
This looks good to me aside from the comment in DataObject.cs

src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/dataobject.cs
vatsan-madhavan: Thanks @miguep and @weltkante. I'll update the comment.

Issue #1307, opened 2019-07-18
Update WPF structs to compute more inclusive hashes


ojhad:
In CoreCLR, the GetHashCode logic for structs with float/double parameters has changed and now only uses the first field for hash computation (see [this](https://github.com/dotnet/corefx/issues/35613) for more details).
As [recommended](https://github.com/dotnet/corefx/issues/35613#issuecomment-467726210), we should go through these structs to override and implement our own hashes. These new hashes should be more inclusive in order to better mirror what the behaviour was like in .NET Framework. There is already an example of this being done [here]() for the MILColorF struct.

rladuca:
@ojhad You marked this Future, do you see any problems arising from this that would potentially need resolution prior to GA?  Should this be a 3.0 work item?

Issue #848, opened 2019-06-04
2D test failing due to change in GetHashCode in Core 3


grubioe:
Copied from VSO ID: 803447

Test:

```
Name="Color API Test" Priority="1" Area="2D" SubArea="API Tests"
```

Fails due to a change in GetHashCode from corefx. What the test code is doing is generating 2 different colors and verifying that the resulting hash codes are not equal, the hashcode is calculated from System.Windows.Media.Color by overriding GetHashCode:

https://github.com/dotnet/wpf/blob/9c5dd1acfe9037f92f6c91b05ab2f7e18edf55b7/src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Media/Color.cs#L214-L220

which essentially only returns the hashcode from its internal struct which is defined as:

https://github.com/dotnet/wpf/blob/9c5dd1acfe9037f92f6c91b05ab2f7e18edf55b7/src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Media/Color.cs#L1158-L1161

in Framework different values for a,r,g,b would result in a different HashCode in Core it seems only the first value is considered for HashCode calculation, our test has the same first value so the hash code is the same.

As per corefx recommendation (https://github.com/dotnet/corefx/issues/35613) we should override and implement a correct HashCode mechanism for our 4 floats in MILColorF, ideally mirroring what Framework used to do

weltkante:
Was it verified in your original ticket source that this isn't actually a bug in .NET Core? Otherweise I think it'd be worth opening an issue with coreclr to verify that this behavior is intended.

I don't mind proper GetHashCode implementations in WPF but if thats an unintended bug the poor hashcode quality will likely impact users who port their applications as well.

weltkante:
@grubioe actually I cannot reproduce this, I took above definition of `MILColor` and created instances which differ only in a single component. Regardless which component differs, the hash code always differs too for me.

```C#
Debug.WriteLine($"Hash = {new MILColor { a = 0xFF, r = 0xAA, g = 0xBB, b = 0xCC }.GetHashCode()}");
Debug.WriteLine($"Hash = {new MILColor { a = 0x00, r = 0xAA, g = 0xBB, b = 0xCC }.GetHashCode()}");
Debug.WriteLine($"Hash = {new MILColor { a = 0xFF, r = 0x00, g = 0xBB, b = 0xCC }.GetHashCode()}");
[...]

PR #1308, opened 2019-07-18
Updated GetHashCode for MILColorF structs

ojhad:
Issue #848 

In CoreCLR, the GetHashCode logic for the MILColorF struct (struct with float values) changed and only uses the first field for hash computation (see [this](https://github.com/dotnet/corefx/issues/35613) for more details). Added custom GetHashCode logic to the MILColorF struct in order to have the GetHashCode computation use all four float values for hash computation (as per recommendation from [here](https://github.com/dotnet/corefx/issues/35613#issuecomment-467726210)).
Opened a [complementary PR in the internal repo](https://dnceng.visualstudio.com/internal/_git/dotnet-wpf-int/pullrequest/2196?_a=overview) to update the MILColorF struct definitions there as well.

- Verified that the failing 2D color test is now passing.
- Ran a full DRT test pass and verified that DRTs are at baseline.
- Logged a new issue for the updating of other structs in the codebase that are affected by the same change [here](https://github.com/dotnet/wpf/issues/1307).

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/Common/Graphics/wgx_core_types.cs
+++ b/src/Microsoft.DotNet.Wpf/src/Common/Graphics/wgx_core_types.cs
@@ -454,6 +454,15 @@ internal struct MilColorF
     internal float g;
     internal float b;
     internal float a;
+
+    public override int GetHashCode()
+    {
+        return a.GetHashCode() ^ r.GetHashCode() ^ g.GetHashCode() ^ b.GetHashCode();
+    }
+    public override bool Equals(object obj)
+    {
+        return base.Equals(obj);
+    }
 };
 
 /// <summary>
--- a/src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Media/Color.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Media/Color.cs
@@ -1158,6 +1158,15 @@ namespace System.Windows.Media
         private struct MILColorF // this structure is the "milrendertypes.h" structure and should be identical for performance
         {
             public float a, r, g, b;
+
+            public override int GetHashCode()
+            {
+                return a.GetHashCode() ^ r.GetHashCode() ^ g.GetHashCode() ^ b.GetHashCode();
+            }
+            public override bool Equals(object obj)
+            {
+                return base.Equals(obj);
+            }
         };
 
         private MILColorF scRgbColor;


src/Microsoft.DotNet.Wpf/src/Common/Graphics/wgx_core_types.cs
```diff
@@ -454,6 +454,15 @@ internal struct MilColorF
     internal float g;
     internal float b;
     internal float a;
+
+    public override int GetHashCode()
+    {
+        return a.GetHashCode() ^ r.GetHashCode() ^ g.GetHashCode() ^ b.GetHashCode();
+    }
+    public override bool Equals(object obj)
```
vatsan-madhavan: Because `MilColorF` is tightly-packed, we'll get byte-wise comparison by default. 

Even if there are minor differences in the float values between different instances of the same color value, I'd imagine that `Equals` would return `false`. Do you think this behavior is probably different between .NET Framework and .NET Core due to CoreClr differences, and needs to be remedied with a better implementation for `Equals` - perhaps something like this: 


```C#
public override bool Equals(object obj)
{
    if (Object.ReferenceEquals(this, obj))
    {
        return true;
    }
    
    if (obj is MilColorF color)
    {
        return Equals(color);
    }

    return base.Equals(obj);
}

public bool Equals(MilColorF color)
{
    return DoubleUtil.AreClose(color.r, r) &&
           DoubleUtil.AreClose(color.g, g) && 
           DoubleUtil.AreClose(color.b, b) &&
           DoubleUtil.AreClose(color.a, a);
}
```

/cc @rladuca

src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Media/Color.cs
```diff
@@ -1158,6 +1158,15 @@ private void ComputeNativeValues(int numChannels)
         private struct MILColorF // this structure is the "milrendertypes.h" structure and should be identical for performance
```
vatsan-madhavan: Not marked as `Pack=1`. Does this really get appropriate treatment from `ValueType.Equals()`  ?

src/Microsoft.DotNet.Wpf/src/Common/Graphics/wgx_core_types.cs
```diff
@@ -454,6 +454,15 @@ internal struct MilColorF
     internal float g;
     internal float b;
     internal float a;
+
+    public override int GetHashCode()
+    {
+        return a.GetHashCode() ^ r.GetHashCode() ^ g.GetHashCode() ^ b.GetHashCode();
```
rladuca: Just to clarify for me, we don't care about uniqueness for these hash codes?

vatsan-madhavan: If two colors compare as equal, hashes would be equal. If two colors are not  equal, the would be different. 

Are you thinking about uniqueness in a different sense ?

src/Microsoft.DotNet.Wpf/src/Common/Graphics/wgx_core_types.cs
vatsan-madhavan: I don’t think we should do this. This will lead to “equal" colors with unequal hashes. We’ll need to redesign GetHashCode if we wanted to use `AreClose` etc.

src/Microsoft.DotNet.Wpf/src/Common/Graphics/wgx_core_types.cs
weltkante: If two colors are different the hash codes could be the same, the XOR can cancel out bits. Theres nothing you can do to prevent that, the hash code range has just 32 bit and four floating point values are definitely above 32 bit even if you restrict them to [0,1] ranges or something along those lines.

MilColorF can't have unique hash codes, you were probably thinking about integer based colors, bat that GetHashCode didn't have a bug in the first place because the struct was just 32 bit to begin with and got hashed correctly by coreclr as one unit.

src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Media/Color.cs
weltkante: Isn't pack just for alignment of fields within the struct when you mix fields of different sizes? I don't think the default packing will leave gaps for 4 floats on any platform.

[doc](https://docs.microsoft.com/en-us/dotnet/api/system.runtime.interopservices.structlayoutattribute.pack) says `Each field must align with fields of its own size (1, 2, 4, 8, etc., bytes) or the alignment of the type, whichever is smaller.` In other words the Pack should only reduce gaps not create them. If all your fields have the same size than Pack is irrelevant as far as I'm aware.

src/Microsoft.DotNet.Wpf/src/Common/Graphics/wgx_core_types.cs
mikedn: Normally I'd expect `Equals` to do `color.r == r && color.g == g ...` and not use something like `AreClose`. Bitwise equality might work but it may be confusing at times because it doesn't account for NaNs and other floating point particularities and as implemented now it's probably pretty slow as well. Though it all depends on how theses structs are used...

mikedn:
What happens in `Color` is pretty confusing. `Color` has custom `Equals` (quite a complicated one) but its `GetHashCode` delegates to `MILColorF`'s `GetHashCode`. It would make more sense to add an explicit `GetHashCode` implementation to `Color`'s `GetHashCode` (because that's a public API and what the test is testing) and leave `MILColorF` alone as that's a private implementation detail.

Same for `MilColorF`, that's a internal struct used to talk to wpfgfx.

src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Media/Color.cs
rladuca: There are checks built into ValueType equality that help determine whether to reflect or use byte-comparisons.  The MethodTable contains the flags for all of this I believe.

See https://github.com/dotnet/coreclr/blob/a64cb0a41a4ebeb4a61b8b0f1f2eeeff8dd539c9/src/System.Private.CoreLib/src/System/ValueType.cs#L44

And follow the chain into the implementations of CanCompareBits and FastEqualsCheck.

These are related to how packed the struct is, see https://github.com/dotnet/coreclr/blob/54d9d9b2698c9628f43494e46f855996b31838bd/src/vm/comutilnative.cpp#L2063.

It's slightly hard to see where all of this gets set and the exact criteria.  The MSDN documentation seems to say as long as none of the fields are reference types, you can use the fast comparisons.  But I am not 100% sure that is true from reading the above.

src/Microsoft.DotNet.Wpf/src/Common/Graphics/wgx_core_types.cs
rladuca: Yeah that makes sense, there is no guaranteeing uniqueness here given the dataset.

src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Media/Color.cs
rladuca: I also see this: https://github.com/dotnet/coreclr/blob/0bdf613f09a3bf409717beb939195b39e40c6828/src/vm/class.cpp#L413

And:
https://github.com/dotnet/coreclr/blob/5885379686cd854361735a09907d654e80f44f05/src/vm/class.h#L1946

So I'm not sure if fast comparisons are turned off all the time or not.

vatsan-madhavan: @ojhad, I approve the PR and this work need to be merge tonight, and since this is about GetHashCod - but I didn't get a good response for this question. As @rladuca points out, the checks for `Pack` are baked into the ValueType equality.... please open an issue to investigate this before merging the PR/closing this comment.

ojhad: Created Issue for follow up:
https://github.com/dotnet/wpf/issues/1338

ojhad:
> What happens in `Color` is pretty confusing. `Color` has custom `Equals` (quite a complicated one) but its `GetHashCode` delegates to `MILColorF`'s `GetHashCode`. It would make more sense to add an explicit `GetHashCode` implementation to `Color`'s `GetHashCode` (because that's a public API and what the test is testing) and leave `MILColorF` alone as that's a private implementation detail.
> 
> Same for `MilColorF`, that's a internal struct used to talk to wpfgfx.

That's a very good point. We had an internal discussion on this alternative and decided to move forward with the current implementation..

PR #1311, opened 2019-07-18
Convert non-seekable PackagePart streams to seekable

### LINKED ISSUE #585: Behavior change regarding seekability of streams returned for ZipArchive

rladuca:
Fixes #585

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/Shared/MS/Internal/IO/Packaging/PackagePartExtensions.cs
+++ b/src/Microsoft.DotNet.Wpf/src/Shared/MS/Internal/IO/Packaging/PackagePartExtensions.cs
@@ -1,0 +1,45 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using MS.Internal.WindowsBase;
+using System.IO;
+using System.IO.Packaging;
+
+namespace MS.Internal.IO.Packaging
+{
+    [FriendAccessAllowed]
+    internal static class PackagePartExtensions
+    {
+        internal static Stream GetSeekableStream(this PackagePart packPart)
+        {
+            return GetSeekableStream(packPart, FileMode.OpenOrCreate, packPart.Package.FileOpenAccess);
+        }
+
+        internal static Stream GetSeekableStream(this PackagePart packPart, FileMode mode)
+        {
+            return GetSeekableStream(packPart, mode, packPart.Package.FileOpenAccess);
+        }
+
+        internal static Stream GetSeekableStream(this PackagePart packPart, FileMode mode, FileAccess access)
+        {
+            var packStream = packPart.GetStream(mode, access);
+
+            // If the stream returned is seekable it meets all requirements and can be used directly.
+            if (packStream.CanSeek)
+            {
+                return packStream;
+            }
+
+            // Non-seekable streams need to be copied out into memory so they are seekable.
+            using (packStream)
+            {
+                var seekableStream = new MemoryStream((int)packStream.Length);
+
+                packStream.CopyTo(seekableStream);
+
+                return seekableStream;
+            }
+        }
+    }
+}
--- a/src/Microsoft.DotNet.Wpf/src/Shared/MS/Internal/IO/Packaging/PackagePartExtensions.cs
+++ b/src/Microsoft.DotNet.Wpf/src/Shared/MS/Internal/IO/Packaging/PackagePartExtensions.cs
@@ -8,19 +8,55 @@ using System.IO.Packaging;
 
 namespace MS.Internal.IO.Packaging
 {
+    /// <summary>
+    /// Extensions to provide wrappers for functionality that no longer exists in System.IO.Packaging.PackagePart
+    /// </summary>
     [FriendAccessAllowed]
     internal static class PackagePartExtensions
     {
+        /// <summary>
+        /// Gets a seekable stream from the PackagePart.
+        /// <see cref="GetSeekableStream(PackagePart, FileMode, FileAccess)"/> for details.
+        /// </summary>
+        /// <param name="packPart"></param>
+        /// <returns>A seekable stream representing the data in the PackagePart.</returns>
         internal static Stream GetSeekableStream(this PackagePart packPart)
         {
             return GetSeekableStream(packPart, FileMode.OpenOrCreate, packPart.Package.FileOpenAccess);
         }
 
+        /// <summary>
+        /// Gets a seekable stream from the PackagePart.
+        /// <see cref="GetSeekableStream(PackagePart, FileMode, FileAccess)"/> for details.
+        /// </summary>
+        /// <param name="packPart"></param>
+        /// <param name="mode">The FileMode to open the PackagePart</param>
+        /// <returns>A seekable stream representing the data in the PackagePart.</returns>
         internal static Stream GetSeekableStream(this PackagePart packPart, FileMode mode)
         {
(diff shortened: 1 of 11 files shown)


src/Microsoft.DotNet.Wpf/src/Shared/MS/Internal/IO/Packaging/PackagePartExtensions.cs
```diff
@@ -0,0 +1,85 @@
...
+using MS.Internal.WindowsBase;
+using System.IO;
+using System.IO.Packaging;
+
+namespace MS.Internal.IO.Packaging
+{
+    /// <summary>
+    /// Extensions to provide wrappers for functionality that no longer exists in System.IO.Packaging.PackagePart
+    /// </summary>
+    [FriendAccessAllowed]
```
vatsan-madhavan: Is this only used by WindowsBase? If so, does this need to be `FriendAccessAllowed`?

src/Microsoft.DotNet.Wpf/src/Shared/MS/Internal/IO/Packaging/PackagePartExtensions.cs
```diff
@@ -0,0 +1,85 @@
...
+            // Non-seekable streams need to be copied out into memory so they are seekable.
+            using (packStream)
+            {
+                var seekableStream = new MemoryStream((int)packStream.Length);
+
+                packStream.CopyTo(seekableStream);
+
+                // Reset the stream to the beginning.  If this is not done, attempts to read the stream 
+                // from the current position will fail.  E.G. XAML/XML parsing.
+                seekableStream.Position = 0;
```
vatsan-madhavan: seekableStream.Seek(0, SeekOrigin.Begin)

src/Microsoft.DotNet.Wpf/src/Shared/MS/Internal/IO/Packaging/PackagePartExtensions.cs
rladuca: It's included by base, used by core and framework.  I wasn't going to add it to shared, but the other utility file is already there (also only included by base and used by core and framework as well) so I figured put it there for now.

src/Microsoft.DotNet.Wpf/src/Shared/MS/Internal/IO/Packaging/PackagePartExtensions.cs
rladuca: This is generally equivalent, is there a specific reason you prefer seek?  There are other places in the code where we use Position over a Seek call.

vatsan-madhavan: Didn’t notice earlier we used Position elsewhere. Yes let’s use that for consistency.

src/Microsoft.DotNet.Wpf/src/WindowsBase/MS/Internal/IO/Packaging/Certificate.cs
```diff
@@ -131,7 +131,7 @@ internal void SetCertificate(X509Certificate2 certificate)
             Byte[] byteArray = _certificate.GetRawCertData();
 
             // FileMode.Create will ensure that the stream will shrink if overwritten
-            using (Stream s = _part.GetStream(FileMode.Create, FileAccess.Write))
+            using (Stream s = _part.GetSeekableStream(FileMode.Create, FileAccess.Write))
```
weltkante: who writes back the writable stream?

src/Microsoft.DotNet.Wpf/src/WindowsBase/MS/Internal/IO/Packaging/XmlDigitalSignatureProcessor.cs
```diff
@@ -937,7 +937,7 @@ private Stream GetRelationshipStream(PartManifestEntry partEntry)
         private void UpdatePartFromSignature(Signature sig)
         {
             // write to stream
-            using (Stream s = SignaturePart.GetStream(FileMode.Create, FileAccess.Write))
+            using (Stream s = SignaturePart.GetSeekableStream(FileMode.Create, FileAccess.Write))
```
weltkante: who writes back the writable stream?

src/Microsoft.DotNet.Wpf/src/Shared/MS/Internal/IO/Packaging/PackagePartExtensions.cs
```diff
@@ -0,0 +1,85 @@
...
+        /// This is, generally, what would be the case in .NET Framework.
+        /// 
+        /// Note that if the stream returned is seekable (the <see cref="PackagePart"/> was opened write or read-write) then we just pass the resulting
+        /// stream back as we're already guaranteed it meets our needs.
+        /// </remarks>
+        /// <param name="packPart"></param>
+        /// <param name="mode">The FileMode to open the PackagePart</param>
+        /// <param name="access">The FileAccess used to open the PackagePart</param>
+        /// <returns>A seekable stream representing the data in the PackagePart.</returns>
+        internal static Stream GetSeekableStream(this PackagePart packPart, FileMode mode, FileAccess access)
```
weltkante: Does it even make sense to specify FileMode/FileAccess, shouldn't it always be open/read? When wrapping a writable stream into a `MemoryStream` who will be forwarding the writes to the original stream?

rladuca:
> After seeing #1363 I realize this PR may not have considered the consequences of wrapping writable streams. If this is just wrapping in a MemoryStream then there needs to be someone who transfers the data to the original stream, otherwise the writes will never happen.

Look at the implementation of the extension method.  A writeable stream is already seekable when it comes back from PackagePart.GetStream.  We just forward those streams onto the caller and do not read them out.  The only case where PackagePart.GetStream returns a non-seekable stream is in read-only scenarios.  That is precisely why the workaround is to open your XPS or Package as write or read-write.

src/Microsoft.DotNet.Wpf/src/Shared/MS/Internal/IO/Packaging/PackagePartExtensions.cs
```diff
@@ -0,0 +1,85 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using MS.Internal.WindowsBase;
```
stevenbrix: Can you move the other `PackagePart` extension method in WindowsBase\MS\Internal\IO\Packaging\PackagingExtensions.cs to this file?

```c#
public static ContentType ValidatedContentType(this PackagePart packagePart)
{
      return new ContentType(packagePart.ContentType);
}
```

PR #1339, opened 2019-07-23
Suppress warnings generated by native-tools installation.

vatsan-madhavan:
See https://github.com/dotnet/arcade/pull/3315


vatsan-madhavan:
@msftbot require 0 signoffs

PR #1387, opened 2019-07-25
Linker hints for C++/CLI dependencies 

### LINKED ISSUE #1385: Trimmed images of WPF applications crash
Project: 
```xml

  <PropertyGroup>
    <TargetFramework>netcoreapp3.0</TargetFramework>
    <UseWPF>true</UseWPF>
    <PublishTrimmed>true</PublishTrimmed>
    <PublishReadyToRun>true</PublishReadyToRun>
    <PublishSingleFile>true</PublishSingleFile>
    <RuntimeIdentifier>win-x64</RuntimeIdentifier>
    <Platforms>x64</Platforms>
    <OutputPath>bin\X64\Release\</OutputPath>
    <AssemblyVersion>1.0.1.7144</AssemblyVersion>
    <FileVersion>1.0.1.7144</FileVersion>
  </PropertyGroup>

</Project>
```
MainWindow.xaml
```xaml
 <Window x:Class="wpf1.MainWindow"
        xmlns:local="clr-namespace:wpf1"
        mc:Ignorable="d"
    <Grid>
        <Button x:Name="hello" Width="100" Height="35"></Button>
    </Grid>
</Window>
````

build this using a recent preview8 SDK like this: `dotnet publish -r win-x64 -c release `, then run the exe produced under `bin\x64\release\netcoreapp3.0\win-x64\publish\`

global.json
```json
{
  "sdk": {
    "version": "3.0.100-preview8-013417"
  }
}
```

Observed: Crash - `FileNotFoundException` for `System.Diagnostics.Debug.dll`

vatsan-madhavan:
Fixes #1385 

When producing ReadyToRun images, the ILLinker is configured to skip C++/CLI images. See mono/linker#651 and mono/linker#658.

In turn, this results in a failure of dependencies of such assemblies (like System.Diagnostics.Debug.dll, which is required by DirectWriteForwarder.dll) from being identified and included in the ReadyToRun images.

These linker hints tell ILLinker to include certain dependencies, which we know to be required by DirectWriteForwarder and System.Printing respectively.

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/PresentationCore/ILLinkTrim.xml
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationCore/ILLinkTrim.xml
@@ -1,0 +1,5 @@
+<linker>
+  <!-- Include dependencies of DirectWriteForwarder, which is not
+       analyzed by the linker. -->
+  <assembly fullname="System.Diagnostics.Debug" />
+</linker>
--- a/src/Microsoft.DotNet.Wpf/src/PresentationCore/ILLinkTrim.xml
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationCore/ILLinkTrim.xml
@@ -2,4 +2,6 @@
   <!-- Include dependencies of DirectWriteForwarder, which is not
        analyzed by the linker. -->
   <assembly fullname="System.Diagnostics.Debug" />
+  <assembly fullname="mscorlib" />
+  <assembly fullname="netstandard" />  
 </linker>
--- a/src/Microsoft.DotNet.Wpf/src/ReachFramework/ILLinkTrim.xml
+++ b/src/Microsoft.DotNet.Wpf/src/ReachFramework/ILLinkTrim.xml
@@ -2,4 +2,6 @@
   <!-- Include dependencies of System.Printing, which is not
        analyzed by the linker. -->
   <assembly fullname="System.Diagnostics.Debug" />
+  <assembly fullname="mscorlib" />
+  <assembly fullname="netstandard" />
 </linker>
--- a/src/Microsoft.DotNet.Wpf/src/ReachFramework/ILLinkTrim.xml
+++ b/src/Microsoft.DotNet.Wpf/src/ReachFramework/ILLinkTrim.xml
@@ -1,0 +1,5 @@
+<linker>
+  <!-- Include dependencies of System.Printing, which is not
+       analyzed by the linker. -->
+  <assembly fullname="System.Diagnostics.Debug" />
+</linker>
--- a/src/Microsoft.DotNet.Wpf/src/PresentationCore/PresentationCore.csproj
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationCore/PresentationCore.csproj
@@ -1448,6 +1448,29 @@
       <XlfInput>false</XlfInput>
       <ManifestResourceName>ColorProfiles</ManifestResourceName>
     </EmbeddedResource>
+    
+    <!-- 
+        Workaround for https://github.com/dotnet/wpf/issues/1385
+        ReadyToRun images of WPF applications crash 
+        
+        When producing ReadyToRun images, the ILLinker is configured to skip 
+        C++/CLI images. See https://github.com/mono/linker/issues/651 and 
+        https://github.com/mono/linker/pull/658. 
+        
+        In turn, this results in the failure of the dependencies of such assemblies from being
+        identified and included in the ReadyToRun images. 
+        
+        This linker hint exists to ensure that such dependencies are included. We need
+        to place these hints in an assembly that has a reference to the corresponding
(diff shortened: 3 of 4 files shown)


vatsan-madhavan:
/cc @fadimounir 

I don't see this fixing the orignal problem @sbomer. The xml files get embedded into PresentationCore & ReachFramework respectively. The published apps continue to fail as before - `FileNotFoundException` missing `System.Diagnostics.Debug`


This is a screenshot of assemblies from within the temporary unzipped location, while running the published app under the debugger. 

image

sbomer:
I confirmed that the linker fix in https://github.com/mono/linker/pull/675 keeps `System.Diagnostics.Debug`. @vatsan-madhavan do you think we should also root `netstandard` and `mscorlib`? I saw them referenced from the following (but haven't actually seen a failure related to these references):

```
from DirectWriteForwarder:
[mscorlib]System.Runtime.CompilerServices.AssemblyAttributesGoHere
[mscorlib]System.Runtime.CompilerServices.AssemblyAttributesGoHereM
[netstandard]System.URi

from System.Printing:
[mscorlib]System.Runtime.CompilerServices.AssemblyAttributesGoHere
[mscorlib]System.Runtime.CompilerServices.AssemblyAttributesGoHereM
[mscorlib]System.Security.SuppressUnmanagedCodeSecurityAttribute
[netstandard]System.IO.FileMode
```

Otherwise the change LGTM.

vatsan-madhavan:
> [mscorlib]System.Runtime.CompilerServices.AssemblyAttributesGoHere
> [mscorlib]System.Runtime.CompilerServices.AssemblyAttributesGoHereM

I believe these are bad type-refs. There are no such types in `mscorlib`. In addition to these, there are a couple more bad type-refs (unrelated to this issue)

```
[System.Runtime]System.Runtime.CompilerServices.SuppressMergeCheckAttribute
[System.Runtime]System.Runtime.CompilerServices.DecoratedNameAttribute
[System.Runtime]System.Runtime.CompilerServices.AssemblyAttributesGoHere
```

sbomer:
Are the netstandard refs also bad? I'm seeing the following in MS.Internal.TrueTypeSubsetter.ComputeSubset:
```
IL_0063:  newobj     instance void [System.IO.Packaging]System.IO.FileFormatException::.ctor(class [netstandard_152]System.Uri)
```

vatsan-madhavan:
> Are the netstandard refs also bad? I'm seeing the following in MS.Internal.TrueTypeSubsetter.ComputeSubset:
> 
> ```
> IL_0063:  newobj     instance void [System.IO.Packaging]System.IO.FileFormatException::.ctor(class [netstandard_152]System.Uri)
> ```

That looks fine to me. I see a TypeForwardedTo(typeof(Uri)) in netstandard, pointing to System.Runtime.

src/Microsoft.DotNet.Wpf/src/PresentationCore/ILLinkTrim.xml
```diff
@@ -0,0 +1,7 @@
+<linker>
+  <!-- Include dependencies of DirectWriteForwarder, which is not
+       analyzed by the linker. -->
+  <assembly fullname="System.Diagnostics.Debug" />
+  <assembly fullname="mscorlib />
+  <assembly fullname="netstandard />
```
rladuca: Missing end quotes

src/Microsoft.DotNet.Wpf/src/ReachFramework/ILLinkTrim.xml
```diff
@@ -0,0 +1,7 @@
+<linker>
+  <!-- Include dependencies of System.Printing, which is not
+       analyzed by the linker. -->
+  <assembly fullname="System.Diagnostics.Debug" />
+  <assembly fullname="mscorlib />
+  <assembly fullname="netstandard />
```
rladuca: Missing end quotes

ericstj:
I'm not a fan of this hack.  What keeps this list up to date?  The linker should be able to traverse the C++/CLI assembly without rewriting it.

sbomer:
Not a fan either, and I definitely hope to analyze them in the future and remove the roots. The alternative that I see to adding these roots now, short of fixing https://github.com/mono/linker/issues/639, is to document a requirement for pretty much every linked WPF app to include these roots. Thoughts?

Also, @vatsan-madhavan, based on discussion with @marek-safar it probably makes sense to make these embedded roots more specific, to only root the types that are needed by these assemblies. I need to check the behavior of type forwards, but I think we'll want to update them.

sbomer:
Another alternative would be to include roots for System.Diagnostics.Debug but not netstandard/mscorlib. System.Diagnostics.Debug was being loaded for every app that renders text (see https://github.com/mono/linker/pull/675#issuecomment-515273404), but I wasn't seeing any failures for the mscorlib/netstandard dependencies, so they may be less common.

PR #1501, opened 2019-08-01
WIP PR: Fixing hit detection in controls in system aware dialogs in mi…

### LINKED ISSUE #968: ComboBox drop down opens in PerMonitor awareness when hosted in System aware UI
Copied from VSO ID:803149

Application Name:  Mathcad Prime

OS: Windows 10 RS4

CPU: X64

.NET Build Number: 4.8.03707

Min Repro Project:
See VSO ID

Repro Steps:

Have .NET 4.7.2+, and 100% primary, 200% secondary
Launch repro proj exe on primary
Move it to secondary
Click “Launch Dialog”
Click ComboBox in dialog (Or tab to it and hit Alt+DownArrow if using keyboard)
 

Expected:

    Popup is aligned correctly, and system scaled

 

Actual:

    Popup is misaligned and PMA scaled

image

arpitmathur:
This is a WIP PR for an issue that was reported with bad hit detection in system aware dialogs hosted in mixed mode DPI

Fixes #968

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/Shared/MS/Internal/CulturePreservingExecutionContext.cs
+++ b/src/Microsoft.DotNet.Wpf/src/Shared/MS/Internal/CulturePreservingExecutionContext.cs
@@ -27,6 +27,9 @@ using System;
 using System.Globalization;
 using System.Security;
 using System.Threading;
+using MS.Utility;
+using MS.Win32;
+using TemporaryTesting;
 
 namespace MS.Internal
 {
@@ -281,7 +284,7 @@ namespace MS.Internal
         private class CultureAndContextManager
         {
             #region Constructor 
-
+            [SecurityCritical]
             private CultureAndContextManager(ContextCallback callback, object state)
             {
                 Callback = callback;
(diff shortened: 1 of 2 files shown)


src/Microsoft.DotNet.Wpf/src/Shared/MS/Internal/CulturePreservingExecutionContext.cs
```diff
@@ -281,7 +284,7 @@ public void Dispose()
         private class CultureAndContextManager
         {
             #region Constructor 
-
+            [SecurityCritical]
```
wjk: Why are we adding CAS annotations in this PR? I thought the plan was to eventually get rid of them all.

rladuca: This is a work in progress port from some ideas that were kicking around in .NET Framework for a fix.

We moved this off to .NET 5 and @arpitmathur is far from done with this in multiple ways.  I wouldn't look at this as anything of note at this point.

Issue #1379, opened 2019-07-25
Two-way binding modifies value of read-only property


pavelovcharov:
* .NET Core Version: 3.0.100-preview7-012821
* Windows version: 1903 (18262.53)
* Does the bug reproduce also in WPF for .NET Framework 4.8?: No

 
 
 **Problem description:**
The two-way binding now changes properties with non-public setters. Is this the expected behavior? .NET Framework 472 throws InvalidOperationException, when you're trying to set two-way binding to read-only property:
```
System.InvalidOperationException: 'A TwoWay or OneWayToSource binding cannot work on the read-only property 'ReadonlyProperty' of type 'Test.ObjectWithReadonlyProperty'.'
```

 **Actual behavior:** 
The two way binding modifies property value even if it has a non-public setter
 **Expected behavior:**
The two way binding shouldn't modify read-only properties. The InvalidOperationException should be thrown.
 **Minimal repro:**
 ```
    class ObjectWithReadonlyProperty : INotifyPropertyChanged {
        string readonlyProperty;
        public ObjectWithReadonlyProperty(string v) {
            readonlyProperty = v;
        }
        public string ReadonlyProperty { get { return readonlyProperty; } private set { readonlyProperty = value; } }
        public event PropertyChangedEventHandler PropertyChanged;
        public void RaisePropertyChanged() {
            PropertyChanged(this, new PropertyChangedEventArgs(null)); //-V3083
        }
    }
    [TestFixture]
    public class TestFixture {
        [Test]
        public void TwoWayBindingToReadonlyProperty() {
            var btn = new Button();
            var data = new ObjectWithReadonlyProperty("12345");
            BindingOperations.SetBinding(btn, Button.ContentProperty, new Binding(nameof(ObjectWithReadonlyProperty.ReadonlyProperty)) { Mode = BindingMode.TwoWay, Source = data });
            Assert.AreEqual("12345", btn.Content);
            btn.Content = "67890";
            Assert.AreEqual("12345", data.ReadonlyProperty);
        }
    }
```

PR #1502, opened 2019-08-01
Removed HandleTwoWayBindingToPropertyWithNonPublicSetter compat flag

ojhad:
Issue number: #1379 

Two-way binding to properties with non-public setters was being allowed because we took on behavior that was caused by a bug in .NET Framework 4.5. Back then, a compatibility flag was introduced due to this 4.5 bug. We no longer want to support this scenario because we want correct behavior in .NET Core and we want the behavior to be on parity with net472/8. So there is no longer need for the compat flag.

Ran a full DRT pass against this change and it was successful.

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/PresentationFramework/MS/Internal/Data/PropertyPathWorker.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationFramework/MS/Internal/Data/PropertyPathWorker.cs
@@ -1684,47 +1684,8 @@ namespace MS.Internal.Data
             }
 
             // if we get here, the property has CanWrite=true, and a non-public
-            // setter.  (We presume that pi.SetValue will actually call that
-            // setter, but there's no way to tell for sure.)   In 4.0, we threw
-            // an exception, but in 4.5 we accidentally allowed writing to such
-            // a property. A compat flag guides how we proceed now.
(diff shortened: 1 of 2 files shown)


weltkante:
Related to Issue #1379 (wasn't auto-linked properly due to missing '#' sign in PR description)

vatsan-madhavan:
@ojhad there is a corresponding sample in Microsoft/wpf-samples that needs a minor-fixup.

vatsan-madhavan:
@weltkante fixed in the main description. Deliberately not using "Fixes " notation in PR's for a couple of weeks. We don't want the linked issues to be auto-closed as soon as the PR merges since the issues are being reviewed offline by .NET Core shiproom...

vatsan-madhavan:
> When we spoke, I thought Vatsan wanted to leave the flag in place and merely change the default to Throw, to ease migration of 4.5 apps that depended on the insecure behavior. But I like this better - it removes the security hole altogether, and forces migrators to actually fix their apps.

That was what we were thinking originally, but after further thought and taking into account your position that this was one of the rare instances where breaking compatibility (which really only applies under a narrow set of conditions - apps targeting .NET 4.5 **and** relying on a bug in a way that is really bad) is probably the right thing to do, we decided to fix it the right way.

PR #1513, opened 2019-08-02
XPS: Use original font when subsetting would grow the font

### LINKED ISSUE #1511: XpsSerialization in presentationcore.dll throws System.IO.FileFormatException
When working with XPS documents WPF will subset fonts in order to reduce their size during embedding.  In certain cases, this can result in the font growing during subsetting.

This results in us returning [ERR_WOULD_GROW](https://github.com/dotnet/wpf/blob/ae1790531c3b993b56eba8b1f0dd395a3ed7de75/src/Microsoft.DotNet.Wpf/src/DirectWriteForwarder/CPP/TrueTypeSubsetter/TtfDelta/ttferror.h):
https://github.com/dotnet/wpf/blob/e4a212b4fd7d3bb6bce7d88cf6f0471a9fc8e6f4/src/Microsoft.DotNet.Wpf/src/DirectWriteForwarder/CPP/TrueTypeSubsetter/TtfDelta/ttfdelta.cpp#L920

We already know what to do in this case (as the comments in the code above suggest), use the original font.  However, we seem to be ignoring that:

https://github.com/dotnet/wpf/blob/70da708a5cb79eaccda5e924687c827393e144ba/src/Microsoft.DotNet.Wpf/src/DirectWriteForwarder/CPP/TrueTypeSubsetter/truetype.cpp#L105-L106

The fix is pretty straightforward, simply copy the font into the result buffer when we detect `ERR_WOULD_GROW` and return as usual.

rladuca:
When a call to CreateDeltaTTF returns ERR_WOULD_GROW, ComputeSubset should not throw FileFormatException.  Instead, it should copy the original font into the result buffer and return that.

This is so a subsetting attempt that would increase the font size can succeed by using the original font, since that is sufficient to display the file and is more compact than the subsetting algorithm can achieve.

Addresses #1511

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/DirectWriteForwarder/CPP/TrueTypeSubsetter/truetype.cpp
+++ b/src/Microsoft.DotNet.Wpf/src/DirectWriteForwarder/CPP/TrueTypeSubsetter/truetype.cpp
@@ -102,8 +102,16 @@ array<System::Byte> ^ TrueTypeSubsetter::ComputeSubset(void * fontData, int file
         Mem_Free(puchDestBuffer);
     }
 
-    if (errCode != NO_ERROR)
+    // If subsetting would grow the font, just use the original one as it's the best we can do.
+    if (errCode == ERR_WOULD_GROW)
+    {
+        retArray = gcnew array<System::Byte>(fileSize);
+        System::Runtime::InteropServices::Marshal::Copy((System::IntPtr)fontData, retArray, 0, fileSize);
+    }
+    else if (errCode != NO_ERROR)
+    {
         throw gcnew FileFormatException(sourceUri);
+    }
 
     return retArray;
 }


rladuca:
Thoughts on changing the throw pattern here to account more accurately for errors that come up from calls?  We'd be throwing much more than FileFormatException here.  Do we want that sort of potential compat break in 3.0?  We could queue an improvement for 5.0 if it's important enough.

rladuca:
Since we're getting a full test pass on this in .NET Framework, I'm going to wait for that to complete before checking this in.

vatsan-madhavan:
Changing _Fixes IssueID_ to _Addresses IssueID_. Please don't auto-close the issue alongwith the PR.

rladuca:
> Changing _Fixes IssueID_ to _Addresses IssueID_. Please don't auto-close the issue alongwith the PR.

My bad, habit.

PR #1515, opened 2019-08-02
Fix for #1198 - fixes potential crashes in high-DPI applications

vatsan-madhavan:
Addresses #1198. 

#### [Watson] clr20r3: CLR_EXCEPTION_System.Collections.Generic.KeyNotFoundException_80131577_PresentationCore.dll!MS.Internal.DpiUtil+DpiAwarenessScope..ctor

--

When a caller passes in `dpiAwarenessContextValue==Invalid`, `new DpiAwarenessContextHandle(dpiAwarenssContextValue)` throws `KeyNotFoundException`.

This happens because we do not have a pre-created pseudo-_HANDLE_  corresponding to `Invalid` in our internal map of pseudo-handles in `DpiAwarenessContextHandle.WellKnownContextValues`. Such a pseudo-handle cannot be created either.

When `dpiAwarenessContextValue=Invalid` is passed, the correct behavior is to treat it as a no-op and simply `return`. The corresponding `Dispose()` will also run benignly and the caller will get the correct behavior - which is that no changes will be made to the thread-state.

Issue #1198, opened 2019-07-09
[Watson] clr20r3: CLR_EXCEPTION_System.Collections.Generic.KeyNotFoundException_80131577_PresentationCore.dll!MS.Internal.DpiUtil+DpiAwarenessScope..ctor


grubioe:
Copied from ADO ID 893177

* Does the bug reproduce also in WPF for .NET Framework 4.8?: Yes

We have a bug here - a missing check. 

\WPF\src\Core\CSharp\MS\Internal\DpiUtil\DpiUtil+DpiAwarenessScope.cs

```
            [SecuritySafeCritical]
            private DpiAwarenessScope(
                DpiAwarenessContextValue dpiAwarenessContextValue,
                bool updateIfThreadInMixedHostingMode,
                bool updateIfWindowIsSystemAwareOrUnaware,
                IntPtr hWnd)
            {
                if (!OperationSupported)
                {
                    return;
                }

                if (updateIfThreadInMixedHostingMode && !this.IsThreadInMixedHostingBehavior)
                {
                    return;
                }

                if (updateIfWindowIsSystemAwareOrUnaware &&
                    (hWnd == IntPtr.Zero || !this.IsWindowUnawareOrSystemAware(hWnd)))
                {
                    return;
                }

                try
                {
                    this.OldDpiAwarenessContext = 
                        UnsafeNativeMethods.SetThreadDpiAwarenessContext(
                            new DpiAwarenessContextHandle(dpiAwarenessContextValue));
                }
                catch (Exception e) when (e is EntryPointNotFoundException || e is MissingMethodException || e is DllNotFoundException)
                {
                    OperationSupported = false;
                }
            }
```

We should either be catch `KeyNotFoundException`, or better yet, checking for `dpiAwarenessContextValue==Invalid` and return with `OperationSupported=false` early in the constructor. 

Found from Watson data:

Call Stack

1. mscorlib.ni!System.Collections.Generic.Dictionary_2[[MS.Utility.DpiAwarenessContextValue,_WindowsBase],[System.IntPtr,_mscorlib]].get_Item
2. at PresentationCore.ni!MS.Internal.DpiUtil in DpiUtil+DpiAwarenessScope.cs
3. at PresentationFramework.ni!System.Windows.SystemResources.CreateResourceChangeListenerWindow in SystemResources.cs
4. at PresentationFramework.ni!System.Windows.SystemResources.EnsureResourceChangeListener in SystemResources.cs
[...]

PR #1525, opened 2019-08-03
Fix for #1198 - support for DPI_AWARENESS_CONTEXT_UNAWARE_GDISCALED

vatsan-madhavan:
Addresses #1198.

#### [Watson] clr20r3: CLR_EXCEPTION_System.Collections.Generic.KeyNotFoundException_80131577_PresentationCore.dll!MS.Internal.DpiUtil+DpiAwarenessScope..ctor

This is a 2nd fix approach for this issue, which is a more complete fix IMO compared to https://github.com/dotnet/wpf/pull/1515. This PR must be paired with the PR to the renderer at https://dev.azure.com/dnceng/internal/_git/dotnet-wpf-int/pullrequest/2498?_a=overview

--

When a caller passes in dpiAwarenessContextValue==Invalid, new DpiAwarenessContextHandle(dpiAwarenssContextValue) throws KeyNotFoundException.

This happens because we do not have a pre-created pseudo-HANDLE corresponding to Invalid in our internal map of pseudo-handles in DpiAwarenessContextHandle.WellKnownContextValues. Such a pseudo-handle cannot be created either.

When dpiAwarenessContextValue=Invalid is passed, the correct behavior is to treat it as a no-op and simply return. The corresponding Dispose() will also run benignly and the caller will get the correct behavior - which is that no changes will be made to the thread-state.

==

Ideally, this problem should not have happened when only dealing with 'valid' or 'good' `DpiAwarenessCotnextValue` enums (or equivalently, 'good' `DpiAwarenessContextHandle` values).

"good" has two meanings here.

i. Values recognized by the OS that are valid (as the OS sees it)
ii. Values recognized by WP as valid (i.e.., WPF has enough intelligence available to process this mode).

WPF only knows how to process the following 4 DPI_AWARENESS_CONTEXT_HANDLE values, which were available during active feature development of .NET 4.8.

- DPI_AWARENESS_CONTEXT_UNAWARE
- DPI_AWARENESS_CONTEXT_SYSTEM_AWARE
- DPI_AWARENESS_CONTEXT_PER_MONITOR_AWARE
- DPI_AWARENESS_CONTEXT_PER_MONITOR_AWARE_V2

It looks like one more value was added in Win10 1809 later in the development cycle which WPF has no understanding of.

- DPI_AWARENESS_CONTEXT_UNAWARE_GDISCALED

In [SystemResources.cs](https://github.com/dotnet/wpf/blob/ac9d1b7a6b0ee7c44fd2875a1174b820b3940619/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/SystemResources.cs#L1146), we query HWND's for their DPI related info like this:

```
    var hwndDpiInfo =
        IsPerMonitorDpiScalingActive ?
        DpiUtil.GetExtendedDpiInfoForWindow(hwndNotify.Handle):
        new DpiUtil.HwndDpiInfo(dpiContextValue, GetDpiScaleForUnawareOrSystemAwareContext(dpiContextValue));
```

`DpiUtil.GetExtendedDpiInfoForWindow(hwndNotify.Handle)` eventually calls into `new HwndDpiInfo`, which in turn attempts to map `DpiAwarenessContextHandle` -> `DpiAwarenessCotnextValue`. This mapping (which is acheived by an overloaded cast operator)
only recognizes `DPI_AWARENESS_CONTEXT_HANDLE` values WPF understands. Anything outside of this set is considered `Invalid`.

I’m reasonably confident that this crash happens because we are getting an `HWND` with `DPI_AWARENESS_CONTEXT_UNAWARE_GDISCALED` or something else outside the recognized spectrum.

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/PresentationCore/MS/internal/DpiUtil/DpiUtil+DpiAwarenessScope.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationCore/MS/internal/DpiUtil/DpiUtil+DpiAwarenessScope.cs
@@ -92,6 +92,11 @@ namespace MS.Internal
                 bool updateIfWindowIsSystemAwareOrUnaware,
                 IntPtr hWnd)
             {
+                if (dpiAwarenessContextValue == DpiAwarenessContextValue.Invalid)
+                {
+                    return;
+                }
+
                 if (!OperationSupported)
                 {
                     return;
--- a/src/Microsoft.DotNet.Wpf/src/PresentationCore/MS/internal/DpiUtil/DpiUtil+DpiAwarenessScope.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationCore/MS/internal/DpiUtil/DpiUtil+DpiAwarenessScope.cs
@@ -173,6 +173,7 @@ namespace MS.Internal
 
                 return
                     dpiAwarenessContext.Equals(DpiAwarenessContextValue.Unaware) ||
+                    dpiAwarenessContext.Equals(DpiAwarenessContextValue.UnawareGdiScaled) ||
                     dpiAwarenessContext.Equals(DpiAwarenessContextValue.SystemAware);
             }
         }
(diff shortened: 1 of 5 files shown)


vatsan-madhavan:
I've started a separate PR to give us an opportunity to weigh the merits vs. risks of taking this change now vs. sticking to #1515 for now and taking this one in .NET 5. 

The main downside I see with this change is testing - its hard to test and therefore well tested. I think its easy to show that it won't regress existing functionality, but its somewhat time-consuming to verify that UnawareGdiScaled child-windows hosted as child-windows inside a PerMonitorV2 application work well. 

Writing test cases or samples that focus on this scenario(related: https://github.com/microsoft/WPF-Samples/issues/92) would help us become more confident IMO. 

We can still take this change now as it is relatively safe, but I want us to have the option of just taking #1515 deferring this...

vatsan-madhavan:
We will take this in .NET 5. Closing for now.

PR #1553, opened 2019-08-06
Dev/sambent/speller1540

### LINKED ISSUE #1540: Crash inserting text in front of Hyperlink when spelling is enabled
Hard crash (FailFast) when inserting text in front of a Hyperlink in a RichTextBox, when spell-checking is enabled.

This is a port of a recent servicing fix in .NETFx.

SamBent:
Addresses #1540 

Ensure that each spell-checking task makes progress before stopping for a timeout.


rladuca:
We're in tell-mode right now, so we don't want to use things like "Fixes" or "Resolves" to auto-close issues.

Issue #1541, opened 2019-08-05
WPF controls in a RowVirtualized DataGrid control lose their value after scrolling


SamBent:
(Port of recent .NETFx servicing fix)
ComboBox (or any Selector) displays the wrong value for SelectedValue (or any selection property) when
1. ComboBox is declared in a cell template of a DataGrid
2. Both ItemsSource and SelectedValue are bound to the row's data item via implicit DataContext (as opposed to explicit Source, ElementName, RelativeSource, etc.)
3. DataGrid row virtualization is enabled.
4. Item is re-virtualized, then de-virtualized.  E.g. scrolling it out of view, then back in.

Root cause:
Re-virtualization causes the item's UI to be disconnected.  This sets DataContext to {DisconnectedItem}, which should set ItemsSource to null and leave SelectedValue unchanged, while removing property-changed listeners and other similar cleanup.  But if ItemsSource happens first, Selector tries to change SelectedValue (to null).   In a DataGrid, this records null as the "proposed value" of SelectedValue, (awaiting writing to the data item when the row-edit is committed).   When the item is later de-virtualized, the cell displays the proposed value (null).

Fix:  Selector should not change its selection properties, when reacting to being disconnected.

SamBent:
Fixed checked in: https://github.com/dotnet/wpf/pull/1557

PR #1557, opened 2019-08-07
WPF controls in a RowVirtualized DataGrid control lose their value after scrolling

SamBent:
Addresses #1541 

Selector should not change its selection properties, when reacting to being disconnected.


SamBent:
Somehow all the line breaks got changed from "unix" to "dos", so it looks like the whole file changed.  (They are "dos" in .NETFx, as are the majority of files in this repo, so reverting them improves consistency.)

The real change occurs in method OnItemsChanged at lines 1113-1130 (in the new file):  
a. move call to base.OnItemsChanged to the beginning
b. adding a check for DisconnectedItem before doing any other work

Issue #1423, opened 2019-07-28
PresentationUI ref-assembly missing: Build fails because cannot find type PresentationUIStyleResources


jbe2277:
* .NET Core SDK: 3.0.100-preview7-012821
* VS 2019: 16.2.0
* Windows version: Win10 1809 (Azure VM, connected via Remote Desktop)
* Does the bug reproduce also in WPF for .NET Framework 4.8?: No

Now I get a build error for a solution which already build with an older .NET Core 3 version.

**It worked with:**
* .NET Core SDK: 3.0.100-preview5-011256
* Microsoft.WindowsDesktop.App 3.0.0-preview5-27610-56

**Reproduce:**
* Clone [https://github.com/jbe2277/waf.git](https://github.com/jbe2277/waf.git)
* Checkout the **netcoreapp3.0** branch
* Open `System.Waf.sln` in VS 2019.
* Build `Samples/Writer/Writer.Presentation`

**Build error:**
```
4>------ Build started: Project: Writer.Presentation, Configuration: Debug Any CPU ------
4>C:\Users\dev\source\repos\waf\src\System.Waf\Samples\Writer\Writer.Presentation\Views\PrintPreviewView.xaml(15,44): 
error MC3066: The type reference cannot find a public type named 'PresentationUIStyleResources'. Line 15 Position 44.
```

**Code line that produces this build error:**
https://github.com/jbe2277/waf/blob/c7773460a1d58256b201df15c197c2b4263ec88e/src/System.Waf/Samples/Writer/Writer.Presentation/Views/PrintPreviewView.xaml#L15

**Workaround:**
Comment this line and it builds.

rladuca:
Well this is strange.  PresentationUI.dll is NOT part of the reference set of .NET Framework.  So this shouldn't build in .NET Framework at all.

However, looking at the binlog from the build, I see it picks this up from the GAC using the full library DLL.

```
    ReferenceAssembly = C:\WINDOWS\Microsoft.Net\assembly\GAC_MSIL\PresentationUI\v4.0_4.0.0.0__31bf3856ad364e35\PresentationUI.dll
```

In Preview7, we removed PresentationUI from our reference set to match .NET Framework's reference set.

@vatsan-madhavan @SamBent What do you think about this?  If picking up references from the GAC is a scenario in .NET Framework I am worried that matching the reference set isn't enough.  Even if this is unintended, it seems like something that is relied on.

rladuca:
[...]

PR #1563, opened 2019-08-07
Adding PresentationUI to RefAssembly list in build-props.

vatsan-madhavan:
Helps with addressing #1423 - adding back PresentationUI ref-assembly.

Adding PUI to RefAssembly list in props.


vatsan-madhavan:
This change must be merged, and the generated transport packages would have to flow into DncEng/dotnet-wpf-int repo, where the actual changes to add back PresentationUI ref assembly will be staged. The two changes (updated package versions from this PR, and the actual PresentationUI ref-assembly additions) must be merged together.

Issue #1564, opened 2019-08-07
Adopt YAML Stages based Publishing and Validation 


vatsan-madhavan:
The Arcade team has guidance available for the adoption of publishing via YAML Stages here: https://github.com/dotnet/arcade/blob/master/Documentation/CorePackages/YamlStagesPublishing.md

They will be tracking the status of repos adopting these templates here: https://github.com/dotnet/arcade/blob/master/Documentation/YamlStagesRepoStatus.md. 

The target completion date for this work is 8/13.

vatsan-madhavan:
PR for DncEng/dotnet-wpf-int merged: https://dev.azure.com/dnceng/internal/_git/dotnet-wpf-int/pullrequest/2567?_a=overview

vatsan-madhavan:
PR for DncEng\dotnet-wpf-int: https://dev.azure.com/dnceng/internal/_git/dotnet-wpf-int/pullrequest/2567?_a=overview (merged)

PR #1565, opened 2019-08-07
Adopt YAML stages based publishing and validation in dotnet/wpf

vatsan-madhavan:
Addresses https://github.com/dotnet/wpf/issues/1564 for [dotnet/wpf](https://github.com/dotnet/wpf). 

We have a couple of successful builds from this branch: 

- https://dev.azure.com/dnceng/internal/_build/results?buildId=299330
- https://dev.azure.com/dnceng/internal/_build/results?buildId=298971

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -77,8 +77,9 @@ stages:
       # enabled back once this issue is resolved: https://github.com/dotnet/arcade/issues/2912
       enableSourceLinkValidation: false
       # This is to enable SDL runs part of Post-Build Validation Stage
+      # Temporarily disable SDK validation until https://github.com/dotnet/arcade/issues/3565 is resolved. 
       SDLValidationParameters:
-        enable: true
+        enable: false
         params: ' -SourceToolsList @("policheck","credscan")
         -TsaInstanceURL $(_TsaInstanceURL)
         -TsaProjectName $(_TsaProjectName)
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -6,13 +6,19 @@
 #
 
 variables:
+    # clean the local repo on the build agents
+  - name: Build.Repository.Clean
+    value: true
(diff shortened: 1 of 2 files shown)


azure-pipelines.yml
```diff
@@ -6,13 +6,19 @@
...
 
 variables:
+    # clean the local repo on the build agents
+  - name: Build.Repository.Clean
+    value: true
+  - name: _PublishUsingPipelines
+    value: true
+  - name: _DotNetArtifactsCategory
+    value: WINDOWSDESKTOP
+  - name: _DotNetValidationArtifactsCategory
```
safern: I guess this will be removed before merging?

eng/pipeline.yml
```diff
@@ -176,3 +177,4 @@ jobs:
         #   (_HelixPipeline && _PublicBuildPipeline && _ContinuousIntegrationTestsEnabled && _BuildConfig == Release)
         #
         condition: and(succeeded(), eq(variables['_HelixPipeline'], 'true') ,eq(variables['_BuildConfig'], 'Release'), eq(variables['_PublicBuildPipeline'], 'true'), eq(variables['_ContinuousIntegrationTestsEnabled'], 'true'))
+
```
safern: Nit: extra whiteline.

azure-pipelines.yml
vatsan-madhavan: I'm assuming the question is about _DotNetValidationArtifactsCategory. _DotNetArtifactsCategory=WINDOWSDESKTOP is required. 

Does it have to be removed? I was going to leave it in. I would imagine it comes into play only when channel=validation.

safern: Yes I was refering to `_DotNetValidationArtifactsCategory` -- it doesn't hurt leaving it in, just to make it cleaner and not have unused variables, however on a second thought, this might come into play when testing new validations or new features for publishing from arcade that need to be adopted.

PR #1576, opened 2019-08-08
Open Sourcing more WPF assemblies

stevenbrix:
This PR open sources the following assemblies:

- PresentationUI
- WindowsFormsIntegration
- System.Printing
- System.Windows.Presentation
- System.Windows.Controls.Ribbon
- Themes
    - PresentationFramework.Aero
    - PresentationFramework.Aero2
    - PresentationFramework.AeroLite
    - PresentationFramework.Classic
    - PresentationFramework.Luna
    - PresentationFramework.Royale

I've attached the [generated IL](https://github.com/dotnet/wpf/files/3483157/ildiffs.zip) from the assemblies, showing what they were previously when building internally and what they are now. There aren't any differences, other than the repo and commit hash which are embedded into the assemblies, which is expected.

Here is the internal PR which removes these assemblies: https://dev.azure.com/dnceng/internal/_git/dotnet-wpf-int/pullrequest/2583?_a=overview

Relates to issue #1579


vatsan-madhavan:
@stevenbrix, we need an issue associated with this for tell-mode tracking

eng/WpfArcadeSdk/tools/Pbt.targets
```diff
@@ -1,104 +1,113 @@
...
+    _WindowsDesktopSdkTargetFrameworkVersionFloor is defined in Microsoft.NET.WindowDesktop.props. 
+    This needs to be defined before Microsoft.NET.Sdk.WindowsDesktop.targets is imported. 
+  -->
+  <PropertyGroup Condition="'$(InternalMarkupCompilation)'=='true' And !Exists('$(LocalMicrosoftWinFXTargets)') ">
+    <_WindowsDesktopSdkTargetFrameworkVersionFloor Condition="'$(_WindowsDesktopSdkTargetFrameworkVersionFloor)' == ''">3.0</_WindowsDesktopSdkTargetFrameworkVersionFloor>
+  </PropertyGroup>
+  <Import Sdk="Microsoft.NET.Sdk.WindowsDesktop"
+          Project="../targets/Microsoft.NET.Sdk.WindowsDesktop.targets"
+          Condition="'$(InternalMarkupCompilation)'=='true' And !Exists('$(LocalMicrosoftWinFXTargets)') "/>
+
```
vatsan-madhavan: Can you add explanations for salient changes? /cc @ryalanms would you mind reviewing this change?

vatsan-madhavan:
I've noticed that Visual Studio has been a bit erratic when making updates to solutions (project add/remove, moving from one solution folder to other etc.) and sometimes updates solution configuration in ways that are a bit unexpected. Have you had a chance to review the Solution and Project configurations carefully for (Release, Debug) x (AnyCPU, x86, x64) x (every project, not just the new or changed ones) to ensure that everything looks correct?

eng/WpfArcadeSdk/tools/Pbt.targets
ryalanms: Why aren't the diffs working correctly?

stevenbrix: yeah, when building the theme dlls we get errors about `WindowsBase` already being in the reference list. i don't know exactly why this wasn't happening internally, but i looked at the binlog and saw that a reference to `WindowsBase` was coming from `Microsoft.NetCore.App.Ref`. this would then manifest itself in an exception being thrown by PBT about `WindowsBase` already being loaded in the `MetadataLoadContext`. So maybe when building internally we weren't picking up the `WindowsBase` that came from GitHub? Regardless, i feel like we never want the `WindowsBase` reference from `Microsoft.NetCore.App.Ref`

vatsan-madhavan: Could it be that the project in question is not using `NetCoreReference` carefully, and thus it is inheriting `WindowsBase `from `Microsoft.NetCore.App.Ref `when it shouldn't?

stevenbrix: i'll add a comment

vatsan-madhavan: When a project uses `NetCoreReference`, only those references that are reqeusted intentionally are used, which means that WindowsBase is not included. It's possible that the project where you had problems forgot to use `NetCoreReference` to enumerate the references it needs, and instead inherits _all_ references from `Microsoft.NetCore.App`. If this is what's happening, then the preferred solution is to fixup the project - identify the smallest set of `NetCoreReference` items and incorporate it into the csproj directly.

stevenbrix:
> Have you had a chance to review the Solution and Project configurations carefully for (Release, Debug) x (AnyCPU, x86, x64) x (every project, not just the new or changed ones) to ensure that everything looks correct?

Good catch, it looks like `System.Printing` wasn't checked to Build for the `AnyCPU` config in Visual Studio's Configuration Manager. The only oddity i see is that `DirectWriteForwarder` has the "Deploy" checkbox checked in x86 builds but not x64. But it shouldn't actually be deployable since it's jut a library anyways.

vatsan-madhavan:
We check the Deploy checkboxes whenever possible.

eng/WpfArcadeSdk/tools/Pbt.targets
stevenbrix: It looks like the project already does that?

```
 <!-- Minimal set of .NET core references -->
  <ItemGroup>
    <NetCoreReference Include="netstandard" />
    <NetCoreReference Include="System.Collections" />
    <NetCoreReference Include="System.Collections.Generic" />
    <NetCoreReference Include="System.Collections.NonGeneric" />
    <NetCoreReference Include="System.ComponentModel" />
    <NetCoreReference Include="System.ComponentModel.Primitives" />
    <NetCoreReference Include="System.Diagnostics.Debug" />
    <NetCoreReference Include="System.Diagnostics.Process" />
    <NetCoreReference Include="System.Diagnostics.Tools" />
    <NetCoreReference Include="System.Diagnostics.TraceSource" />
    <NetCoreReference Include="System.Drawing.Common" />
    <NetCoreReference Include="System.Drawing.Primitives" />
    <NetCoreReference Include="System.IO" />
    <NetCoreReference Include="System.IO.FileSystem" />
    <NetCoreReference Include="System.ObjectModel" />
    <NetCoreReference Include="System.Resources.ResourceManager" />
    <NetCoreReference Include="System.Runtime.Extensions" />
    <NetCoreReference Include="System.Runtime.InteropServices" />
    <NetCoreReference Include="System.Security.Cryptography.Algorithms" />
    <NetCoreReference Include="System.Security.Cryptography.Primitives" />
    <NetCoreReference Include="System.Security.Cryptography.X509Certificates" />
    <NetCoreReference Include="System.Threading" />
    <NetCoreReference Include="System.Threading.Thread" />
    <NetCoreReference Include="System.Threading.ThreadPool" />

    <!-- Additional references beyond the minimal set for PresentationFramework.Aero -->
    <NetCoreReference Include="System.ComponentModel.TypeConverter" />
    <NetCoreReference Include="System.Collections.Concurrent" />
  </ItemGroup>
```

stevenbrix:
> We check the Deploy checkboxes whenever possible.

Ok, I'll check deploy for x64 then. What does checking deploy do for non executable projects?

eng/WpfArcadeSdk/tools/Pbt.targets
```diff
@@ -101,4 +101,23 @@
   <Import Sdk="Microsoft.NET.Sdk.WindowsDesktop"
           Project="../targets/Microsoft.NET.Sdk.WindowsDesktop.targets"
           Condition="'$(InternalMarkupCompilation)'=='true' And !Exists('$(LocalMicrosoftWinFXTargets)') "/>
+
+  <!--
+  WindowsBase comes from NetCoreApp.Reference. When we try to compile the themes dlls, which reference WindowsBase,
```
rladuca: Maybe I am late on this, but do we have a binlog from a prior build that proves they were getting the wrong WindowsBase in wpf-int?  If the answer is no, then I don't think we should make this change, but fix the underlying reason why this has changed in the port to GitHub.

eng/WpfArcadeSdk/tools/Pbt.targets
rladuca: Maybe I am late on this, but do we have a binlog from a prior build that proves they were getting the wrong WindowsBase in wpf-int? If the answer is no, then I don't think we should make this change, but fix the underlying reason why this has changed in the port to GitHub.

eng/WpfArcadeSdk/tools/ReferenceAssembly.targets
```diff
@@ -1,88 +1,86 @@
...
+      <Output PropertyName="BaseReferenceAssemblyDir" TaskParameter="Value"/>
+    </CreateProperty>
+
+    <CreateProperty Condition="'$(ReferenceAssemblyDir)'==''" Value="$(BaseReferenceAssemblyDir)$(Configuration)\$(TargetFramework)\">
+      <Output PropertyName="ReferenceAssemblyDir" TaskParameter="Value"/>
+    </CreateProperty>
+  </Target>
+
+  <Target Name="CopyIntellisenseXmls"
+          Condition="'$(ReferenceProjectRepoLocation)'=='External' and '$(RepositoryName)'=='wpf'"
```
rladuca: Shouldn't we just do this now, do we need the condition at all?

eng/WpfArcadeSdk/tools/ReferenceAssembly.props
```diff
@@ -11,7 +11,6 @@
 
   <!-- Categorize the repo location based on the hand crafted assembly location. -->
   <PropertyGroup>
-    <ReferenceProjectRepoLocation Condition="$(InternalHandCraftedReferenceProjects.Contains('$(NormalizedMSBuildProjectName);'))">Internal</ReferenceProjectRepoLocation>
     <ReferenceProjectRepoLocation Condition="$(ExternalHandCraftedReferenceProjects.Contains('$(NormalizedMSBuildProjectName);'))">External</ReferenceProjectRepoLocation>
```
rladuca: Where is this used?  Does this make sense anymore considering everything that has a ref assembly has moved?

eng/WpfArcadeSdk/tools/Pbt.targets
stevenbrix: Something must've broken here, because it looks like all the references from `Micrososft.NetCore.App.Ref` are being passed to the compiler, not just the ones specified by `NetCoreReference`.

I'd prefer to just keep this change, rather than try to unwind what all of these targets are trying/supposed to do and figure out what exactly is going on. I've spent 20-30 minutes looking through the binlogs and nothing really stands out other than things don't seem to be working the way they should. I'm not even convinced that this is an incorrect fix, it's simple and straightforward.

rladuca: I don't agree, if the targets work 100% on GitHub and on wpf-int and this change breaks them, then this change should fix them directly.

Adding stuff on top of the engineering system without truly knowing the root of the break isn't something I'm willing to sign off on personally.  I'll defer to @vatsan-madhavan on this if he thinks it's correct, but to me this feels like doing more work on top of an issue we already handled historically and doesn't exist outside of this PR.

If I'm characterizing this error wrong, someone please explain to me what I am missing.

eng/WpfArcadeSdk/tools/Pbt.targets
stevenbrix: Ahh ok. i just looked at one of the binlogs in wpf-int and it looks like `WindowsBase` gets referenced from the `Microsoft.DotNet.Wpf.GitHub` package:

> Encountered conflict between 'Platform:WindowsBase.dll' and 'Reference:F:\workspace\_work\1\s\.packages\runtime.win-x64.microsoft.dotnet.wpf.github\4.8.0-preview9.19408.5\lib\netcoreapp3.0\WindowsBase.dll'. Choosing 'Reference:F:\workspace\_work\1\s\.packages\runtime.win-x64.microsoft.dotnet.wpf.github\4.8.0-preview9.19408.5\lib\netcoreapp3.0\WindowsBase.dll' because file version '4.800.19.40805' is greater than '4.700.19.40503'.

So WindowsBase from GitHub Nuget was being chosen over WindowsBase from NetCore.App due to the file version being greater. Now that we are referencing WindowsBase as a ProjectReference, this doesn't happen and so we now get build breaks because there already is a reference to WindowsBase.

At this point we should be able to delete/refactor a lot of the WpfArcadeSdk now that all the managed code is on GitHub.  However,I don't believe this is the time or place for that. I'd prefer a small, simple, and targeted fix for right now that unblocks this PR rather than trying to engineer something more complicated. I don't think fix is technically incorrect either,

eng/WpfArcadeSdk/tools/ReferenceAssembly.targets
stevenbrix: the condition ensures this target only runs for reference assemblies, which i think we still want

eng/WpfArcadeSdk/tools/ReferenceAssembly.props
stevenbrix: see ReferenceAssembly.targets

I think we still need this because we only want that target to run for reference assembly projects

eng/WpfArcadeSdk/tools/ReferenceAssembly.targets
rladuca: Yeah, you're right, thanks.

eng/WpfArcadeSdk/tools/Pbt.targets
rladuca: Ok that makes some more sense I think.  But shouldn't the version from the project reference beat NetCore.App?  This could be a quirk with RAR that we have to have another workaround on top for.

Is this happening elsewhere?  Like won't everyone using WindowsBase as a project reference have this issue?  Why does just PBT encounter this?

eng/WpfArcadeSdk/tools/Pbt.targets
vatsan-madhavan: Let's try to understand what's really going on. I'd like a root-cause for why Microsoft.NetCore.App.Ref is being involved when we expect it to be filtered out.

stevenbrix: See my comment here https://github.com/dotnet/wpf/pull/1576#discussion_r312260083

eng/WpfArcadeSdk/tools/Pbt.targets
stevenbrix: There is a known issue around ProjectReference not being considered, see https://github.com/dotnet/sdk/issues/2674

Since there are two WindowsBase references, PBT is trying to load both into the same MetadataLoadContext, and so an exception is thrown since this isn't allowed. Our test projects don't actually ProjectReference WindowsBase, they use a regular Reference instead*. 

*They actually do the same thing as I've done here with removing WindowsBase from NetCore.App but it isn't necessary since referencing the DLL will overwrite the one that comes from NetCore.App.

eng/WpfArcadeSdk/tools/Pbt.targets
vatsan-madhavan: Not quite.

Please cherry-pick https://github.com/dotnet/wpf/commit/e00c96ad09c93306fcd59214883deec025f742d9 and undo your Pbt.targets changes; then try again.

eng/WpfArcadeSdk/tools/Pbt.targets
stevenbrix: I believe the reason we are hitting it now is because the themes dlls explicitly reference WindowsBase in the markup like this:

```
xmlns:base="clr-namespace:System.Windows;assembly=WindowsBase"
```

which isn't a common practice. do you know why they do that?

vatsan-madhavan: ProjectReference not being factored in conflict-resolution during ResolveAssemblyReferences is indeed a problem. The solution for that in our builds is to side-step the problem by keeping NetCore.app.Ref copy of Windowsbase out of the reference-list. This is done by way of using NetCoreReference.

rladuca: Just to clarify, we reference WindowsBase for flags.  The themes have to know about various compat flags for accessibility (or they did in .NET Framework).  The best place to generate those was in Base because we set all that up in the Dispatcher's static constructor.

That's probably not needed right now, but that's a large change to generated code that we don't really want to make without the appropriate generators enabled.  So, for now, this sort of has to stick around.

vatsan-madhavan: > I believe the reason we are hitting it now is because the themes dlls explicitly reference WindowsBase in the markup like this:

This might be the first-order cause in a manner of speaking, but I'm doubtful that this is the problem. Please follow up on my [suggestion ](https://github.com/dotnet/wpf/pull/1576#discussion_r312287808) and report back - we'll figure out next steps if that doesn't work.

rladuca: Let me correct myself, those can't be removed regardless, they're needed for 4.8 compat.  5.0 is the first candidate for removing them as they'll break compat expectations with .NET 4.8.

eng/WpfArcadeSdk/tools/Pbt.targets
```diff
@@ -101,4 +101,23 @@
...
+  we then get an error during markup compilation that looks like this:
+  src\Microsoft.DotNet.Wpf\src\Themes\PresentationFramework.Aero\Themes\Aero.NormalColor.xaml(26,27): error MC1000: Unknown build error,
+  'The assembly 'WindowsBase, Version=4.0.0.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35' has already loaded been loaded into this
+  MetadataLoadContext. Line 26 Position 27.'  [src\Microsoft.DotNet.Wpf\src\Themes\PresentationFramework.Aero\PresentationFramework.Aero.csproj]
+
+  We always want to reference the WindowsBase that comes from this repository, so we'll go ahead and remove the reference to the one that comes
+  from Microsoft.NetCore.App.Ref
+  -->
+  <Target Name="RemoveWindowsBaseNetCoreAppReference"
```
rladuca: Can you remove this?

jeffbog:
yay

jeffbog
425.269.9314

vatsan-madhavan:
If XAML compilation works now, that's great. I haven't had a chance to look at the rest yet - will do later tonight or tomorrow.

rladuca:
Validated ApiCompat is working as expected against the new assemblies (both ref and lib).

vatsan-madhavan:
lgtm! 

This PR and the corresponding one that deletes this code from our internal repo must be merged in sync. I don't mean 'roughly at the same time' - they must be merged in a specific way. 

- Start by merging this PR. Let the rolling build complete. 
- A DARC PR will be generated into dotnet-wpf-int immediately after this rolling build completes. 
- Retarget your [dotnet-wpf-int PR](https://dev.azure.com/dnceng/internal/_git/dotnet-wpf-int/pullrequest/2583?_a=files) from **master** into the DARC PR branch. 
  - Let the DARC generated PR restart its PR-build. Once it succeeds, make sure it is merged (not sure whether auto-merge will work once you 'touch' it with your own changes). 

Some things to note. 

- Once you merge this PR, you'll have to start watching the build-pipelines to ensure that no other PR that is merged subsequently into dotnet/wpf _races_ against your builds to reach dotnet-wpf-int. 
  - Depending on the nature of the change, this may not be a problem - best to avoid these sorts of situations altogether. 
- Your PR into dotnet-wpf-int should be merged only after validating that the DARC PR has put out the right version numbers. Open up the corresponding NuPkg's and make sure that the new binaries recently added into dotnet/wpf are present. 
  - If one were to inadvertently merge the dotnet-wpf-int PR against an older version of dotnet/wpf dependencies, the build will very likely succeed anyway. It will end up producing a bad core-sdk though (one containing fewer WPF assemblies than it should), and cause a bunch of failure-reports from customers who consume daily core-sdk builds. **You must ensure that this does not happen.** 
- Once the DARC PR containing the the changes from dotnet/wpf have been initiated onto dotnet-wpf-int repo, you'll have limited time to ensure that you complete your work (retarget your own PR onto the DARC generated branch etc.). 

Overall, all of these steps are best done by (a) constantly watching both the public and internal build-pipelines for both repos and (b) ensuring that no other PR's are being merged at the same time you are trying to complete your work. You'll probably want to do this when the code-flow traffic is low and when you are able to get help if there are problems - I recommend completing this before 10 am PST when it won't block other active work, and will give plenty of time to troubleshoot problems if any were to arise during the day. 

Once this end-to-end merge is completed and a successful rolling build is generated from dotnet-wpf-int, trigger subscriptions to ensure that core-setup, and then subsequently core-sdk, are also built. We'd want to get a core-sdk build in hand by tomorrow that contains these changes. This should allow us to run some tests on that core-sdk build immediately. From past experience, I expect this could take 3-5 hours of babysitting builds across different repos. We always do this anyway after big repo changes (like project moves for open-sourcing) to ensure that end-to-end product composition is not inadvertently broken in some way. 
 
Also trigger the reverse subscription from dotnet-wpf-int -> dotnet/wpf to validate that route works without any problems.

Issue #1567, opened 2019-08-08
Remove .NETCore <3.0 as SupportedTargetFrameworks for WinForms and WPF


vatsan-madhavan:
This issue tracks WPF's portion of the work. 

-- 

@etbyrd commented on [Thu Jul 18 2019](https://github.com/dotnet/sdk/issues/3438)

The Project System needs a way to know the correct SupportedTargetFrameworks for WinForms and WPF projects so we can display the correct TFM's in the property pages. 

Is it possible that this can be done on the SDK side?

https://github.com/dotnet/project-system/issues/5010

cc @nguerrera 

---

@nguerrera commented on [Thu Jul 18 2019](https://github.com/dotnet/sdk/issues/3438#issuecomment-513017079)

cc @vatsan-madhavan @rladuca

---

@nguerrera commented on [Thu Jul 18 2019](https://github.com/dotnet/sdk/issues/3438#issuecomment-513017129)

cc @dsplaisted 

---

@nguerrera commented on [Thu Jul 18 2019](https://github.com/dotnet/sdk/issues/3438#issuecomment-513018706)

Today we have a static list of supported TFMs here:

https://github.com/dotnet/sdk/blob/master/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.SupportedTargetFrameworks.props

What I'm thinking is that the WindowsDesktop SDK could override this somehow to remove the target frameworks that it doesn't support.

Simplest form of that would be to do some `Removes` but then I think we want no versions of .NET Standard to be supported, and then we'd have a maintenance problem when say .NETStandard 2.2 came out and wasn't yet removed. 

Another way would be for WindowsDesktop SDK to remove all and have its own fixed allow-list. But then it is in the business of taking an update for every version.

Finally, we could make the list above dynamic based on whether the WindowsDesktop SDK has set certain props.

What do you all think would be best?

---

@vatsan-madhavan commented on [Thu Jul 18 2019](https://github.com/dotnet/sdk/issues/3438#issuecomment-513019026)

I don't understand the request for the SDK. 

We already have logic for handling  TFM < 3.0 && TargetFrameworkIdentifier==.NETCoreApp. What specifically is being requested here? 

[...]

PR #1578, opened 2019-08-08
 Remove unsupported TFM's from SupportedTargetFramework.

vatsan-madhavan:
- Addresses #1567
- Completes work started in https://github.com/dotnet/sdk/issues/3438
- Removes unsupported TFM's from SupportedTargetFramework.

### DIFF
--- a/packaging/Microsoft.NET.Sdk.WindowsDesktop/targets/Microsoft.NET.Sdk.WindowsDesktop.props
+++ b/packaging/Microsoft.NET.Sdk.WindowsDesktop/targets/Microsoft.NET.Sdk.WindowsDesktop.props
@@ -170,7 +170,16 @@
     <_UnsupportedNETStandardTargetFramework Include="@(SupportedNETStandardTargetFramework)" />
 
     <!--
-        Windows Forms and WPF are supported on .NET Framework 3.0+
+        Windows Forms was supported since .NET Framework 1.0, and is currently supported on 
+        .NET Framework 2.0+. 
+        
+        WPF is supported on .NET Framework and WPF are supported on .NET Framework 3.0+
+        
+        In practice, the WindowsDesktop SDK is only supported on .NET Framework 3.0+ - this is controlled
+        by $(_WindowsDesktopSdkTargetFrameworkVersionFloor), defined as 3.0, which applies to both .NETFramework
+        and .NETCore.
+        
+        Here, we will encode .NET Framework 3.0 as the lowest supported version for both Windows Forms and WPF. 
         
         The SDK does not define versions < 2.0 in @(SupportedNETFrameworkTargetFramework) list, so none of those
         need to be excluded here - removing 2.0 would suffice.
--- a/packaging/Microsoft.NET.Sdk.WindowsDesktop/targets/Microsoft.NET.Sdk.WindowsDesktop.props
+++ b/packaging/Microsoft.NET.Sdk.WindowsDesktop/targets/Microsoft.NET.Sdk.WindowsDesktop.props
@@ -132,4 +132,45 @@
     <_SDKImplicitReference Include="WindowsFormsIntegration"
                        Condition=" ('$(UseWindowsForms)' == 'true') And ('$(UseWPF)' == 'true') "/>
   </ItemGroup>
+  
+  
+  <!-- 
+        Supported (and unsupported) TargetFrameworks 
+        
+        Visual Studio Project System determines the list of valid TargetFrameworks to show 
+        in the Project properties by querying SupportedTargetFramework values. 
+        
+        The Project System does not refer to this list at this time for .NET Framework TFM's.
+  -->
+
+  <!-- 
+    When WindowsDesktop SDK is used without setting UseWPF or UseWinForms, it shows a (suppressible) warning and functions much
+    like Microsoft.NET.Sdk
+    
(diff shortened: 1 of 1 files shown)


vatsan-madhavan:
I decided to start this PR before the definition for `SupportedNETStandardTargetFramework` etc. flows into dotnet/wpf - I figured the change would simply be benign if those are not defined. 

@peterhuene I can give you a privately build WindowsDesktop SDK  based on this change - would you able to help test this with VS?

peterhuene:
I've tested it in VS with my original suggestion (without the condition you added to the group) with a manually modified WindowsDesktop SDK.  I can retest with this PR's contents too.

vatsan-madhavan:
> I've tested it in VS with my original suggestion (without the condition you added to the group) with a manually modified WindowsDesktop SDK. I can retest with this PR's contents too.

That's really the interesting portion. Take a WPF or Windows Forms app, remove `UseWPF `or `UseWindowsForms `and see if it behaves ok, set TFM to `netcoreapp2.0 `and see if it behaves ok etc. 

[Microsoft.NET.Sdk.WindowsDesktop.3.0.0-dev.zip](https://github.com/dotnet/wpf/files/3483339/Microsoft.NET.Sdk.WindowsDesktop.3.0.0-dev.zip)

peterhuene:
I'll test that out.

peterhuene:
Without `UseWpf`/`UseWindowsForms`, I get `NETSDK1106: Microsoft.NET.Sdk.WindowsDesktop requires 'UseWpf' or 'UseWindowsForms' to be set to 'true'`.  I get the full list of .NET Core App target frameworks.  With it, I get just 3.0 (I have to reload the project, though; the project properties doesn't seem to re-evaluate the supported frameworks upon project changes).

vatsan-madhavan:
> Without `UseWpf`/`UseWindowsForms`, I get `NETSDK1106: Microsoft.NET.Sdk.WindowsDesktop requires 'UseWpf' or 'UseWindowsForms' to be set to 'true'`. I get the full list of .NET Core App target frameworks. 

That's what I would expect. The warning is suppressible and we designed it that way to enable certain scenarios. 

> With either property set, I get just 3.0 (I have to reload the project, though; the project properties doesn't seem to re-evaluate the supported frameworks upon project changes).

great! 

Would you mind testing to see what happens when you set TFM to netcoreapp2.2. in your project by hand? This is clearly not natural scenario, but I think it can happen when using MSBuild SDK Extras etc. indirectly.

peterhuene:
`NETSDK1105: Windows desktop applications are only supported on .NET Core 3.0 or higher.`	

I get the full list of .NET Core App target frameworks in the project properties (after reload).

vatsan-madhavan:
> `NETSDK1105: Windows desktop applications are only supported on .NET Core 3.0 or higher.` and the expected build errors due to missing references.
> 
> I get the full list of .NET Core App target frameworks in the project properties (after reload).

Perfect! 

Thank you!

peterhuene:
I'll review as soon as Nick gets back to me regarding what we want to do with the .NET Framework values.

vatsan-madhavan:
@peterhuene, I've updated incorporating exclusions for netfx.

packaging/Microsoft.NET.Sdk.WindowsDesktop/targets/Microsoft.NET.Sdk.WindowsDesktop.props
```diff
@@ -132,4 +132,55 @@
...
+    <_UnsupportedNETCoreAppTargetFramework Include=".NETCoreApp,Version=v2.1" />
+    <_UnsupportedNETCoreAppTargetFramework Include=".NETCoreApp,Version=v2.2" />
+
+    <!--
+        Windows Forms and WPF are not supported an any .NET Standard targets
+    -->
+    <_UnsupportedNETStandardTargetFramework Include="@(SupportedNETStandardTargetFramework)" />
+
+    <!--
+        Windows Forms and WPF are supported on .NET Framework 3.0+
```
nguerrera: I would expect Winforms to work in 2.0.

vatsan-madhavan: I missed that - fixing...

vatsan-madhavan: oh, but just realized that we made another design decision before - that when using WindowsDesktop SDK, we don't support net20... 

that's how _WindowsDesktopSdkTargetFrameworkVersionFloor gets defined as 3.0 - as a common floor for netfx and netcore.

nguerrera: I don't really care about 2.0, but if it were free to support it, it would be more accuratet.

vatsan-madhavan: See https://github.com/dotnet/wpf/pull/1027#discussion_r296308910

vatsan-madhavan: I improved the comment about what is supported, and why we are setting 3.0 as the lowest supported version in WindowsDesktop SDK. Still not adding 2.0 though.

Issue #1604, opened 2019-08-12
Update private C++ toolset to 0.0.0.7


rladuca:
A new drop of private C++ tooling is available and tested on the WPF repositories (both internal and external).

We should update to these tools in both repositories.

vatsan-madhavan:
After this is fixed, the following dangling type-refs would still be left: 


- Release Builds: 
  - DirectWriteForwarder:
    - `[mscorlib] System.Runtime.CompilerServices.AssemblyAttributesGoHereM`
    - `[mscorlib] System.Runtime.CompilerServices.AssemblyAttributesGoHere`  
  - System.Printing
    - All the same problems observed in DirectWriteForwarder
    - Additional:`[System.Runtime.CompilerServices.VisualC] System.Runtime.CompilerServices.IsVolatile`
- Debug Builds:
  - DirectWriteForwarder
    - `[mscorlib] System.Runtime.CompilerServices.SuppressMergeCheckAttribute` 
    - `[System.Runtime] System.Runtime.CompilerServices.AssemblyAttributesGoHere`
    - `[mscorlib] System.Runtime.CompilerServices.AssemblyAttributesGoHereM`
    - `[mscorlib] System.Runtime.CompilerServices.AssemblyAttributesGoHere`
  - System.Printing
    - All the same problems observed in  Debug/DirectWriteForwarder
    - Additional: `[System.Runtime.CompilerServices.VisualC`

grubioe:
Reviewed during 8/13 .NET Tactics

PR #1605, opened 2019-08-12
Update c++ toolset to version 0.0.0.7

rladuca:
Addresses #1604


rladuca:
There are broader changes to make here so abandoning this in favor of @vatsan-madhavan 's branches.

PR #1606, opened 2019-08-12
Update private C++ toolset to 0.0.0.7

rladuca:
Addresses #1604 

From @vatsan-madhavan 's update branch.


vatsan-madhavan:
Note: We have DRT validation on {Release, x86}., {Release, x64} on these tools. They also improve dangling type-ref errors in DirectWriteForwarder and System.Printing significantly.

PR #1658, opened 2019-08-19
Add RaisePropertyChanged extension for INotifyPropertyChanged

Jogge:
This extension makes it much easier to raise `PropertyChanged`.

You also save a lot of time writing your own `RaisePropertyChanged` method for each class or invoking the `PropertyChanged` event with the right arguments.

**How to use the `RaisePropertyChanged` extension:**
```C#
using System.ComponentModel;

public class Foo : INotifyPropertyChanged
{
    private object _bar;

    public object Bar
    {
        get => _bar;
        set
        {
            _bar = value;
            this.RaisePropertyChanged();
        }
    }
}
```

**Some examples of how to raise `PropertyChanged` the old way:**

1. Saves you a lot of writing. Also, if you forget the conditional access an exception will be thrown if there is no subscribers to `PropertyChanged`.
 
```C#
using System.ComponentModel;

public class Foo : INotifyPropertyChanged
{
    private object _bar;

    public object Bar
    {
        get => _bar;
        set
        {
            _bar = value;
            RaisePropertyChanged();
        }
    }

    public void RaisePropertyChanged([CallerMemberName] string propertyName = null)
    {
        PropertyChanged?.Invoke(this, new PropertyChangedEventArgs(propertyName));
    }
}
```

2. Saves you even more writing, if you have a lot of properties. Also, if you forget the conditional access an exception will be thrown if there is no subscribers to `PropertyChanged`.

```C#
using System.ComponentModel;

public class Foo : INotifyPropertyChanged
{
    private object _bar;

    public object Bar
    {
        get => _bar;
        set
        {
            _bar = value;
            PropertyChanged?.Invoke(this, new PropertyChangedEventArgs(nameof(Bar)));
        }
    }
}
```

3. Saves a garbage reference to `GalaSoft.MvvmLight` and avoids inheritance.
```C#
using GalaSoft.MvvmLight;

public class Foo : ViewModelBase
{
    private object _bar;

    public object Bar
    {
        get => _bar;
        set
        {
            _bar = value;
            RaisePropertyChanged();
        }
    }
}
```

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/WindowsBase/System/ComponentModel/PropertyChangedExtensions.cs
+++ b/src/Microsoft.DotNet.Wpf/src/WindowsBase/System/ComponentModel/PropertyChangedExtensions.cs
@@ -1,0 +1,16 @@
+﻿namespace System.ComponentModel
+{
+    using System.Reflection;
+    using System.Runtime.CompilerServices;
+
+    public static class PropertyChangedExtensions
+    {
+        public static void RaisePropertyChanged(this INotifyPropertyChanged @this, [CallerMemberName] string propertyName = null)
+        {
+            var declaringType = @this.GetType().GetEvent(nameof(INotifyPropertyChanged.PropertyChanged)).DeclaringType;
+            var propertyChangedFieldInfo = declaringType.GetField(nameof(INotifyPropertyChanged.PropertyChanged), BindingFlags.Instance | BindingFlags.NonPublic);
+            var propertyChangedEventHandler = propertyChangedFieldInfo.GetValue(@this) as PropertyChangedEventHandler;
+            propertyChangedEventHandler?.Invoke(@this, new PropertyChangedEventArgs(propertyName));
+        }
+    }
+}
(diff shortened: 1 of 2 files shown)


weltkante:
This should be added where the `INotifyPropertyChanged` interface is defined (corefx?) not in some UI framework which just consumes the interface.

dotMorten:
I don't really get why this should be part of WPF?
First of all any app could just add these few lines of code, second it should be part of System.ComponentModel where this event lives (ie CoreFX), and I can't say I love the idea of using reflection for this, nor having external callers be able to raise events on other classes.

jnm2:
Allowing events to be raised from outside the model is scary.

This could be added as a protected nonvirtual method on INotifyPropertyChanged using C# 8's Default Interface Methods which ship in 3.0:
https://devblogs.microsoft.com/dotnet/default-implementations-in-interfaces/
https://docs.microsoft.com/en-us/dotnet/csharp/whats-new/csharp-8#default-interface-members

This is my favorite pattern:

```cs

private int foo;
public int Foo { get => foo; set => Set(ref foo, value); }

private string bar;
public string Bar { get => bar; set => Set(ref bar, value); }

// Boilerplate
private bool Set<T>(ref T field, T value, [CallerMemberName] string propertyName = null)
{
    // Is there any way this change could be observable? Reference equality for reference types,
    // underlying binary equality for value types.
    if (RuntimeHelpers.Equals(field, value)) return false;

    field = value;
    OnPropertyChanged(propertyName);
    return true;
}

private void OnPropertyChanged([CallerMemberName] string propertyName = null)
{
    PropertyChanged?.Invoke(this, new PropertyChangedEventArgs(propertyName));
}
```

stevenbrix:
Hey @Jogge, thanks for the PR! Unfortunately, all PRs need a corresponding issue that has been approved before we'll merge them. Please take a look at our [contributing guidlines](https://github.com/dotnet/wpf/blob/master/Documentation/contributing.md#contribution-bar) for a better understanding of how you can help contribute in the future.

PR #1692, opened 2019-08-22
Updating issue and feature request templates

diverdan92:
We've observed some tooling related issues filed in the repo that take some additional time to get routed to the right Visual Studio teams. By suggesting developers to file issues on VS directly (for tools related issues), hopefully we can expedite triage and ultimately fixes.


diverdan92:
@grubioe FYI

vatsan-madhavan:
Looks reasonable to me. We had asked for a GitHub repo to point customers to, esp. for keeping track of VS issues that impact WPF scenarios. I think many of these are likely in that category. Do we have one set-up yet that we can reference here? If not, we can set up one now ?

Issue #1651, opened 2019-08-16
Invalid comparison in MSBuild's ignore-conditions mode in outer build of multitargeted project


rainersigwald:
Originally reported by @batzen at https://github.com/dotnet/cli/issues/12267#issuecomment-522067627

* .NET Core Version: `3.0.100-preview9-013744`
* Windows version: `10.0.18362.239`
* Does the bug reproduce also in WPF for .NET Framework 4.8?: No

 **Problem description:**

A multitargeted WPF project throws an exception when evaluated in MSBuild's "ignore conditions" mode, which is often used in design-time scenarios.

This is because while the conditions here are sufficient at build time for real projects, when conditions are ignored MSBuild thinks the project is invalid (microsoft/msbuild#4622)

https://github.com/dotnet/wpf/blob/d6c7ae5e7034fb19cc60d4ead2ea92b3210e6b83/packaging/Microsoft.NET.Sdk.WindowsDesktop/targets/Microsoft.NET.Sdk.WindowsDesktop.props#L98-L121

**Actual behavior:** 

```
error: A numeric comparison was attempted on "$(_TargetFrameworkVersionWithoutV)" that evaluates to "" instead of a number, in condition "'$(_TargetFrameworkVersionWithoutV)' >= '4.5'".  c:\program files\dotnet\sdk\3.0.100-preview8-013656\Sdks\Microsoft.NET.Sdk.WindowsDesktop\targets\Microsoft.NET.Sdk.WindowsDesktop.props
```
 
 **Expected behavior:**
 
No error

 **Minimal repro:**
 
- dotnet new wpf
- Replace `TargetFramework` with `<TargetFrameworks>netcoreapp3.0;net45</TargetFrameworks>`
- dotnet restore
- dotnet list package

rainersigwald:
I believe this could be worked around by moving that `ItemGroup` into a conditionally-imported file, and ensuring that when the condition on the import is `true`, `_TargetFrameworkVersionWithoutV` will be defined.

If this doesn't meet the bar for 3.0 (I wouldn't push for it, personally), it might be better to use microsoft/msbuild#3212, but that's not yet implemented.

batzen:
This also affects loading projects/folders in VS code because omnisharp fails with
```
[...]

PR #1707, opened 2019-08-23
Fixes problems in WindowsDesktop SDK due to _TargetFrameworkVersionWithoutV being undefined sometimes

vatsan-madhavan:
Addresses https://github.com/dotnet/wpf/issues/1651 - Invalid comparison in MSBuild's ignore-conditions mode in outer build of multitargeted project

This change ensures that the WindowsDesktop SDK never deals with undefined numeric values related to TFM.

This solution leverages the fact that `Items` are evaluated after `Properties` by MSBuild. See [Comparing properties and items](https://docs.microsoft.com/en-us/visualstudio/msbuild/comparing-properties-and-items?view=vs-2019)

`_TargetFrameworkVersionValue` - a new Property which is defined for exclusive use within the WindowsDesktop SDK, and which is intended to act as a proxy for `_TargetFrameworkVersionWithoutV` - is defined in `Microsoft.NET.Sdk.WindowsDesktop.targets` _after_ `TargetFrameworkVersionWithoutV` is guaranteed to be defined by `Microsoft.NET.Sdk.targets`. 

Even here, we ensure that a fallback default (`0.0`) is provided - guaranteeing that `_TargetFrameworkVersionValue` will always be numeric. 

When the `Condition`s in the various `Items` in `Microsoft.NET.WindowsDesktop.props` are evaluated (which will happen strictly after all `Properties` are evaluated - whether they appear in `.props` or `.targets`), `_TargetFrameworkVersionValue` would have been (a) well-defined and (b) guaranteed to be a numeric value. This would in turn guarantee that errors of the kind `"A numeric comparison was attempted on "$(_TargetFrameworkVersionValue)" that evaluates to "" instead of a number,"` will no longer appear in outer-builds or in any other context. 

PS: This is a fix for a regression introduced by https://github.com/dotnet/wpf/pull/1027 (which shipped in Preview 6) as part of a set of fixes for improving multi-targeting experience, improved warnings and error messages during build, and improvements to project-system integration.

### DIFF
--- a/eng/WpfArcadeSdk/tools/Pbt.targets
+++ b/eng/WpfArcadeSdk/tools/Pbt.targets
@@ -92,11 +92,20 @@
           Condition="'$(InternalMarkupCompilation)'=='true' And Exists('$(LocalMicrosoftWinFXTargets)') "/>
 
   <!-- 
-    _WindowsDesktopSdkTargetFrameworkVersionFloor is defined in Microsoft.NET.WindowDesktop.props. 
-    This needs to be defined before Microsoft.NET.Sdk.WindowsDesktop.targets is imported. 
+    _WindowsDesktopSdkTargetFrameworkVersionFloor, _UndefinedTargetFrameworkVersion and _TargetFrameworkVersionValue are 
+    defined in Microsoft.NET.WindowDesktop.props. 
+    
+    These need to be defined before Microsoft.NET.Sdk.WindowsDesktop.targets is imported. 
   -->
   <PropertyGroup Condition="'$(InternalMarkupCompilation)'=='true' And !Exists('$(LocalMicrosoftWinFXTargets)') ">
     <_WindowsDesktopSdkTargetFrameworkVersionFloor Condition="'$(_WindowsDesktopSdkTargetFrameworkVersionFloor)' == ''">3.0</_WindowsDesktopSdkTargetFrameworkVersionFloor>
+    
+    <!--  Represents an undefined TFV value. -->
+    <_UndefinedTargetFrameworkVersion>0.0</_UndefinedTargetFrameworkVersion>
+
+    <!-- Initial/Default value set to 'undefined'. Updated in Microsoft.NET.WindowsDesktop.targets -->
+    <_TargetFrameworkVersionValue>$(_UndefinedTargetFrameworkVersion)</_TargetFrameworkVersionValue>
+    
   </PropertyGroup>
   <Import Sdk="Microsoft.NET.Sdk.WindowsDesktop"
           Project="../targets/Microsoft.NET.Sdk.WindowsDesktop.targets"
--- a/packaging/Microsoft.NET.Sdk.WindowsDesktop/targets/Microsoft.NET.Sdk.WindowsDesktop.props
+++ b/packaging/Microsoft.NET.Sdk.WindowsDesktop/targets/Microsoft.NET.Sdk.WindowsDesktop.props
@@ -15,12 +15,8 @@
     using the SDK style projects. For e.g., see https://github.com/microsoft/msbuild/issues/1333
     
     Irrespective of whether '$(TargetFrameworkIdentifier)' is '.NETCoreApp' or '.NETFramework', 
-    the minimum value of $(_TargetFrameworkVersionWithoutV) we will be testing for is '3.0'
+    the minimum value of $(_TargetFrameworkVersionValue) we will be testing for is '3.0'
     
-    Note:
-      Please see https://github.com/microsoft/msbuild/issues/3212 for a discussion about the use of 
-      the private $(_TargetFrameworkVersionWithoutV) property - which will likely remain supported and
-      is safe to use here. 
   -->
     <_WindowsDesktopSdkTargetFrameworkVersionFloor>3.0</_WindowsDesktopSdkTargetFrameworkVersionFloor>
 
@@ -32,7 +28,7 @@
   </PropertyGroup>
 
   <ItemGroup Condition=" ('$(EnableDefaultItems)' == 'true') And ('$(UseWPF)' == 'true') And 
-                         ('$(_TargetFrameworkVersionWithoutV)' != '') And ('$(_TargetFrameworkVersionWithoutV)' >= '$(_WindowsDesktopSdkTargetFrameworkVersionFloor)')">
+                         ('$(_TargetFrameworkVersionValue)' != '$(_UndefinedTargetFrameworkVersion)') And ('$(_TargetFrameworkVersionValue)' >= '$(_WindowsDesktopSdkTargetFrameworkVersionFloor)')">
     <ApplicationDefinition Include="App.xaml"
                            Condition="'$(EnableDefaultApplicationDefinition)' != 'false' And Exists('$(MSBuildProjectDirectory)/App.xaml') And '$(MSBuildProjectExtension)' == '.csproj'">
       <Generator>MSBuild:Compile</Generator>
@@ -66,8 +62,8 @@
   
   <ItemGroup Condition="('$(DisableImplicitFrameworkReferences)' != 'true') And
                       ('$(TargetFrameworkIdentifier)' == '.NETCoreApp') And
-                      ('$(_TargetFrameworkVersionWithoutV)' != '') And
-                      ('$(_TargetFrameworkVersionWithoutV)' >= '$(_WindowsDesktopSdkTargetFrameworkVersionFloor)')">
+                      ('$(_TargetFrameworkVersionValue)' != '$(_UndefinedTargetFrameworkVersion)') And
+                      ('$(_TargetFrameworkVersionValue)' >= '$(_WindowsDesktopSdkTargetFrameworkVersionFloor)')">
 
     <FrameworkReference Include="Microsoft.WindowsDesktop.App" IsImplicitlyDefined="true"
                         Condition="('$(UseWPF)' == 'true') And ('$(UseWindowsForms)' == 'true')"/>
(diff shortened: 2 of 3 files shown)


eng/WpfArcadeSdk/tools/Pbt.targets
```diff
@@ -92,12 +92,18 @@
...
-    _WindowsDesktopSdkTargetFrameworkVersionFloor is defined in Microsoft.NET.WindowDesktop.props. 
-    This needs to be defined before Microsoft.NET.Sdk.WindowsDesktop.targets is imported. 
+    _WindowsDesktopSdkTargetFrameworkVersionFloor, _InvalidTargetFrameworkVersion & _TargetFrameworkVersionValue are defined in 
+     Microsoft.NET.WindowDesktop.props. 
+    
+     These need to be defined before Microsoft.NET.Sdk.WindowsDesktop.targets is imported. 
   -->
   <PropertyGroup Condition="'$(InternalMarkupCompilation)'=='true' And !Exists('$(LocalMicrosoftWinFXTargets)') ">
+    <_InvalidTargetFrameworkVersion>0.0</_InvalidTargetFrameworkVersion>
+    <_TargetFrameworkVersionValue Condition="'$(_TargetFrameworkVersionValue)' == ''">$([MSBuild]::ValueOrDefault($(_TargetFrameworkVersionWithoutV), $(_InvalidTargetFrameworkVersion)))</_TargetFrameworkVersionValue>
```
vatsan-madhavan: Ignore the changes in Pbt.targets. 

I tried defining `_TargetFrameworkVersionValue` similar to here in `Microsoft.NET.Sdk.WindowsDesktop.props` (instead of defining an item `_WindowsDesktopTargetFrameworkVersion`), but it always ended up being `0.0`. 

Turns out that property evaluation order bites, and because SDK props are imported early in a project's life, `_TargetFrameworkVersionWithoutV` is never defined at the start when properties are being evaluated.

packaging/Microsoft.NET.Sdk.WindowsDesktop/targets/Microsoft.NET.Sdk.WindowsDesktop.props
```diff
@@ -15,18 +26,19 @@
     using the SDK style projects. For e.g., see https://github.com/microsoft/msbuild/issues/1333
     
     Irrespective of whether '$(TargetFrameworkIdentifier)' is '.NETCoreApp' or '.NETFramework', 
-    the minimum value of $(_TargetFrameworkVersionWithoutV) we will be testing for is '3.0'
+    the minimum value of @(_WindowsDesktopTargetFrameworkVersion) we will be testing for is '3.0'
```
rainersigwald: This comment suggests to me that it's not just a Desktop TF version. Is that wrong?

eng/WpfArcadeSdk/tools/Pbt.targets
rainersigwald: Why should these be ignored? Your reasoning makes sense and defining this in targets seems reasonable.

packaging/Microsoft.NET.Sdk.WindowsDesktop/targets/Microsoft.NET.Sdk.WindowsDesktop.targets
```diff
@@ -16,7 +17,9 @@
   <Target Name="CheckForDuplicatePageItems" 
           BeforeTargets="_CheckForInvalidConfigurationAndPlatform;CoreCompile"
           DependsOnTargets="CheckForDuplicateItems"
-          Condition="('$(UseWPF)' == 'true') And ('$(_TargetFrameworkVersionWithoutV)' != '') And ('$(_TargetFrameworkVersionWithoutV)' >= '$(_WindowsDesktopSdkTargetFrameworkVersionFloor)')">
+          Condition="('$(UseWPF)' == 'true') And 
+                     ('@(_WindowsDesktopTargetFrameworkVersion)' != '$(_InvalidTargetFrameworkVersion)') And 
+                     ('@(_WindowsDesktopTargetFrameworkVersion)' >= '$(_WindowsDesktopSdkTargetFrameworkVersionFloor)')">
```
rainersigwald: These targets will only run for inner builds so it's not strictly necessary to change the property references here, but I agree that you should for consistency.

packaging/Microsoft.NET.Sdk.WindowsDesktop/targets/Microsoft.NET.Sdk.WindowsDesktop.props
```diff
@@ -1,9 +1,20 @@
...
     <EnableDefaultApplicationDefinition Condition="'$(EnableDefaultApplicationDefinition)' == ''">true</EnableDefaultApplicationDefinition>
   </PropertyGroup>
+  
+  <PropertyGroup>
+    <_InvalidTargetFrameworkVersion>0.0</_InvalidTargetFrameworkVersion>  
+  </PropertyGroup>
+  
+  <ItemGroup>
+    <_WindowsDesktopTargetFrameworkVersion Remove="@(_WindowsDesktopTargetFrameworkVersion)" />
+    <_WindowsDesktopTargetFrameworkVersion Include="$([MSBuild]::ValueOrDefault($(_TargetFrameworkVersionWithoutV), $(_InvalidTargetFrameworkVersion)))" />
```
rainersigwald: Why is this becoming an item? Is it just to get later evaluation? I'd prefer the define-the-property-in-targets approach.

packaging/Microsoft.NET.Sdk.WindowsDesktop/targets/Microsoft.NET.Sdk.WindowsDesktop.props
```diff
@@ -15,18 +15,26 @@
...
+    <!-- 
+      Represents an undefined TFV value. This will be used in comparisons of _TargetFrameworkVersionValue (defined in Microsoft.NET.WindowsDesktop.targets)
+      to identify when a TFV is undefined
+    -->
+    <_UndefinedTargetFrameworkVersion>0.0</_UndefinedTargetFrameworkVersion>
+    
+    <!--
+      Initial/Default value set to 'undefined'. Updated in Microsoft.NET.WindowsDesktop.targets
+    --> 
+    <_TargetFrameworkVersionValue>$(_UndefinedTargetFrameworkVersion)</_TargetFrameworkVersionValue>
```
rainersigwald: Is this needed? Are there circumstances where this file will be imported but `Microsoft.NET.WindowsDesktop.targets` won't?

eng/WpfArcadeSdk/tools/Pbt.targets
```diff
@@ -92,11 +92,20 @@
...
+    _WindowsDesktopSdkTargetFrameworkVersionFloor, _UndefinedTargetFrameworkVersion and _TargetFrameworkVersionValue are 
+    defined in Microsoft.NET.WindowDesktop.props. 
+    
+    These need to be defined before Microsoft.NET.Sdk.WindowsDesktop.targets is imported. 
   -->
   <PropertyGroup Condition="'$(InternalMarkupCompilation)'=='true' And !Exists('$(LocalMicrosoftWinFXTargets)') ">
     <_WindowsDesktopSdkTargetFrameworkVersionFloor Condition="'$(_WindowsDesktopSdkTargetFrameworkVersionFloor)' == ''">3.0</_WindowsDesktopSdkTargetFrameworkVersionFloor>
+    
+    <!--  Represents an undefined TFV value. -->
+    <_UndefinedTargetFrameworkVersion>0.0</_UndefinedTargetFrameworkVersion>
```
rainersigwald: I don't think I understand this change. When is this file used vs the other .targets file?

packaging/Microsoft.NET.Sdk.WindowsDesktop/targets/Microsoft.NET.Sdk.WindowsDesktop.props
vatsan-madhavan: Clearly we support such a pattern, like this `<Import Sdk="Microsoft.NET.Sdk.WindowsDesktop" Project="Sdk.props" />`. Whether someones custom-build setup will actually do something like that is anybody's guess - I figured we should be defensive.

eng/WpfArcadeSdk/tools/Pbt.targets
vatsan-madhavan: This doesn't ship - it's used only for building dotnet/wpf repo itself. Some WPF product assemblies require XAML compilation, but we do leverage the WindowsDesktop SDK in our build. Pbt.targets is part of the XAML compilation orchestration of those product assemblies. It tries to reuse as much of the actual WindowsDesktop SDK targets/logic as possible directly.

rainersigwald: ah, thanks! 👍

packaging/Microsoft.NET.Sdk.WindowsDesktop/targets/Microsoft.NET.Sdk.WindowsDesktop.props
rainersigwald: I wouldn't bother supporting it personally. Lots of things break if you don't have props/targets in matched pairs (despite the failed "common.targets imports common.props if it hasn't already been" attempt). But no strong objection.

dsplaisted:
Why doesn't the existing `'$(_TargetFrameworkVersionWithoutV)' != ''` check short-circuit the subsequent clauses and avoid this problem?

rainersigwald:
@dsplaisted It does, but only in the cases where it's in the same condition. It doesn't apply when the is-defined check is applied at the ItemGroup level.

vatsan-madhavan:
> Why doesn't the existing `'$(_TargetFrameworkVersionWithoutV)' != ''` check short-circuit the subsequent clauses and avoid this problem?

I'm not the expert - @rainersigwald should confirm; it does, but in this way I think: 
``` xml
 <ItemGroup Condition="$(A) != '' And $(A) >= 3.0> <!-- works ok --->
  <!-- You'd think that A != '' is  a given, but it needs to be repeated here - and everywhere else involving an expression containing 'A' - otherwise it doesn't count --> 
   <Foo Condition="$(A) >= 4.0" Include="Bar" /> 
</ItemGroup>

dsplaisted:
@vatsan-madhavan Assuming that's correct, would it not be simpler to just repeat the nonempty check in all the conditions that need to do comparisons on the version?

vatsan-madhavan:
> Assuming that's correct, would it not be simpler to just repeat the nonempty check in all the conditions that need to do comparisons on the version?

That's not an approach particularly suited to _maintainability_ IMO. It's simple to explain - yes, but it makes for worse readability, and sets up a system where small changes could incur complex and redundant conditions that should otherwise have been guaranteed by an enclosing scope already. 

It's somewhat arguable that such a change would even be even simpler. The proposed change effectively leaves all the conditional constructs intact (modulo small well defined transformations from `$(A) != ''` to`$(Y) != '0.0'`; `$(A)` -> `$(Y)`). Changing conditions all over the place also changes the semantics of how everything is evaluated - only just a bit - but enough to require more testing, more careful reviews etc. At this point in 3.0, lower the entropy of changes, the better it seems to me.

PR #1716, opened 2019-08-24
Enabling keyboard focus for page control buttons and zoom buttons on the FlowDocumentViewer toolbar

### LINKED ISSUE #1471: A11y_.NETCore3.0(WPF)_WPFAnnotationsStyling_Keyboard: The keyboard users are not able to perform various actions using keyboard
Copied from VSO [956777](https://devdiv.visualstudio.com/DevDiv/_queries/edit/956777/)

Environment Details:
Application Name: .NET Core 3.0 (WPF)

Application version: VS 2019 Preview 4.7.03190

Windows Version: Windows10

ScreenReader Version: Narrator, NVDA 2019.1.1

Pre-requisite:

Repro Steps:

1) Run exe file for "AnnotationsStyling" from the build app folder provided.
2) Main Window for "AnnotationsStyling" solution will appear.
3) Start screen reader.
4) Navigate to the contents in the main window.

Actual:

When keyboard users navigate using the "Tab" key within the main window they are unable to access the "Next Page",  "Decrease Zoom" and "Increase Zoom" buttons.

The same issue is observed using the both the screen readers i.e. Narrator and NVDA.

Expected:

All the actions that a mouse users is able to perform the same action keyboard and screen reader users  should be available to perform using keyboard.

User Impact:

If the screen reader and keyboard users are not able to interact with the interactive elements using the keyboard then they might not be able to perform all the functions as a mouse user is able to access.

Recommendations:

MAS Reference:

MAS 1.3.1

https://microsoft.sharepoint.com/:w:/r/teams/msenable/_layouts/15/WopiFrame.aspx?sourcedoc={54f28d1f-a2d1-4dcd-84e1-5c9b87e8aba4}

MAS 2.1.1

https://microsoft.sharepoint.com/:w:/r/teams/msenable/_layouts/15/WopiFrame.aspx?sourcedoc={8492c4eb-c179-40ae-8777-cd044ed725a2}

arpitmathur:
This change enables a user to keyboard focus on the page navigation controls and on the zoom controls in the toolbar in FlowDocumentPageViewer.

Fixes #1471

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/PresentationUI/Themes/Generic.xaml
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationUI/Themes/Generic.xaml
@@ -8449,7 +8449,7 @@ style should be placed here.
                                               ToolTip="Previous Page"
                                               Command="NavigationCommands.PreviousPage"
                                               IsEnabled="{TemplateBinding CanGoToPreviousPage}"
-                                              Focusable="False">
+                                              Focusable="True">
                                     <Viewbox x:Uid="Viewbox_5" Height="{TemplateBinding FontSize}">
                                         <Path x:Uid="Path_15" Fill="{TemplateBinding Foreground}">
                                             <Path.Data>
@@ -8472,7 +8472,7 @@ style should be placed here.
                                               ToolTip="Next Page" 
                                               Command="NavigationCommands.NextPage" 
                                               IsEnabled="{TemplateBinding CanGoToNextPage}"
-                                              Focusable="False">
+                                              Focusable="True">
                                     <Viewbox x:Uid="Viewbox_6" Height="{TemplateBinding FontSize}">
                                         <Path x:Uid="Path_16" Fill="{TemplateBinding Foreground}">
                                             <Path.Data>
@@ -8501,7 +8501,7 @@ style should be placed here.
                                               BorderBrush="{TemplateBinding BorderBrush}"
                                               Width="{TemplateBinding FontSize}" 
                                               Height="{TemplateBinding FontSize}" 
-                                              Focusable="False">
+                                              Focusable="True">
                                     <RepeatButton.LayoutTransform>
                                         <ScaleTransform x:Uid="ScaleTransform_12" ScaleX="2.2" 
                                                         ScaleY="2.2"/>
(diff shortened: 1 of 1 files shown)


rladuca:
`PresentationUI` has a theme generator and generic.xaml is one of the inputs.  Are you sure you don't need to run the generator and/or make similar changes to what the generator would do?

Can you check what the generator does in .NET Framework and understand the mechanism so we all can understand why editing just generic.xaml is the correct change?

rladuca:
> `PresentationUI` has a theme generator and generic.xaml is one of the inputs. Are you sure you don't need to run the generator and/or make similar changes to what the generator would do?
> 
> Can you check what the generator does in .NET Framework and understand the mechanism so we all can understand why editing just generic.xaml is the correct change?

It looks like all the generator does is strip comments and make an SDK folder with the PUI themes.

@vatsan-madhavan @SamBent do either of you have any idea what this SDK folder is used for?  It's not part of .NET Framework's installation.  Maybe this feeds into VS supplied theme files?  You can take a look on the latest drops of .NET Framework, it's under the wpf bin directory.

rladuca:
> > `PresentationUI` has a theme generator and generic.xaml is one of the inputs. Are you sure you don't need to run the generator and/or make similar changes to what the generator would do?
> > Can you check what the generator does in .NET Framework and understand the mechanism so we all can understand why editing just generic.xaml is the correct change?
> 
> It looks like all the generator does is strip comments and make an SDK folder with the PUI themes.
> 
> @vatsan-madhavan @SamBent do either of you have any idea what this SDK folder is used for? It's not part of .NET Framework's installation. Maybe this feeds into VS supplied theme files? You can take a look on the latest drops of .NET Framework, it's under the wpf bin directory.

I don't think the generator is anything to worry about with respect to the actual DLL functionality, but we should understand what it was for.  That's ancillary to this PR though.

src/Microsoft.DotNet.Wpf/src/PresentationUI/Themes/Generic.xaml
```diff
@@ -8449,7 +8449,7 @@ style should be placed here.
                                               ToolTip="Previous Page"
                                               Command="NavigationCommands.PreviousPage"
                                               IsEnabled="{TemplateBinding CanGoToPreviousPage}"
-                                              Focusable="False">
+                                              Focusable="True">
```
rladuca: Annotate .NET Framework's file and see if there was ever an explicit reason these were not focusable.  You can also search some of the older bug databases for something.

It doesn't feel like these should have been configured this way, but sometimes there are good reasons.

Issue #1549, opened 2019-08-06
WPF controls render slower on .NET Core 3 compared to .NET Framework 4.8


AlexChuev:
* .NET Core 3.0 Preview 7
* Windows version 1809
* Does the bug reproduce also in WPF for .NET Framework 4.8?: No
 
 **Problem description:**  
 According to our performance tests, visual operations take longer to execute in .NET Core 3 compared to .NET Framework 4.8. We were unable to narrow down the issue so far, but it is certainly not connected to the performance issue with the GetCustomAttributes method reported [here](https://github.com/dotnet/wpf/issues/94), since you can see the difference in performance even when all XAML has been parsed (e.g., when scrolling).
 
 **Minimal repro:**  
 You can find a small sample with the Data Grid control here: https://github.com/AlexChuev/NetCorePerformanceTest
In the sample, it takes about 30% more time to scroll the Data Grid vertically in a .NET Core 3 project compared to an identical .NET Framework 4.8 configuration.

rladuca:
First off @AlexChuev thank you very much for reporting this problem.  From this we've gotten a much better idea of some areas we need to investigate and ensure we have performance metrics and testing for.

Second, here is an update on where we are with this investigation.

We've narrowed down a few areas where we are seeing performance regressions.
* There are small layout regressions where we are occasionally doing more layouts
* There is a larger issue with text rendering and layout

We are currently focusing on this text rendering and layout case as it is the larger contributor to the problem and we have the best shot at addressing it in a short time frame.

From some experimentation, it looks as if a strong possibility is that an early on split of PresentationCore into two assemblies (PresentationCore and DirectWriteForwarder) might be responsible for this text rendering and layout regression.  Replacing the latest split binaries with a merged version from Preview 1, erases the difference.  When this is done and the repro app is changed slightly to warm up code paths (as these older binaries are being fully JITed) we are actually faster than .NET Framework.

We're looking into multiple avenues for fixing this, including going back to a single assembly that contains both C++/CLI and C# code (as it is in .NET Framework).  We've also engaged .NET JIT teams to investigate if there is something that can be done to improve performance in the case where the assemblies are split.

I'm going to continue updating here and as we move into other perf investigations I will file separate issues.

weltkante:
Nice work on the analysis, if there's any public issues on coreclr or corefx related to this it'd be great to link them.

rladuca:
[...]

PR #1727, opened 2019-08-26
Re-merge DirectWriteForwarder into PresentationCore

rladuca:
Addresses #1549 

Reverting to using a NetModule and merging with a C++/CLI object file in order to build the C++/CLI and C# components of PresentationCore into a single module.

Tested:
* DRTs have passed on x86 and x64
* Some basic performance testing applications run without issue
* Performance is about the same as prior tests with Preview1 binaries and hand-merged IL.
* DRTs have passed on x86 and x64 crossgen assemblies
* Microsuite has passed on crossgen assemblies on RS5.  Full pass still in progress.
* Full test pass has passed on crossgen assemblies across our standard OS installs.
* Crossgen team has said there should be no issue with crossgen and the merged assembly

In Progress:
* IL diffs of various sorts (Merged vs Unmerged vs .NET Framework Merge)


rladuca:
We have a better pathway to fix this, closing.  See #1744

weltkante:
Thanks anyways for doing this in the open, I appreciate having an example of how to build and merge netmodules, this is pretty underdocumented. I've had to do this in the past and it was a lot of trial and error and I don't think I ever got it entirely right. I'm [preserving the branch](https://github.com/weltkante/wpf/pull/1) for future reference, if thats ok with you.

rladuca:
> Thanks anyways for doing this in the open, I appreciate having an example of how to build and merge netmodules, this is pretty underdocumented. I've had to do this in the past and it was a lot of trial and error and I don't think I ever got it entirely right. I'm [preserving the branch](https://github.com/weltkante/wpf/pull/1) for future reference, if thats ok with you.

Preserving it is completely fine.  Just to make sure you realize, currently the private C++/CLI tools we use are necessary to build this merged module.  Otherwise you will hit a linker error during metadata merging of the individual modules.  These changes will eventually make their way to preview/public versions of VS, but they are not there just yet.

There are other uses for this as well within the WPF build and we may yet come back to some version of this to accomplish other things.

If you have any questions about parts of this, feel free to ask!

PR #1744, opened 2019-08-29
Fix DebuggableAttribute in InjectModuleInitializer post-build targets

rladuca:
Addresses #1549 

In the `ReconstituteDll` target, we call ILAsm to merge a module constructor into the assembly we previously disassembled.  We need to enable JIT optimizations here so that we get an appropriate `DebuggableAttribute` on the assembly and don't cause perf problems.

@vatsan-madhavan who helped debug and find this fix.

### DIFF
--- a/eng/WpfArcadeSdk/tools/InjectModuleInitializer.targets
+++ b/eng/WpfArcadeSdk/tools/InjectModuleInitializer.targets
@@ -404,6 +404,11 @@
       <ILFile>$([System.IO.Path]::GetFullPath('$(IntermediateOutputPath)$(AssemblyName).il'))</ILFile>
     </PropertyGroup>
 
+    <PropertyGroup>
+      <DebugEnableJitOptimization>false</DebugEnableJitOptimization>
+      <DebugEnableJitOptimization Condition="'$(Configuration)' == 'Release'">true</DebugEnableJitOptimization>
+    </PropertyGroup>
+
     <!-- 
       Make a backup before overwriting $(TargetFile) 
     -->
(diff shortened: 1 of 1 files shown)


ericstj:
Curious why this needs to be specified and it isn't just persisted from the source.  Is that a bug in ILAsm or ILDasm?  Have we filed that issue?

rladuca:
> Curious why this needs to be specified and it isn't just persisted from the source. Is that a bug in ILAsm or ILDasm? Have we filed that issue?

I'll have to check with later versions of .NET IL(D)Asm/ILAsm.  We're pegged to .NET 4.6.2 due to an issue with the MSBuild tool locator, but also due to [this bug](https://developercommunity.visualstudio.com/content/problem/545431/ildasmexe-regression-with-infinum-floating-point-v.html) with round-tripping in .NET 4.8.

We only just found out about these in the last couple of days, but we'll be sure to file the appropriate issues on both teams.

rladuca:
> > Curious why this needs to be specified and it isn't just persisted from the source. Is that a bug in ILAsm or ILDasm? Have we filed that issue?
> 
> I'll have to check with later versions of .NET IL(D)Asm/ILAsm. We're pegged to .NET 4.6.2 due to an issue with the MSBuild tool locator, but also due to [this bug](https://developercommunity.visualstudio.com/content/problem/545431/ildasmexe-regression-with-infinum-floating-point-v.html) with round-tripping in .NET 4.8.
> 
> We only just found out about these in the last couple of days, but we'll be sure to file the appropriate issues on both teams.

In the disassembled IL, this attribute is commented out and it indicates this will be added automatically.  When I did a recombination of `DirectWriteForwarder` and `PresentationCore` with IL(D)Asm, we saw the perf increase as well.  The difference here, I believe, is passing the debug flag to get PDB output.  When you do that (as we do in WPF builds) I think ILAsm overrides the attribute, requiring you to manually specify optimizations to be enabled.

ericstj:
I see, makes sense.

rladuca:
Approved by tactics, merging.

Issue #1758, opened 2019-08-31
EnableDefaultApplicationDefinition to false prevents Page's from working


onovotny:
* .NET Core Version: (e.g. 3.0 Preview1, or daily build number, use `dotnet --info`) 3.0 preview 8

 
 
 **Problem description:**

If you set `EnableDefaultApplicationDefinition` to `false`, the other page XAML items get set to None instead of Page.
 
 **Actual behavior:** 
 
 **Expected behavior:**
Page items should still be working 

 **Minimal repro:**
 
New WPF app, add a page.

Set `EnableDefaultApplicationDefinition` to `false`.

rladuca:
Thanks for reporting this @onovotny !  This was a regression from a prior change to support Visual Studio's reading of None items (see #685).

This fix was merged in both master and release/3.0 branches.

PR #1768, opened 2019-09-04
Fixing removal of None from Page under certain conditions

rladuca:
Addresses #1758 

Fixing removal of None from Page under certain conditions (EnableDefaultApplicationDefinition or EnableDefaultPageItems is false).

Further adding a fix to ensure this doesn't result in ApplicationDefinition being included in Page as well when EnableDefaultApplicationDefinition is false.

@nguerrera Can you take a look at this as well?  Thanks!

### DIFF
--- a/packaging/Microsoft.NET.Sdk.WindowsDesktop/targets/Microsoft.NET.Sdk.WindowsDesktop.targets
+++ b/packaging/Microsoft.NET.Sdk.WindowsDesktop/targets/Microsoft.NET.Sdk.WindowsDesktop.targets
@@ -27,11 +27,26 @@
 
   <ItemGroup Condition=" ('$(EnableDefaultItems)' == 'true') And ('$(UseWPF)' == 'true') And ('$(_TargetFrameworkVersionValue)' != '$(_UndefinedTargetFrameworkVersion)') And 
                          ('$(_TargetFrameworkVersionValue)' >= '$(_WindowsDesktopSdkTargetFrameworkVersionFloor)')">
-    
-    <!-- In the WindowsDesktop .props, we globbed all .xaml files as Page items.  If any of those files are included
-         as Resource, Content, or None items, then remove them from the Page items. -->
-    <Page Remove="@(Resource);@(Content);@(None)"
+
+    <!--
+         In the WindowsDesktop .props, we globbed all .xaml files as Page items.  If any of those files are included
+         as Resource, Content then remove them from the Page items.
+         
+         If EnableDefaultApplicationDefinition is false, ApplicationDefinition must also be removed from Page as
+         PresentationBuildTasks will include the generated code for the ApplicationDefinition twice otherwise.
+         We cannot do this in all cases in the props file as ApplicationDefinition won't be available
+         at that point in the evaluation when EnableDefaultApplicationDefinition is false.
+    -->
+    <Page Remove="@(Resource);@(Content);@(ApplicationDefinition)"
           Condition="'$(EnableDefaultPageItems)' != 'false'" />
+
+    <!--
+         Only remove None from Page in the event that we disable both the default ApplicationDefinition and the DefaultPageItems.
+         Otherwise we can enter a situation where we are explicitly removing page items that should otherwise be included (such as
+         when (EnableDefaultApplicationDefinition xor EnableDefaultPageItems) is true).
+    -->
+    <Page Remove="@(None)"
+          Condition="'$(EnableDefaultApplicationDefinition)' != 'false' and '$(EnableDefaultPageItems)' != 'false'" />
   </ItemGroup>
 
   <!-- Generate error if there are duplicate page items.  The task comes from the .NET SDK, and this target follows
--- a/packaging/Microsoft.NET.Sdk.WindowsDesktop/targets/Microsoft.NET.Sdk.WindowsDesktop.targets
+++ b/packaging/Microsoft.NET.Sdk.WindowsDesktop/targets/Microsoft.NET.Sdk.WindowsDesktop.targets
@@ -32,7 +32,7 @@
          In the WindowsDesktop .props, we globbed all .xaml files as Page items.  If any of those files are included
          as Resource, Content then remove them from the Page items.
     -->
-    <Page Remove="@(Resource);@(Content);@(ApplicationDefinition)"
+    <Page Remove="@(Resource);@(Content)"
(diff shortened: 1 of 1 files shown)


packaging/Microsoft.NET.Sdk.WindowsDesktop/targets/Microsoft.NET.Sdk.WindowsDesktop.targets
```diff
@@ -27,11 +27,26 @@
...
+    <!--
+         In the WindowsDesktop .props, we globbed all .xaml files as Page items.  If any of those files are included
+         as Resource, Content then remove them from the Page items.
+         
+         If EnableDefaultApplicationDefinition is false, ApplicationDefinition must also be removed from Page as
+         PresentationBuildTasks will include the generated code for the ApplicationDefinition twice otherwise.
+         We cannot do this in all cases in the props file as ApplicationDefinition won't be available
+         at that point in the evaluation when EnableDefaultApplicationDefinition is false.
+    -->
+    <Page Remove="@(Resource);@(Content);@(ApplicationDefinition)"
```
vatsan-madhavan: We should remove `@(ApplicationDefinition)` unconditionally - otherwise project-supplied value for `@(ApplicationDefinition)` will not be removed here. I would propose something like this: 

 ```xml

<Page Remove="@(Resource);@(Content)" Condition="'$(EnableDefaultPageItems)'!='false'" />
<Page Remove="@(ApplicationDefinition)" />
```

**Comment by nguerrera** [comment]
@dsplaisted

File: packaging/Microsoft.NET.Sdk.WindowsDesktop/targets/Microsoft.NET.Sdk.WindowsDesktop.targets (hunk 1 above)
Comment (rladuca) [reply packaging/Microsoft.NET.Sdk.WindowsDesktop/targets/Microsoft.NET.Sdk.WindowsDesktop.targets]: Forgot to make that change this morning.

**Comment by rladuca** [comment]
@nguerrera @dsplaisted Any input on this?

File: packaging/Microsoft.NET.Sdk.WindowsDesktop/targets/Microsoft.NET.Sdk.WindowsDesktop.targets [hunk 2]
```diff
@@ -27,11 +27,30 @@
...
+        If ApplicationDefinition must also be removed from Page as PresentationBuildTasks will include the
+        generated code for the ApplicationDefinition twice otherwise.  We cannot do this in all cases in the
+        props file as ApplicationDefinition won't be available at that point in the evaluation when
+        EnableDefaultApplicationDefinition is false.  We unconditionally remove it here as ApplicationDefinition
+        is considered a special singleton Item that WPF ensures is only where it needs to be.
+    -->
+    <Page Remove="@(ApplicationDefinition)" />
+    
+    <!--
+         Only remove None from Page in the event that we disable both the default ApplicationDefinition and the DefaultPageItems.
```
dsplaisted: I think that this comment doesn't accurately reflect the code (though it's hard to tell with all the negatives).  I think it would be correct to replace "disable" with "enable".

Also I'm not following why EnableDefaultApplicationDefinition should affect whether `None` is removed from `Page`.

vatsan-madhavan:
@rladuca, Marking as milestone=5.0 since this targets _master_. We need another PR targeting _release/3.0_

rladuca:
> Looks good, though I have a question about the last bit.
> 
> Also, I would really love to see test coverage of the different cases here.

That's fair, we'd have to add this via manual tests on the backend since we won't have access to a full up SDK with these enabled.  We can probably get this working by replacing the files in the builds, but for now, a manual set should suffice.

I am creating a set of tests that can easily be run for various switch combinations, I'll put them up on GitHub and we can use them in validations going forward.

packaging/Microsoft.NET.Sdk.WindowsDesktop/targets/Microsoft.NET.Sdk.WindowsDesktop.targets
rladuca: If you set EnableDefaultApplicationDefinition to false, then we don't remove anything from none here: https://github.com/dotnet/wpf/blob/f0c10e1364f1230b053ce6f2c9832817f88a9758/packaging/Microsoft.NET.Sdk.WindowsDesktop/targets/Microsoft.NET.Sdk.WindowsDesktop.props#L64

This means None will contain page items like MainWindow.xaml.  If you then remove None from Page, you'll remove items that should be there.

The change to none was intentional (see the comment above) since VS couldn't handle more specific globbing fixups to None.  In order to preserve that and fix the issue from this bug, we have to condition removing None from Page.

vatsan-madhavan:
> I am creating a set of tests that can easily be run for various switch combinations, I'll put them up on GitHub and we can use them in validations going forward.

It's possible to re-use the tests (the infrastructure portion) from dotnet/sdk (or pehraps it is from dotnet/core-sdk) by consuming them as NuGet package and bootstrap testing for WindowsDesktop SDK. It's the sort of thing we should do in .NET 5/master (and not create a disparate test-framework/system for this).

packaging/Microsoft.NET.Sdk.WindowsDesktop/targets/Microsoft.NET.Sdk.WindowsDesktop.targets
rladuca: Updated comments to better explain the conditions.

rladuca:
> > I am creating a set of tests that can easily be run for various switch combinations, I'll put them up on GitHub and we can use them in validations going forward.
> 
> It's possible to re-use the tests (the infrastructure portion) from dotnet/sdk (or pehraps it is from dotnet/core-sdk) by consuming them as NuGet package and bootstrap testing for WindowsDesktop SDK. It's the sort of thing we should do in .NET 5/master (and not create a disparate test-framework/system for this).

Yeah, that makes complete sense in the long term.  The tests I have are basic project builds that I scripted to test these changes myself.  I'll give these to our test team for manual verification for immediate previews and daily builds so we have some coverage until we can get a real infrastructure set up and running.

rladuca:
I uploaded test cases I used locally to verify this change (via replacing the relevant code in a local install).  They can be found [here](https://github.com/rladuca/WindowsDesktopSwitchTests).  If you have any suggestions on extra coverage or different switch combinations, let me know.

I'll give these to our test team with instructions to run against our daily builds once these changes make it to the SDK.

PR #1790, opened 2019-09-05
Fixing removal of None from Page under certain conditions

rladuca:
Addresses #1758

# Description 

Fixing removal of None from Page under certain conditions 

# Issue Details 

Under default circumstances, WindowsDesktop props will remove “**\*.xaml” from the None MSBuild Item in order to support Visual Studio.  Items that are left in None are then removed from Page in the WindowsDesktop targets.  This was a change made to address complaints here about more specific removal criteria that Visual Studio was unable to handle. 

When a developer sets EnableDefaultApplicationDefinition to false, WindowsDesktop props will not remove “**\*.xaml” from None as it cannot only remove the automatically globbed XAML (as we’re no longer globbing all XAML).  This causes issues in the WindowsDesktop targets since removing None from Page now causes compilation errors. 

The solution is to guard the removal of None from Page with the same conditional as the removal of “**\*.xaml” from None so we only do this work in the same fully default configuration.  This change exposes another issue with ApplicationDefinition being multiply included in compilation so we also unconditionally remove ApplicationDefinition from Pages as WPF’s PresentationBuildTasks treats ApplicationDefinition as a special singleton that shouldn’t exist in any other XAML compilation items. 

## Customer Impact

*Severity*:  Medium 

Customers using non-default ApplicationDefinitions (named something other than App.xaml or Application.xaml) will be unable to compile without explicitly adding back removed Page items.  Essentially, this means turning off default ApplicationDefinitions turns off default Page includes as well.  

## Regression? 

Yes, from the work to support VS in https://github.com/dotnet/wpf/issues/685 

### Mitigations: 

* The None item hasn’t changed in semantics so VS still will see the same behavior there 

* The Page item has the same behavior in the default case and fixed behavior in the currently broken case 

* The Page item will no longer carry ApplicationDefinition under any scenario, which is desirable. 

### Validations 

I’ve run builds against combinations of flags using a modified .NET Core 3 RC1 installation (See these tests).  All pass as expected.  I’ll add these tests to our manual test team validations for now and further work can be queued up for .NET 5 to automate these.

### DIFF
--- a/packaging/Microsoft.NET.Sdk.WindowsDesktop/targets/Microsoft.NET.Sdk.WindowsDesktop.targets
+++ b/packaging/Microsoft.NET.Sdk.WindowsDesktop/targets/Microsoft.NET.Sdk.WindowsDesktop.targets
@@ -27,11 +27,26 @@
 
   <ItemGroup Condition=" ('$(EnableDefaultItems)' == 'true') And ('$(UseWPF)' == 'true') And ('$(_TargetFrameworkVersionValue)' != '$(_UndefinedTargetFrameworkVersion)') And 
                          ('$(_TargetFrameworkVersionValue)' >= '$(_WindowsDesktopSdkTargetFrameworkVersionFloor)')">
-    
-    <!-- In the WindowsDesktop .props, we globbed all .xaml files as Page items.  If any of those files are included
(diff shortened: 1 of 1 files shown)


rladuca:
Setting auto-merge as I rebased to resolve merge conflicts and this has been approved by tactics.

PR #1895, opened 2019-09-16
Ensure that markup compilation is run for all design time builds

davidwengier:
Fixes https://github.com/dotnet/project-system/issues/2488

In SDK-style projects the project system will combine design time builds together to improve performance, but depending on timing, this can cause issues. If something like the designer asks for output groups, then the build for that could be combined with the build that produces the intellisense file, but getting output groups sets BuildingProject to true.

This change works around the issue by allowing the intellisense file to generate even if `BuildingProject` is true, as long we're still doing a design time build.

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/PresentationBuildTasks/Microsoft.WinFx.targets
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationBuildTasks/Microsoft.WinFx.targets
@@ -106,7 +106,7 @@
   <Target Name="DesignTimeMarkupCompilation">
 
         <!-- Only if we are not actually performing a compile i.e. we are in design mode -->
-        <CallTarget Condition="'$(BuildingProject)' != 'true'"
+        <CallTarget Condition="'$(BuildingProject)' != 'true' Or $(DesignTimeBuild) == 'true'"
                 Targets="MarkupCompilePass1" />
   </Target>
 


nguerrera:
I assume we'd want to take this in 3.1 so that the experience is fixed for 16.4? Is the process for WPF like corefx? Check in to master than get it approved and cherry picked to 3.1?

nguerrera:
Also, worth noting that https://github.com/dotnet/project-system/issues/2488 was filed a long while ago while using the framework's in-box Microsoft.WinFx.targets. To get this fix, folks would have to switch to WindowsDesktop SDK, which I assume is OK as we never officially supported WPF in CPS in a classic project.

davidwengier:
> To get this fix, folks would have to switch to WindowsDesktop SDK, which I assume is OK as we never officially supported WPF in CPS in a classic project.

Certainly this fix is only meant to avoid a bad experience for people updating to SDK projects, but @davkean and I were discussing this yesterday and it looks like we could in a future change the MSBuild targets so that classic projects opening in CPS use the WindowsDesktop SDK instead of the ones from the framework. Obviously that would depend on the WPF team sign-off.

src/Microsoft.DotNet.Wpf/src/PresentationBuildTasks/Microsoft.WinFx.targets
```diff
@@ -106,7 +106,7 @@
   <Target Name="DesignTimeMarkupCompilation">
 
         <!-- Only if we are not actually performing a compile i.e. we are in design mode -->
-        <CallTarget Condition="'$(BuildingProject)' != 'true'"
+        <CallTarget Condition="'$(BuildingProject)' != 'true' Or $(DesignTimeBuild) == 'true'"
```
davkean: I think we should just remove the "BuildingProject" directive completely - it just adds confusion to this block as to when it is set/not set. This used to be used to detect legacy design-time builds, but this targets file will not be run in legacy.

vatsan-madhavan: Is `BuildingProject` not set anywhere ?

vatsan-madhavan:
> > To get this fix, folks would have to switch to WindowsDesktop SDK, which I assume is OK as we never officially supported WPF in CPS in a classic project.
> 
> Certainly this fix is only meant to avoid a bad experience for people updating to SDK projects, but @davkean and I were discussing this yesterday and it looks like we could in a future change the MSBuild targets so that classic projects opening in CPS use the WindowsDesktop SDK instead of the ones from the framework. Obviously that would depend on the WPF team sign-off.

The current design (of PBT, WinFX.targets) has been done with build support for .NET Framework and multi-targeting in mind. We can certainly look into this once WindowsDesktop SDK + .NET Core experience has matured a bit and we have had a chance to find and fix any more lurking bugs.

src/Microsoft.DotNet.Wpf/src/PresentationBuildTasks/Microsoft.WinFx.targets
vatsan-madhavan: ok, we'll remove it in a separate pass (https://github.com/dotnet/wpf/issues/1914).

PR #1929, opened 2019-09-19
Cleanup DirectWriteForwarder

wjk:
This PR is the first in a series of PRs that I am going to create that clean up (according to my, admittedly subjective, opinions) the code in this repo. There will be one PR created per project/group of closely related projects (e.g. Extensions, Themes). I will endeavor to cause zero functional changes and zero API surface changes in these PRs. I am perfectly OK with these being delayed until after 3.0 release, or if any of my cleanup ideas is contrary to what the core WPF maintainers desire. This is just really, really hairy code, and I thought giving it a shave would be an interesting use of my time.

In this PR, I only went over a very few files, as the others were not listed in the Solution Explorer, and due to the complex build process, IntelliSense was not working in the C++ files. Thanks!

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/Shared/cpp/Utils.cxx
+++ b/src/Microsoft.DotNet.Wpf/src/Shared/cpp/Utils.cxx
@@ -5,10 +5,6 @@
 #include "Utils.hxx"
 #include <shlwapi.h>
 
-#if _MANAGED
-using namespace System::Security;
-#endif
-
 // These constants are cloned in
 // wpf\src\Shared\MS\Internal\Registry.cs
 // Should these reg keys change the above file should be also modified to reflect that.
@@ -30,8 +26,6 @@ namespace WPFUtils {
 // If the function succeeds, the return value is ERROR_SUCCESS.
 // If the function fails, the return value is a nonzero error code defined in Winerror.h
 //
-#if _MANAGED
-#endif
 LONG ReadRegistryString(__in HKEY rootKey, __in LPCWSTR keyName, __in LPCWSTR valueName,
                                      __out LPWSTR value, size_t cchMax)
 {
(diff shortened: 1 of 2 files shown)


wjk:
I have since discovered (while working on cleaning up System.Xaml) exactly how tall an order this task really is. Therefore, I may not actually get around to creating any other cleanup PRs. However, this PR is still ready for review. Thanks for bearing with!

src/Microsoft.DotNet.Wpf/src/Shared/cpp/Utils.cxx
```diff
@@ -5,10 +5,6 @@
 #include "Utils.hxx"
 #include <shlwapi.h>
 
-#if _MANAGED
```
vatsan-madhavan: A copy of this file is used by PresentationNative (i.e., not managed :-)). 

Since this is under Shared, it gets packaged by `Microsoft.DotNet.Arcade.Wpf.Sdk` transport-package, and flows into the private `dotnet-wpf-int` repository, where it is consumed by the build for PresentationNative_cor3.dll. 

https://github.com/dotnet/wpf/blob/55815c9287682c1621cd5f89ffdd75ae770f4ea6/packaging/Microsoft.DotNet.Arcade.Wpf.Sdk/Microsoft.DotNet.Arcade.Wpf.Sdk.ArchNeutral.csproj#L38-L41



In short, you can't do this.

wjk: I'm afraid I don't quite understand. What I'm removing has no effect when the file is _not_ managed, as the `#if` would evaluate to false then. Furthermore, all that these blocks contain is an unused `using namespace System.Security`. This using statement is meaningless when not compiling managed, and on top of that it isn't even used. Uses of the types inside this namespace were removed in e4a212b4fd7d3bb6bce7d88cf6f0471a9fc8e6f4, but the using statement (and the empty `#if` block below) were not removed at the same time.

vatsan-madhavan: You're right - I wasn't looking too carefully. I was still thinking this had the old system::security CAS declarations. You can definitely do this. I'll build your branch against the internal branch just to be sure everything builds ok and let you know.

vatsan-madhavan:
thanks for the PR!

PR #1933, opened 2019-09-20
Move System.Xaml.sln into System.Xaml directory

wjk:
This was done to align with the solution files in the `PresentationBuildTasks` and `WindowsBase` directories.


vatsan-madhavan:
thanks!

PR #1934, opened 2019-09-20
Remove outdated README files

wjk:
There is no longer any need to warn users about modifying the code in these directories, as it has long since been open-sourced. I have kept the warning file in the Shared directory for now, as code in the remaining native projects may well reference that code.


vatsan-madhavan:
thanks!

PR #1947, opened 2019-09-24
Fix for issue 1883

### LINKED ISSUE #1883: Copy file procedure fails in copy-wpf.ps1 file
.NET Core Version: **3.0.100-preview9-014004**
Windows version: **1903 (build SO 18362.356)**
Does the bug reproduce also in WPF for .NET Framework 4.8?: **not applicable**
 
 **Problem description:**
The macro doesn't copy the files because the source directory is targeted 3.0 yet.
 
 **Actual behavior:**
The procedure exit without any message about fail.
 
 **Expected behavior:**
 All WPF files are copyed or a message that tells the source directory doesn't exists.

Maybe I was wrong but I've commit a solution to this bug and it appears in the pull request of another bug [1761](https://github.com/dotnet/wpf/pull/1761)
 Please tell me how I should behave

Simon-IT:
More command-line messages are shown into the prompt window when the WPF file copy procedure is used. It has been modified the source directory of the binary files from 3.0 to 5.0.

Fixes #1883

### DIFF
--- a/eng/copy-wpf.ps1
+++ b/eng/copy-wpf.ps1
@@ -27,13 +27,21 @@ function Print-Usage()
     Write-Host ""
 }
 
+Write-Host "*** Copy WPF files procedure ***"
+
 $RepoRoot = Resolve-Path (Join-Path $PSScriptRoot "..")
 $Config = if ($release) { "Release" } else { "Debug" }
+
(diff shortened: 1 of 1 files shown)


stevenbrix:
Thanks for this PR @Simon-IT! I'm re-running the failed checks, just for precaution, even though there isn't anything wrong with this PR

PR #1976, opened 2019-10-01
Adding border to ToggleButton when checked in high-contrast themes

arpitmathur:
In non-high contrast themes, when you check a ToggleButton and navigate to another control, we display the checked status by highlighting the button with a lighter shade of the selection color.

We don't follow this behavior in high contrast themes. The text is slightly-depressed upon checking but that is not visually apparent and would definitely not be to a visually impaired person.
highcontrastwhitetogglebutton_without_change

In the image above, the 'S', 'C' and 'I' buttons are checked. However, it's barely noticeable that 'S' and 'C' are checked. The checked status of the 'I' button is also not noticeable when it receives focus.

The fix here is to check for a toggle button to be checked when in a high-contrast theme, we then add a border around the button. When a checked button also has keyboard focus, we add a border in the background-color. When the user navigates away, the border color is the same as the selection-color.

In the image below, the 'S', 'C' and 'I' buttons are checked again. Here, it's visually apparent that they are currently checked.
highcontrastwhitetogglebutton_with_change

In this pull request, relevant changes are in lines 673:693 in Button.xaml, the remaining files are generated off those changes.

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/Themes/XAML/Button.xaml
+++ b/src/Microsoft.DotNet.Wpf/src/Themes/XAML/Button.xaml
@@ -1,5 +1,7 @@
 <!--=================================================================
-Copyright (C) Microsoft Corporation.  All rights reserved.
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
 ==================================================================-->
 
 <!-- [[Aero.NormalColor]] -->
--- a/src/Microsoft.DotNet.Wpf/src/Themes/XAML/Button.xaml
+++ b/src/Microsoft.DotNet.Wpf/src/Themes/XAML/Button.xaml
@@ -680,6 +680,7 @@
             <MultiDataTrigger.Conditions>
                 <Condition Binding="{Binding Path=(SystemParameters.HighContrast)}" Value="true"/>
                 <Condition Binding="{Binding Path=IsChecked, RelativeSource={RelativeSource Self}}" Value="true"/>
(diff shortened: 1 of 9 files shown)


src/Microsoft.DotNet.Wpf/src/Themes/XAML/Button.xaml
```diff
@@ -1,40 +1,40 @@
 <!--=================================================================
-Licensed to the .NET Foundation under one or more agreements.
-The .NET Foundation licenses this file to you under the MIT license.
-See the LICENSE file in the project root for more information.
+// Licensed to the .NET Foundation under one or more agreements.
```
rladuca: Why are you adding these?

rladuca: Oh, did these come from the generator?  That makes sense.

src/Microsoft.DotNet.Wpf/src/Themes/XAML/Button.xaml
```diff
@@ -59,632 +59,648 @@ See the LICENSE file in the project root for more information.
...
+            Value="Center"/>
+    <Setter Property="Template">
+        <Setter.Value>
+            <ControlTemplate TargetType="{x:Type ButtonBase}">
+                <theme:ButtonChrome x:Name="Chrome"
+                                    BorderBrush="{TemplateBinding BorderBrush}"
+                                    Fill="{TemplateBinding Background}"
+                                    RenderDefaulted="{TemplateBinding Button.IsDefaulted}"
+                                    RenderMouseOver="{TemplateBinding IsMouseOver}"
+
```
rladuca: Explicitly add IsKeyboardFocused false check here, otherwise you're depending on evaluation order of the triggers to get the correct BorderBrush.

PR #1977, opened 2019-10-01
Removing separators from SizeOfSet and PositionInSet counts

### LINKED ISSUE #1467: A11y_.NETCore3.0(WPF)_WPFConcepts_MenuSample_Screenreader: Screen reader is announcing wrong ordering for the file /edit / Project menuitem
Copied from VSO[ 956410](https://devdiv.visualstudio.com/DevDiv/_queries/edit/956410/)

Environment Details:

Application Name: .NET Core 3.0 WPF

.NET Core SDK version:  3.0.100-preview8-013442
Windows Version: Windows10

Screen Reader Vesion : Narrator, NVDA 2019.1.1

Pre-requisite:

Repro Steps:

1) Run exe file for "Concepts" from the build app folder provided. 
2) Main Window for "Concepts" solution will appear. 
3) Select "Samples" tab item. 
4) From right navigation under "Controls" select "Menu Sample". 
5) Navigate to the preview tab and the to "File"/"Edit" menu item.
6) start screen reader and observe the screen reader announcement.

Actual:

Screen reader is announcing wrong ordering for the file /edit / Project menu item.
for example there are four sub menu inside the Edit menu item but screen reader is announcing for first sub menu item as 1 of 5.
Expected:

Screen reader should announce proper ordering for sub menu item.

User Impact:

Screen reader user may get confuse with ordering inside the menuitem.

Recommendations:

MAS Reference:

MAS1.3.1

https://microsoft.sharepoint.com/teams/msenable/_layouts/15/WopiFrame.aspx?sourcedoc={54f28d1f-a2d1-4dcd-84e1-5c9b87e8aba4}

arpitmathur:
Screen readers were reading incorrect values of the number of menu items and their position since we were counting separators as MenuItems in the automation tree. 

I filtered for separators in the current MenuItems to fix the SizeofSet count.

I filtered for separators again for PositionInSet but added a short-circuit to break when you're at the element whose position you're trying to find. This short-circuit was added to not remove separators below the relevant MenuItem from the PositionInSet count.

Fixes #1467

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Automation/Peers/MenuItemAutomationPeer.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Automation/Peers/MenuItemAutomationPeer.cs
@@ -94,13 +94,20 @@ namespace System.Windows.Automation.Peers
         override protected int GetSizeOfSetCore()
         {
             int sizeOfSet = base.GetSizeOfSetCore();
+            MenuItem owner = (MenuItem)Owner;
+            ItemsControl parent = ItemsControl.ItemsControlFromItemContainer(owner);
 
             if (sizeOfSet == AutomationProperties.AutomationSizeOfSetDefault)
             {
-                MenuItem owner = (MenuItem)Owner;
-                ItemsControl parent = ItemsControl.ItemsControlFromItemContainer(owner);
-
                 sizeOfSet = ItemAutomationPeer.GetSizeOfSetFromItemsControl(parent, owner);
+            }
+
+            foreach (var item in parent.Items)
+            {
+                if (item is Separator)
+                {
+                    sizeOfSet -= 1;
+                }
             }
 
             return sizeOfSet;
@@ -118,12 +125,24 @@ namespace System.Windows.Automation.Peers
         override protected int GetPositionInSetCore()
         {
             int positionInSet = base.GetPositionInSetCore();
+            MenuItem owner = (MenuItem)Owner;
+            ItemsControl parent = ItemsControl.ItemsControlFromItemContainer(owner);
 
             if (positionInSet == AutomationProperties.AutomationPositionInSetDefault)
             {
-                MenuItem owner = (MenuItem)Owner;
-                ItemsControl parent = ItemsControl.ItemsControlFromItemContainer(owner);
                 positionInSet = ItemAutomationPeer.GetPositionInSetFromItemsControl(parent, owner);
+            }
+
+            foreach (var item in parent.Items)
+            {
+                if (item == owner)
+                {
+                    break;
+                }
+                if (item is Separator)
+                {
+                    positionInSet -= 1;
+                }
             }
 
             return positionInSet;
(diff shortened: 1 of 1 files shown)


weltkante:
> Iterating over the items again seems like the wrong solution here.
> 
> We should be able to influence the original iterations to return the correct values. Maybe supplying a temporary filter during this operation can give us the correct values without adding (much) work? We should look into this further.

The code for finding the position uses public API and eventually `IndexOf` on a user provided collection, you have no chance of extending this, so if you want to provide a filter you have to give up on the current implementation and just iterate over the `IEnumerable`.

If you want to do that I'd suggest adding an overload with filter delegate to `ItemAutomationPeer.GetPositionInSetFromItemsControl` because that is still internal. Leaving the original in place means other callers can still take advantage of `IndexOf` optimizations if they don't need to filter. (You can't take advantage because your filter is iterating unconditionally anyways, so any optimization in IndexOf gets nullified by the filter loop after the IndexOf call.)

rladuca:
> > Iterating over the items again seems like the wrong solution here.
> > We should be able to influence the original iterations to return the correct values. Maybe supplying a temporary filter during this operation can give us the correct values without adding (much) work? We should look into this further.
> 
> This doesn't look like a good idea, you are merging the position from two sources (`base.GetPositionInSetCore` and `ItemAutomationPeer.GetPositionInSetFromItemsControl`) and then use a third set to filter the position (`parent.Items`). Applying a filter to that looks nontrivial.

I am not sure we are thinking of the same thing here.

My plan would be to keep the original calls that were already there and simply use a different filter to ensure they return correctly (saving the original filter callback for restoration).

The flow I had in mind is:
* Set callback (this invalidates caching, sets up everything to create new `Enumerators`.
* Call `ItemAutomationPeer.GetPositionInSetFromItemsControl`
* ItemCollection.IndexOf
* CollectionView.IndexOf
* EnumerableWrapper.IndexOf
  * This is actually an IndexedEnumerable, returning FilteredEnumerator, which should give us the "skips" we need.

This works similarly for `ItemAutomationPeer.GetSizeOfSetFromItemsControl` since the calls to `Count` will forward down to the `FilteredEnumerator`.

This would still have to iterate through items twice (once for a Position and once for Size), but it's only doing the iterations needed.  We might be able to cache these results somehow so one call gets both, but it's probably not worth doing.

I haven't tested this specifically, but it seems like we should be able to use the mechanism.

weltkante:
Oh, I thought you wanted to create new method overloads taking a filter callback. If you are actually wanting to set a filter on a user-visible collection view thats even worse, you are going to break the user UI by invalidating cached instances. For example DataTemplates will be reconstructed, losing transient state in the controls like Focus and scroll positions.

For this to work you would have to build infrastructure which allows you to save caches, change filter callback to perform your operations while supressing events, then restore the old caches.

rladuca:
> Oh, I thought you wanted to create new method overloads taking a filter callback. If you are actually wanting to set a filter on a user-visible collection view thats even worse, you are going to break the user UI by invalidating cached instances. For example DataTemplates will be reconstructed, losing transient state in the controls like Focus and scroll positions.
> 
> For this to work you would have to build infrastructure which allows you to save caches, change filter callback to perform your operations while supressing events, then restore the old caches.

You're right.  After I typed it up, I started getting suspicious of what I was saying but I didn't context switch back and look.

arpitmathur:
To avoid the time cost we're paying here, we discussed an approach to fixing this issue that involved us counting the separators in the loop in GetChildrenCore and storing the number of preceding separators and total separators in each menu item.

We decided not to go with that approach because individual WPF menus generally do not contain a large number of menu items while an app might contain a large number of menus themselves and therefore, a large number of menu items, overall in the app.  If we chose the alternate approach, we would always be paying the memory cost for each menu item as long as the app is open.

Issue #1964, opened 2019-09-26
WindowsBase ref-assembly is missing some type forwards


airbreather:
* .NET Core Version: **3.0.100**
* Windows version: **Windows 10 v1809 (OS Build 17763.737)**
* Does the bug reproduce also in WPF for .NET Framework 4.8?: **No**
* Is this bug related specifically to tooling in Visual Studio (e.g. XAML Designer, Code editing, etc...)? **No**

 
 
 **Problem description:**

Cannot build when referencing very old WPF libraries that look for `System.Collections.ObjectModel.ObservableCollection<T>` in `WindowsBase.dll`

 **Actual behavior:**
![image](https://user-images.githubusercontent.com/3792815/65722212-cad2fb80-e079-11e9-8df1-21a80342fbbb.png)
 
 **Expected behavior:**

Builds fine.
 
 **Minimal repro:**

1. Unzip: [SampleSolution.zip](https://github.com/dotnet/wpf/files/3659810/SampleSolution.zip)
1. Load the solution in Visual Studio 2019 (tested on Enterprise 16.3.1)
1. Try to build it.
1. Note that changing `SampleApp`'s `TargetFramework` to `net472`, it builds just fine.

**My exploration**:

Shot in the dark, browsing the MSIL of `C:\Program Files\dotnet\packs\Microsoft.WindowsDesktop.App.Ref\3.0.0\ref\netcoreapp3.0\WindowsBase.dll`, it's missing these forwards that are present in `C:\Program Files\dotnet\shared\Microsoft.WindowsDesktop.App\3.0.0\WindowsBase.dll`:

```
.class extern forwarder System.Collections.ObjectModel.ObservableCollection`1
{
	.assembly extern System.ObjectModel
}
.class extern forwarder System.Collections.ObjectModel.ReadOnlyObservableCollection`1
{
	.assembly extern System.ObjectModel
}
.class extern forwarder System.Collections.Specialized.INotifyCollectionChanged
{
	.assembly extern System.ObjectModel
}
.class extern forwarder System.Collections.Specialized.NotifyCollectionChangedAction
{
	.assembly extern System.ObjectModel
}
.class extern forwarder System.Collections.Specialized.NotifyCollectionChangedEventArgs
{
	.assembly extern System.ObjectModel
}
.class extern forwarder System.Collections.Specialized.NotifyCollectionChangedEventHandler
{
	.assembly extern System.ObjectModel
}
```

vatsan-madhavan:
[...]

PR #1986, opened 2019-10-01
Fix missing type forwards in reference assemblies

rladuca:
Relates to #1964 

Some reference assemblies (WindowsBase, System.Xaml) were missing type forwards.  This adds them back and rebaselines the ApiCompat (especially against 4.8) to be correct.  This also fixes some issues in the PresentationFramework reference assembly that might have allowed instances of this problem to show up in the future.

Steps taken to find and fix these:

- Find all type forwards deeper in the code tree than OtherAssemblyAttrs.cs
- Ensure they are all hoisted into the corresponding OtherAssemblyAttrs.cs
- Ensure all ref assemblies directly use the lib assemblies OtherAssemblyAttrs.cs so they maintain parity.

Further investigation is needed into why reference API compat failed to find these issues (4.8 API compat did but I incorrectly baselined them).


rladuca:
> Can we (a) port this to 3.1 and (b) port to 3.0 and bring this up as a servicing candidate?

Yep that's my intent.  I'll create PRs for both and get them brought up for Tuesday's meeting.

PR #1987, opened 2019-10-02
Use  Array.Empty

YoshihiroIto:
This change reduces heap allocation.

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/PresentationFramework/Microsoft/Win32/SaveFileDialog.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationFramework/Microsoft/Win32/SaveFileDialog.cs
@@ -75,7 +75,7 @@ namespace Microsoft.Win32
 
             // Extract the first filename from the FileNamesInternal list.
             // We can do this safely because FileNamesInternal never returns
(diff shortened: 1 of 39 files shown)


weltkante:
Is this the only occurence where this can be done? Otherwise it might be worth doing a more structured approach than doing one PR per file when you happen to come across such a change?

YoshihiroIto:
@weltkante 

Thank you for your comment.
What you are saying is correct.
With this unit of PR, a large number of PRs can be created.

OK.
I would like to make the same kind of changes in this PR.

**From**
```
new ***[0]; 
```

**To**
```
Array.Empty<***>();
```

src/Microsoft.DotNet.Wpf/src/PresentationFramework/Microsoft/Win32/SaveFileDialog.cs
```diff
@@ -75,7 +75,7 @@ public Stream OpenFile()
 
             // Extract the first filename from the FileNamesInternal list.
             // We can do this safely because FileNamesInternal never returns
-            // null - if _fileNames is null, FileNamesInternal returns new string[0];
+            // null - if _fileNames is null, FileNamesInternal returns Array.Empty<string>();
```
stevenbrix: thank you for updating this comment :)

Issue #1998, opened 2019-10-03
Visual Studio and MSBuild net4x TFM builds use legacy PresentationBuildTasks.dll when using WindowsDesktop SDK


vatsan-madhavan:
When building WPF projects targeting net4x TFM and using the WindowsDesktop SDK in Visual Studio or MSBuild, legacy `PresentationBuildTasks.dll` (i.e, the one that shipped with .NET Framework, and installed in GAC) is used for Markup Compilation. 

This behavior is wrong - and not what is intended. The correct behavior is to use the copy of PresentationBuildTasks.dll that ships with the .NET Core SDK under `3.0.100\SDKs\Microsoft.NET.Sdk.WindowsDesktop\tools\net472\PresentationBuildTasks.dll`

Repro:

```
# Launch Visual Studio 2019 Developer Command Prompt
dotnet new globaljson --sdk-version 3.0.100
dotnet new wpf --target-framework-override net472
msbuild /bl 
// Inspect the logs; Search for $task MarkupCompilePass1
// Observed: PresentationBuildTasks, Version 4.0.0.0 is used from the .NET Framework GAC  
// Expected: 3.0.100\SDKs\Microsoft.NET.Sdk.WindowsDesktop\tools\net472\PresentationBuildTasks.dll from .NET Core 3.0 SDK should be used
```

nguerrera:
I think the description here should be "When building WPF projects **targeting .NET Framework** using the WindowsDesktop SDK", correct?

vatsan-madhavan:
Yes, that’s it.

vatsan-madhavan:
@nguerrera I was right the first time around. This applies to all TFM's - not just netfx TFM's. I just verified it experimentally. I don't know why I emphasized `net472` TFM in my repro instructions - but that was certainly misleading... the only thing that matters is that `$(MSBuildRuntimeType)==Full`, which in turn ought to act as in index into `net472\PresentationBuildTasks.dll`. 

When building in MSBuild, .NET Framework's copy of WinFX.targets comes into play, which muddies the waters, along with the unexpected 'first one wins' behavior of `UsingTask`.. and so here we are.

nguerrera:
I don't know how that could be because PresentationBuildTasks.dll doesn't work targeting core. Checking.

nguerrera:
My understanding was that the .NETFramework Microsoft.WinFX.targets would not be imported in a Core project.

nguerrera:
```
[...]

PR #1999, opened 2019-10-03
Use correct PresentationBuildTasks.dll for VS and MSBuild builds

vatsan-madhavan:
The PresentationBuildTasks.dll built out of the .NET Core codebase is not being used for msbuild based builds (i.e., when `$(MSBuildRuntimeType)==Full`) of WPF projects that use WindowsDesktop SDK. Instead, the PresentationBuildTasks.dll from GAC, i.e., the DLL that shipped with .NET Framework, is being used for builds instead.

This is because the *first* occurance of an `UsingTask` element that applies to a `TaskName` will always be used - it can not be overridden by subsequent `UsingTask` entries (this is unlike `Property` and `Item` behavior). See note in https://github.com/MicrosoftDocs/visualstudio-docs/blob/master/docs/msbuild/usingtask-element-msbuild.md immediate after the **Syntax** section (Note: The msbuild team added this note after identifying this behavior as part of investiaging this PresentationBuildTasks.dll issue).

The fix ensures that the `UsingTask` declarations supplied by the WindowsDesktop SDK precede those supplied by .NET Framework's copy of `Microsoft.WinFX.targets` by introducing a new `.props` file - `Microsoft.WinFX.props` - and moving a small number of `Property` and `UsingTask` declartions into it. Since `.props` are imported before `targets`, the `UsingTask` declarations supplied by WindowsDesktop SDK will thus take precedence.

Note that `Pbt.props` is used only for building this repo - it doesn't ship. 

Addresses #1998 
This needs to be ported to release/3.1 and release/3.0 (servicing).

### DIFF
--- a/eng/WpfArcadeSdk/tools/Pbt.props
+++ b/eng/WpfArcadeSdk/tools/Pbt.props
@@ -1,5 +1,6 @@
 <Project>
   <PropertyGroup>
+    <LocalMicrosoftWinFXProps>$(RepoRoot)src\Microsoft.DotNet.Wpf\src\PresentationBuildTasks\Microsoft.WinFX.props</LocalMicrosoftWinFXProps>
     <LocalMicrosoftWinFXTargets>$(RepoRoot)src\Microsoft.DotNet.Wpf\src\PresentationBuildTasks\Microsoft.WinFX.targets</LocalMicrosoftWinFXTargets>
   </PropertyGroup>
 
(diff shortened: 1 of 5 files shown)


nguerrera:
I don't know how that happened because the old PBT was breaking hard on System.Runtime as core assembly.

nguerrera:
We had workarounds in place just to be able to use the old PBT

nguerrera:
Oh, when the target framework is net4x!

vatsan-madhavan:
Fixed issue description. Yeah - this is limited to netfx TFM’s.

vatsan-madhavan:
> Fixed issue description. Yeah - this is limited to netfx TFM’s.

Sorry, that's not right. This is **not** limited to netfx TFM's. See https://github.com/dotnet/wpf/issues/1998#issuecomment-538482213.

nguerrera:
I think that it was right. See my comments.

vatsan-madhavan:
> I think that it was right. See my comments.

you're right!

Issue #1966, opened 2019-09-27
OnApplyTemplate not called on ContentControl


mysteryx93:
Version: .NET Core 3
Windows 10 x64
It works in .NET Framework 4.8

I have a WPF user control deriving from ContentControl. It's working fine in .NET Framework 4.8. In .NET Core 3, however, OnApplyTemplate never gets called. (MediaPlayerWpf inherits from ContentControl)

```
public class MpvMediaPlayer : MediaPlayerWpf
{
    public override void OnApplyTemplate()
    {
        base.OnApplyTemplate();
        // This never gets called.
    }
}
```

weltkante:
Works for me in both 3.0 GA (3.0.0-rc1-19456-20) and an older preview 8 I had lying around (3.0.0-preview8-28376-27). I'm using this code and can put a breakpoint into OnApplyTemplate which gets hit correctly. Can you provide a reproduction example?

<details><summary>working code</summary>

```C#
namespace WpfApp1
{
    public partial class MainWindow : Window
    {
        public MainWindow()
        {
            InitializeComponent();
        }
    }

    public class MediaPlayerWpf : ContentControl
    {
    }

    public class MpvMediaPlayer : MediaPlayerWpf
    {
        public override void OnApplyTemplate()
        {
            base.OnApplyTemplate();
            // This never gets called.
        }
    }
}
```

```XAML
<Window x:Class="WpfApp1.MainWindow"
        xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
        xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
        xmlns:local="clr-namespace:WpfApp1"
        Title="MainWindow" Height="450" Width="800">
    <Grid>
        <local:MpvMediaPlayer />
    </Grid>
</Window>
```

</details>

mysteryx93:
Reproduced the problem. The line causing OnApplyTemplate to never trigger is DefaultStyleKeyProperty. Comment that line and it gets triggered

```
public class MpvTestBase : ContentControl {
    static MpvTestBase() {
        DefaultStyleKeyProperty.OverrideMetadata(typeof(MpvTestBase), new FrameworkPropertyMetadata(typeof(MpvTestBase)));
    }

    public override void OnApplyTemplate() {
        base.OnApplyTemplate();
    }
}
```

weltkante:
[...]

PR #2006, opened 2019-10-05
Add `ThemeInfoAttribute` to application project templates

### LINKED ISSUE #1699: Regression in application template, generic.xaml doesn't work out of the box
* .NET Core Version: latest nightly (5.0.0-alpha1.19415.3)

 **Problem description:**
Creating a new WPF application for .NET Core the `generic.xaml` does not work out of the box when added manually (e.g. copy pasted from an existing project), it used to work on Desktop. It does work when using "new item" templates but since adding custom controls is just inheriting from a framework class it is often done without using the "new item" template and instead just writing code.
 
 **Actual behavior:**
Creating a new application with the WPF for .NET Core template does not add a ThemeInfo attribute, I need to add a CustomControl template to get the attribute. This is a regression from Desktop Framework template behavior.

 **Expected behavior:**
The WPF for .NET Core application template should already add the ThemeInfo attribute.
 
 **Minimal repro:**
* create a new WPF application for .NET Core
* manually add code for a Custom Control and generic.xaml

vatsan-madhavan:
Fixes #1699, #1966 

### Description 

Adding `generic.xaml` to a .NET Core application doesn't work out of the box because the default application template doesn't include `ThemeInfoAttribute` the way it used to in .NET Framework.

This can create subtle problems like https://github.com/dotnet/wpf/issues/1966

This change adds back `ThemeInfoAttribute` that was inadvertently left out of the .NET Core application templates.

### Customer Impact

`generic.xaml` is used to supply default styles, often for custom controls etc. For it to function so, the application must have the `ThemeInfoAttribute` set. This was always the case in .NET Framework given the fact that WPF's application template set this attribute by default. 

In .NET core, we inadvertently dropped this attribute from the application project-templates (they are still correctly retained by the .NET Core _User Control_ and _Custom Control_ project templates). 

This can certainly be worked-around by a developer by adding an attribute in a `.cs` files, but identifying the cause of app misbehavior (owing to the missing `ThemeInfo` attribute in itself would be a confusing and difficult process. This is a confusion we would very much like to prevent in the ecosystem. 

### Regression 

Yes, from .NET Framework.

### Risk

Low. (a) Does not impact WPF product (isolated to project templates), and (b) easy to test the locally built project templates. 

Note that the fix to the VB project template in this branch is for completeness' sake only. VB project templates are not enabled in this branch at this time.

### DIFF
--- a/packaging/Microsoft.Dotnet.Wpf.ProjectTemplates/content/WpfApplication-CSharp/AssemblyInfo.cs
+++ b/packaging/Microsoft.Dotnet.Wpf.ProjectTemplates/content/WpfApplication-CSharp/AssemblyInfo.cs
@@ -1,0 +1,10 @@
+using System.Windows;
+
+[assembly:ThemeInfo(
+    ResourceDictionaryLocation.None, //where theme specific resource dictionaries are located
+                                     //(used if a resource is not found in the page,
+                                     // or application resource dictionaries)
+    ResourceDictionaryLocation.SourceAssembly //where the generic resource dictionary is located
+                                              //(used if a resource is not found in the page,
+                                              // app, or any theme specific resource dictionaries)
+)]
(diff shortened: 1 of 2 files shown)


vatsan-madhavan:
Marking as `*NO MERGE*` until shiproom approves merging into release/3.1

wtgodbe:
Approved by shiproom/tactics/Steve, will merge after the CI re-runs

PR #2051, opened 2019-10-15
Update config toolset with guards

rladuca:
https://github.com/dotnet/arcade/pull/3975 changed the include semantics for configure_toolset.ps1.  It now is included from tools.ps1 instead of build.ps1 directly.  This means the properties variable may not be defined in some instances.  We need to guard against these call paths.


rladuca:
Depends on https://github.com/dotnet/arcade/pull/4123

rladuca:
Merging as build passed.  Something is up with the checks here, but this build: https://github.com/dotnet/wpf/runs/264828833 will fix the Arcade PR.

PR #2065, opened 2019-10-18
Adding more guards in configure-toolset.ps1 to fix issues in publish steps

rladuca:


rladuca:
Internal requested build shows this fixes the issue: https://dnceng.visualstudio.com/internal/_build/results?buildId=393501

PR #2076, opened 2019-10-22
[release/3.0] Use correct PresentationBuildTasks.dll for VS and MSBuild builds

### LINKED ISSUE #1998: Visual Studio and MSBuild net4x TFM builds use legacy PresentationBuildTasks.dll when using WindowsDesktop SDK

vatsan-madhavan:
Port of https://github.com/dotnet/wpf/pull/1999
Fixes #1998 

## Description 

The PresentationBuildTasks.dll built out of the .NET Core codebase is not being used for msbuild based builds (i.e., when `$(MSBuildRuntimeType)==Full`) of WPF projects that use WindowsDesktop SDK. Instead, the PresentationBuildTasks.dll from GAC, i.e., the DLL that shipped with .NET Framework, is being used for builds instead.

This is because the *first* occurance of an `UsingTask` element that applies to a `TaskName` will always be used - it can not be overridden by subsequent `UsingTask` entries (this is unlike `Property` and `Item` behavior). See note in https://github.com/MicrosoftDocs/visualstudio-docs/blob/master/docs/msbuild/usingtask-element-msbuild.md immediate after the **Syntax** section (Note: The msbuild team added this note after identifying this behavior as part of investiaging this PresentationBuildTasks.dll issue).

This fix ensures that the `UsingTask` declarations supplied by the WindowsDesktop SDK precede those supplied by .NET Framework's copy of `Microsoft.WinFX.targets` by introducing a new `.props` file - `Microsoft.WinFX.props` - and moving a small number of `Property` and `UsingTask` declartions into it. Since `.props` are imported before `targets`, the `UsingTask` declarations supplied by WindowsDesktop SDK will thus take precedence.

Note that `Pbt.props` is used only for building this repo - it doesn't ship.

## Customer Impact 

Multi-targeted builds targeting .NET Framework will be produced incorrectly under some conditions, specifically when building using VS. There may be no problems initially due to the strong backwards compatibility between .NET Core and .NET Framework WPF assemblies - but errors in XAML/markup compilation may lead to hard-to-identify/debug problems. 

## Regression 

Not a regression. This was a missed corner case not caught until now. 

## Risk 

- The changes are scoped and  safe. 
- They have been tested by building a corpus of WPF samples 
- The fix has been in .NET 5 for 2 weeks now without any adverse reports.

### DIFF
--- a/eng/WpfArcadeSdk/tools/Pbt.props
+++ b/eng/WpfArcadeSdk/tools/Pbt.props
@@ -1,5 +1,6 @@
 <Project>
   <PropertyGroup>
+    <LocalMicrosoftWinFXProps>$(RepoRoot)src\Microsoft.DotNet.Wpf\src\PresentationBuildTasks\Microsoft.WinFX.props</LocalMicrosoftWinFXProps>
     <LocalMicrosoftWinFXTargets>$(RepoRoot)src\Microsoft.DotNet.Wpf\src\PresentationBuildTasks\Microsoft.WinFX.targets</LocalMicrosoftWinFXTargets>
   </PropertyGroup>
 
@@ -34,6 +35,8 @@
             Project="../targets/Microsoft.NET.Sdk.WindowsDesktop.props"
             Condition="!Exists('$(LocalMicrosoftWinFXTargets)') And '$(InternalMarkupCompilation)'=='true'"/>
             
+    We need Microsoft.WinFx.props though, so that's imported here specifically. 
+            
     The use of Microsoft.NET.Sdk.WindowsDesktop doesn't break the source-build promise. 
       - Microsoft.NET.Sdk.WindowsDesktop is built from sources
       - It is used for compilation of assemblies in dotnet-wpf-int (the internal WPF repo), and serves as a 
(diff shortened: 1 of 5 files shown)


rladuca:
@vatsan-madhavan looks like a NuGet feed is down, re-queuing in case it was intermittent.

vatsan-madhavan:
Not fixing this in 3.0 - only fixing this in 3.1

PR #2098, opened 2019-10-23
[release/3.1] Context Menus are sometimes not shown in High-DPI applications

vatsan-madhavan:
Addresses #2088 
.NET 5 PR: #2097 

## Description (Summary)

As part of a previous fix (that shipped originally as part of .NET 4.8 in), a change was made to `Popup` that involved the destruction and recreation of the underlying `HWND`. This was done to ensure that the `HWND` was always created with the correct monitor (~=DPI) affinity.

This previous fix depended upon a private helper-method (`Popup.DestroyWindow`) that had side-effects that were not accounted for in the original fix. One of the side effects is that the `ContextMenu` is not sown consistently. 

The solution modifies (refactors) the private method to extract the useful portion and uses it to improve the previous fix. 

## Customer Impact

This is a forward-port from from .NET 4.8. This was reported by a customer, and has also been discovered by Visual Studio internally. 

## Regression 

Regression introduced by .NET 4.8. .NET Core 3.0 shipped with this bug. 

## Risk 

Low - this has been well tested internally and validated by multiple customers. The fix is well understood and small/scoped. 

## Details 
As part of a previous fix (that shipped originally as part of .NET 4.8 in), a change was made to `Popup` that involved the destruction and recreation of the underlying `HWND`. This was done to ensure that the `HWND` was always created with the correct monitor (~=DPI) affinity.

In order to acheive this, `DestroyWindow()` and `BuildWindow()` - private methods in `Popup` - were used.

We are finding that `DestroyWindow()` has side-effects that can lead to incorrect behavior. The incorrect beahvior works as follows:

  - The call into `Popup.CreateWindow` is usually a consequence `Popup.IsOpenChanged` (`false -> true`)
  - Within `Popup.CreateWindow`, `DestroyWindow()` is called (when high-dpi mode is detected).
  - `DestroyWindow()` destroys the underlying `HWND`, releases the capture, raises the *OnClosed* event and clears the placement-target.
    - At the end of `DestroyWindow()`, we get `IsOpen == false`.
  - After `DestroyWindow()`, we call into `BuildWindow()` and `CreateNewPopupRoot()` etc., which go on to build the `Popup` again (and also instnatiate a new `HWND`).
    - Unfortunately, there is no mechanism here for resetting `IsOpen` back to `true` (without also leading to an undesirable infinite-recursion that calls back into `CreateWindow`).

If these calls to show the context-menu arise from `ContextMenu.OnIsOpenChanged`, and flow through `ContextMenu.HookupParentPopup -> Popup.CreateRootPopup`, then `IsOpen` gets reset (there is a direct call into `SetBinding(IsOpenProperty)`, in `Popup.CreateRootPopupInternal`) and the popup is shown correctly. Until then, the context-menu is "stuck" not being able to be shown.

The solution is to stop using `DestroyWindow()` as-is, which does more than what we need for it to accomplish. Our original intent in calling `DestroyWindow()` was simply destroy and recreate the `HWND`. This fix refactors `DestroyWindow()` to suit this need and uses the newly introduced `DestroyWindowImpl()` to destroy the `HWND`, and then recreate just that. The rest of the state is retained intact, and the `Popup/ContextMenu` continues to function well as before.

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Controls/Primitives/Popup.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Controls/Primitives/Popup.cs
@@ -1510,7 +1510,7 @@ namespace System.Windows.Controls.Primitives
             // cascading failure.
             if (PopupInitialPlacementHelper.IsPerMonitorDpiScalingActive)
             {
-                DestroyWindow();
+                DestroyWindowImpl();
                 _positionInfo = null;
                 makeNewWindow = true;
             }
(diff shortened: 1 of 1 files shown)


fabriceleal:
Is there any way to mitigate this until the next version of .NET Framework?

Issue #2089, opened 2019-10-22
NULL HWND's attempted to be parented under SystemResources listener window cause unhanded exception


vatsan-madhavan:
When an `HwndHost `receives `SourceChanged `event, it goes through `BuildOrReparentWindow`. When the hosted window is invisible, it is usually reparented under a temporary windows maintained by WPF in the `SystemResources `class, until later on the window can be rebuilt and parented back to a valid parent. 

There is a latent bug in this logic where in `NULL ` `HWND's `are attempted to be parented to `SystemResources `managed temporary windows. This bug goes back quite a while (.NET 4.5 likely). WPF seems to ignore the return value from `kernel32!SetParent` and not deal with this failure. This has not been a crashing failure until now. 

Starting .NET 4.8, there have been some changes to this codepath that have resulted in this bug becoming a crashing bug. In addition to calling `kernel32!SetParent `on a `NULL HWND`, WPF attempts to perform several other operations on the `HWND`, including querying its DPI characteristics (which returns invalid information). Once of the subsidiary steps returns an invalid value that in turn results in a crash with the following stack:

```
System.Collections.Generic.KeyNotFoundException: 'The given key was not present in the dictionary.'
mscorlib.dll!System.Collections.Generic.Dictionary<System.__Canon, System.__Canon>.this[System.__Canon].get(System.__Canon key) Line 183	C#
 	PresentationFramework.dll!System.Windows.SystemResources.GetDpiAwarenessCompatibleNotificationWindow(System.Runtime.InteropServices.HandleRef hwnd) Line 1731	C#
 	PresentationFramework.dll!System.Windows.Interop.HwndHost.BuildOrReparentWindow() Line 1196	C#
 	PresentationFramework.dll!System.Windows.Interop.HwndHost.OnSourceChanged(object sender, System.Windows.SourceChangedEventArgs e) Line 1077	C#
 	PresentationCore.dll!System.Windows.SourceChangedEventArgs.InvokeEventHandler(System.Delegate genericHandler, object genericTarget) Line 142	C#
 	PresentationCore.dll!System.Windows.RoutedEventArgs.InvokeHandler(System.Delegate handler, object target) Line 341	C#
 	PresentationCore.dll!System.Windows.RoutedEventHandlerInfo.InvokeHandler(object target, System.Windows.RoutedEventArgs routedEventArgs) Line 81	C#
[...]

PR #2102, opened 2019-10-23
[release/3.0] Prevent NULL HWND's from being parented under SystemResources listener windows

vatsan-madhavan:
Addresses #2089 
.NET 5 PR: #2100 
.NET Core 3.1 PR: #2101 

## Description (Summary)

When `HwndHost` hosted `HWND`'s need to be "parked" while the `HwndHost` is not being actively shown, WPF will reparent such `HWND`'s under temporary message-only windows maintained inside the `SystemResources` class. 

There is a bug that causes null `HWND`'s to be parented under such message-only windows. When this happens in High-DPI applications that use mixed-mode DPI capabilities introduced by WPF in .NET 4.8 (parent and child windows with different `DPI_AWARENESS_CONTEXT `values),  a crash ensues. 

## Customer Impact 

This is a fix for a crash affecting several applications including Visual Studio, and Azure Information Protection Add-in for Office 

This was fixed recently in .NET 4.8, and is being forwarded ported to .NET Core for consistency. 

## Regresssion 

Not a regression in .NET Core, but this was a regression introduced by .NET 4.8. 

## Risk 

The fix is small and well understood, and has been tested well. The .NET Framework version of this fix has been validated by Visual Studio (the codebases are identical in this area and .NET Framework testing is a reliable proxy for this change in .NET Core). 

## Details 

When an `HwndHost `receives `SourceChanged `event, it goes through `BuildOrReparentWindow`. When the hosted window is invisible, it is usually reparented under a temporary windows maintained by WPF in the `SystemResources `class, until later on the window can be rebuilt and parented back to a valid parent.

There is a latent bug in this logic where in `NULL ` `HWND's `are attempted to be parented to `SystemResources `managed temporary windows. This bug goes back quite a while (.NET 4.5 likely). WPF seems to ignore the return value from `kernel32!SetParent` and not deal with this failure. This has not been a crashing failure until now.

Starting .NET 4.8, there have been some changes to this codepath that has resulted in the current bug becoming a crash. In addition to calling `kernel32!SetParent` on a `NULL` `HWND`, WPF attempts to obtain a DPI-specific parking-window. This process of querying a DPI-specific parking window fails because WPF is unable to use the `DPI_AWARENESS_CONTEXT` value returned by the system for `(HWND)nullptr`.

The only necessary part of this fix is in `HwndHost`: WPF should not attempt to reparent the hosted window under a parking-window if the hosted window is `(HWND)nullptr`. This only requires a simple check : `else if (_hwnd.Handle != IntPtr.Zero)`). All other changes in `SystemResources` and `HwndHost` are defensive improvements.

`SystemResources.EnsureResourceChangeListener(HwndDpiInfo)` can attempt to create a parking-window corresponding to `DPI_AWARENESS_CONTEXT_VALUE` that is invalid/meaningless. This should not be allowed. A few additional checks are added to ensure this. Further, `GetDpiAwarenessCompatibleNotificationWindow` is augmented to be more defensive.

Also, variant of `EnsureResourceChangeListener`  is dead code - it is being removed.

If for some unknown reason `SystemResources.GetDpiAwarenessCompatibleNotificationWindow`  fails and returns `null` to `HwndHost.BuildOrReparentWindow`, WPF will fail to reparent the hosted window, and it will be 'lost'. This seems very unlikely - I have added a Trace to ensure that we can debug this situation if it does occur.

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Interop/HwndHost.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Interop/HwndHost.cs
@@ -976,23 +976,28 @@ namespace System.Windows.Interop
                         UnsafeNativeMethods.SetParent(_hwnd, new HandleRef(null,hwndParent));
                     }
                 }
-                else
+                else if (_hwnd.Handle != IntPtr.Zero)
                 {
                     // Reparent the window to notification-only window provided by SystemResources
                     // This keeps the child window around, but it is not visible.  We can reparent the 
                     // window later when a new parent is available
                     var hwnd = SystemResources.GetDpiAwarenessCompatibleNotificationWindow(_hwnd);
-                    UnsafeNativeMethods.SetParent(_hwnd, new HandleRef(null, hwnd.Handle));
-                    // ...But we have a potential problem: If the SystemResources listener window gets 
-                    // destroyed ahead of the call to HwndHost.OnDispatcherShutdown(), the HwndHost's window
-                    // will be destroyed too, before the "logical" Dispose has had a chance to do proper
-                    // shutdown. This turns out to be very significant for WebBrowser/ActiveXHost, which shuts
-                    // down the hosted control through the COM interfaces, and the control destroys its
-                    // window internally. Evidently, the WebOC fails to do full, proper cleanup if its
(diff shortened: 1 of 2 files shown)


vatsan-madhavan:
We are not forward porting .NET Framework fixes to release/3.0 - this will go into release/3.1 only.

PR #1761, opened 2019-09-03
Hit testing by geometry fails if more than one matrix trasformation is applied to a geometry.

Simon-IT:
This is related to my issue [642](https://github.com/dotnet/wpf/issues/642).  In the hit testing by geometry algotithm, the order in which the transformations are crosses is wrong. In the issue [642](https://github.com/dotnet/wpf/issues/642) there is a project to test in very simple manner the problem.

Issue #642, opened 2019-05-03
Hit test fails if mouse is over a text object


vsfeedback:
<p>Hello Microsoft Support,</p><p>I've drawn a text by a DrawingVisual object, but the hit test fails when the mouse pointer is on it. For hit testing I'm using a GeometryHitTestParameters and EllipseGeometry objects. I've notice that if I use a PointHitTestParameters object, so all works well.</p><p>I've attached a complete solution that show the problem. Let me point out, that this issue is very serious for us because it doesn't allows to go on in our work.<br>Thank you in advance.<br>Best regards,<br><br>Simone</p>

_This issue has been moved from https://developercommunity.visualstudio.com/content/problem/533445/hit-test-fails-if-mouse-is-over-a-text-object.html
VSTS ticketId: 847713_
_These are the original issue comments:_

Visual Studio Feedback System on 4/15/2019, 07:36 PM (17 days ago): <p>We have directed your feedback to the appropriate engineering team for further evaluation. The team will review the feedback and notify you about the next steps.</p>

Jeff Schwartz [MSFT] on 4/16/2019, 04:20 PM (16 days ago): <p>Thanks for sharing feedback <a rel="user" href="https://developercommunity.visualstudio.com/users/86125/82408540-96cb-425a-a759-0107c4e509f6.html" nodeid="86125">@Simone Padovan</a>.  Can you elaborate more on what is not happening that you would expect?</p><p>In the attached sample I am seeing the proper events fire and callbacks fire (by setting break points and mousing over the text block).  Thanks</p>

Jeff Schwartz [MSFT] on 4/17/2019, 02:49 PM (15 days ago): <p>Thanks for the details.  I am going to route to the WPF team.</p>

_These are the original issue solutions:_

Simone Padovan solved on 4/17/2019, 10:14 AM (16 days ago), 0 votes: 
<p>Hello Jeff,</p>
[...]

PR #2103, opened 2019-10-23
[Release/3.1] Hit testing by geometry fails if more than one matrix trasformation is applied to a geometry.

rladuca:
On behalf of @Simon-IT :

This is related to my issue 642. In the hit testing by geometry algotithm, the order in which the transformations are crosses is wrong. In the issue 642 there is a project to test in very simple manner the problem.

3.1 port related to https://github.com/dotnet/wpf/pull/1761

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Media/HitTestWithGeometryDrawingContextWalker.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Media/HitTestWithGeometryDrawingContextWalker.cs
@@ -276,7 +276,7 @@ namespace System.Windows.Media
             else if ((_currentTransform != null) && !_currentTransform.IsIdentity)
             {
                 // Both the current transform and the new one are nontrivial, combine them
-                Matrix combined = _currentTransform.Value * transform.Value;
+                Matrix combined =  transform.Value * _currentTransform.Value;
                 transform = new MatrixTransform(combined);
             }
 


SamBent:
Related to #642

5.0 PR: #1761

### Description
When using a DrawingVisual to render lightweight graphics, pushing two transforms causes hit-testing to fail, if the two transforms don't commute (e.g. a scale and a translate).

### Customer Impact
User clicks (or moves mouse) over a drawing created as above, but doesn't get the expected response.   Or user moves/clicks mouse and gets the response for a different part of the drawing.

### Regression
No.  This bug has been lurking since WPF's initial release (.NET 3.0, 2006).

### Risk - Medium
The fix is in HitTestWithGeometryDrawingContextWalker.PushTransform, and affects all subsequent calls to that class.  This is an internal class, but the method is an override of a public base class (DrawingContext).  The normal usage is internal:  when a hit-test result is requested, internal code creates an instance of the class, asks it to walk the drawing instructions, and extracts the result.  This usage is low-risk.  The only difference visible to the public is that hit-test queries now return the correct result.  It's difficult to imagine an app relying on the incorrect result;  to do so would require intimate knowledge of the transforms being applied, essentially re-implementing the entire hit-testing pipeline.

A greater risk is an app that never noticed the hit-test results were wrong.  Now the app will react when it didn't before.  User's clicks and moves may incite behavior that the app never tested.  An app that goes to the trouble of declaring behavior related to this kind of hit-testing should have noticed that it wasn't working, but oversights like this sometimes happen.

There's also a risk with an app that interacts directly with the internal class, using the public methods of the public base class DrawingContext.  The app could easily see different behavior.  However this presumes that the app can obtain a reference to the internal instance.  I don't see any obvious way for that to happen, but I'm not 100% sure it can't.

### Workarounds
Hit-testing works correctly if the app combines the transforms correctly in the original drawing instructions, replacing two calls of PushTransform with one.   But this is awkward to do when the app builds the drawing from separate components, each of which may contribute its own transform.  This is the customer's situation.

PR #2138, opened 2019-10-30
Increase .NET version to ensure we are using higher versions of NuGet

rladuca:
We've been seeing build failures due to a recent update from Arcade (https://github.com/dotnet/arcade/issues/4235) because we are not using a high enough version of .NET in our builds.

Therefore, bumping the version to ensure the publishing step is not using incorrect versions of NuGet.

### DIFF
--- a/eng/WpfArcadeSdk/Sdk/Sdk.props
+++ b/eng/WpfArcadeSdk/Sdk/Sdk.props
@@ -61,8 +61,10 @@
     <!--
       Temporarily suppress NU3027
       https://github.com/dotnet/arcade/issues/2304
+      
+      NU5131, NU5128 - Our transport packages do not conform to the layout NuGet wants.
     -->
-    <NoWarn>$(NoWarn);NU3027</NoWarn>
+    <NoWarn>$(NoWarn);NU3027;NU5131;NU5128</NoWarn>
     
     <!-- 
         Suppress NU1605 (Package downgrade warnings) when building inside Visual Studio
--- a/global.json
+++ b/global.json
@@ -1,6 +1,6 @@
 {
   "tools": {
-    "dotnet": "5.0.100-alpha1-014065",
+    "dotnet": "5.0.100-alpha1-015515",
     "runtimes": {
       "dotnet": [
         "2.1.7",


rladuca:
@vatsan-madhavan FYI, we had to suppress NU5128 and NU5131 as our transport packages do not conform.

rladuca:
Internal PR: https://dnceng.visualstudio.com/internal/_git/dotnet-wpf-int/pullrequest/4052?_a=overview

eng/WpfArcadeSdk/Sdk/Sdk.props
```diff
@@ -61,8 +61,10 @@
     <!--
       Temporarily suppress NU3027
       https://github.com/dotnet/arcade/issues/2304
+      
+      NU5131, NU5128 - Our transport packages do not conform to the layout NuGet wants.
```
vatsan-madhavan: In what way are we nonconformant ?

vatsan-madhavan: Can we fix the transport package layout?

vatsan-madhavan: Are you sure we are not hitting https://github.com/NuGet/Home/issues/8467 or similar?

rladuca: We just directly add files to the package.  I believe we'd need to create the NuSpec in such a fashion that we have a reference group directly calling out all the reference assemblies.  That way NuGet will not expect a corresponding reference assembly for every lib assembly.

So maybe layout is not the correct term to use, but it's more about the NuSpec not satisfying the guidance (see https://docs.microsoft.com/en-gb/nuget/create-packages/select-assemblies-referenced-by-projects).

rladuca: I think we're running afoul of this directly (and not hitting some bug) since we're definitely not doing what is suggested for the NuSpec and we're also getting actual binary names in the warnings.  That's contrary to https://github.com/NuGet/Home/issues/8467.

rladuca:
Merging this to address the immediate issue, I'll file an issue to investigate the NuSpec problems.

Issue #381, opened 2019-02-24
NRE in System.Windows.Controls.PopupControlService.RaiseToolTipClosingEvent(bool reset)


clairernovotny:
Windows 10 17134

WPF version: `4.8.27325.3 @BuiltBy: dlab14-DDVSOWINAGE058 @Branch: master @Commit: 24b4ae9332545f7677a8ce2f8916cc2284e31b75`

This resulted in an app crash as it was unhandled

```
System.NullReferenceException: Object reference not set to an instance of an object.
   at System.Windows.Controls.PopupControlService.RaiseToolTipClosingEvent(bool reset)
   at System.Windows.Controls.PopupControlService.InspectElementForToolTip(DependencyObject o, ToolTipTrigger triggerAction)
   at System.Windows.Controls.PopupControlService.OnFocusChanged(object sender, KeyboardFocusChangedEventArgs e)
   at System.Windows.Input.KeyboardNavigation+<>c__DisplayClass22_0.<NotifyFocusChanged>b__0(object item)
   at System.Windows.Input.KeyboardNavigation+WeakReferenceList.Process(Func<T, TResult> action)
   at System.Windows.Input.KeyboardNavigation.NotifyFocusChanged(object sender, KeyboardFocusChangedEventArgs e)
   at System.Windows.FrameworkElement.OnGotKeyboardFocus(object sender, KeyboardFocusChangedEventArgs e)
   at System.Windows.Input.KeyboardFocusChangedEventArgs.InvokeEventHandler(Delegate genericHandler, object genericTarget)
   at System.Windows.RoutedEventArgs.InvokeHandler(Delegate handler, object target)
   at System.Windows.RoutedEventHandlerInfo.InvokeHandler(object target, RoutedEventArgs routedEventArgs)
   at System.Windows.EventRoute.InvokeHandlersImpl(object source, RoutedEventArgs args, bool reRaised)
   at System.Windows.UIElement.RaiseEventImpl(DependencyObject sender, RoutedEventArgs args)
   at System.Windows.UIElement.RaiseTrustedEvent(RoutedEventArgs args)
   at System.Windows.Input.InputManager.ProcessStagingArea()
   at System.Windows.Input.InputManager.ProcessInput(InputEventArgs input)
   at System.Windows.Input.KeyboardDevice.ChangeFocus(DependencyObject focus, int timestamp)
[...]

PR #2167, opened 2019-11-08
Fix NRE when tooltip closing(#381)

YoshihiroIto:
Reproduction
https://github.com/dotnet/wpf/issues/381#issuecomment-547206598

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Controls/PopupControlService.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Controls/PopupControlService.cs
@@ -578,8 +578,12 @@ namespace System.Windows.Controls
                             if (_ownToolTip)
                                 BindingOperations.ClearBinding(_currentToolTip, ToolTip.ContentProperty);
                         }
-                        _currentToolTip.FromKeyboard = false;
-                        _currentToolTip = null;
+
+                        if (_currentToolTip != null)
+                        {
+                            _currentToolTip.FromKeyboard = false;
+                            _currentToolTip = null;
+                        }
                     }
                 }
             }


weltkante:
I'm not sure this is the right fix, this might just supressing the symptoms and leave the bug unfixed.

`RaiseToolTipOpeningEvent` seems to be the only event where `_currentToolTip` is set to null, and it does already have a `if (_currentToolTip != null)` guard further up the method. So the only reasonable explanation is that you are triggering reentrancy in `RaiseToolTipOpeningEvent` and execute it twice as a side effect of something done in that method. Maybe `_currentToolTip.ClearValue(OwnerProperty);` can trigger reentrancy?

I'd like to get the problem getting looked at closer before suppressing it with a seemlingly redundant check. If the method is reentrant there might be more that needs to be done to make sure the logic executes correctly. For example you might want to make sure not to call `element.RaiseEvent(new ToolTipEventArgs(false));` twice and that `BindingOperations.ClearBinding(_currentToolTip, ToolTip.ContentProperty);` correctly clears the content property and doesn't just no-op because its null.

In either case it definitely needs adding a comment (preferrably after debugging and understanding whats happening here), otherwise you are at risk that someone removes the seemingly redundant `if (_currentToolTip != null)` you are adding.

YoshihiroIto:
@weltkante 

Thank you.

As you said, this method is re-entered during processing. I recognized this and created this pull request.
This is because the original code makes it clear.
Here.
https://github.com/dotnet/wpf/blob/28a25ccaac84b36007a5ae7e44bf246945226ced/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Controls/PopupControlService.cs#L556

Your suggestion will be a re-implementation of this method, so I want someone to do it in another pull request. I'm doing it if I can, but I'm not yet familiar with the WPF code.

This PR is intended to correct a small part of WPF that has been stable for a long time. There is only one revision.

I agree that comments are needed. I will add a comment.

As you can see, I am not good at English. I will give you a little time.

weltkante:
The problem with your fix is that it still will raise the event twice, so while it may not crash anymore you haven't fixed the problem of `RaiseToolTipOpeningEvent` not being able to deal with reentrancy correctly. We can wait for someone from the WPF team to decide the direction to take with the fix, I'm just commenting.

> Your suggestion will be a re-implementation of this method, so I want someone to do it in another pull request.

I don't think a reimplementation will be necessary, but I didn't debug it so I can't tell for sure. The problem with doing it in another PR is that if there is too much time between fixing the crash people may start accidentally depending on the broken event order and then it'll be too late to fix the event. The WPF project isn't exactly seeing a lot of action so I want to make sure this is fixed correctly while its impossible for people to depend on the reentrant event ordering.

jbhensley:
Stripping out all but the relevant parts of that `finally` block that starts on 547, you have this:

```C#
finally
{
	if (isOpen)
	{
		_currentToolTip.IsOpen = false;

		// Setting IsOpen makes call outs to app code. So it is possible that 
		// the _currentToolTip is deleted as a result of an action there. If that 
		// were the case we do not need to set off the timer to close the tooltip.
		if (_currentToolTip != null)
		{
			...
		}

		...
	}
	else
	{
		...
	}
	_currentToolTip.FromKeyboard = false;
	_currentToolTip = null;
}
```

So it's pretty clear that `_currentToolTip.IsOpen = false` can cause side-effects that lead to  `_currentToolTip == null`. Once we leave the if/else `_currentToolTip` is accessed again without being checked.

Suggested fix of

```C#
if (_currentToolTip != null)
{
	_currentToolTip.FromKeyboard = false;
	_currentToolTip = null;
}
```

would appear to be right.

singhashish-wpf:
Thanks @YoshihiroIto for the PR, However the fix is already applied for this issue in latest releases.
https://github.com/dotnet/wpf/blob/d4797f1ec6de9347c3725150b5dcacc307859749/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Controls/PopupControlService.cs#L623

PR #2191, opened 2019-11-14
Update PreReleaseVersionLabel to RTM

wtgodbe:
Non-shipping packages should be labeled RTM for GA. We need to take this fix for 3.1 GA


vatsan-madhavan:
@msftbot hold this pr for 1 minutes

Issue #2201, opened 2019-11-18
Add internal/* to PR trigger list


vatsan-madhavan:
Add internal/* branches to PR trigger list in yml file

PR #2204, opened 2019-11-18
add internal/release/3.* to PR trigger list

vatsan-madhavan:
Addresses #2201

Adds internal/3.* to PR trigger list

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -50,6 +50,7 @@ pr:
     - master
     - release/3.0
     - release/3.0-preview9 
+    - internal/release/3.*
     - release/3.1
   paths:
     exclude:


azure-pipelines.yml
```diff
@@ -50,6 +50,7 @@ pr:
     - master
     - release/3.0
     - release/3.0-preview9 
+    - internal/release/3.*
```
RussKie: Why do you need it on the master branch?
And why not have `- release/*`?

vatsan-madhavan: In master, it should have been `release/*` and not `release/3.*`. I'll fix this up again when I merge https://github.com/dotnet/wpf/pull/2205. 

`internal/*` branches exist for non-public development, for e.g., security fixes. They are not intended to be part of the GitHub repo - they exist to be used from the the internal AzDo mirror repo.

RussKie: > `internal/*` branches exist for non-public development, for e.g., security fixes.

My question was more about why you added `- internal/release/3.*` to the master branch. It would make sense to have it as `- internal/release/*`

PR #2237, opened 2019-11-25
Fix typo

filipnavara:


vatsan-madhavan:
thanks for the fixup @filipnavara!

PR #2270, opened 2019-12-02
SharedSizeGroup loop

### LINKED ISSUE #2222: Hang during layout when using SharedSizeGroup
SharedSizeGroup does redundant layout - requesting remeasure of Grids that don't need it.  This can lead to a hang in which the app is unresponsive and has to be killed from TaskManager, but isn't marked as "(Not Responding)".  This can happen if
  a. the Grids are within the scope of a ScrollViewer with VerticalScrollbarVisibility="Auto"
  b. there are more than 157 Grids (the threshold for coalescing pending measure tasks into one task at the root)
  c. the minimum width of a non-long-pole Grid increases, away from the shared column 

[Port of servicing fix for .NET 4.5-4.8]

SamBent:
Ask Mode Template:  

Description 

After every layout pass, a SharedSizeGroup re-evaluates its  shared size and updates the state of its constituent ColumnDefinitions.  The  goal is to mark the "long-pole" definitions (the largest ones in the group), and  to identify Grids that need to be remeasured.  The logic for this is flawed in  several ways;  both goals can fail in either direction (false positive or false  negative).   These flaws usually result in redundant layout work - the Grids get  measured twice, with the mistakes in the second pass cancelling those in in the  first - but correct layout results.  But in some cases they lead to an infinite  cycle of re-layout;  the app hangs.   In the customer's scenario this happens  when  
a. the Grids are within the scope of a ScrollViewer with  VerticalScrollbarVisibility="Auto"  
b. there are more than 157 Grids (the  threshold for coalescing pending measure tasks into one task at the root)   
c. the minimum width of a non-long-pole Grid increases, away from the shared column

Customer Impact 

Hang. 

Regression 

No.   (Port of servicing fix for .NET 4.5.2 - 4.8)

Risk 

Low 

Fixes #2222
Hang during layout when using SharedSizeGroup

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Controls/DefinitionBase.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Controls/DefinitionBase.cs
@@ -83,7 +83,7 @@ namespace System.Windows.Controls
         {
             if (_sharedState == null)
             {
-                //  start with getting SharedSizeGroup value. 
+                //  start with getting SharedSizeGroup value.
                 //  this property is NOT inhereted which should result in better overall perf.
                 string sharedSizeGroupId = SharedSizeGroup;
                 if (sharedSizeGroupId != null)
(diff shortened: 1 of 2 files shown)


vatsan-madhavan:
Marking as NO-MERGE - we will merge this sometime in Jan '20.

vivmishra:
Moved to Mar as per Tactics. Will need to be explicitly approved for Feb, if required.

PR #2288, opened 2019-12-06
Handle incremental build when XAML files are removed from the Page glob

### LINKED ISSUE #2287: Handle incremental build when XAML files are removed from the Page glob
Related to https://github.com/microsoft/msbuild/issues/4963

WPF needs to update its handling of incremental building in PBT (likely in [IncrementalCompileAnalyzer](https://github.com/dotnet/wpf/blob/f4e04b7a34f8c490b7d351b44d2825293462e8a5/src/Microsoft.DotNet.Wpf/src/PresentationBuildTasks/MS/Internal/Tasks/IncrementalCompileAnalyzer.cs)) so that it detects invalid items in the cache and ignores them during markup compilation.

Otherwise, we potentially attempt to compile files that are no longer present on disk and the build will fail.

MSBuild has no mechanism to do this automatically as of yet (see https://github.com/microsoft/msbuild/issues/2470), so this must be handled at the WPF level.

rladuca:
Ensure that we don't include XAML files that no longer exist on disk as part of the LocalMarkupPages when we load from the compiler cache.

This allows SDK style projects to react to files deleted from the glob between two incremental builds.

Fixes #2287

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/PresentationBuildTasks/MS/Internal/Tasks/IncrementalCompileAnalyzer.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationBuildTasks/MS/Internal/Tasks/IncrementalCompileAnalyzer.cs
@@ -389,13 +389,22 @@ namespace MS.Internal.Tasks
                 {
                     numLocalTypeXamls = CompilerLocalReference.LocalMarkupPages.Length;
 
+                    // Under incremental builds of SDK projects, we can have a state where the cache contains some XAML files but the Page blob
+                    // no longer contains them.  To avoid attempting to recompile a file that no longer exists, ensure that any cached XAML file
+                    // still exists in the Page blob prior to queuing it up for recompilation.
+                    HashSet<string> localMarkupPages = new HashSet<string>(_mcPass1.PageMarkup.Select(x => x.GetMetadata("FullPath")));
+
                     for (int i = 0; i < CompilerLocalReference.LocalMarkupPages.Length; i++)
                     {
                         LocalReferenceFile localRefFile = CompilerLocalReference.LocalMarkupPages[i];
-                        recompiledXaml.Add(new FileUnit(
-                                                localRefFile.FilePath, 
-                                                localRefFile.LinkAlias, 
-                                                localRefFile.LogicalName));
+
+                        if (localMarkupPages.Contains(localRefFile.FilePath))
+                        {
+                            recompiledXaml.Add(new FileUnit(
+                                                    localRefFile.FilePath,
+                                                    localRefFile.LinkAlias,
+                                                    localRefFile.LogicalName));
+                        }
                     }
                 }
 
--- a/src/Microsoft.DotNet.Wpf/src/PresentationBuildTasks/MS/Internal/Tasks/IncrementalCompileAnalyzer.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationBuildTasks/MS/Internal/Tasks/IncrementalCompileAnalyzer.cs
@@ -393,7 +393,7 @@ namespace MS.Internal.Tasks
                     // Under incremental builds of SDK projects, we can have a state where the cache contains some XAML files but the Page blob
                     // no longer contains them.  To avoid attempting to recompile a file that no longer exists, ensure that any cached XAML file
                     // still exists in the Page blob prior to queuing it up for recompilation.
-                    HashSet<string> localMarkupPages = new HashSet<string>(_mcPass1.PageMarkup.Select(x => x.GetMetadata(SharedStrings.FullPath)));
+                    HashSet<string> localMarkupPages = new HashSet<string>(_mcPass1.PageMarkup.Select(x => x.GetMetadata(SharedStrings.FullPath)), StringComparer.OrdinalIgnoreCase);
 
                     for (int i = 0; i < CompilerLocalReference.LocalMarkupPages.Length; i++)
                     {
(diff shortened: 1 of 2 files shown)


src/Microsoft.DotNet.Wpf/src/PresentationBuildTasks/MS/Internal/Tasks/CompilerLocalReference.cs
```diff
@@ -1,347 +1,352 @@
...
+
+                    if (lrf != null)
+                    {
+                        alMarkupPages.Add(lrf);
+                    }
+                }
+
+                if (alMarkupPages.Count > 0)
+                {
+                    // With SDK style projects, a file may be removed from the Page glob (via deleting from disk).
```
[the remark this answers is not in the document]
rladuca: Yep, that seems to work just fine, the added UserControl is picked up.

src/Microsoft.DotNet.Wpf/src/PresentationBuildTasks/MS/Internal/Tasks/CompilerLocalReference.cs
```diff
@@ -1,347 +1,352 @@
...
+                    // When incremental build is triggered after that, we can potentially try to
+                    // compile a file that is no longer on disk.  So ensure that we don't queue any files
+                    // that no longer exist.
+                    LocalMarkupPages = alMarkupPages.Cast<LocalReferenceFile>().Where(x => File.Exists(x.FilePath)).ToArray();
```
vatsan-madhavan: whoa! does this touch every markup file on the disk for every incremental build? that can't be good for build-perf 😢 

Can we instead use a [`FileSystemWatcher`](https://docs.microsoft.com/en-us/dotnet/api/system.io.filesystemwatcher?view=netcore-3.0) and let the OS tell us when something is deleted? (i.e., don't poll - use interrupts?)

/cc @nguerrera, @dsplaisted can you think of a more efficient way to do this? i'm afraid the answer is going to be no, but figured i'd check in anyway.

rladuca: The watcher would only run during build, the delete is prior to the build, so that can't help us here.  

I think any perf issue would be a one-shot cost that is of similar disk cost to just doing 2 back to back incremental builds.  In the incremental build case, we're still asking MarkupCompilePass1 to touch xaml files and re-generate BAML.

rladuca: I suppose we can (if we have access to it) examine the Page blob for files that no longer exist in it and skip them.  I can look into plumbing that down into this part of PBT.

rladuca: Correction above, the Page change is what was invalidating the markup and touching all XAML.  I got confused between some binlogs.  In any case, I have a candidate fix I am messing with right now that should just rely on an intersection between the set of current pages passed into MarkupCompilePass1 vs. the cached values.

rladuca: Resolving as I've force pushed a change that just hits the current Page blob.

src/Microsoft.DotNet.Wpf/src/PresentationBuildTasks/MS/Internal/Tasks/IncrementalCompileAnalyzer.cs
```diff
@@ -389,13 +389,22 @@ private void UpdateFileListForIncrementalBuild(List<FileUnit> modifiedXamlFiles)
                 {
                     numLocalTypeXamls = CompilerLocalReference.LocalMarkupPages.Length;
 
+                    // Under incremental builds of SDK projects, we can have a state where the cache contains some XAML files but the Page blob
+                    // no longer contains them.  To avoid attempting to recompile a file that no longer exists, ensure that any cached XAML file
+                    // still exists in the Page blob prior to queuing it up for recompilation.
+                    HashSet<string> localMarkupPages = new HashSet<string>(_mcPass1.PageMarkup.Select(x => x.GetMetadata("FullPath")));
```
rladuca: Add string to SharedStrings class and use it from there.

vatsan-madhavan: Instantiate `HashSet<string>` with `StringComparer.OrdinalIgnoreCase`, something like this:
```C#
HashSet<string> localMarkupPages = new HashSet<string>(_mcPass1.PageMarkup.Select(x => x.GetMetadata("FullPath")), StringComparer.OrdinalIgnoreCase);

```

src/Microsoft.DotNet.Wpf/src/PresentationBuildTasks/MS/Internal/Tasks/IncrementalCompileAnalyzer.cs
```diff
@@ -389,13 +390,22 @@ private void UpdateFileListForIncrementalBuild(List<FileUnit> modifiedXamlFiles)
                 {
                     numLocalTypeXamls = CompilerLocalReference.LocalMarkupPages.Length;
```
SamBent: Surely this variable was meant to be used in the for-loop two lines down.

rladuca: Yeah probably, no real reason for it that I can see, I'll just remove it.

SamBent: Don't remove it.  Use it.

PR #2293, opened 2019-12-08
Updated target framework path to 3.1

dotMorten:
Fixed the reference path


vatsan-madhavan:
Don't think we'll take this change in release/3.1 branch. 

Changes to release/* branches are limited to what is absolutely essential, and almost all such changes require shiproom review and app;roval. Fixing this in master would have to do - sorry ☹️ 

Thanks for catching this and offering to fix - we appreciate it!

Issue #2298, opened 2019-12-10
Remove dependency on .NET SDK installs for IL Tools


SamBent:
* .NET Core Version: (e.g. 3.0 Preview1, or daily build number, use `dotnet --info`) 5.0.100-alpha1-015914
* Windows version: (`winver`) 1909 (10.0.18363)
* Does the bug reproduce also in WPF for .NET Framework 4.8?: No
* Is this bug related specifically to tooling in Visual Studio (e.g. XAML Designer, Code editing, etc...)?  No

 **Problem description:** Build failure when suitable .NET SDK is not installed

**Repro steps:**
1. New machine with Win10 1909 (hence .NET 4.8).
2. Setup machine per [WPF developer guide](https://github.com/dotnet/wpf/blob/master/Documentation/developer-guide.md).
3. Fork/clone master branch of dotnet/wpf.
4. Obtain fix for #2297 
5. build 
 
 **Actual behavior:** 
 D:\repos\wpf\eng\WpfArcadeSdk\tools\InjectModuleInitializer.targets(372,5): error MSB4044: The "ILDasmTask" task was not given a value for the required parameter "ILDAsm". [D:\repos\wpf\src\Microsoft.DotNet.Wpf\src\PresentationCore\PresentationCore.csproj]

 **Expected behavior:** No failure
 
This failure goes away after installing .NET 4.6.1 SDK, which makes a version of ildasm.exe visible to
PresentationCore.csproj.  Note that ilasm.exe is already visible (at C:\Windows\Microsoft.NET\Framework64\v4.0.30319\ILAsm.exe) before installing a .NET SDK

vatsan-madhavan:
@rladuca Didn't you find that (a) there is a bug in newer versions .NET Framework's of ILDASM and (b) we can't use .NET Core ILDASM/ILASM just yet due to https://github.com/dotnet/coreclr/issues/2982?

rladuca:
.NET 4.8 had the [floating point round-tripping issue](https://developercommunity.visualstudio.com/content/problem/545431/ildasmexe-regression-with-infinum-floating-point-v.html) and, yes, .NET Core still doesn't save debugging information.

[...]

PR #2301, opened 2019-12-10
Add .NET SDK to developer guide

SamBent:
Document dependency on .NET SDK.  

The build of PresentationCore uses IL tools.  ILdasm.exe is obtained from a .NET SDK, either 4.6.1 or 4.6.2 (but not higher).

This is a mitigation for #2298.

### DIFF
--- a/Documentation/developer-guide.md
+++ b/Documentation/developer-guide.md
@@ -16,6 +16,7 @@ WPF requires the following workloads and  components be selected when installing
 * Required Individual Components:
   * C++/CLI support
   * Windows 10 SDK
+  * .NET 4.6.1 SDK or .NET 4.6.2 SDK (see [#2298](https://github.com/dotnet/wpf/issues/2298))
 
 ## Workflow
 


vatsan-madhavan:
The best way to articulate required components is by using Visual Studio [workload component ID's](https://docs.microsoft.com/en-us/visualstudio/install/workload-component-id-vs-enterprise?view=vs-2019). 

Our current documentation is a far cry from that 😞 - it is incomplete, and doesn't use a standardized format. We require repo users to install Visual Studio 2019 (now we want them to use 16.4+ since there is a known Xeon bug in older tools), yet we don't enforce it (there is a simple way to do that - see below) or document it in a standard format. 

For e.g., .NET 4.6.2 SDK is `Microsoft.Net.Component.4.6.2.SDK` (unless you meant `Microsoft.Net.ComponentGroup.4.6.2.DeveloperTools` or something else related). 

Ideally, we want this list **not** in the documentation at all, but instead in `global.json` inside the [`"vs.components"` section](https://github.com/dotnet/arcade/blob/3d22e08e45bc87afd6dc5e6724dec28c78acb086/Documentation/ArcadeSdk.md). 

Something like this would be both self-documenting also:

```json
"vs": {
  "version": "16.4",
  "components": ["Microsoft.Net.Component.4.6.2.SDK"] // and more...
}
```

If a local install of VS fails to match these requirements, Arcade build scripts would failover to an xcopy deployable version of MSBuild. 

If we also specify an invalid version for `xcopy-msbuild` (since we don't want to use that), it will surely result in the build failing immediately (due to unsatisfied dependencies). Something like this:

```json
{
  "tools": {
    "vs": {
      "version": "16.4",
      "components": ["Microsoft.Net.Component.4.6.2.SDK"] // and more...
    },
    "xcopy-msbuild": "0.0"
  }
}
```

SamBent:
@vatsan-madhavan - your ideas don't work in practice.  

After changing global.json, the build fails immediately:

> (NETCORE_ENGINEERING_TELEMETRY=InitializeToolset) No package matching package identifier RoslynTools.MSBuild and version 0.0 was found in the given feed.

Removing the "xcopy-msbuild" line gets farther, but still fails immediately:

> (NETCORE_ENGINEERING_TELEMETRY=InitializeToolset) No package matching package identifier RoslynTools.MSBuild and version 16.4.0-alpha was found in the given feed.

(The VS installed on my machine is 16.5.0 Preview 1.0)

Even if we get past these problems, the "components" line requires the 4.6.2 SDK, which is more restrictive than necessary (we need either 4.6.1 or 4.6.2, but not higher - I suspect there's no way to express that in global.json).   [Note that this requirement may go away, pending a fix to #2998.]

Also, IMO the list _should_ be in the doc, so that people can get started right the first time.  This bit me recently when I included "Windows 10 SDK" in my VS install.  The doc didn't say which version, so I chose the latest one.  But the build failed, fortunately with a message that deciphered to the real requirement of Windows 10 SDK (10.0.17134.0) - the 3rd largest version on VS Installer's list.  I'd fix that in the doc now if I knew it was the right version for everyone and wasn't likely to change soon.  Is it?

Yes, it would be nice if all this worked automagically, but the current PR is the best I can do today.

vatsan-madhavan:
BTW this works quite well: 

https://github.com/wpfcontrib/wpf/blob/2e18cd5679f2f904b15a8851b86f699a0203c667/global.json#L10-L44

I think https://github.com/wpfcontrib/wpf/commit/2e18cd5679f2f904b15a8851b86f699a0203c667 may be a reasonable first-pass at the set of components we need to build.

PR #2305, opened 2019-12-11
Remove dependency on .NET SDK installs for IL Tools

### LINKED ISSUE #2298: Remove dependency on .NET SDK installs for IL Tools

rladuca:
Fixes #2298

Use a set of matched, private IL tools in order to inject a module constructor into `PresentationCore`.  These are copied binaries from a .NET Framework 4.7.2 installation and SDK.

Also fixing some issues with utilizing `ILAsm` outside of the installed Framework directory and some bugs that went unnoticed in our injection targets.

# Pending
- [x] IL comparison between `PresentationCore` from this change and current .NET 5.0 `PresentationCore`.

### DIFF
--- a/eng/WpfArcadeSdk/tools/InjectModuleInitializer/AddModuleConstructorTask.cs
+++ b/eng/WpfArcadeSdk/tools/InjectModuleInitializer/AddModuleConstructorTask.cs
@@ -41,31 +41,33 @@ namespace WpfArcadeSdk.Build.Tasks
                 var moduleConstructorIL = File.ReadAllText(ModuleConstructorIL);
                 var msCorLibAssemblySectionIL = File.ReadAllText(MsCorLibAssemblySectionIL);
                 var sourceIL = File.ReadAllText(ILFile);
+                string ilToInject = msCorLibAssemblySectionIL + Environment.NewLine + moduleConstructorIL;
 
                 using (var outputIL = new StreamWriter(File.Open(ILFile, FileMode.Truncate)))
                 {
-                    if (Regex.Match(sourceIL, @"\.assembly extern mscorlib", RegexOptions.Singleline) == Match.Empty)
-                    {
-                        outputIL.WriteLine(msCorLibAssemblySectionIL);
-                    }
-
                     if (Regex.Match(sourceIL, @"\.class.+?ModuleInitializer.+?\.method.+?static.+?Initialize\(\).+?end of class ModuleInitializer", RegexOptions.Singleline) == Match.Empty)
                     {
                         Log.LogError("Inserting a module initializer requires the assembly to implement a class named ModuleInitializer with a static parameterless method named Initialize.");
                         return false;
                     }
 
-                    if (Regex.Match(sourceIL, @"\.class private auto ansi '\<Module\>'.+?\.method private hidebysig specialname rtspecialname static void \.cctor \(\) cil managed ", RegexOptions.Singleline) == Match.Empty)
-                    {
-                        outputIL.WriteLine(moduleConstructorIL);
-                    }
-                    else
+                    if (Regex.Match(sourceIL, @"\.class private auto ansi '\<Module\>'.+?\.method private hidebysig specialname rtspecialname static void \.cctor \(\) cil managed ", RegexOptions.Singleline) != Match.Empty)
                     {
                         Log.LogError("Cannot insert a module initializer into an assembly that already contains one.");
                         return false;
                     }
 
-                    outputIL.WriteLine(sourceIL);
+                    // Attempt to replace an existing mscorlib reference with the reference + module initializer.
+                    string modifiedIL = Regex.Replace(sourceIL, @"\.assembly extern mscorlib.+};", ilToInject, RegexOptions.Singleline);
+
+                    // If we are given the same string back, there was nothing to replace.
+                    // In that case write out the reference and initializer from scratch.
+                    if (object.ReferenceEquals(modifiedIL, sourceIL))
+                    {
+                        outputIL.WriteLine(ilToInject);
+                    }
+
+                    outputIL.WriteLine(modifiedIL);
                 }
 
                 return true;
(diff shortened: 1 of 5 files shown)


vatsan-madhavan:
You have to undo @SamBent's documentation update.

vatsan-madhavan:
Instead of inventing new targets for ILAsm/ILDasm, you can reuse those supplied by `Microsoft.NET.Sdk.IL.targets` that comes as part of `Microsot.NET.Sdk.il` package, and override `ILAsmToolPath` ? 

As long as the commandline usage have remained largely compatible between .NET Framework and .NET Core versions of ILAsm/ILDasm, I suspect that this would be doable.

rladuca:
> Instead of inventing new targets for ILAsm/ILDasm, you can reuse those supplied by `Microsoft.NET.Sdk.IL.targets` that comes as part of `Microsot.NET.Sdk.il` package, and override `ILAsmToolPath` ?
> 
> As long as the commandline usage have remained largely compatible between .NET Framework and .NET Core versions of ILAsm/ILDasm, I suspect that this would be doable.

I don't think [those targets ](https://github.com/dotnet/runtime/blob/4f9ae42d861fcb4be2fcd5d3d55d5f227d30e723/src/coreclr/src/.nuget/Microsoft.NET.Sdk.IL/targets/Microsoft.NET.Sdk.IL.targets) are fully featured.  While you can override the IL(D)AsmDir properties to choose non-package based tools, it seems there is only a target to decompile resources and a target to compile an IL file.  They also don't accept some switches that we use.  Generally, they seem created for a specific purpose and not for general use unless I am missing something.

vatsan-madhavan:
>Generally, they seem created for a specific purpose and not for general use 

 😞

rladuca:
IL comparison seems fine, no major differences there between corresponding release DLLs.

eng/WpfArcadeSdk/tools/InjectModuleInitializer/AddModuleConstructorTask.cs
```diff
@@ -1,80 +1,80 @@
...
+        {
+            try
+            {
+                var moduleConstructorIL = File.ReadAllText(ModuleConstructorIL);
+                var msCorLibAssemblySectionIL = File.ReadAllText(MsCorLibAssemblySectionIL);
+                var sourceIL = File.ReadAllText(ILFile);
+
+                using (var outputIL = new StreamWriter(File.Open(ILFile, FileMode.Truncate)))
+                {
+                    if (Regex.Match(sourceIL, @"\.assembly extern mscorlib", RegexOptions.Singleline) == Match.Empty)
```
SamBent: If sourceIL already contains mscorlib reference, shouldn't you fail now?   If I understand what you've told me, that situation will produce IL where the module initializer precedes the mscorlib reference, which leads ilasm to fail silently (by not including the module init).  So catching this now when we can complain loudly seems good.

rladuca: Yeah, when we first did this we wanted to keep it general.  However, given the failure mode is silent, we should be as loud as possible when things seem incorrect.

rladuca: I decided just to make it work by replacing the mscorlib reference when we find it with the combined set of what we mean to inject (assembly reference included).  If we do not find it, we inject both from scratch at the beginning.

eng/WpfArcadeSdk/tools/InjectModuleInitializer/AddModuleConstructorTask.cs
```diff
@@ -41,31 +41,33 @@ public override bool Execute()
...
-                    else
+                    if (Regex.Match(sourceIL, @"\.class private auto ansi '\<Module\>'.+?\.method private hidebysig specialname rtspecialname static void \.cctor \(\) cil managed ", RegexOptions.Singleline) != Match.Empty)
                     {
                         Log.LogError("Cannot insert a module initializer into an assembly that already contains one.");
                         return false;
                     }
 
-                    outputIL.WriteLine(sourceIL);
+                    // Attempt to replace an existing mscorlib reference with the reference + module initializer.
+                    string modifiedIL = Regex.Replace(sourceIL, @"\.assembly extern mscorlib.+};", ilToInject, RegexOptions.Singleline);
```
SamBent: You're now looking for a multi-line pattern, from ".assembly extern mscorlib" to the matching "};" several lines later.
So SingleLine isn't right.

SamBent: I take that back.  "Singleline" does not mean what I thought it meant.  Never start a land war in Asia.

rladuca: That's not correct, there are multiple multi-line patterns in the checks, all of which use SingleLine.

SingleLine means it match across newline characters, see: https://docs.microsoft.com/en-us/dotnet/api/system.text.regularexpressions.regexoptions?view=netframework-4.8

It's counter-intuitive (at least to me).

PR #2366, opened 2019-12-28
Ensure the WindowsDesktopSdk RepoLocation overrides all others.

### LINKED ISSUE #2363: Microsoft.Net.Sdk.WindowsDesktop missing Mifcrosoft.WinFx.props
It seems the latest package versions of Microsoft.Net.Sdk.WindowsDesktop are missing a few files, like Microsoft.WinFx.props and Microsoft.WinFx.targets and the tools folder.

We caught this because with these files missing it is not possible to build windows desktop apps.

I turned the tests for these off in core-sdk to unblock the flow of dependencies there and linked to this issue.

rladuca:
Fixes #2363 

Ensure the WindowsDesktopSdk RepoLocation overrides all others.  This is key to packaging PresentationBuildTasks and associated files.

### DIFF
--- a/eng/WpfArcadeSdk/tools/ShippingProjects.props
+++ b/eng/WpfArcadeSdk/tools/ShippingProjects.props
@@ -185,7 +185,7 @@
     <RepoLocation Condition="'$(RepoLocation)' == '' And $(ExternalShippingProjects.Contains('$(NormalizedMSBuildProjectName)'))">External</RepoLocation>
  
     <!-- WindowsDesktopSdk overrides Internal, External etc. -->
-    <RepoLocation Condition="'$(RepoLocation)' == '' And $(WindowsDesktopSdkProject.Contains('$(NormalizedMSBuildProjectName)'))">WindowsDesktopSdk</RepoLocation>
+    <RepoLocation Condition="$(WindowsDesktopSdkProject.Contains('$(NormalizedMSBuildProjectName)'))">WindowsDesktopSdk</RepoLocation>
 
     <IncludeLibFilesInPackaging Condition="$(ShippingLibProjects.Contains('$(NormalizedMSBuildProjectName)'))">true</IncludeLibFilesInPackaging>
   </PropertyGroup>


eng/WpfArcadeSdk/tools/ShippingProjects.props
```diff
@@ -185,7 +185,7 @@
     <RepoLocation Condition="'$(RepoLocation)' == '' And $(ExternalShippingProjects.Contains('$(NormalizedMSBuildProjectName)'))">External</RepoLocation>
  
     <!-- WindowsDesktopSdk overrides Internal, External etc. -->
-    <RepoLocation Condition="'$(RepoLocation)' == '' And $(WindowsDesktopSdkProject.Contains('$(NormalizedMSBuildProjectName)'))">WindowsDesktopSdk</RepoLocation>
+    <RepoLocation Condition="$(WindowsDesktopSdkProject.Contains('$(NormalizedMSBuildProjectName)'))">WindowsDesktopSdk</RepoLocation>
```
rladuca: @vatsan-madhavan I was thinking of adding something here to specifically check for overriding internal or external, but I felt that was overkill and we always want the SDK to package properly.

rladuca:
Looks like the bot just up and merged this.  @vatsan-madhavan if there are any comments you have on this I'll update with a new PR, for now I think this should just solve the immediate problem.

vatsan-madhavan:
Why was targets being picked up but not props - what makes props different?

rladuca:
> Why was targets being picked up but not props - what makes props different?
> 
> https://github.com/dotnet/wpf/blob/d53c1b914676303e6bd32b0c1c2981329eed5e46/src/Microsoft.DotNet.Wpf/src/PresentationBuildTasks/PresentationBuildTasks.csproj#L42-L49

Nothing in PBT was packaged properly.  If you look at the WindowsDesktop SDK package, you'll see everything PBT (including WinFX) was missing.  They were landing in our private transport package.

vatsan-madhavan:
I just changed your code back and packaged it locally - windowsdesktop.nupkg built correctly for me.

Issue #2425, opened 2020-01-14
Microsoft Security Advisory CVE-2020-0606 : .NET Core Remote Code Execution Vulnerability


blowdart:
# Microsoft Security Advisory CVE-2020-0606 : .NET Core Remote Code Execution Vulnerability

## <a name="executive-summary"></a>Executive summary

Microsoft is releasing this security advisory to provide information about a vulnerability in .NET Core. This advisory also provides guidance on what developers can do to update their applications to remove this vulnerability.

Microsoft is aware of a remote code execution vulnerability exists in .NET software when the software fails to check the source markup of a file. An attacker who successfully exploited the vulnerability could run arbitrary code in the context of the current user.

Exploitation of the vulnerability requires that a user open a specially crafted file with an affected version of .NET Core. In an email attack scenario, an attacker could exploit the vulnerability by sending the specially crafted file to the user and convincing the user to open the file.

The security update addresses the vulnerability by correcting how .NET Core checks the source markup of a file.

## Announcement

The original announcement for this issue can be found at https://github.com/dotnet/announcements/issues/149

## <a name="mitigation-factors"></a>Mitigation factors

Microsoft has not identified any mitigating factors for this vulnerability.

## <a name="affected-software"></a>Affected software

Any .NET Core 3.0 application running on .NET Core 3.0.0, 3.0.1 or 3.1.0.

### <a name="how-affected"></a>How do I know if I am affected?

If you have a runtime or SDK with a version listed in [affected software](#affected-software) you are exposed to the vulnerability.

### <a name="how-fix"></a>How do I fix the issue?

To fix the issue please install the latest version of .NET Core. If you have multiple versions of .NET Core installed you will need to install multiple runtimes, or SDKs depending on what you have installed.

You can list the versions you have installed by running the `dotnet --info` command. You will see output like the following;

```
.NET Core SDK (reflecting any global.json):
 Version:   3.1.100
 Commit:    cd82f021f4

Runtime Environment:
 OS Name:     Windows
 OS Version:  10.0.18363
 OS Platform: Windows
 RID:         win10-x64
 Base Path:   C:\Program Files\dotnet\sdk\3.1.100\

Host (useful for support):
  Version: 3.1.0
  Commit:  65f04fb6db

.NET Core SDKs installed:
  3.0.101 [C:\Program Files\dotnet\sdk]
  3.1.100 [C:\Program Files\dotnet\sdk]

.NET Core runtimes installed:
[...]

PR #2428, opened 2020-01-14
Fixing Microsoft Security Advisory CVE-2020-0605 : .NET Core Remote Code Execution Vulnerability (3.1 Fixed PR)

rladuca:
This PR addresses issues #2424, #2425


vatsan-madhavan:
@mmitche I think this is better - it preserves the original commit from the internal branch. we'll recreate the tags after merging. 

as far as i can see, i don't think we can do a branding update here  - our versioning doesn't lend itself to updates for servicing releases. 

https://github.com/dotnet/wpf/blob/c86d4b9e7e16c3de7b65990f44a52bd630e33dca/eng/Versions.props#L3-L10

vatsan-madhavan:
**note**: do not squash when merging

vatsan-madhavan:
@rladuca  - empty change - so no azp run. You can just merge it directly IMO. Remember to create a merge-commit and not a squash-commit.

PR #2521, opened 2020-02-08
Add support for publishing Debug packages separately

vatsan-madhavan:
This is used for testing and will be consumed in an upcoming repo-refactoring. 

Release transport-packages will remain unchanged. Debug packages will get a `.Debug` suffix to the `PackageName`.


vatsan-madhavan:
@rladuca please review when you are able to - i'm merging now since I need this for further use immediately.

PR #2553, opened 2020-02-12
Open Sourcing WpfGfx and PenImc

vatsan-madhavan:
Open Sourcing Native Binaries

- Migrates WpfGfx and PenImC from internal repo (dotnet-wpf-int)
- Has several related adjustments to build props/targets.

The end result is that `Microsoft.DotNet.Wpf.GitHub` produced will contain everything - all of WPF's runtime content.

/cc @dotnet/wpf-developers


lindexi:
Awesome

[WineHQ Bugzilla – Bug 24501 – wine-mono does not support WPF](https://bugs.winehq.org/show_bug.cgi?id=24501 )

[A developers guide to running WPF apps on Linux with .NET Core and Wine | github.com/ccifra](https://ccifra.github.io/PortingWPFAppsToLinux/Overview.html )

PR #2573, opened 2020-02-13
Update win10 sdk used by WPF builds to 18362

vatsan-madhavan:
/cc @wjk fyi.


vatsan-madhavan:
@rladuca mind looking at these lines a bit carefully: 

https://github.com/wpfcontrib/wpf/blob/a25e8ab1242d73899a0305508b3a2f1e64ea2134/src/Microsoft.DotNet.Wpf/src/WpfGfx/common/DirectXLayer/XMath/extensions_xm.hpp#L65-L74

PR #2605, opened 2020-02-18
Include additional headers in the transported Microsoft.DotNet.Arcade.Wpf.Sdk used by internal dotnet-wpf-int repo

vatsan-madhavan:
Include additional headers in the transported Microsoft.DotNet.Arcade.Wpf.Sdk used by internal dotnet-wpf-int repo.

### DIFF
--- a/eng/WpfArcadeSdk/tools/FolderPaths.targets
+++ b/eng/WpfArcadeSdk/tools/FolderPaths.targets
@@ -14,6 +14,10 @@
     <!-- Consume from $(WpfTestArcadeWpfSdkPath) -->
     <WpfTransportedCommonDir Condition="'$(WpfTransportedCommonDir)'=='' And Exists('$(WpfArcadeSdkRoot)..\..\src\Microsoft.DotNet.Wpf\src\Common\')">$(WpfArcadeSdkRoot)..\..\src\Microsoft.DotNet.Wpf\src\Common\</WpfTransportedCommonDir>
 
+    <WpfTransportedGraphicsDir Condition="'$(WpfTransportedGraphicsDir)'=='' And Exists('$(WpfArcadeSdkRoot)src\WpfGfx\')">$(WpfArcadeSdkRoot)src\WpfGfx\</WpfTransportedGraphicsDir>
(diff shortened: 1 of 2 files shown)


eng/WpfArcadeSdk/tools/FolderPaths.targets
```diff
@@ -14,6 +14,10 @@
     <!-- Consume from $(WpfTestArcadeWpfSdkPath) -->
     <WpfTransportedCommonDir Condition="'$(WpfTransportedCommonDir)'=='' And Exists('$(WpfArcadeSdkRoot)..\..\src\Microsoft.DotNet.Wpf\src\Common\')">$(WpfArcadeSdkRoot)..\..\src\Microsoft.DotNet.Wpf\src\Common\</WpfTransportedCommonDir>
 
+    <WpfTransportedGraphicsDir Condition="'$(WpfTransportedGraphicsDir)'=='' And Exists('$(WpfArcadeSdkRoot)src\WpfGfx\')">$(WpfArcadeSdkRoot)src\WpfGfx\</WpfTransportedGraphicsDir>
+    <WpfTransportedGraphicsDir Condition="'$(WpfTransportedGraphicsDir)'=='' And Exists('$(WpfArcadeSdkRoot)..\..\src\Microsoft.DotNet.Wpf\src\WpfGfx\')">$(WpfArcadeSdkRoot)..\..\src\Microsoft.DotNet.Wpf\src\WpfGfx\</WpfTransportedGraphicsDir>
+
+    <WpfTransportedGraphicsPath>$(WpfTransportedGraphicsDir)</WpfTransportedGraphicsPath>
```
rladuca: Why do we need a Dir property and a Path property when they are the same value?

vatsan-madhavan: We really don't; when I was building this I noticed that the local convention here was `YYYDir`, but the convention elsewhere tends to be `YYYPath`. I figured I'd just respect both and invent an alias.  😞

rladuca: Yeah I don't think it's a huge deal, was just making sure there wasn't something I was missing about it.

PR #2650, opened 2020-02-24
https://www.facebook.com/shenkate.liyag

jhasmine15:
https://www.facebook.com/shenkate.liyag********


karelz:
@vatsan-madhavan use the "..." in right top corner of the post - choose "Report content". You then have option for "Block user".
I did it in this case ...

PR #2668, opened 2020-02-28
Fix CGlyphRunResource::GetDWriteRenderingMode to choose correct DWriteRenderingMode under ClearType

### LINKED ISSUE #2025: ClearType TextRenderingMode uses incorrect DWrite rendering mode
* .NET Core Version: All
* Windows version: All

When rendering text, `CGlyphRunResource::GetDWriteRenderingMode` is called in order to get the appropriate rendering mode to pass to DWrite in` CGlyphRunResource::CreateRealization` for a call to `pIDWriteFactoryNoRef->CreateGlyphRunAnalysis`.

The analysis is what gives WPF the glyph bitmap(s) it uses to render to the screen.  The problem comes in some special cases in `CGlyphRunResource::GetDWriteRenderingMode`.  In this function, WPF overrides what the recommended rendering mode would be from DWrite with `DWRITE_RENDERING_MODE_CLEARTYPE_NATURAL`. DWrite, if given the choice, would use `DWRITE_RENDERING_MODE_NATURAL_SYMMETRIC` for this font and emsize.

What this means is the algorithm used to render the font is not symmetric, as in the anti-aliasing algorithm is only performed in the horizontal and not the vertical direction.  Due to this, fine features in the horizontal can be sampled out if the font itself is optimized for symmetric anti-aliasing.  

This bug was originally found while investigating an issue with the rendering of `MS Mincho` after the installation of [KB4482887](https://support.microsoft.com/en-us/help/4482887/windows-10-update-kb4482887).  The application was setting `TextOptions.TextRenderingMode` to `ClearType`.  According to the [documentation](https://docs.microsoft.com/en-us/dotnet/api/system.windows.media.textrenderingmode?view=netframework-4.8):

```
Text is rendered with the most appropriate `ClearType` rendering algorithm based on the layout mode that was used to format the text.
```

This is simply not true for all fonts as described above.  WPF should, in this case, simply ask DWrite what the appropriate rendering mode is and use the closest `ClearType` equivalent to that ([see here](https://docs.microsoft.com/en-us/windows/win32/api/dwrite/ne-dwrite-dwrite_rendering_mode)).  There are other special cases that should be preserved, but this particular one needs fixing.

Note that applications that do not set `TextOptions.TextRenderingMode` to `ClearType` will not see this problem as the default mode is `Auto`.  In `Auto`, WPF will ask DWrite what the appropriate mode to use is.

EDIT:  Since `wpfgfx_cor3` has been open sourced, here is the offending code:
https://github.com/dotnet/wpf/blob/c3b1ecefd9debf850b4757bb8d657b15af79469d/src/Microsoft.DotNet.Wpf/src/WpfGfx/core/resources/glyphrunslave.cpp#L948

[Original internal bug for reference](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/979556)

rladuca:
Fixes #2025

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/WpfGfx/core/resources/glyphrunslave.cpp
+++ b/src/Microsoft.DotNet.Wpf/src/WpfGfx/core/resources/glyphrunslave.cpp
@@ -966,38 +966,54 @@ CGlyphRunResource::GetDWriteRenderingMode(__in IDWriteFontFace *pIDWriteFontFace
             *pDWriteRenderingMode = DWRITE_RENDERING_MODE_CLEARTYPE_NATURAL_SYMMETRIC;
         }
         else
-        {      
-            if (   (textRenderingMode == MilTextRenderingMode::Grayscale)
+        {   
+            // We defer to DWrite for this decision in some cases
+            // The scaleFactor is used in conjunction with the muSize to calculate the
+            // actual rendered size of this glyph run.
+            if (FAILED((pIDWriteFontFace->GetRecommendedRenderingMode(
+                m_muSize,
+                scaleFactor,
+                m_measuringMethod,
+                pDisplaySettings->pIDWriteRenderingParams,
+                pDWriteRenderingMode
+                ))))
+            {
+                //
+                // Default to CT natural/ideal in failure case, since failure of this
+                // call is non fatal for our purposes.
+                //
+                *pDWriteRenderingMode = IsDisplayMeasured() ? DWRITE_RENDERING_MODE_CLEARTYPE_GDI_CLASSIC :
+                    DWRITE_RENDERING_MODE_CLEARTYPE_NATURAL;
+            }
+
+            if ((textRenderingMode == MilTextRenderingMode::Grayscale)
                 || (textRenderingMode == MilTextRenderingMode::ClearType))
             {
-                *pDWriteRenderingMode = IsDisplayMeasured() ? 
-                                        DWRITE_RENDERING_MODE_CLEARTYPE_GDI_CLASSIC :
-                                        DWRITE_RENDERING_MODE_CLEARTYPE_NATURAL;
+                if (IsDisplayMeasured())
+                {
+                    *pDWriteRenderingMode = DWRITE_RENDERING_MODE_CLEARTYPE_GDI_CLASSIC;
+                }
+                else
+                {
+                    // If DWrite choose a symmetric anti-aliasing algorithm and the developer
+                    // has explicitly chosen ClearType rendering, choose the corresponding
+                    // symmetric ClearType algorithm.
+                    if (*pDWriteRenderingMode == DWRITE_RENDERING_MODE_CLEARTYPE_NATURAL_SYMMETRIC
+                        || *pDWriteRenderingMode == DWRITE_RENDERING_MODE_NATURAL_SYMMETRIC)
+                    {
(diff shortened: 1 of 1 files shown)


vatsan-madhavan:
Do we still support Windows Server 2008 R2?

rladuca:
> Do we still support Windows Server 2008 R2?

Can you explain what that has to do with this change?  We will only choose symmetric if DWrite itself chooses symmetric.  It's not obvious, but `DWRITE_RENDERING_MODE_NATURAL_SYMMETRIC` and `DWRITE_RENDERING_MODE_CLEARTYPE_NATURAL_SYMMETRIC` are the same.  As in, they have the exact same value in the enumeration.

vatsan-madhavan:
> Can you explain what that has to do with this change? 

I rushed into the question thinking it was new call and may not be supported on some platforms; and then realized it has been used before  all along.

rladuca:
> > Can you explain what that has to do with this change?
> 
> I rushed into the question thinking it was new call and may not be supported on some platforms; and then realized it has been used before all along.

All good, there are other things I am looking into that actually do have a Win8+ requirement.  But those would have to be a separate PR and would be much more extensive and invasive.  If that actually becomes a thing we need to have a discussion about what we're willing to change in the rendering process to fully support them.  I'll file a different issue altogether.

vatsan-madhavan:
LGTM; low pri - can you add some comments about how a sample app might exercise each of the branches in GetDWriteRenderingMode (including the last dangling else with just an assert)? It would be helpful when writing tests for this, I think (or debugging problems...)

rladuca:
Sure, I can add that.

rladuca:
I am going to wait on merging this anyway until I get some more info on how effective this change will be.

src/Microsoft.DotNet.Wpf/src/WpfGfx/core/resources/glyphrunslave.cpp
```diff
@@ -943,6 +943,20 @@ CGlyphRunResource::CreateRealization(
...
+// TextRenderingMode | TextFormattingMode | DWRITE_RENDERING_MODE
+// --------------------------------------------------------------
+// Auto              | Display            | DWrite's Choice
+// Auto              | Ideal              | DWrite's Choice
+// Aliased           | Display            | DWRITE_RENDERING_MODE_ALIASED
+// Aliased           | Ideal              | DWrite's Choice
+// ClearType         | Display            | DWRITE_RENDERING_MODE_CLEARTYPE_GDI_CLASSIC
+// ClearType         | Ideal              | DWRITE_RENDERING_MODE_CLEARTYPE_NATURAL_SYMMETRIC if DWrite chose symmetric, otherwise DWRITE_RENDERING_MODE_CLEARTYPE_NATURAL
+// Grayscale         | Display            | DWRITE_RENDERING_MODE_CLEARTYPE_GDI_CLASSIC
+// Grayscale         | Ideal              | DWRITE_RENDERING_MODE_CLEARTYPE_NATURAL
```
SamBent: The code doesn't handle this case.  Appears to fall through to ask DWrite.

rladuca:
I did a run of all text DRTs and microsuites and everything passed.  I'm suspicious that we don't have certain parts of this coverage ported for .NET Core.  Before I merge this I am going to make the equivalent changes to .NET Framework and get a full pass done there.

rladuca:
Full test run is back and everything looks good, merging.

PR #2691, opened 2020-03-02
Fix handling of wildcard AssemblyVersion properties during markup compilation

### LINKED ISSUE #2517: For non-deterministic builds WPF application crashes on resource dictionary load
If project is set to be non-deterministic and AssemblyVersion uses wildcards, the application fails while loading resource dictionaries. It seem to try to parse Version object from wildcard string from meta (?) instead of determining concrete running assembly version, either that or there is a bug in Version parser. 
Note: I understand there are benefits to deterministic, but that is not the topic, non-deterministic should not be a showstopper for wpf core, especially when porting existing applications and follow-up processes. 

*__Steps to reproduce:__*

1. Create .net core WPF application
2. Modify csproj and switch to wildcard non-deterministic build

```xml
  <PropertyGroup>
    <TargetFramework>netcoreapp3.1</TargetFramework>
    <UseWPF>true</UseWPF>
    <!-- non-deterministic with wildcards -->
    <Deterministic>False</Deterministic>
    <AssemblyVersion>1.0.*</AssemblyVersion>
  </PropertyGroup>
</Project>
```

2. Add any resource file to project (e.g. test.xaml), contents is irrelevant.
3. Load resource from App.xaml

```xml
<Application x:Class="WpfApp1.App"
             xmlns:local="clr-namespace:WpfApp1"
             StartupUri="MainWindow.xaml">
    <Application.Resources>
        <ResourceDictionary>
            <ResourceDictionary.MergedDictionaries>
                <ResourceDictionary Source="Test.xaml" />
            </ResourceDictionary.MergedDictionaries>
        </ResourceDictionary>
    </Application.Resources>
</Application>
```
4. Build&Run app

*__Expected behavior__*
A working application?

*__Actual behavior__*
Application crashes with unhandled exception immediately on start. Stack trace is:

```
Application: WpfApp1.exe
CoreCLR Version: 4.700.19.56402
.NET Core Version: 3.1.0
Description: The process was terminated due to an unhandled exception.
Exception Info: System.FormatException: Input string was not in a correct format.
   at System.Number.ThrowOverflowOrFormatException(ParsingStatus status, TypeCode type)
   at System.Version.TryParseComponent(ReadOnlySpan`1 component, String componentName, Boolean throwOnFailure, Int32& parsedComponent)
   at System.Version.ParseVersion(ReadOnlySpan`1 input, Boolean throwOnFailure)
   at System.Version.Parse(String input)
   at System.Version..ctor(String version)
   at System.Windows.Navigation.BaseUriHelper.GetLoadedAssembly(String assemblyName, String assemblyVersion, String assemblyKey)
   at MS.Internal.AppModel.ResourceContainer.GetResourceManagerWrapper(Uri uri, String& partName, Boolean& isContentFile)
   at MS.Internal.AppModel.ResourceContainer.GetPartCore(Uri uri)
   at System.IO.Packaging.Package.GetPartHelper(Uri partUri)
   at System.IO.Packaging.Package.GetPart(Uri partUri)
   at System.Windows.Application.GetResourceOrContentPart(Uri uri)
   at System.Windows.Application.LoadComponent(Object component, Uri resourceLocator)
   at WpfApp1.App.InitializeComponent() in C:\playground\WpfApp1\WpfApp1\App.xaml:line 1
   at WpfApp1.App.Main()
```

rladuca:
fixes #2517


lindexi:
Maybe we should fix the line break.

rladuca:
> Maybe we should fix the line break.

I think we need to sweep the repo and set some attributes to ensure we're all using the same line endings.  I don't want to make single changes in this PR, for now I'll keep this as is and we'll make a choice and fixup the repo.

src/Microsoft.DotNet.Wpf/src/PresentationBuildTasks/MS/Internal/MarkupCompiler/MarkupCompiler.cs
```diff
@@ -2596,7 +2596,20 @@ private void GenerateInitializeComponent(bool isApp)
...
 
-            string version = String.IsNullOrEmpty(AssemblyVersion) ? String.Empty : COMPONENT_DELIMITER + VER + AssemblyVersion;
+            // If we have an actual version string, try to detect wildcards.  Note that this is required to be at least major.minor.
+            // Prior to markup compilation, this will have been validated by the C# compiler.
+            bool wildcardUsed = false;
+
+            if (!string.IsNullOrEmpty(AssemblyVersion))
+            {
+                var splitVersion = AssemblyVersion.Split('.');
+                wildcardUsed = splitVersion[splitVersion.Length - 1] == "*";
```
vatsan-madhavan: You need index checks here. IIRC you mentioned that csc would do additional checks for us, but we can not rely on that for all languages. This needs to be robust for VB, as well as any future language support we might add. We should not rely on C# compilers behavior here - we should do our own independent validation that the version is well-formed

rladuca: VB is a good point.  I can do the full version check like Roslyn does, but we'll have to add errors and localization.  Let me look into adding all of that.

rladuca: Side note, the [VB compiler](https://github.com/dotnet/roslyn/tree/master/src/Compilers/VisualBasic) is under Roslyn as well.  It uses the same version checks.

src/Microsoft.DotNet.Wpf/src/PresentationBuildTasks/MS/Internal/MarkupCompiler/MarkupCompiler.cs
```diff
@@ -2596,7 +2596,20 @@ private void GenerateInitializeComponent(bool isApp)
...
 
-            string version = String.IsNullOrEmpty(AssemblyVersion) ? String.Empty : COMPONENT_DELIMITER + VER + AssemblyVersion;
+            bool hasWildcard = false;
+
+            // Attempt to parse out the AssemblyVersion if it exists.  This validates that we can either use an empty version string (wildcards exist)
+            // or we can utilize the passed in string (valid parse).
+            if (!string.IsNullOrEmpty(AssemblyVersion)
+                && !VersionHelper.TryParseAssemblyVersion(AssemblyVersion, allowWildcard: true, out Version _, out hasWildcard))
+            {
+                throw new Exception(SR.Get(SRID.InvalidAssemblyVersion, AssemblyVersion));
```
rladuca: Still investigating if throwing or using the OnError mechanism is more appropriate here.

With OnError we signal via an event and return, throwing *should* make its way up to a catch that does OnError at a higher level.  OnError is odd here since we may end up in a state where we continue to parse but haven't generated the file.  The throw is intended to short circuit that and stop immediately.

I am still testing this, so I marked it WIP and will investigate further.

src/Microsoft.DotNet.Wpf/src/PresentationBuildTasks/MS/Internal/MarkupCompiler/MarkupCompiler.cs
weltkante: You might want to be robust against IL postprocessing tools editing the assembly version as well, depending on how they are embedded in the msbuild chain they may be running before you are looking at the attributes.

src/Microsoft.DotNet.Wpf/src/PresentationBuildTasks/MS/Internal/MarkupCompiler/MarkupCompiler.cs
rladuca: The error looks like this in VS (with a throw) and it comes up per XAML file.  The line numbers, etc, basically point to the first EndElement (this is where we generate `InitializeComponent` function).

I'm sort of on the fence about this, I might have to find a better way to plumb things all the way up beyond the markup compiler error paradigm.

image

rladuca: Also, FYI, `MC1005` is just the next number available in a series of errors that looks more akin to project configuration issues rather than actual parsing errors.  So I felt it better to group it there.

src/Microsoft.DotNet.Wpf/src/PresentationBuildTasks/MS/Internal/MarkupCompiler/MarkupCompiler.cs
vatsan-madhavan: That's why the note about robustness against future language support 😛. C++/CLI, F# etc. support are both plausible in near future; F# esp. seems to be growing in popularity and we've heard some requests. 

In this case, the validation isn't very hard to implement locally either, so IMHO we should just add it here. We could even borrow it from Roslyn.

rladuca: I already borrowed the Roslyn code and modified it for our purposes.

Resolving this against that.

src/Microsoft.DotNet.Wpf/src/PresentationBuildTasks/MS/Internal/MarkupCompiler/MarkupCompiler.cs
vatsan-madhavan: MC1005 seems fine to me.

vatsan-madhavan: It would be better if the errors mapped to the target of the task though.

rladuca: I can take a look at doing that, we'd have to have the plumbing to get the data back to the task output.  The MarkupCompiler is down a ways in the stack and EndElement is basically a parser event callback.

What we could do is throw a new exception type and, instead of trapping and converting to a MarkupException, allow it to bubble to the task and trap it specifically there, resulting in the output.

I'll think about it.

vatsan-madhavan: I think that will work - just invent a new internal-only `Exception` type, and then when you catch it inside `Execute`...

src/Microsoft.DotNet.Wpf/src/PresentationBuildTasks/MS/Internal/MarkupCompiler/MarkupCompiler.cs
```diff
@@ -762,6 +762,13 @@ static void ThrowCompilerExceptionImpl(string message)
 
         internal void OnError(Exception e)
         {
+            // Don't treat an AssemblyVersion parsing error as a XamlParseException.
+            // Throw it back to the task execution.
+            if(e is AssemblyVersionParseException)
+            {
+                throw e;
```
weltkante: Should this be `System.Runtime.ExceptionServices.ExceptionDispatchInfo.Capture(e).Throw();` in order to add to the stack trace instead of overwriting it?

src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Markup/XamlParser.cs
```diff
@@ -203,6 +203,13 @@ public bool ReadXaml(bool singleRecordMode)
                 }
                 else
                 {
+                    // Don't treat an AssemblyVersion parsing error as a XamlParseException.
+                    // Throw it back to the task execution.
+                    if(e is AssemblyVersionParseException)
+                    {
+                        throw e;
```
weltkante: ditto

src/Microsoft.DotNet.Wpf/src/PresentationBuildTasks/MS/Internal/MarkupCompiler/MarkupCompiler.cs
rladuca: Yeah should just be `throw`.

rladuca: Or rather, just throw in the XamlParser.

rladuca:
Moving this from WIP to actual, I think it's working pretty well right now.

rladuca:
I've checked the following scenarios:

- Normal build (Deterministic + non-wildcard AssemblyVersion)
- WildCard Framework-Style (Non-Deterministic + wildcard AssemblyVersionAttribute)
- WildCard Core-Style (Non-Deterministic + wildcard AssemblyVersion)
- Various permutations of bad AssemblyVersion properties (1.1.x, 1.1.1x, etc)

Everything works as expected.  The error is shown on a bad AssemblyVersion (assuming XAML compilation occurs).  Without the fix, wildcard applications using AssemblyVersion crash as in the report.  With the fix, they no longer crash.  Behavior is the same in both normal scenarios and in AssemblyVersionAttribute scenarios (as far as I can tell).

C# does give a warning about wildcard AssemblyVersion properties:
```
 obj\Debug\netcoreapp5.0\PbtWildCardTest.AssemblyInfo.cs(16,59): warning CS7035: The specified version string does not
 conform to the recommended format - major.minor.build.revision [d:\wpf\PbtWildCardTest\PbtWildCardTest.csproj]
```
That seems to be the only difference between using the attribute and the msbuild property.

src/Microsoft.DotNet.Wpf/src/PresentationBuildTasks/MS/Internal/MarkupCompiler/MarkupCompiler.cs
```diff
@@ -2596,7 +2603,20 @@ private void GenerateInitializeComponent(bool isApp)
 
             string uriPart = string.Empty;
 
-            string version = String.IsNullOrEmpty(AssemblyVersion) ? String.Empty : COMPONENT_DELIMITER + VER + AssemblyVersion;
+            bool hasWildcard = false;
```
vatsan-madhavan, line 2606: Declare `hasWildcard` as part of `VersionHelper.TryParseAssemblyVersion` call.

src/Microsoft.DotNet.Wpf/src/PresentationBuildTasks/MS/Internal/MarkupCompiler/MarkupCompiler.cs
```diff
@@ -2596,7 +2603,20 @@ private void GenerateInitializeComponent(bool isApp)
...
+
+            // Attempt to parse out the AssemblyVersion if it exists.  This validates that we can either use an empty version string (wildcards exist)
+            // or we can utilize the passed in string (valid parse).
+            if (!string.IsNullOrEmpty(AssemblyVersion)
+                && !VersionHelper.TryParseAssemblyVersion(AssemblyVersion, allowWildcard: true, out Version _, out hasWildcard))
```
vatsan-madhavan, line 2611: s/`out Version _`/`out var _` => discard doesn't really need its type called out either 😉

src/Microsoft.DotNet.Wpf/src/PresentationBuildTasks/MS/Internal/MarkupCompiler/MarkupCompiler.cs
```diff
@@ -2596,7 +2603,20 @@ private void GenerateInitializeComponent(bool isApp)
...
+            {
+                throw new AssemblyVersionParseException(SR.Get(SRID.InvalidAssemblyVersion, AssemblyVersion));
+            }
+
+            // If a developer explicitly sets the AssemblyVersion build property to a wildcard version string, we would use that as part of the URI here.
```
vatsan-madhavan, line 2616: [nit]: A detailed explanation might be helpful, esp. one that explains the context in which `AssemblyVersion` vs. `AssemblyVersionAttribute` come into play, old-style vs. SDK-style projects etc. Puzzling out the nature of the bug and the fix was minor brain-teaser, and though the extent of these comments are strictly sufficient for the next person to reason about the behavior completely, a little bit more detail might be beneficial.

src/Microsoft.DotNet.Wpf/src/PresentationBuildTasks/MS/Internal/MarkupCompiler/MarkupCompiler.cs
```diff
@@ -2596,7 +2603,20 @@ private void GenerateInitializeComponent(bool isApp)
...
+            // This results in an error in Version.Parse during InitializeComponent's call tree.  Instead, do as we would have when the developer sets a
+            // wildcard version string via AssemblyVersionAttribute and use an empty string.
+            string version = (hasWildcard || String.IsNullOrEmpty(AssemblyVersion)) ? String.Empty : COMPONENT_DELIMITER + VER + AssemblyVersion;
```
vatsan-madhavan, line 2619: outer parens may be unnecessary 

nit: for better readability, consider folding ternary operator into multiple lines, like this: 

```c#
            string version = hasWildcard || String.IsNullOrEmpty(AssemblyVersion) 
                ? String.Empty 
                : COMPONENT_DELIMITER + VER + AssemblyVersion;
```

src/Microsoft.DotNet.Wpf/src/PresentationBuildTasks/MS/Internal/MarkupCompiler/VersionHelper.cs
```diff
@@ -0,0 +1,215 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+// Modified from https://github.com/dotnet/roslyn/blob/master/src/Compilers/Core/Portable/VersionHelper.cs
```
vatsan-madhavan, line 5: Refer to the sha that formed the source (instead of blob/master/... which is ever changing)

src/Microsoft.DotNet.Wpf/src/PresentationBuildTasks/Resources/xlf/Strings.cs.xlf
```diff
@@ -152,6 +152,11 @@
         <target state="translated">Třída InternalTypeHelper není pro tento projekt požadována, vyprázdněte soubor {0}.</target>
         <note />
       </trans-unit>
+      <trans-unit id="InvalidAssemblyVersion">
```
vatsan-madhavan, line 155: We need to notify the localizers about this.

src/Microsoft.DotNet.Wpf/src/PresentationBuildTasks/MS/Internal/MarkupCompiler/MarkupCompiler.cs
lindexi: How about `out _`  https://docs.microsoft.com/en-us/dotnet/csharp/discards

src/Microsoft.DotNet.Wpf/src/PresentationBuildTasks/MS/Internal/MarkupCompiler/VersionHelper.cs
lindexi: Maybe we should use commit instead of master strings

src/Microsoft.DotNet.Wpf/src/PresentationBuildTasks/MS/Internal/MarkupCompiler/VersionHelper.cs
```diff
@@ -0,0 +1,215 @@
...
+
+using System;
+using System.Diagnostics;
+using System.Globalization;
+
+namespace MS.Internal
+{
+    internal static class VersionHelper
+    {
+        static Version NullVersion = new Version(0, 0, 0, 0);
```
lindexi, line 17: And I think the NullVersion is a readonly field

lindexi: And the original code is `readonly` 

https://github.com/dotnet/roslyn/blob/284a67e9e9e1b33370aadf15ff082389f9dedb50/src/Compilers/Core/Portable/MetadataReference/AssemblyIdentity.cs#L248

src/Microsoft.DotNet.Wpf/src/PresentationBuildTasks/MS/Internal/MarkupCompiler/MarkupCompiler.cs
rladuca: I prefer to at least have some at a glance context on what parameter is being discarded.  So I'll add an explicit argument name as well.

src/Microsoft.DotNet.Wpf/src/PresentationBuildTasks/Resources/xlf/Strings.cs.xlf
rladuca: The localization team has automation that should detect and update based on these changes (I have an email to that affect:

```
Currently all 3 locations are monitored daily by automation and we keep weekly cycles of localization. In case of new string is added we deliver PR to repository.
```

In any case, I will ping them as a heads up once we merge.

HClausing:
Thanks @vatsan-madhavan !

Now I understand the correlation between these tags on csproj file and Application.LoadComponent() method.

vatsan-madhavan:
@dotnet/wpf-developers Please write a nice doc for this and make it discoverable through public documentation so that the next person can learn about it more easily. Right now the only way to discover it is for someone in the know to make the correlation and point community-members to this PR and the associated issue.

PR #2719, opened 2020-03-05
Preview 3 Branding

rladuca:


vatsan-madhavan:
We should port https://github.com/dotnet/wpf/pull/2439 over soon.

PR #2721, opened 2020-03-06
Add .gitattributes and renormalize line endings

rladuca:
Added a copy of (mostly) the .gitattributes from dotnet/runtime.

Also ran `git add --renormalize .` to fixup any issues with line endings.

I tested this out with autocrlf enabled and disabled and it all seems to work fine.  Files I was editing earlier today that resulted in large whitespace diffs no longer show these issues.

With autocrlf off and a fresh clone, I ran a search (via `file`) to find everything with CR/LF endings and this renormalized list seems to match that.  So I believe things are all fixed up.


.gitattributes
```diff
@@ -0,0 +1,67 @@
+# Set default behavior to automatically normalize line endings.
+* text=auto
+
+*.doc  diff=astextplain
+*.DOC	diff=astextplain
```
vatsan-madhavan: Shouldn't .doc be binary?

.gitattributes
```diff
@@ -0,0 +1,67 @@
...
+*.docx	diff=astextplain
+*.DOCX	diff=astextplain
+*.dot	diff=astextplain
+*.DOT	diff=astextplain
+*.pdf	diff=astextplain
+*.PDF	diff=astextplain
```
vatsan-madhavan: .dot, .pdf also binary perhaps?

vatsan-madhavan:
Also, consider creating a merge commit instead of squashing this PR.

vatsan-madhavan:
One more suggestion - I'd really like this change ported to release/3.1. I'm sure we can get permission to merge the change as an infra-update.

.gitattributes
rladuca: Everywhere I look, these are all `astextplain`.  Consensus isn't correctness, but I have to dig deeper into why that's the way it is.

rladuca: Interesting, so it seems there is a common [shell script](https://github.com/msysgit/msysgit/blob/master/bin/astextplain) that is used to do conversions into text from word and PDF docs.  The funny thing is, dotnet/runtime doesn't have this script at all and no one mentioned `msysgit`.  

In any case, I don't see a real reason to not mark these as binary for our purposes, I'll change them up.

vatsan-madhavan:
Once this is ported to release/3.1, do we need at last @dotnet/wpf-developers to zap their clones (and clones of any forks) and re-clone again ?

rladuca:
> Once this is ported to release/3.1, do we need at least @dotnet/wpf-developers to zap their clones (and clones of any forks) and re-clone again ?

That all depends, you could have files sitting in old branches that are not re-normalized.  I had this issue with XLF generation and merging with master earlier.  I think a pull from master to master should be fine and all will be well.  However, branches with recent changes may need a `git add --renormalize .` to correct any problems post-pull.  Rebasing might be just fine.

PR #2869, opened 2020-04-09
Adjust properties to allow muti-target

wli3:
without special Sdk="Microsoft.NET.Sdk.WindowsDesktop"

The project could look like

```xml

  <PropertyGroup>
    <TargetFramework>net5.0</TargetFramework>
    <UseWPF>true</UseWPF>
  </PropertyGroup>

</Project>
```

### DIFF
--- a/packaging/Microsoft.NET.Sdk.WindowsDesktop/targets/Microsoft.NET.Sdk.WindowsDesktop.props
+++ b/packaging/Microsoft.NET.Sdk.WindowsDesktop/targets/Microsoft.NET.Sdk.WindowsDesktop.props
@@ -1,40 +1,6 @@
 <Project>
-  <PropertyGroup>
-    <_MicrosoftNetSdkWindowsDesktop>true</_MicrosoftNetSdkWindowsDesktop>
-    <EnableDefaultPageItems Condition="'$(EnableDefaultPageItems)' == ''">true</EnableDefaultPageItems> 
-    <EnableDefaultApplicationDefinition Condition="'$(EnableDefaultApplicationDefinition)' == ''">true</EnableDefaultApplicationDefinition>
-  </PropertyGroup>
 
-  <PropertyGroup>
-    <!-- 
-    WindowsDesktop SDK supports WPF and WinForms on 
-      - .NET Core 3.0 and greater
-      - .NET Framework 3.0 and greater
-   
-    Note that on .NET Framework versions < 4.0, additional workarounds may be required to build applications 
-    using the SDK style projects. For e.g., see https://github.com/microsoft/msbuild/issues/1333
-    
-    Irrespective of whether '$(TargetFrameworkIdentifier)' is '.NETCoreApp' or '.NETFramework', 
-    the minimum value of $(_TargetFrameworkVersionValue) we will be testing for is '3.0'
-    
-  -->
-    <_WindowsDesktopSdkTargetFrameworkVersionFloor>3.0</_WindowsDesktopSdkTargetFrameworkVersionFloor>
-
-    <!-- 
-      Represents an undefined TFV value. This will be used in comparisons of _TargetFrameworkVersionValue (defined in Microsoft.NET.WindowsDesktop.targets)
-      to identify when a TFV is undefined
-    -->
-    <_UndefinedTargetFrameworkVersion>0.0</_UndefinedTargetFrameworkVersion>
-    
-    <!--
-      Initial/Default value set to 'undefined'. Updated in Microsoft.NET.WindowsDesktop.targets
-    --> 
-    <_TargetFrameworkVersionValue>$(_UndefinedTargetFrameworkVersion)</_TargetFrameworkVersionValue>
-  </PropertyGroup>
-
-  <ItemGroup Condition=" ('$(EnableDefaultItems)' == 'true') And ('$(UseWPF)' == 'true') And 
-                         ('$(_TargetFrameworkVersionValue)' != '$(_UndefinedTargetFrameworkVersion)') And 
-                         ('$(_TargetFrameworkVersionValue)' >= '$(_WindowsDesktopSdkTargetFrameworkVersionFloor)')">
+  <ItemGroup Condition=" '$(_EnableWindowsDesktopGlobbing)' == 'true' ">
     <ApplicationDefinition Include="App.xaml"
                            Condition="'$(EnableDefaultApplicationDefinition)' != 'false' And Exists('$(MSBuildProjectDirectory)/App.xaml') And '$(MSBuildProjectExtension)' == '.csproj'">
       <Generator>MSBuild:Compile</Generator>
@@ -66,10 +32,7 @@
   </ItemGroup>
 
   
-  <ItemGroup Condition="('$(DisableImplicitFrameworkReferences)' != 'true') And
-                      ('$(TargetFrameworkIdentifier)' == '.NETCoreApp') And
-                      ('$(_TargetFrameworkVersionValue)' != '$(_UndefinedTargetFrameworkVersion)') And
-                      ('$(_TargetFrameworkVersionValue)' >= '$(_WindowsDesktopSdkTargetFrameworkVersionFloor)')">
+  <ItemGroup Condition=" '$(_EnableWindowsDesktopNetCoreFrameworkReferences)' == 'true' ">
 
     <FrameworkReference Include="Microsoft.WindowsDesktop.App" IsImplicitlyDefined="true"
                         Condition="('$(UseWPF)' == 'true') And ('$(UseWindowsForms)' == 'true')"/>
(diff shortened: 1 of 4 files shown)


packaging/Microsoft.NET.Sdk.WindowsDesktop/Sdk/Sdk.props
```diff
@@ -13,13 +13,11 @@ Copyright (c) .NET Foundation. All rights reserved.
 
   <Import Sdk="Microsoft.NET.Sdk" Project="Sdk.props" />
```
wli3, line 15: Most of the sdk.props and sdk.targets are empty now

packaging/Microsoft.NET.Sdk.WindowsDesktop/targets/Microsoft.NET.Sdk.WindowsDesktop.props
```diff
@@ -1,40 +1,6 @@
 <Project>
-  <PropertyGroup>
```
wli3, line 2: Moved to target

packaging/Microsoft.NET.Sdk.WindowsDesktop/targets/Microsoft.NET.Sdk.WindowsDesktop.props
```diff
@@ -1,40 +1,6 @@
...
-    <!--
-      Initial/Default value set to 'undefined'. Updated in Microsoft.NET.WindowsDesktop.targets
-    --> 
-    <_TargetFrameworkVersionValue>$(_UndefinedTargetFrameworkVersion)</_TargetFrameworkVersionValue>
-  </PropertyGroup>
-
-  <ItemGroup Condition=" ('$(EnableDefaultItems)' == 'true') And ('$(UseWPF)' == 'true') And 
-                         ('$(_TargetFrameworkVersionValue)' != '$(_UndefinedTargetFrameworkVersion)') And 
-                         ('$(_TargetFrameworkVersionValue)' >= '$(_WindowsDesktopSdkTargetFrameworkVersionFloor)')">
+  <ItemGroup Condition=" '$(_EnableWindowsDesktopGlobbing)' == 'true' ">
```
wli3, line 3: Extract a new property in target to reduce the surface area, since props will always be imported

src/Microsoft.DotNet.Wpf/src/PresentationBuildTasks/Microsoft.WinFx.props
```diff
@@ -1,16 +0,0 @@
-<Project>
-  <PropertyGroup>
```
wli3, line 2: Moved to WinFx.target

packaging/Microsoft.NET.Sdk.WindowsDesktop/targets/Microsoft.NET.Sdk.WindowsDesktop.targets
```diff
@@ -22,6 +51,25 @@
   <PropertyGroup>
     <_TargetFrameworkVersionValue>$([MSBuild]::ValueOrDefault('$(_TargetFrameworkVersionWithoutV)', '$(_UndefinedTargetFrameworkVersion)'))</_TargetFrameworkVersionValue>
   </PropertyGroup>
+
+    <!-- PropertyGroup flags control if Itemgroups in Microsoft.NET.Sdk.WindowsDesktop.props file is being imported -->
+  <PropertyGroup>
+      <_EnableWindowsDesktopGlobbing Condition=" ('$(EnableDefaultItems)' == 'true') And ('$(UseWPF)' == 'true') And
```
wli3, line 57: Please make sure the name looks good @dsplaisted

wli3:
@dsplaisted also I realized later the first insertion to SDK will fail without SDK side of change. However, I think that is fine, we can merge the SDK side of change with the insertion

wli3:
@vatsan-madhavan let's setup a time to discuss that. I am surprised that .NET Core SDK is not the only public facing dependent.

vatsan-madhavan:
Microsoft.DotNet.Arcade.Wpf.Sdk  is internal to WPF repos and it’s used in managing our dual-repo (or 3 repo, if you include test) build orchestration. Think of it as an augmentation to Arcade that includes all of WPF’s build intelligence and overrides, including the ability to do markup compilation from sources before WindowsDesktop SDK is built. That last part is what’s coming from Pbt.*. 

This project is going to require some  experimenting and getting things right. I’d recommend spending some time building the 3 repos and getting to know Microsoft.DotNet.Arcade.Wpf.Sdk, how local markup compilation works and how we build other wpf repos etc. MSBuild Binlogs are your friend. 

Also spend time understanding the packaging architecture - it’s pretty powerful and unique to our repo. There is nothing magic about it, but there is a lot of convention-based stuff going on.

vatsan-madhavan:
Could you add properties to detect multiple inclusion, and simply include WindowsDesktop SDK from with the base Microsoft.NET.Sdk ? 

Then if Net.Sdk.WindowsDesktop includes Microsoft.NET.Sdk, then Microsoft.NET.Sdk has to detect this condition (Net.Sdk.WindowsDesktop would have already set a property) and avoid re-inclusion. 

And if Microsoft.NET.Sdk is above, and includes Net.Sdk.WindowsDesktop (owning to UseWpf or UseWindowsForms), then Net.Sdk.WindowsDesktop has to detect this condition (probably by leveraging a property set by Microsoft.NET.Sdk) and avoid re-inclusion of Microsoft.NET.Sdk

wli3:
@dsplaisted @vatsan-madhavan 
> Could you add properties to detect multiple inclusion, and simply include WindowsDesktop SDK from with the base Microsoft.NET.Sdk ?

Actually I don't think there is a case multiple inclusion will occur. There is graph of import

image

wli3:
I updated my PR. This should be a "refactoring". So there should be no visible change from outside (no SDK import method change)

rladuca:
> Actually I don't think there is a case multiple inclusion will occur. There is graph of import
Can someone still FrameworkReference?  People have been using that idiom to get WPF/WinForms binaries in console apps.  This change shouldn't break those projects.

dsplaisted:
> Can someone still FrameworkReference? People have been using that idiom to get WPF/WinForms binaries in console apps. This change shouldn't break those projects.

Yes, adding an explicit FrameworkReference should not be impacted by these changes.

src/Microsoft.DotNet.Wpf/src/PresentationBuildTasks/Microsoft.WinFx.props
Nirmal4G: Instead of moving the code, just add import into WinFX targets, if they are not already imported.

OR Is that pattern not preferred?

packaging/Microsoft.NET.Sdk.WindowsDesktop/targets/Microsoft.NET.Sdk.WindowsDesktop.props
Nirmal4G: Instead of moving the code, just add import into WinFX targets, if they are not already imported.

OR Is that pattern not preferred?

src/Microsoft.DotNet.Wpf/src/PresentationBuildTasks/Microsoft.WinFx.props
wli3: In this case, props will be imported regardless if the workload is installed or not. So we need to move everything that could be moved.

For more see the description, linked issue. As well as the doc related to the net5.0 design.

Nirmal4G: You could keep the logic in the props file and import it in the Sdk targets. No need to move the logic and risk losing history. -- Just a suggestion

wli3:
Any update?

wli3:
Ping. It is been more than a month

ryalanms:
I'll be testing this today and tomorrow.  Thanks.

ryalanms:
This will require an update to the SDK in all projects in https://github.com/microsoft/WPF-Samples.

dsplaisted:
> This will require an update to the SDK in all projects in https://github.com/microsoft/WPF-Samples.

How will they need to update?

The intent is that existing projects should continue to work.  It would be good to update the samples to the new project format once we ship an SDK that supports it.  However I wouldn't expect the samples to break with this change.  Was there a break you ran into?

ryalanms:
Please disregard.  I haven't seen anything break, and it looks like the samples are still building against 3.1.

ryalanms:
All samples from the wpf-samples repo I've tested are building successfully against these changes.  The updates to the props/targets files did not cause any regressions with building existing .NET 5 applications.    

To clarify the goal and staging of this feature: @wli3 is enabling WPF applications to build with Project SDK set to `Microsoft.NET.Sdk`, but this will come in two separate parts.  This is part 1, which refactors `Microsoft.NET.Sdk.WindowsDesktop\targets` files (Microsoft.WinFX.props/targets, Microsoft.NET.Sdk.WindowsDesktop.props/targets).  WPF applications will require `<Project = Sdk="Microsoft.NET.Sdk.WindowsDesktop"> ` until part 2 of @wli3's changes come in.

wli3:
@vatsan-madhavan could you dismiss your "change required"? So I can merge

wli3:
also looks like WIP bot get stuck (it was removed across the org). Maybe I need to create a new PR just to workaround that

wli3:
Continue here https://github.com/dotnet/wpf/pull/3111

PR #2871, opened 2020-04-10
SpellCheck incorrectly detects common abbreviations like "e.g." as misspellings

vatsan-madhavan:
WPF's spell-checker on Win8+ ignores punctuations like `.` (period) at the end of words like `etc.` This results in well-formed words like `etc.`, `Mr.`, `Mrs.` being marked as spelling-errors.

Further, it is impossible to _correct_ these words in WPF.

Because of the way word-breaking is performed by `MS.Data.Text.WordsSegmenter` under the hood, these punctuations are always dropped during word-breaking. The older spell-checker (NaturalLanguage6.dll) used on Win7 does not have this defect.

`MS.Data.Text.WordsSegmenter` has declined to 'fix'/'modify this behavior, so we are adding a workaround in WPF.

This change introduces a wrapper on top of the word-breaker and recalculates the tokens, taking into account dropped text that occurs immediately after a token that is likely to fail spell-checking.

**Note:** See #2753 for the main discussion


src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Documents/WinRTSpellerInterop.cs
```diff
@@ -854,14 +854,25 @@ private SpellChecker CurrentSpellChecker
 
         #region Private Types
 
-        private struct TextRange: SpellerInteropBase.ITextRange
+        internal struct TextRange: SpellerInteropBase.ITextRange
```
lindexi, line 857: Can we set it to readonly struct?

src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Documents/WinRTSpellerInteropExtensions.cs
```diff
@@ -0,0 +1,184 @@
...
+                        }
+                    }
+                }
+
+                allTokens.Add(
+                    new SpellerSegment(
+                        text,
+                        new WinRTSpellerInterop.TextRange(
+                            nextTokenStartPosition,
+                            (int)tokens[i].SourceTextSegment.Length),
```
lindexi, line 101: I think we should introduce the `tokens[i].SourceTextSegment` variable.  This makes the code neat

src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Documents/WinRTSpellerInteropExtensions.cs
```diff
@@ -0,0 +1,184 @@
...
+        ///       
+        ///     - Also see https://github.com/dotnet/wpf/pull/2753#issuecomment-602120768 for a discussion related to this. 
+        /// </remarks>
+        public static IReadOnlyList<SpellerSegment> ComprehensiveGetTokens(
+            this WordsSegmenter segmenter,
+            string text,
+            SpellChecker spellChecker,
+            WinRTSpellerInterop owner)
+        {
+            IReadOnlyList<WordSegment> tokens = segmenter?.GetTokens(text) ?? new List<WordSegment>().AsReadOnly();
```
lindexi, line 62: When the number of tokens array elements is empty, I think the empty allTokens array should be returned immediately

lindexi: When the number of tokens array elements is 0,  the code will be skip

https://github.com/dotnet/wpf/blob/749da527494e53cb677fdc7f0df9f8049a0d27a8/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Documents/WinRTSpellerInteropExtensions.cs#L66

https://github.com/dotnet/wpf/blob/749da527494e53cb677fdc7f0df9f8049a0d27a8/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Documents/WinRTSpellerInteropExtensions.cs#L107

And then it will return the empty allTokens

And I think you should return immediately when the `tokens` is empty and use the `Array.Empty` to return

src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Documents/NLGSpellerInterop.cs
```diff
@@ -643,6 +643,16 @@ private void EnumerateSubSegments()
 
             #region SpellerInteropBase.ISpellerSegment
 
+            /// <summary>
+            /// <inheritdoc/>
```
lindexi, line 647: The other code only uses the `/// <inheritdoc/>` without `/// <summary>` see:

https://github.com/dotnet/wpf/blob/7afb994e07336fc4bfb1c74c5c9669e0c464d2e2/src/Microsoft.DotNet.Wpf/src/Shared/System/Windows/Markup/Replacements/TypeUriConverter.cs#L21

src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Documents/WinRTSpellerInteropExtensions.cs
vatsan-madhavan: @lindexi thanks for the feedback!

src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Documents/NLGSpellerInterop.cs
vatsan-madhavan: @lindexi thanks for the suggestions - i've incorporated them.

src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Documents/WinRTSpellerInteropExtensions.cs
```diff
@@ -0,0 +1,191 @@
...
+            var allTokens = new List<SpellerSegment>();
+
+            if (tokens.Count == 0)
+            {
+                return allTokens.AsReadOnly();
```
lindexi, line 67: Thank you @vatsan-madhavan 

And I think this code may be a little improved

            if (tokens.Count == 0)
            {
                return Array.Empty<SpellerSegment>();
            }

            var allTokens = new List<SpellerSegment>();

src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Documents/WinRTSpellerInteropExtensions.cs
```diff
@@ -0,0 +1,191 @@
...
+        ///       
+        ///     - Also see https://github.com/dotnet/wpf/pull/2753#issuecomment-602120768 for a discussion related to this. 
+        /// </remarks>
+        public static IReadOnlyList<SpellerSegment> ComprehensiveGetTokens(
+            this WordsSegmenter segmenter,
+            string text,
+            SpellChecker spellChecker,
+            WinRTSpellerInterop owner)
+        {
+            IReadOnlyList<WordSegment> tokens = segmenter?.GetTokens(text) ?? new List<WordSegment>().AsReadOnly();
```
lindexi, line 62: I think when segmenter is null and the tokens can be set as the Array.Empty<WordSegment>

            IReadOnlyList<WordSegment> tokens = segmenter?.GetTokens(text) ?? Array.Empty<WordSegment>();

JesikaDG:
This issue seems like it is contributing to massive performance issues with the July .Net update, documented in https://github.com/dotnet/wpf/issues/3350. This is one of very few issue that touched the relevant files. On other I found was https://github.com/dotnet/wpf/pull/3092
There was mention in https://github.com/dotnet/wpf/pull/2753 about performance concerns, but no discussion here about managing those concerns.

PR #2890, opened 2020-04-15
Fix PenIMC Activation Context Handling

### LINKED ISSUE #2393: Issue with touch being disabled in certain scenarios
I found an issue where touch would be disabled.
If you want to create a custom splash screen and creates a new `Thread` for it, and then create the main window in the Main Thread, touch will no longer work.
This same code works fine in .NET Framework.

Project for reproducing the issue can be found here including description: https://github.com/mbendtsen/touch_issue_net_core
Requires a touch enabled device.

```csharp
public partial class App : Application
{
   private Thread _splashThread;
   
   protected async override void OnStartup(StartupEventArgs e)
   {
       base.OnStartup(e);
       
       ShowSplash();
       
       // Simulate long running work.
       await Task.Delay(2000);
       
       _splashViewModel.SetInitialized();
       
       var window = new MainWindow()
       {
          DataContext = new MainWindowViewModel()
       };
       window.Show();
   }
   
   private void ShowSplash()
   {
       _splashViewModel = new SplashViewModel();
       
       _splashThread = new Thread(() =>
       {
          new Splash { DataContext = _splashViewModel }.Show();
          Dispatcher.Run();
       });
       _splashThread.SetApartmentState(ApartmentState.STA);
       _splashThread.IsBackground = true;
       _splashThread.Name = "Splash";
       _splashThread.Start();
   }
}
```

rladuca:
Related #2393 (so we keep the bug open)

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/PenImc/dll/SxSCOMRegistration.cpp
+++ b/src/Microsoft.DotNet.Wpf/src/PenImc/dll/SxSCOMRegistration.cpp
@@ -46,6 +46,3 @@ extern "C" ULONG_PTR WINAPI RegisterDllForSxSCOM()
     // Return the context cookie : caller is responsible for deactivating the context.
     return activationContextCookie; 
 }
-
-
-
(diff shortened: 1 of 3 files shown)


rladuca:
We should port this to 3.1 as well.

rladuca:
@ryalanms You should seriously consider porting this change to 3.1 as well.

Issue #2072, opened 2019-10-21
TouchMove intermediate points are not DPI aware


dotMorten:
* .NET Core Version: 3.1 Preview1
* Windows version: (`winver`) 1809
* Does the bug reproduce also in WPF for .NET Framework 4.8?: Yes
* Is this bug related specifically to tooling in Visual Studio (e.g. XAML Designer, Code editing, etc...)? No

 **Problem description:**
Using the TouchMove event in a multi-dpi scenario, the interpolated points returned are not correctly adjusting for DPI, whereas the current point is.
When using the event to draw a path, you can see the discrepancy between these two inputs:
![image](https://user-images.githubusercontent.com/1378165/67240439-571cd800-f406-11e9-8b49-6baaad9a1a52.png)
In this case my main monitor is scale factor 1.5x, whereas my touch-monitor where I'm running the app on is 1.0x.
The app has been configured to support multi-monitor DPI via the app.manifest.

 
 **Actual behavior:** 
Touch events (red line) are about 30% off from the correct location. If I multiple the location with 1.5, the location is correct (apart from the distance of the window top-bar). This is however weird, as the screen the app is running on is 1.0x screen scale (but main monitor is 1.5x). I'd guess the app registers it needs to run at 1.5 screen scale, then realizes the app is now running at 1.x and but forgot intermediate points should no longer be divided by 1.5x.

 **Expected behavior:**
 Intermediate and current touch points coincide.

 **Minimal repro:**
Xaml:
```xml
    <Grid>
        <Canvas Margin="30" x:Name="canvas" TouchMove="canvas_TouchMove" IsManipulationEnabled="True" Background="White" />
    </Grid>
```
xaml.cs code-behind:
```cs
     private PathFigure intermediatePoints;
        private PathFigure currentPoint;

        private void canvas_TouchMove(object sender, TouchEventArgs e)
        {
            foreach (var p in e.GetIntermediateTouchPoints(canvas))
                intermediatePoints = AddVertex(p.Position, intermediatePoints, Colors.Red, 3);
            currentPoint = AddVertex(e.GetTouchPoint(canvas).Position, currentPoint, Colors.Blue, 1);
        }

        private PathFigure AddVertex(Point point, PathFigure pf, Color color, double width)
        {
            if (pf == null)
            {
                pf = new PathFigure(point, new PathSegment[] { }, false);
                PathGeometry pg = new PathGeometry(new PathFigure[] { pf });
                Path p = new Path() { Data = pg };
[...]

PR #2891, opened 2020-04-15
Fix raw stylus data to support per-monitor DPI

rladuca:
Related #2072


rladuca:
Should port this to 3.1 as well.

rladuca:
This can also be considered for a 4.8 fix (it's a per-mon DPI miss).  We should discuss that.

src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Input/Stylus/Common/StylusLogic.cs
```diff
@@ -180,6 +180,9 @@ internal enum FlickScrollDirection
         // Caches the pointer stack enabled state
         private static bool? _isPointerStackEnabled = null;
 
+        // Caches TransformToDevice matrices per DpiScale2
+        Dictionary<DpiScale2, Matrix> _transformToDeviceMatrices = new Dictionary<DpiScale2, Matrix>();
```
lindexi, line 184: Should we add private readonly ?

rladuca:
@ryalanms Just a ping on this, it blocks inking from working properly in per-monitor DPI scenarios.

dotMorten:
Thanks for the merge!

> This can also be considered for a 4.8 fix (it's a per-mon DPI miss). We should discuss that.

Was this discussed / done? If so I need to do a version-check and remove my workaround for newer frameworks.

PR #2877, opened 2020-04-13
Fix casing for Microsoft.WinFx.targets.

PaulEremeeff:
Fix incorrect casing for imports of wpf/src/Microsoft.DotNet.Wpf/src/PresentationBuildTasks/Microsoft.WinFx.targets file, which caused evaluation errors under Unix (unable to find import because the file is not found).

PR #2975, opened 2020-05-10
Fix 'WinFx' -> 'WinFX' casing

Nirmal4G:
As the title suggests, This PR renames **`WinFx`** to **`WinFX`** where possible except in the sources (e.g.: `GetWinFxCallback`).

I've separated out the new-line and white-space changes to a separate commits.
Only files that were touched with `WinFX` casing changes are updated with formatting changes.
If it's not possible to merge and/or if you want the line & space changes applied to the whole repo, please let me know, I'll separate these into a new PR.


Nirmal4G:
Supersedes #2877

moljac:
This is due to testing on case insensitive filesystems like Windows and Mac default AFS.

I think this should be fixed in the templates in the MsBuild.Extras.

And BTW you are breaking framework design guidelines in respect to naming. 
WinFx can never be WinFX. 
(Sorry, but I had to be mean/cynical/sarcastic).

Nirmal4G:
@moljac

> And BTW you are breaking framework design guidelines in respect to naming.

It's like WinRT, WinAPI, etc... There are special cases. Casing is important, It represents the product.

The `Microsoft.NET.Sdk.WindowsDesktop` is owned by the WPF team. The original casing is 'WinFX' and not `WinFx`. It was only changed in the NETCLR 4 builds. I don't know how or why but all target files are jumbled up in casing for comments, properties and filenames. They might've used a formatter but did not consider for the special cases.

> I think this should be fixed in the templates in the MSBuild.Sdk.Extras.

Extras SDK merely extends the preinstalled SDKs. There is no direct reference to these files in the Extras SDK.

Nirmal4G:
It seems this also fixes the case where people using case-sensitive file-system that would result in file not found error because of the incorrect casing in the files on disk.

ryalanms:
Thank you for the fix, @Nirmal4G.  Your casing commits have been separated and moved to another PR.

Nirmal4G:
@ryalanms Is there any issue with this PR, not being mergable?

PR #3001, opened 2020-05-14
Add links in "new issue" flow to other related repos

danmosemsft:
This .yml adds quick links to other repos below the existing issue template buttons in your new issue flow.

For an example of what this looks like, see https://github.com/danmosemsft/runtime/issues/new/choose (the white buttons below your existing green buttons)

We are adding this in dotnet/runtime (https://github.com/dotnet/runtime/pull/36431). Reasoning: it's sometimes confusing which is the correct repo to open an issue in. The goal here is to make it easier to open an issue in the correct repo first time so we need to move fewer issues and they more quickly get attention from the right people.

### DIFF
--- a/.github/ISSUE_TEMPLATE/config.yml
+++ b/.github/ISSUE_TEMPLATE/config.yml
@@ -1,0 +1,21 @@
+blank_issues_enabled: true
+contact_links:
+  - name: Issue with .NET runtime or core .NET libraries
+    url:  https://github.com/dotnet/runtime/issues/new/choose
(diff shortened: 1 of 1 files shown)


danmosemsft:
@SamBent does this look OK to you? It doesn't affect code, doesn't affect build - seems like merging this is just goodness. I've done it for all the other major dotnet repos.

danmosemsft:
@SamBent wanna merge ? 😸

danmosemsft:
Fixed typo.

PR #3004, opened 2020-05-15
Update branding for 3.1.5

ryalanms:


vatsan-madhavan:
Good job following the historical pattern like we discussed I’m email. 

As a convention, we’ve been diligent about **not** creating personal branches in this repo. Please take note of this and decide what you want your policies to be. I recommend being very strict about using forks. Also never have this repo’s url as a push-url in any of your remotes I’m clones (since you may have wrote permissions).

ryalanms:
Thanks for helping with this, Vatsan.  The forking policy makes sense and will be followed in the future.

RussKie:
I have 2 remotes:
- `origin` which points to my fork
- `remote` which points to the main trunk

You can also configure protections in the main trunk (e.g. for the master branch) to prevent destructive actions (e.g. force push, delete, etc.). These won't prevent from accidental push a new commit to the master though.
I'm quite certain this is already done, but worth calling this out as well.

PR #3008, opened 2020-05-17
Use `.Length == 0` instead of `== string.Empty`

Youssef1313:
Checking `.Length == 0` is a bit faster than `== string.Empty`.
For more information, see [CA1820](https://docs.microsoft.com/en-us/visualstudio/code-quality/ca1820?view=vs-2019#rule-description).


miloush:
Also smells of `NullReferenceException`. Did you check all the places you updated guarantee the string is non-null?

Youssef1313:
@miloush, Yes I checked that. But it's better to double check in case I missed something.

src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Input/Command/CommandValueSerializer.cs
```diff
@@ -117,7 +117,7 @@ public override object ConvertFromString(string value, IValueSerializerContext c
         {
             if (value != null)
             {
-                if (value != String.Empty)
+                if (value.Length != 0)
```
lindexi, line 120: I think this code can merge two if.

if (!string.IsNullOrEmpty(value))

src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Markup/XmlnsDictionary.cs
```diff
@@ -347,8 +347,7 @@ public string LookupNamespace(string prefix)
                 for (int thisDecl = _lastDecl-1; thisDecl >= 0; thisDecl--)
                 {
                     if ((_nsDeclarations[thisDecl].Prefix == prefix) && 
-                        _nsDeclarations[thisDecl].Uri != null && 
-                        _nsDeclarations[thisDecl].Uri != String.Empty)
+                        !String.IsNullOrEmpty(_nsDeclarations[thisDecl].Uri))
```
lindexi, line 350: I recommend you use keyword `string`

lindexi:
@miloush  I checked all too. 

Thank you @Youssef1313 and it looks good for me.

src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Input/Command/CommandValueSerializer.cs
Youssef1313: There is an `else` clause for the nested if.

```
if (value != null)
{
    if (value.Length != 0)
    {
        // ......
    }
    else
    {
        // ....
    }
}
```

I can't imagine how can they both be combined into one condition.

src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Markup/XmlnsDictionary.cs
Youssef1313:
```suggestion
                        !string.IsNullOrEmpty(_nsDeclarations[thisDecl].Uri))
```

src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Input/Command/CommandValueSerializer.cs
Nirmal4G: you could do:

```cs
if (string.IsNullOrEmpty(value)) return null;

// inner logic here...

return base.ConvertFromString(value, context);
```

Youssef1313: @Nirmal4G, if it's null, you'll return null by your code, but the original code will `return base.ConvertFromString(value, context);`.
I'd keep it as-is for this PR. I believe there might be a simplification for this nested if, but can't get it well in my head.

Nirmal4G: OK. I see now, the `ValueSerializer`'s `ConvertFromString` method throws null exception, if the value is null. So, we're hooking up with `base`'s error checking.

> Better way is to separate the exception helpers into separate static class and use it across the codebase.

deeprobin:
Short question: 
Is `value.Length != 0` faster than `value.Length > 0`?

lindexi:
> 
> 
> Short question:
> Is `value.Length != 0` faster than `value.Length > 0`?

I think both of these are very fast. Sorry, I can't answer your question directly.

deeprobin:
@lindexi 
I wrote a little benchmark in LinqPad:
```cs
const int Iterations = 10000000;

void Main()
{
	BenchNotEqual();
	BenchGreaterThan();
}

void BenchNotEqual() {
	var list = new List<bool>();
	var watch = new Stopwatch();
	watch.Start();
	
	for(var i = 0; i < Iterations; i++) {
		list.Add(RunNotEqual(i));
	}
	watch.Stop();

	Console.WriteLine($"Not equal took: {watch.ElapsedMilliseconds}ms");
}

void BenchGreaterThan()
{
	var list = new List<bool>();
	var watch = new Stopwatch();
	watch.Start();


	for (var i = 0; i < Iterations; i++)
	{
		list.Add(RunGreaterThan(i));
	}
	watch.Stop();

	Console.WriteLine($"Greater than took: {watch.ElapsedMilliseconds}ms");
}

[MethodImplAttribute(MethodImplOptions.NoInlining | MethodImplOptions.NoOptimization)]
bool RunNotEqual(int number) {
	return number != 0;
}

[MethodImplAttribute(MethodImplOptions.NoInlining | MethodImplOptions.NoOptimization)]
bool RunGreaterThan(int number)
{
	return number > 0;
}
```

I ran this script about 10 times and greater than was faster.
```
Not equal took: 156ms
Greater than took: 132ms
```

Feel free to try, maybe the results are not the same everywhere.

lindexi:
@deeprobin  Sorry, I do not think your benchmark is correct.

ThomasGoulet73:
The easiest way to check which is fastest is to check the generated native code from the JIT. In my example I'll use x86 for simplicity. Both checks generate almost the same native code with the exception of the instruction for the comparison (`setne` for != and `setg` for >). So it's just the difference of both of these instruction. I think it is fair to say that both instruction are pretty close to each other because if they weren't, someone would probably already have modified the JIT to use the fastest. Also, given the speed of modern processor, I wouldn't bother comparing two of these similar instruction because the difference is likely to small to be even measurable (Probably not 156 ms vs 132 ms).

I would also recommend using BenchmarkDotNet to do micro-benchmark like this, instead of using LinqPad because the actual code that you want to compare gets watered down by other code that are way slower which ends up corrupting your results.

Use sharplab.io to check native code:
https://sharplab.io/#v2:C4LghgzgtgPgAgJgIwFgBQcDMACR2DC2A3utmbjgEYD21ANtgHLXACiAjgK5h0AUcSAAzYAbj04BTAJSlyJNOUW4A7KPESAdABkJAOwDmwABbYAhAF5sggNyyyAXzvYnWbDXrYA4gCcJYYBLeACpGYLr8Qmp0kjIKck6KcKpi0Zo6BsbYAHxWtnEO6PZAA==

src/Microsoft.DotNet.Wpf/src/WindowsBase/System/IO/Packaging/PackageDigitalSignatureManager.cs
```diff
@@ -470,7 +470,7 @@ public PackageDigitalSignature Sign(IEnumerable<Uri> parts, X509Certificate cert
             VerifySignArguments(parts, certificate, relationshipSelectors, signatureId, signatureObjects, objectReferences);
 
             // substitute default id if none given
-            if ((signatureId == null) || (signatureId == String.Empty))
+            if (String.IsNullOrEmpty(signatureId))
```
RussKie, line 473: It has to be `string` and not `String`
```suggestion
            if (string.IsNullOrEmpty(signatureId))
```

Applies universally.

src/Microsoft.DotNet.Wpf/src/WindowsBase/System/IO/Packaging/PackageDigitalSignatureManager.cs
```diff
@@ -1006,7 +1006,7 @@ private void VerifyPartsExist(IEnumerable<Uri> parts)
             }
 
             // ensure id is legal Xml id
-            if ((signatureId != null) && (signatureId != String.Empty))
+            if (!String.IsNullOrEmpty(signatureId))
```
Youssef1313, line 1009:
```suggestion
            if (!string.IsNullOrEmpty(signatureId))
```

src/Microsoft.DotNet.Wpf/src/WindowsBase/System/Windows/NameScope.cs
```diff
@@ -118,7 +118,7 @@ public void UnregisterName(string name)
         /// <returns>corresponding Context if found, else null</returns>
         public object FindName(string name)
         {
-            if (_nameMap == null || name == null || name == String.Empty)
+            if (_nameMap == null || String.IsNullOrEmpty(name))
```
Youssef1313, line 121:
```suggestion
            if (_nameMap == null || string.IsNullOrEmpty(name))
```

RussKie (approved):

Youssef1313:
@RussKie can we get this merged? Thanks!

RussKie:
I'm afraid I can't do that as I'm not a member of WPF team.
________________________________

Youssef1313:
@RussKie Do you know who can give a second review and merge?

singhashish-wpf:
@Youssef1313 This is under review by @dotnet/wpf-developers

dipeshmsft (approved):

dipeshmsft:
@Youssef1313 , thank you for your contribution.

PR #3009, opened 2020-05-17
Remove always-true condition

Youssef1313:


src/Microsoft.DotNet.Wpf/src/WindowsBase/System/Windows/Input/KeyConverter.cs
```diff
@@ -142,7 +142,7 @@ public override object ConvertTo(ITypeDescriptorContext context, CultureInfo cul
 
         private object GetKey(string keyToken, CultureInfo culture)
         {
-            if (keyToken == String.Empty)
+            if (keyToken.Length == 0)
```
lindexi, line 145: NullReferenceException?

Could you use string.IsNullOrEmpty ?

lindexi:
Thank you @Youssef1313 

And I think the `if (keyToken.Length == 0)` code doesn't match the subject. Your subject is `Remove always-true condition` but this code is  to improve performance.

src/Microsoft.DotNet.Wpf/src/WindowsBase/System/Windows/Input/KeyConverter.cs
Youssef1313: The caller code is:

```
                string fullName = ((string)source).Trim();
                object key = GetKey(fullName, CultureInfo.InvariantCulture);
```

Trim won't return null, so `fullName` is going to be a non-null value.

Youssef1313: If you looked at the old code and traced it if the passed `keyToken` is null, You'll find that it will throw NullReferenceException at:

```
                keyToken = keyToken.ToUpper(culture);
```

But, anyway the passed value isn't null. So, it's okay.

Youssef1313:
> Your subject is Remove always-true condition but this code is to improve performance.

That's correct. I just did both because they're in the same file and the change is really small.

ryalanms (approved):

ryalanms:
Thanks, @Youssef1313.

Issue #96, opened 2018-12-05
Allow registering xmlns prefixes for foreign assemblies


taori:
As pointed out in https://github.com/dotnet/wpf/issues/45, using

```cs
[assembly: XmlnsPrefix("http://schemas.dotMorten.com/2018", "custom")]
[assembly: XmlnsDefinition("http://schemas.dotMorten.com/2018", "MyClassNamespace")]
```

only works for assemblies you own. 

It would be nice if you could declare those attributes at application level, to provide controls of an assembly with custom namespaces, if they lack those attributes, or even override them. 

While the later one isn't essential, i think the first one could be useful, if people share their controls on nuget and don't know about custom namespace prefixes.

At least i've had situations where i was annoyed with the awfully verbose prefixes others chose, or the lack of them alltogether.

Another scenario would be that you are supplied with a viewmodels assembly and are going to use those in your views. Being able to group those namespaces would be neat.

mstrobel:
The prefixes are only suggestions, and you can always substitute your own at the import site.  Being able to add `XmlnsDefinition` mappings for external assemblies is the far more interesting part of this proposal.

Interestingly, `XmlnsDefinitionAttribute` already has an `Assembly` property, but as far as I can tell, `System.Xaml` simply ignores it.  Perhaps this is a capability that Microsoft intended to implement, but simply didn’t.

mstrobel:
The prefixes are only suggestions, and you can always substitute your own at the import site.  Being able to add `XmlnsDefinition` mappings for external assemblies is the far more interesting part of this proposal.

Interestingly, `XmlnsDefinitionAttribute` already has an `Assembly` property, but as far as I can tell, `System.Xaml` simply ignores it.  Perhaps this is a capability that Microsoft intended to implement, but simply didn’t.

rrelyea:
From: Rob Relyea
Sent: Tuesday, January 19, 2010 8:50 AM
To: Mark;Michael
Cc: David
Subject: RE: Is the AssemblyName property on XmlnsDefinitionAttribute supposed to have any effect?

Yes, that property is kindof a noop…that we discovered late in the game.
The idea was to let WindowsFormsIntegration.dll provide an XmlnsDefinitionAttribute for v2 windorms in v3.
 
No scenarios have been blocked with the noop.

dotMorten:
[...]

Issue #1, opened 2018-11-02
Use nameof(SRID.PropertyName) syntax instead of depending on GenerateResxSource.GenerateResourcesCodeAsConstants


vatsan-madhavan:
The codebase currently uses `SRID.PropertyName` syntax pervasively. This is supplied by an Arcade build task [GenerateResxSource](https://github.com/dotnet/arcade/blob/946bdc7e51961d64a22a00d1d3613816630f4386/src/Microsoft.DotNet.Arcade.Sdk/src/GenerateResxSource.cs) to which I have added an additional support for this particular case via `GenerateResourcesCodeAsConstants`. 

There are ~7k instances across the codebase, and ~1k instances of these are likely coming from generated code. Once we have all the code generators ported over to .NET Core, we need to clean up everything and switch them over to use `nameof(SRID.PropertyName)` syntax.

PR #3066, opened 2020-05-29
Issue 96: Use AssemblyName property in XmlnsDefinitionAttribute

dotMorten:
I believe this addresses the issue mentioned in #96, by using the `AssemblyName` property on `XmlnsDefinitionAttribute` which has always been a no-op property and not worked right.

I've been able to test the parsing at runtime, but [the readme still lacks documentation](https://github.com/dotnet/wpf/blob/master/Documentation/developer-guide.md#testing-presentationbuildtasks) on how to test `PresentationBuildTasks`.

Example:
```cs
[assembly: System.Windows.Markup.XmlnsPrefix("http://schemas.xaml.dev/dotMorten/2020", "sysText")]
[assembly: System.Windows.Markup.XmlnsDefinition("http://schemas.xaml.dev/dotMorten/2020", "System.Text", AssemblyName = "System.Runtime")]

// ...

StringReader stringReader = new StringReader("<Grid xmlns=\"http://schemas.microsoft.com/winfx/2006/xaml/presentation\" xmlns:x=\"http://schemas.microsoft.com/winfx/2006/xaml\" xmlns:sysText=\"http://schemas.xaml.dev/dotMorten/2020\" ><Grid.Resources><sysText:Rune x:Key=\"rune\" /></Grid.Resources></Grid>");
XmlReader xmlReader = XmlReader.Create(stringReader);
Grid grid = (Grid)System.Windows.Markup.XamlReader.Load(xmlReader);

```

This PR comes with a "Works On My Machine" guarantee :-)
image

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/System.Xaml/System/Xaml/MS/Impl/XmlNsInfo.cs
+++ b/src/Microsoft.DotNet.Wpf/src/System.Xaml/System/Xaml/MS/Impl/XmlNsInfo.cs
@@ -203,7 +203,7 @@ namespace System.Xaml.MS.Impl
         {
             if (String.IsNullOrEmpty(xmlns) || clrns == null)
             {
-                throw new XamlSchemaException(SR.Get(SRID.BadXmlnsDefinition, assemblyName));
+                throw new XamlSchemaException(SR.Format(SR.BadXmlnsDefinition, assemblyName));
             }
 
             result.Add(new XmlNsDefinition { ClrNamespace = clrns, XmlNamespace = xmlns, AssemblyName = assemblyName });


src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Markup/XmlnsCache.cs
```diff
@@ -314,8 +325,8 @@ private void LoadClrnsToAssemblyNameMappingCache(Assembly[] asmList)
                 {
                     string xmlns = null;
                     string clrns = null;
-
-                    GetNamespacesFromDefinitionAttr(attributes[attrIdx], out xmlns, out clrns);
+                    string assmname = null;
```
miloush, line 328: That is a confusing name, I suggest `customAssemblyName` or `assemblyNameOverride` or similar

src/Microsoft.DotNet.Wpf/src/System.Xaml/System/Xaml/XamlSchemaContext.cs
```diff
@@ -1176,7 +1176,14 @@ bool UpdateNamespaceByUriList(XmlNsInfo nsInfo)
             IList<XmlNsInfo.XmlNsDefinition> xmlnsDefs = nsInfo.NsDefs;
             foreach (XmlNsInfo.XmlNsDefinition xmlnsDef in xmlnsDefs)
             {
-                AssemblyNamespacePair pair = new AssemblyNamespacePair(nsInfo.Assembly, xmlnsDef.ClrNamespace);
+                Assembly assembly = nsInfo.Assembly;
+                if (!string.IsNullOrEmpty(xmlnsDef.AssemblyName))
+                {
+                    assembly = ResolveAssembly(xmlnsDef.AssemblyName);
+                    if(assembly == null)
```
miloush, line 1183: space

src/Microsoft.DotNet.Wpf/src/System.Xaml/System/Xaml/XamlSchemaContext.cs
```diff
@@ -1176,7 +1176,14 @@ bool UpdateNamespaceByUriList(XmlNsInfo nsInfo)
...
+                        return false;
```
miloush, line 1184: Is this the correct action here rather than `continue` or `throw`?

src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Markup/XmlnsCache.cs
```diff
@@ -253,6 +254,15 @@ private void GetNamespacesFromDefinitionAttr(CustomAttributeData data, out strin
                 else
                     throw new ArgumentException(SR.Get(SRID.ParserAttributeArgsHigh, "XmlnsDefinitionAttribute"));
             }
+            IList<CustomAttributeNamedArgument> namedArguments = data.NamedArguments;
+            for (int i = 0; i < namedArguments.Count; i++)
+            {
+                var namedArgument = namedArguments[i];
+                if (namedArgument.MemberName == "AssemblyName")
```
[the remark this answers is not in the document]
dotMorten: Yes... I almost did that too, but was trying to follow the pattern of current code - I don't really see the use of nameof much in the codebase, if at all.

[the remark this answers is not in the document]
vatsan-madhavan: Most of the codebase predates nameof, that’s all. It’s not a deliberate convention to avoid use of it etc.. Newer code does use it. 

Use of nameof should be fine except where translatable strings are involved (they typically appear as SRID.Name).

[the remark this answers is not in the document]
dotMorten: There's another issue here though: You can only use nameof(...) for types that are referenced, but the `XmlnsDefinitionAttribute` type isn't referenced in this project.

src/Microsoft.DotNet.Wpf/src/System.Xaml/System/Xaml/MS/Impl/XmlNsInfo.cs
```diff
@@ -191,6 +199,16 @@ void LoadNsDefHelper(IList<XmlNsDefinition> result, string xmlns, string clrns,
             result.Add(new XmlNsDefinition { ClrNamespace = clrns, XmlNamespace = xmlns });
         }
 
+        void LoadNsDefHelper(IList<XmlNsDefinition> result, string xmlns, string clrns, string assemblyName)
+        {
+            if (String.IsNullOrEmpty(xmlns) || clrns == null)
+            {
+                throw new XamlSchemaException(SR.Get(SRID.BadXmlnsDefinition, assemblyName));
```
ThomasGoulet73, line 206: Hey @dotMorten, since you pulled main, your PR build is now breaking here which is caused by my changes as part of dotnet/wpf#1. SRID is being replaced with SR which contains the same properties but they now return the resource instead of the resource name. For now, only a couple of projects uses SR instead of SRID, here's the list: https://github.com/dotnet/wpf/blob/main/eng/WpfArcadeSdk/tools/SystemResources.props#L15-L19.

```suggestion
                throw new XamlSchemaException(SR.Format(SR.BadXmlnsDefinition, assemblyName));
```

dotMorten: Nice catch! Thanks!

ThomasGoulet73:
> I've been able to test the parsing at runtime, but [the readme still lacks documentation](https://github.com/dotnet/wpf/blob/master/Documentation/developer-guide.md#testing-presentationbuildtasks) on how to test `PresentationBuildTasks`.

When I want to test PresentationBuildTasks, I usually do a normal build and in the project I want to test, I override [_PresentationBuildTasksAssembly](https://github.com/dotnet/wpf/blob/main/src/Microsoft.DotNet.Wpf/src/PresentationBuildTasks/Microsoft.WinFX.targets#L6) with a path pointing to the assembly built in the artifacts folder. This is the simplest way to test it if you don't have API changes to an MSBuild task, which you don't.

dotMorten:
@dipeshmsft  any chance you could have a look at this one?

PR #3078, opened 2020-06-02
Update the docs link

lindexi:


lindexi:
@vatsan-madhavan Thank you.  Could you help me create the aka.ms type link? I don't have access to create it

vatsan-madhavan:
>Could you help me create the aka.ms type link? I don't have access to create it

You wouldn't but MSFT folks can. But creating an aka.ms link doesn't help in itself. _Maintaining_ the link is where the value comes. Usually, the owners of the target-URL are the ownes vested in maintaining these short URL's. 

Links that use the go.microsoft.com/fwlink/?LinkId= system, for e.g., are permalinks maintained by the respective owners or docs teams that automatically forwards to the right language based on detected browser locale etc. In general, wherever possible, using locale agnostic permalinks are preferable, if they can be found. 

If one of us creates an aka.ms or go.microsoft.com link newly for docs/typography, then it would suffer from the original problem. When the target url becomes defunct (sometime in the future, hypotheticallly), the short-url would also become useless (and would have to be updated). But if we could discover a pre-existing fwLink URL for docs/typography that's already being maintained by docs folks (or the typography folks etc.), then that would be more suitable for long-term use in this type of a PR. 

All that said, its probably not worth slowing down/blocking too much here. If @fabiant3 can find a substitute easily, then great. Otherwise I think what you have here is pretty good and we should just merge it 👍

fabiant3:
Agreed with Vatsan here, this is good enough, I'm merging this in. Congrats on this pull request @lindexi .

PR #3084, opened 2020-06-03
Fix unknown character

lindexi:


ryalanms (approved):

miloush:
I just want to point out that what needed to be fixed here is the encoding of the files.

The characters were valid quotation marks in 1252 codepage:
image

(so this might potentially break code that parses the error messages, although I agree that should not be encouraged)

lindexi:
@miloush Thank you. But I do not think we should use this character .

ryalanms:
We can revert this for now and reconsider for Preview 2.  Any objections?

lindexi:
@ryalanms Thank you. Actually I found out that this problem was caused by garbled characters appearing on my device

miloush:
I only wanted to point out that at least some of these were not some randomly garbled characters the way GitHub makes it look like. They don't seem to be applied consistently though (i.e. for specific languages) and also, after further checking, the XML files contain UTF-8 BOM which makes them mis-encoded, so it's just the txt file that had a reasonably working characters.

Furthermore, these are all `ObjectReaderTypeCannotRoundtrip` exception messages only which I find unlikely to be thrown for most application users, so if you are asking us to asses the risks, I would lean towards keeping this merged in and see if someone complains, especially if it's a preview.

PR #3085, opened 2020-06-03
Reduce memory usage by Array.Empty

lindexi:
The Empty property are only used in TypeReferences

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/System.Xaml/System/Windows/Markup/ValueSerializer.cs
+++ b/src/Microsoft.DotNet.Wpf/src/System.Xaml/System/Windows/Markup/ValueSerializer.cs
@@ -92,8 +92,6 @@ namespace System.Windows.Markup
             throw GetConvertFromException(value);
         }
 
-        static List<Type> Empty = new List<Type>();
-
         /// <summary>
         /// Returns an enumeration of the types referenced by the value serializer. If the value serializer asks for
         /// a value serializer for System.Type, any types it asks to convert should be supplied in the returned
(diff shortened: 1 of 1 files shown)


miloush:
Note that this is a public API now returning a fixed array instead of a list, though an array seems more appropriate here.

There is also `System.Windows.Markup.Primitives.ElementPropertyBase.EmptyTypes`.

Rand-Random:
The title totally got my expectations high, but the actual code change destroyed my expectations pretty fast.

vatsan-madhavan:
I was just looking up `List` yesterday for something else and noticed that `new List()` seems to have an [optimization](https://github.com/dotnet/corefx/blob/e99ec129cfd594d53f4390bf97d1d736cff6f860/src/Common/src/CoreLib/System/Collections/Generic/List.cs#L40) similar to `Array.Empty()` in that it returns an empty pre-constructed array by default.

```C#
            _items = s_emptyArray;
```

In situations where an empty `List` is converted again into an `IReadOnlyList` etc. that occurs [another object allocation](https://github.com/dotnet/corefx/blob/e99ec129cfd594d53f4390bf97d1d736cff6f860/src/Common/src/CoreLib/System/Collections/Generic/List.cs#L247), the slight improvement of `Array.Empty()` is clearer.   

```C#
        public ReadOnlyCollection<T> AsReadOnly()
            => new ReadOnlyCollection<T>(this);
```

But straight-up swapping of an empty `List` with `Array.Empty()` (esp. for cases where the desired interface type is implemented by either of those, like `IEnumerable`) - this I'm not sure I understand the benefit of.

lindexi:
@ vatsan-madhavan Yeah, but we create a static field will cause unnecessary memory allocation, although there is only very little memory.  

- a static field
- a List object
  - object meta
  - `_items` field
  - `_size` field
  - `_version` field

vatsan-madhavan:
Makes sense.

ryalanms (approved):

PR #3092, opened 2020-06-05
Remove dependency on built-in WinRT support

jkoritzinsky:
Remove dependency on built-in WinRT support by copying over portions of C#/WinRT and slimming down the projected interfaces.

Ryland and I did ad-hoc testing of the spell checker component and it worked.

Once we believe this is ready for merge, I'll cherry-pick it for release/5.0-preview6.

cc: @rladuca @fabiant3 @ryalanms @AaronRobinsonMSFT @jeffschwMSFT


ryalanms:
Thanks @jkoritzinsky for the fast response!

src/Microsoft.DotNet.Wpf/src/Themes/PresentationFramework.Aero2/PresentationFramework.Aero2_dacxb2zy_wpftmp.csproj
```diff
@@ -0,0 +1,107 @@
+<!-- PresentationFramework.Aero2 -->
+<Project Sdk="Microsoft.NET.Sdk">
+  <PropertyGroup>
+    <TargetFramework>netcoreapp5.0</TargetFramework>
```
AaronRobinsonMSFT, line 4: I think this should be `net5.0`.

src/Microsoft.DotNet.Wpf/src/Themes/PresentationFramework.Aero2/PresentationFramework.Aero2_dacxb2zy_wpftmp.csproj
```diff
@@ -0,0 +1,107 @@
...
+    <NetCoreReference Include="System.Security.Cryptography.X509Certificates" />
+    <NetCoreReference Include="System.Threading" />
+    <NetCoreReference Include="System.Threading.Thread" />
+    <NetCoreReference Include="System.Threading.ThreadPool" />
+    <!-- Additional references beyond the minimal set for PresentationFramework.Aero2 -->
+    <NetCoreReference Include="System.ComponentModel.TypeConverter" />
+    <NetCoreReference Include="System.Collections.Concurrent" />
+  </ItemGroup>
+  <ItemGroup>
+    <ReferencePath Include="C:\Users\jekoritz\.nuget\packages\microsoft.netcore.app.ref\5.0.0-preview.6.20279.7\ref\net5.0\mscorlib.dll" />
```
AaronRobinsonMSFT, line 74: This seems bad...

src/Microsoft.DotNet.Wpf/src/PresentationCore/MS/internal/WindowsRuntime/Windows/UI/ViewManagement/InputPane.cs
```diff
@@ -62,21 +57,26 @@ internal class InputPane : IDisposable
             [System.Diagnostics.CodeAnalysis.SuppressMessage("Microsoft.Design", "CA1031:DoNotCatchGeneralExceptionTypes")]
             static InputPane()
             {
+                // WinRT types are only available on Win10+
+                if (Environment.OSVersion.Platform == PlatformID.Win32NT && Environment.OSVersion.Version < new Version(10, 0))
```
lindexi, line 61: Can it check the win10 system version?

> For applications that have been manifested for Windows 8.1 or Windows 10. Applications not manifested for Windows 8.1 or Windows 10 will return the Windows 8 OS version value (6.2). To manifest your applications for Windows 8.1 or Windows 10, refer to Targeting your application for Windows.

See https://docs.microsoft.com/en-us/windows/win32/sysinfo/operating-system-version?wt.mc_id=MVP

AaronRobinsonMSFT:
A few additional notes.

- Banners on the generated code?
- Thoughts on how this is reconciled with an official CsWinRT generated assembly? (e.g. function pointers, new assembly, code integration into build process, etc).
- The `DllImport` all seem to point at api-sets. Are they known to exist down to supported OS's? (e.g. 8.1?). I want to make sure we don't break some version support with this approach.

src/Microsoft.DotNet.Wpf/src/Themes/PresentationFramework.Aero2/PresentationFramework.Aero2_dacxb2zy_wpftmp.csproj
jkoritzinsky: This is a generated file I didn't mean to commit. I'll remove it.

src/Microsoft.DotNet.Wpf/src/PresentationCore/MS/internal/WindowsRuntime/Windows/UI/ViewManagement/InputPane.cs
jkoritzinsky: Yes, `Environment.OSVersion` was recently updated to get the real OS version, not the one calculated based on the manifest.

src/Microsoft.DotNet.Wpf/src/PresentationFramework/MS/Internal/WindowsRuntime/Generated/WinRT.cs
```diff
@@ -0,0 +1,518 @@
...
+                    return value;
+                }
+            }
+        }
+    }
+
+    internal class WinrtModule
+    {
+        readonly IntPtr _mtaCookie;
+        static Lazy<WinrtModule> _instance = new Lazy<WinrtModule>();
```
lindexi, line 207: Maybe this field can be set to readonly

jkoritzinsky: This is part of the generated code. I'd prefer to limit stylistic changes while we get everything else in.

src/Microsoft.DotNet.Wpf/src/PresentationFramework/MS/Internal/WindowsRuntime/Generated/WinRT/ComWrappersSupport.cs
```diff
@@ -0,0 +1,187 @@
...
+
+            return (
+                new InspectableInfo(type, iids),
+                interfaceTableEntries);
+        }
+
+        internal class InspectableInfo
+        {
+            public Guid[] IIDs { get; }
+            // WPF's usage of WinRT doesn't create any ccws, so we can return an empty string.
```
lindexi, line 177: What is the `ccws`?

src/Microsoft.DotNet.Wpf/src/PresentationFramework/MS/Internal/WindowsRuntime/Generated/WinRT/ComWrappersSupport.cs
```diff
@@ -0,0 +1,187 @@
...
+
+            Type type = obj.GetType();
+
+            if (type.FullName.StartsWith("ABI."))
+            {
+                type = Projections.FindCustomPublicTypeForAbiType(type) ?? type.Assembly.GetType(type.FullName.Substring("ABI.".Length)) ?? type;
+            }
+
+            return (
+                new InspectableInfo(type, iids),
```
lindexi, line 170: In fact, we did not use the type variable. So is it possible to remove the code that creates the type variable

src/Microsoft.DotNet.Wpf/src/PresentationFramework/MS/Internal/WindowsRuntime/Generated/WinRT/ComWrappersSupport.net5.cs
```diff
@@ -0,0 +1,156 @@
...
+{
+    internal static partial class ComWrappersSupport
+    {
+        public static readonly ConditionalWeakTable<object, InspectableInfo> InspectableInfoTable = new ConditionalWeakTable<object, InspectableInfo>();
+
+        private static ComWrappers ComWrappers { get; } = new WpfWinRTComWrappers();
+
+        public static unsafe InspectableInfo GetInspectableInfo(IntPtr pThis)
+        {
+            var _this = FindObject<object>(pThis);
```
lindexi, line 21: Maybe we can use `@this` 

```suggestion
            var @this = FindObject<object>(pThis);
```

jkoritzinsky:
> A few additional notes.
> 
> * Banners on the generated code?

I can add these for the files I copied over if you prefer.

> * Thoughts on how this is reconciled with an official CsWinRT generated assembly? (e.g. function pointers, new assembly, code integration into build process, etc).

Once the Windows SDK is shipping along with .NET as part of targeting `net5.0-windows`, WPF should be able to take a dependency on the SDK's projection directly and delete all of this code. The only part they need to keep is the custom-defined `WordsSegmenter.Create` method. In the meantime, I don't know if it's worth updating the generated code each time something changes. The previous implmentation was extremely reflection-heavy (every method call was done via reflection), so even if we don't update this implementation with function pointers support when it becomes available, I would be shocked if it performed worse than the previous implementation.

> * The `DllImport` all seem to point at api-sets. Are they known to exist down to supported OSs? (e.g. 8.1?). I want to make sure we don't break some version support with this approach.

Yes, the API sets exist on downlevel OSes. The ones I'm using were introduced on Win8.

src/Microsoft.DotNet.Wpf/src/PresentationFramework/MS/Internal/WindowsRuntime/Generated/WinRT/ComWrappersSupport.cs
jkoritzinsky: Com callable wrapper. Basically WPF doesn't pass any managed objects to native, so we don't need full support here.

src/Microsoft.DotNet.Wpf/src/PresentationFramework/MS/Internal/WindowsRuntime/Generated/WinRT/ComWrappersSupport.net5.cs
```diff
@@ -0,0 +1,156 @@
...
+
+        static partial void PlatformSpecificInitialize()
+        {
+            IUnknownVftbl = WpfWinRTComWrappers.IUnknownVftbl;
+        }
+    }
+
+    internal class WpfWinRTComWrappers : ComWrappers
+    {
+        private static ConditionalWeakTable<object, VtableEntriesCleanupScout> ComInterfaceEntryCleanupTable = new ConditionalWeakTable<object, VtableEntriesCleanupScout>();
```
lindexi, line 58: Maybe this field can be set to readonly

src/Microsoft.DotNet.Wpf/src/PresentationFramework/MS/Internal/WindowsRuntime/Generated/WinRT/ComWrappersSupport.net5.cs
jkoritzinsky: As mentioned previously, I'd rather avoid making style changes to the generated or imported code unless absolutely required. Keeping it the same when possible makes it easier to maintain.

src/Microsoft.DotNet.Wpf/src/PresentationFramework/MS/Internal/WindowsRuntime/Generated/WinRT/EventRegistrationToken.cs
```diff
@@ -0,0 +1,43 @@
...
+using System.Collections.Generic;
+using System.Text;
+
+namespace WinRT
+{
+    internal struct EventRegistrationToken : IEquatable<EventRegistrationToken>
+    {
+        public long Value;
+
+        public EventRegistrationToken(long _Value)
```
lindexi: It seems that the variable `_Value` naming does not meet the specifications.

AaronRobinsonMSFT:
@jkoritzinsky All my concerns are address as long as the WPF team is okay. Thank you. I concur with your sentiment about not making any style changes to the generated code. Especially since this code is autogenerated and should be mostly deleted as soon as WPF can directly consume the CsWinRT solution.

src/Microsoft.DotNet.Wpf/src/PresentationFramework/MS/Internal/WindowsRuntime/Generated/WinRT/ComWrappersSupport.net5.cs
AaronRobinsonMSFT: @lindexi This is all generated code and intended to be removed as soon as the CsWinRT tool can be consumed. I agree with @jkoritzinsky that updating this with stylistic changes doesn't help actually address the reason for this change.

src/Microsoft.DotNet.Wpf/src/PresentationFramework/MS/Internal/WindowsRuntime/Generated/WinRT/ExceptionHelpers.cs
```diff
@@ -0,0 +1,396 @@
...
+
+            return e;
+        }
+    }
+
+    internal static class ExceptionExtensions
+    {
+        public static void SetHResult(this Exception ex, int value)
+        {
+            ex.GetType().GetProperty("HResult").SetValue(ex, value);
```
lindexi, line 353: Why do we need to use reflection technology?  We can set this property.

ex.HResult = value;

See https://docs.microsoft.com/en-us/dotnet/api/system.exception.hresult?wt.mc_id=MVP

jkoritzinsky: This is copied from C#/WinRT, which targets .NET Standard 2.0 (the setter isn't public in .NET Standard 2.0). This code is currently unused here anyway, so I don't want to change it unless someone from the WPF team asks for it.

src/Microsoft.DotNet.Wpf/src/PresentationFramework/MS/Internal/WindowsRuntime/Generated/WinRT/ExceptionHelpers.cs
```diff
@@ -0,0 +1,396 @@
...
+        }
+
+        internal static Exception GetExceptionForHR(this Exception innerException, int hresult, string messageResource)
```
lindexi, line 356: I recommend this parameter to be named `hResult`

```suggestion
        internal static Exception GetExceptionForHR(this Exception innerException, int hResult, string messageResource)
```

rladuca:
@vatsan-madhavan FYI

src/Microsoft.DotNet.Wpf/src/PresentationFramework/MS/Internal/WindowsRuntime/Generated/WinRT/ExceptionHelpers.cs
```diff
@@ -0,0 +1,396 @@
...
+            e.SetHResult(hresult);
+            return e;
+        }
+    }
+
+    internal class ErrorStrings
+    {
+        internal static string Format(string format, params object[] args) => String.Format(format, args);
+
+        internal static readonly string Arg_IndexOutOfRangeException = "Index was outside the bounds of the array.";
```
lindexi, line 377: Does the code here need to consider multi-language support. Such as Chinese, Japanese and so on

src/Microsoft.DotNet.Wpf/src/PresentationFramework/MS/Internal/WindowsRuntime/Generated/WinRT/GuidGenerator.cs
```diff
@@ -0,0 +1,143 @@
...
+
+            if (type.IsDelegate())
+            {
+                return "delegate({" + GetGUID(type) + "})";
+            }
+
+            return "{" + type.GUID.ToString() + "}";
+        }
+
+        private static Guid encode_guid(byte[] data)
```
lindexi, line 101: The naming of this method does not seem to conform to the code specification

src/Microsoft.DotNet.Wpf/src/PresentationFramework/MS/Internal/WindowsRuntime/Generated/WinRT/ExceptionHelpers.cs
jkoritzinsky: This code never considered multi-language support when it used to live in CoreCLR, so I don't believe it needs to here. Someone on the WPF team can correct me if that is not the case.

src/Microsoft.DotNet.Wpf/src/PresentationFramework/MS/Internal/WindowsRuntime/Generated/WinRT/GuidGenerator.cs
```diff
@@ -0,0 +1,143 @@
...
+
+            if (type == typeof(object))
+            {
+                return "cinterface(IInspectable)";
+            }
+
+            if (type.IsGenericType)
+            {
+                var args = type.GetGenericArguments().Select(t => GetSignature(t));
+                return "pinterface({" + GetGUID(type) + "};" + String.Join(";", args) + ")";
```
lindexi, line 46: I recommend using `string` instead of `String`

In [runtime/coding-style.md at master · dotnet/runtime](https://github.com/dotnet/runtime/blob/master/docs/coding-guidelines/coding-style.md ) NO.11, we use language keywords instead of BCL types

src/Microsoft.DotNet.Wpf/src/PresentationFramework/MS/Internal/WindowsRuntime/Generated/WinRT/GuidGenerator.cs
```diff
@@ -0,0 +1,143 @@
...
+                t = data[6];
+                data[6] = data[7];
+                data[7] = (byte)((t & 0x0f) | (5 << 4));
+                // encode rfc clock/reserved field
+                data[8] = (byte)((data[8] & 0x3f) | 0x80);
+            }
+            return new Guid(data.Take(16).ToArray());
+        }
+
+        private static Guid wrt_pinterface_namespace = new Guid("d57af411-737b-c042-abae-878b1e16adee");
```
lindexi, line 126: The naming of this field does not seem to conform to the code specification. And this field may be set to readonly.

See [runtime/coding-style.md at master · dotnet/runtime](https://github.com/dotnet/runtime/blob/master/docs/coding-guidelines/coding-style.md ) NO.3, we use `_camelCase` for internal and private fields.

lindexi:
@jkoritzinsky Thank you and sorry, I didn't notice this is the generated code

src/Microsoft.DotNet.Wpf/src/PresentationFramework/MS/Internal/WindowsRuntime/Generated/WinRT/ComWrappersSupport.cs
lindexi: @jkoritzinsky Thank you.

src/Microsoft.DotNet.Wpf/src/PresentationFramework/MS/Internal/WindowsRuntime/Generated/WinRT/Projections.cs
```diff
@@ -0,0 +1,232 @@
...
+                    return true;
+                }
+                return false;
+            }
+            return CustomTypeToAbiTypeNameMappings.ContainsKey(type)
+                || type.IsPrimitive
+                || type == typeof(string)
+                || type == typeof(Guid)
+                || type == typeof(object)
+                || type.GetCustomAttribute<WindowsRuntimeTypeAttribute>() is object;
```
lindexi, line 137: Is the code `type.GetCustomAttribute<WindowsRuntimeTypeAttribute>() is object` here to determine whether this type has added WindowsRuntimeTypeAttribute

I think we can replace `type.IsDefined<WindowsRuntimeTypeAttribute>` to `type.GetCustomAttribute<WindowsRuntimeTypeAttribute>() is object` to improve performance

src/Microsoft.DotNet.Wpf/src/Themes/PresentationFramework.Aero2/PresentationFramework.Aero2_dacxb2zy_wpftmp.csproj
```diff
@@ -0,0 +1,107 @@
...
+    <ReferencePath Include="C:\Users\jekoritz\.nuget\packages\microsoft.netcore.app.ref\5.0.0-preview.6.20279.7\ref\net5.0\System.Security.Cryptography.Primitives.dll" />
+    <ReferencePath Include="C:\Users\jekoritz\.nuget\packages\microsoft.netcore.app.ref\5.0.0-preview.6.20279.7\ref\net5.0\System.Security.Cryptography.X509Certificates.dll" />
+    <ReferencePath Include="C:\Users\jekoritz\.nuget\packages\microsoft.netcore.app.ref\5.0.0-preview.6.20279.7\ref\net5.0\System.Threading.dll" />
+    <ReferencePath Include="C:\Users\jekoritz\.nuget\packages\microsoft.netcore.app.ref\5.0.0-preview.6.20279.7\ref\net5.0\System.Threading.Thread.dll" />
+    <ReferencePath Include="C:\Users\jekoritz\.nuget\packages\microsoft.netcore.app.ref\5.0.0-preview.6.20279.7\ref\net5.0\System.Threading.ThreadPool.dll" />
+    <ReferencePath Include="D:\source\wpf\artifacts\bin\PresentationUI\Debug\netcoreapp5.0\PresentationUI.dll" />
+    <ReferencePath Include="D:\source\wpf\artifacts\bin\System.Xaml\Debug\netcoreapp5.0\System.Xaml.dll" />
+    <ReferencePath Include="D:\source\wpf\artifacts\bin\WindowsBase\Debug\netcoreapp5.0\WindowsBase.dll" />
```
lindexi: This file should not be uploaded because it is a generated file

src/Microsoft.DotNet.Wpf/src/PresentationCore/MS/internal/WindowsRuntime/Windows/UI/ViewManagement/InputPane.cs
vatsan-madhavan: In WPF code base, we do more precise version checks that overcomes appcompat manifest lies. Environment.OSVersion should not be used. 

Use this instead https://github.com/dotnet/wpf/blob/8c7ab1592b6c70e732106aeabf58f215d4b71c22/src/Microsoft.DotNet.Wpf/src/Shared/System/Windows/InterOp/OSVersionHelper.cs

src/Microsoft.DotNet.Wpf/src/PresentationCore/MS/internal/WindowsRuntime/Windows/UI/ViewManagement/InputPaneRcw.cs
```diff
@@ -19,6 +19,36 @@ namespace Windows.UI.ViewManagement
         /// </summary>
         internal static class InputPaneRcw
         {
+            [DllImport("api-ms-win-core-winrt-string-l1-1-0.dll", CallingConvention = CallingConvention.StdCall)]
```
vatsan-madhavan, line 22: DllImport definitions should be moved over to common files. They are named xxxNativeMethods.cs.

src/Microsoft.DotNet.Wpf/src/PresentationCore/MS/internal/WindowsRuntime/Windows/UI/ViewManagement/InputPaneRcw.cs
```diff
@@ -19,6 +19,36 @@ namespace Windows.UI.ViewManagement
...
+            private static extern unsafe int WindowsCreateString([MarshalAs(UnmanagedType.LPWStr)] string sourceString,
+                                                  int length,
+                                                  out IntPtr hstring);
+
+            [DllImport("api-ms-win-core-winrt-string-l1-1-0.dll", CallingConvention = CallingConvention.StdCall)]
```
vatsan-madhavan, line 27: Assembly names used in DllImport attributes should be defined as costants in a common file. 

https://github.com/dotnet/wpf/blob/ac9d1b7a6b0ee7c44fd2875a1174b820b3940619/src/Microsoft.DotNet.Wpf/src/Shared/RefAssemblyAttrs.cs

src/Microsoft.DotNet.Wpf/src/PresentationFramework/MS/Internal/WindowsRuntime/Generated/WinRT.cs
```diff
@@ -0,0 +1,289 @@
+// This file was generated by cswinrt.exe version 0.0.0-private.0
```
vatsan-madhavan, line 1: - Add copyright
- if it is generated then can it be generated during build and consumed from OBJ dir? Checking in a generated file should be avoided if possible. 
- describe generation steps in comments in more detail if possible

vatsan-madhavan: Same comments for all generated files

src/Microsoft.DotNet.Wpf/src/PresentationFramework/MS/Internal/WindowsRuntime/Generated/WinRT/ComWrappersSupport.cs
vatsan-madhavan: If this is generated code how come there are WPF spec fix comments here? I'm confused....

src/Microsoft.DotNet.Wpf/src/PresentationFramework/MS/Internal/WindowsRuntime/Generated/WinRT/ComWrappersSupport.net5.cs
vatsan-madhavan: Since this is in master and not p6, why not introduce cswinrt now instead of adding quasi-generated sources? Is the tool not ready?

src/Microsoft.DotNet.Wpf/src/PresentationFramework/MS/Internal/WindowsRuntime/Windows/Data/Text/WordsSegmenter.cs
```diff
@@ -93,6 +42,10 @@ private WordsSegmenter(string language)
             /// <exception cref="PlatformNotSupportedException">The OS platform is not supported</exception>
             public static WordsSegmenter Create(string language, bool shouldPreferNeutralSegmenter = false)
             {
+                if (Environment.OSVersion.Platform == PlatformID.Win32NT && Environment.OSVersion.Version < new Version(10, 0))
```
vatsan-madhavan, line 45: See prev note about os checks

vatsan-madhavan: We certainly support WordsSegmenter on OS < 10. This check is semantically wrong. It should be checked against win8.1 to prevent regressions.

vatsan-madhavan:
> Once the Windows SDK is shipping along with .NET as part of targeting net5.0-windows, WPF should be able to take a dependency on the SDK's projection directly

Isn't this something that can be done via native toolset bootstrapping in Arcade, or through a NuGet package, so that tools can be consumed before they fully ship officially ? 

Is there a preview SDK with CsWinRT in current DeV16 preview builds?

src/Microsoft.DotNet.Wpf/src/PresentationFramework/MS/Internal/WindowsRuntime/Generated/WinRT/ComWrappersSupport.cs
jkoritzinsky: I've made a small number of functional changes here to remove functionality (this allowed me to remove 100s of lines of code)

jkoritzinsky: I've tried to make as few changes as possible. The only changes I have made have been to enable removing significant amounts of code since this might be going into preview 6.

src/Microsoft.DotNet.Wpf/src/PresentationFramework/MS/Internal/WindowsRuntime/Generated/WinRT/ComWrappersSupport.net5.cs
jkoritzinsky: The correct manner is for the SDK to be shipped as part of the `net5.0-windows` tfm. We aren't quite at the point for that change yet, so for now we're doing this (embedding the sources) for master and p6 if this gets cherry-picked.

src/Microsoft.DotNet.Wpf/src/PresentationFramework/MS/Internal/WindowsRuntime/Generated/WinRT.cs
jkoritzinsky: These files cannot be generated during build since they've needed at least partial modifications to internal-ize public surface area and remove unused types. These should be considered a one-off generation. We'll remove this once the Windows SDK is ready to ship as part of the TFM.

I'll add the license header.

src/Microsoft.DotNet.Wpf/src/PresentationFramework/MS/Internal/WindowsRuntime/Windows/Data/Text/WordsSegmenter.cs
jkoritzinsky: I'll change this to check against 8.1 instead of 10.

jkoritzinsky:
> > Once the Windows SDK is shipping along with .NET as part of targeting net5.0-windows, WPF should be able to take a dependency on the SDK's projection directly
> 
> Isn't this something that can be done via native toolset bootstrapping in Arcade, or through a NuGet package, so that tools can be consumed before they fully ship officially ?
> 
> Is there a preview SDK with CsWinRT in current DeV16 preview builds?

I don't believe that the Windows SDK team is ready to ship the SDK in-box with .NET 5 quite yet, so I don't want to add a dependency directly on the new Microsoft.Windows.SDK.NET package and bring it inbox with WPF and accidentally force their hand.

src/Microsoft.DotNet.Wpf/src/PresentationCore/MS/internal/WindowsRuntime/Windows/UI/ViewManagement/InputPaneRcw.cs
weltkante: does this require deploying `api-ms-win-core-winrt-string-l1-1-0.dll` for Win7 (which is still supported as target by .NET Core) ?

src/Microsoft.DotNet.Wpf/src/PresentationCore/MS/internal/WindowsRuntime/Windows/UI/ViewManagement/InputPane.cs
```diff
@@ -33,19 +32,15 @@ internal class InputPane : IDisposable
             #region Fields
 
             /// <summary>
-            /// The name of the InputPane WinRT runtime class
+            /// Bool to check if the WinRT input pane is supported
             /// </summary>
-            private static readonly string s_TypeName = "Windows.UI.ViewManagement.InputPane, Windows, ContentType=WindowsRuntime";
+            private static readonly bool _isSupported;
```
rladuca, line 37: This change seems to turn a capability check into an OS version check.  That's the opposite of what we'd consider best practice.  Is there an overriding reason why we cannot retain the check for the presence of the type itself in some other fashion?

rladuca: Oh I see, is this covered by the EntryPointNotFound below?

src/Microsoft.DotNet.Wpf/src/PresentationCore/MS/internal/WindowsRuntime/Windows/UI/ViewManagement/InputPaneRcw.cs
rladuca: In Windows 7, the WPF InputPane wrapper classes are irrelevant.  WPF touch keyboard support works in Windows 7 without any intervention.  When focus tracking was changed to use WM_POINTER in Win8+, WPF touch keyboard support broke because WISP (which underlies all WPF touch/stylus support) redirects input messages to a separate thread and does not allow WM_POINTER* messages to flow to the actual message loop.  This was added in response to that change (although much later) so that WPF could explicitly request the touch keyboard to show.

src/Microsoft.DotNet.Wpf/src/PresentationCore/MS/internal/WindowsRuntime/Windows/UI/ViewManagement/InputPane.cs
rladuca: It seems to me that the EntryPointNotFoundException should be what determines support (a capability check) and not an OS version check.  Is there a reason why we absolutely require a version check here?

src/Microsoft.DotNet.Wpf/src/PresentationCore/MS/internal/WindowsRuntime/Windows/UI/ViewManagement/InputPaneRcw.cs
rladuca: In terms of Windows 10, the APIs needed are not only Win10 only, but 1607+:
https://docs.microsoft.com/en-us/windows/win32/api/inputpaneinterop/nn-inputpaneinterop-iinputpaneinterop

They were added (initially) to support this WPF mechanism, though they're useful for all Win32 applications.

weltkante: I was talking about the DLL reference to `api-ms-win-core-winrt-string-l1-1-0.dll` on Windows 8.1 (or whatever the version check will let through) resolving correctly in general, not about the particular InputPaneRCW

rladuca: Right, I am saying a DLLNotFoundException is acceptable on Win8.1 as the platform cannot support the InputPane usage anyway.  We'll have to ensure that exception is caught though (@jkoritzinsky) and appropriately handled to mark the InputPane mechanisms as not supported.

src/Microsoft.DotNet.Wpf/src/PresentationCore/MS/internal/WindowsRuntime/Windows/UI/ViewManagement/InputPane.cs
jkoritzinsky: Mainly for performance reasons since exceptions are slow. I can remove the OS checks and we can use exceptions to detect lack of support.

src/Microsoft.DotNet.Wpf/src/PresentationCore/MS/internal/WindowsRuntime/Windows/UI/ViewManagement/InputPaneRcw.cs
jkoritzinsky: I'm updating the initialization to handle the `DllNotFoundException` and have the code go down the same "mark as unsupported" path.

src/Microsoft.DotNet.Wpf/src/PresentationCore/MS/internal/WindowsRuntime/Windows/UI/ViewManagement/InputPane.cs
rladuca: I don't believe there is much of a perf impact here.  This constructor is only running on demand the first time we attempt to deal with the InputPane (which [TipTsfHelper](https://github.com/dotnet/wpf/blob/master/src/Microsoft.DotNet.Wpf/src/PresentationCore/MS/internal/Interop/TipTsfHelper.cs) only does when necessary).  After that we know the answer to whether or not the platform is supported and future calls are basically noops or useful work.

I'd prefer we stick to the capability check (since this OS check is actually not specific enough anyway, it needs to ensure 1607+).

src/Microsoft.DotNet.Wpf/src/PresentationFramework/MS/Internal/WindowsRuntime/Generated/WinRT/ExceptionHelpers.cs
rladuca: @ryalanms @fabiant3 should decide this, but seeing as this is a temporary change and the original CoreCLR code isn't localized, I agree we shouldn't do anything special here.

src/Microsoft.DotNet.Wpf/src/PresentationCore/MS/internal/WindowsRuntime/Windows/UI/ViewManagement/InputPane.cs
jkoritzinsky: Yes, this is covered by the `EntryPointNotFoundException` below and by the exception that `RoGetActivationFactory` will throw when it can't find the class.

src/Microsoft.DotNet.Wpf/src/PresentationCore/MS/internal/WindowsRuntime/Windows/UI/ViewManagement/NativeMethods.cs
```diff
@@ -0,0 +1,35 @@
...
+        {
+            [DllImport(DllImport.ApiSetWinRTString, CallingConvention = CallingConvention.StdCall)]
+            internal static extern unsafe int WindowsCreateString([MarshalAs(UnmanagedType.LPWStr)] string sourceString,
+                                                  int length,
+                                                  out IntPtr hstring);
+
+            [DllImport(DllImport.ApiSetWinRTString, CallingConvention = CallingConvention.StdCall)]
+            internal static extern int WindowsDeleteString(IntPtr hstring);
+
+            [DllImport(DllImport.ApiSetWinRT)]
```
vatsan-madhavan, line 31: Why no CallingConvention here?

jkoritzinsky: No particular reason. Looks like I missed it when I copied it over. The default is correct, but I'll add it explicitly.

src/Microsoft.DotNet.Wpf/src/PresentationFramework/MS/Internal/WindowsRuntime/Generated/WinRT/ExceptionHelpers.cs
vatsan-madhavan: I don't think this needs multi-lang support right now - as this is temporary/transient. 

In general, WPF localizes exception strings. Consuming an unlocalized exception string would break the norm. 

That said, WPF also links against other DLL's (that can in turn throw their own exceptions) that have different localization norms, so this is no different. 

My 2c. New tools/software like CsWinRT should be designed with globalization needs in mind. Hopefully hard-coding en-us strings is just indicative of the fact that the tool is an early prototype.

vatsan-madhavan:
Thanks for the changes and explanations @jkoritzinsky. Also thanks for explanations offline @AaronRobinsonMSFT. 

I understand that checking in the CsWinRT generated files is unavoidable for an interim-period since the tooling, build-integration etc. aren't quite ready, but removing WinRT support from the runtime needs to happen now. So this is a bootstrapping problem that must be solved with a little bit of a creative scotch-taping, I guess 😉 

The rest of the changes look good to me. I would wait for @ryalanms & @rladuca to signoff as well.

ryalanms:
Tests are still being run on the version of this change from this morning.

jeffschwMSFT:
@mmitche

jkoritzinsky:
@ryalanms can we merge this in and cherry-pick it to the preview 6 branch?

jkoritzinsky:
Or I can work on cherry-picking it myself and I'll put out the PR.

jkoritzinsky:
I've cherry picked this to Preview 6 in https://github.com/dotnet/wpf/pull/3094

ryalanms:
Thank you @lindexi for your help on this PR.

src/Microsoft.DotNet.Wpf/src/PresentationCore/MS/internal/WindowsRuntime/Windows/UI/ViewManagement/InputPane.cs
```diff
@@ -65,18 +60,17 @@ static InputPane()
...
-                    s_WinRTType = Type.GetType(s_TypeName);
-
                     // If we cannot get a new activation factory, then we cannot support
                     // this platform.  As such, null out the type to guard instantiations.
                     if (GetWinRtActivationFactory(forceInitialization: true) == null)
                     {
-                        s_WinRTType = null;
+                        _isSupported = false;
                     }
+                    _isSupported = true;
```
KalleOlaviNiemitalo, line 69: This looks wrong. If `GetWinRtActivationFactory` returns `null`, then the code first assigns `_isSupported = false` but immediately overwrites that with `_isSupported = true`.

KalleOlaviNiemitalo: So, if `GetWinRtActivationFactory` returns `null`, then `_isSupported` ends up being `true` even though InputPane is not actually supported. That then causes `private InputPane(IntPtr? hwnd)` constructor to call `GetWinRtActivationFactory` again, instead of throwing `PlatformNotSupportedException` right away. However, that call presumably returns `null` again, and the constructor then throws `PlatformNotSupportedException` anyway. I think this bug will not cause any real problem then, although it may cause `RoGetActivationFactory` to be unnecessarily retried and thus perhaps slow things down a bit.

PR #3101, opened 2020-06-08
Fix casing for Microsoft.WinFx.targets. (Nirmal4G, PaulEremeeff)

ryalanms:
This includes the first two commits from Nirmal4g and PaulEremeeff to fix WinFX casing.  (Thank you!)


vatsan-madhavan:
I highly recommend taking a look at this patch by @daviddenis-stx and integrating some of the changes related to path-separator chars into this PR: 

https://github.com/systemathics/wpf/commit/a9b28ed2998f13ee3c9bce37f405ff820cf43b08

ryalanms:
Thanks, @vatsan-madhavan.  I wasn't aware of @daviddenis-stx's change.

ryalanms:
I will be testing this today and tomorrow.  Thanks.

src/Microsoft.DotNet.Wpf/src/PresentationBuildTasks/MS/Internal/MarkupCompiler/MarkupCompiler.cs
```diff
@@ -50,7 +50,7 @@ internal sealed class MarkupCompiler
         public string TargetPath
         {
             get { return _targetPath; }
-            set { _targetPath = value; }
+            set { _targetPath = value;}
```
lindexi, line 53: Format issues

src/Microsoft.DotNet.Wpf/src/PresentationBuildTasks/MS/Internal/MarkupCompiler/MarkupCompiler.cs
```diff
@@ -712,7 +712,7 @@ private SourceFileInfo OnSourceFileResolve(FileUnit file)
                 if (sourceFileInfo.IsXamlFile)
                 {
                     int fileExtIndex = file.Path.LastIndexOf(DOT, StringComparison.Ordinal);
-
+
```
lindexi, line 715: Extra space

src/Microsoft.DotNet.Wpf/src/PresentationBuildTasks/Microsoft/Build/Tasks/Windows/MarkupCompilePass1.cs
```diff
@@ -1080,8 +1080,8 @@ private string GetResolvedFilePath(string filePath, ref string newSourceDir)
                 // and put the deepest directory that file is in as the new
                 // SourceDir.
                 //
-                int pathEndIndex = fullFilePath.LastIndexOf("\\", StringComparison.Ordinal);
-
+                int pathEndIndex = fullFilePath.LastIndexOf(string.Empty + Path.DirectorySeparatorChar, StringComparison.Ordinal);
+
```
lindexi, line 1084: Extra space

src/Microsoft.DotNet.Wpf/src/PresentationBuildTasks/Microsoft/Build/Tasks/Windows/MarkupCompilePass1.cs
```diff
@@ -1466,7 +1466,7 @@ private void GenerateOutputItems( )
 
                     codeItem = new TaskItem();
                     codeItem.ItemSpec = genLangFilePath;
-
+
```
lindexi, line 1469: Extra space

vatsan-madhavan:
With these changes in place, can you now build from with WSL?

vatsan-madhavan:
I know this is obvious but I’ll say it just so it doesn’t get forgotten - please don’t accidentally squash the PR :smile:

src/Microsoft.DotNet.Wpf/src/PresentationBuildTasks/MS/Internal/MarkupCompiler/MarkupCompiler.cs
ryalanms: @lindexi: Thank you again for the review.  These formatting changes will be taken care of with the other half of @Nirmal4g's PR.  Thanks.

ryalanms:
@vatsan-madhavan: It looks like more changes are required for WSL.  

`
/usr/share/dotnet/sdk/5.0.100-preview.4.20258.7/Sdks/Microsoft.NET.Sdk/targets/Microsoft.NET.Sdk.FrameworkReferenceResolution.targets(59,5): error NETSDK1100: Windows is required to build Windows desktop applications. 
`

daviddenis-stx:
Hello all :)

If you read the thread thoroughly you'll find that cross compilation support was found to be something dotnet team wasn't eager to support in the short term (or not at all ?)

The message "Windows is required... " was added after this small commit I made merely to test and see. In my experiments it produced working WPF assemblies but in the thread you'll find that some corner cases were expected to hurt (some of them depending on windows executables handling resx and/or resources if I recollect well). It's probably one of the reasons why the target stops with this message ever since. Otherwise it would potentially have created mangled outputs and the team probably didn't want to add more problems than solutions.

Long story short cross compiling was not broken just because of case sensitivity in targets files.

I still wish cross compilation was supported, in our current CI setup, we build WPF repos using docker on windows with nanoserver 2004 base images, because we have no, other choice. Each and every other repo we have builds in both Ubuntu and Windows containers and overall the time to completion is 2x to 3x longer on Windows for the exact same code ... painful. Also, we cannot run the tests that reach inside WPF parts (view models and stuff) in nano server containers, we even have to use servercore images for that. 5GB nightmare. Docker on Windows is far less efficient than on Linux, alas, the extra time wasting seems to be occurring when docker publishes to layers, you wait 2 minutes after a stage containing a simple "dotnet build" finishes to run (that is, you're not out of trouble when dotnet build returns as a lot of copy operations are occurring between some temp hcs to a windowsfilter sub folder, and it's 33.6 modem old times again... )

Cross compilation please :)




Get Outlook for Android<https://aka.ms/ghei36>

________________________________

vatsan-madhavan:
Yup, Microsoft.NET.Sdk.FrameworkReferenceResolution.targets has  NETSDK1100 error - “Windows is required to build Windows desktop applications”. 

The existing toolchain will probably work largely fine under cross-compilation - module problems like path-separators etc. Somebody should remove that error and try it to be sure. Any minor fixups needed from there can probably be figured out. 

Earlier in .NET Core 3.0, there were still pieces of the toolchain that were still msbuild only and hadn’t been ported to ‘dotnet’. IIRC an example was the WinForms  ResXFileCodeGenerator. 

The one thing I’d pay attention to is the BAML generated under cross compilation. I’d suggest comparing it against reference from Windows based builds.

vatsan-madhavan:
> Otherwise it would potentially have created mangled outputs

I don’t recall that part of the history. Would you mind elaborating?

daviddenis-stx:
I meant incomplete or different than what comes out of a Windows build.

In my tests I remember that BAML was not binary identical between Windows build and Linux build, but decompiled to identical XAML with ILSPY. If I remember well, there were some binary values here and there in the binary stream (guids?) that were random (that would be useful to also compare 2 rebuilds from the same platfom to confirm).

We could remove the switch that stops the build when non Windows is detected and try again the same recipe to see where it lands :)

D

Get Outlook for Android<https://aka.ms/ghei36>

________________________________

dsplaisted:
We'd like to remove it eventually.  It's there right now because we wanted to avoid shipping the targeting packs (or downloading them during restore) for non-Windows versions of the SDK.  When we support [optional SDK workloads](https://github.com/dotnet/designs/blob/master/accepted/2020/workloads/workloads.md) we could have the Windows support be a workload and remove the error.

dsplaisted:
Also FYI you can suppress the NETSDK1100 error with something like this:

```xml
<ItemGroup>
  <FrameworkReference Update="@(FrameworkReference)" IsWindowsOnly="false" />
</ItemGroup>
```

vatsan-madhavan:
You should consider making it so that cross-compilation can be tried out more easily than that (for e.g., a propertygroup that can be set in commandline).

Nirmal4G:
@ryalanms @vatsan-madhavan Will this be back ported into 3.1 as it is an LTS release?

Ref: dotnet/sdk#11108

vatsan-madhavan:
No idea what the team would choose to do but I wouldn't recommend porting it back to 3.1. 

This is a developer experience improvement, and "LTS" is a runtime quality/maintenance concept. NET 5 SDK can be used to target 3.1 runtime IIRC, and developers needing this fix could adopt a newer SDK. Besides, this hasn't been an adoption blocker for 3.1 in general.

worldbeater:
Anyway, it's definitely disappointing that we are still unable to use `MSBuild.Sdk.Extras` with .NET Core 3.1 due to this issue. The good news is, however, that a [workaround suggested by Jerome](https://github.com/unoplatform/docker/pull/4/commits/8e13b5b5f68ccc082048913dac88bc86ffe6006b) finally made using `MSBuild.Sdk.Extras` possible on Linux, even with .NET Core 3.1.

PR #3117, opened 2020-06-09
Remove custom copy of PtrToStringChars

GrabYourPitchforks:
This PR removes WPF's custom implementation of `PtrToStringChars` in favor of the built-in version. The custom version was originally introduced to work around CAS performance issues. However, since Core did not bring any of the CAS infrastructure forward, there's no need for the custom copy.

Related: https://github.com/dotnet/runtime/issues/31406


ryalanms (approved):

fabiant3 (commented):
There should be no more instances of our custom implementation now, given wpfvcclr.h is also being deleted. We have yet to enable our test collateral to be run in the open, but this looks safe. Thanks for the PR @GrabYourPitchforks!

fabiant3 (approved):

ryalanms:
Thank you for your contribution, @GrabYourPitchforks.

Issue #1718, opened 2019-08-24
.NET Core 3 Preview 8 WPF Build error on custom BaseIntermediateOutputPath 


TFTomSun:
* .NET Core Version: 3.0 Preview 8
* Windows version: Windows 10
* Does the bug reproduce also in WPF for .NET Framework 4.8?: No

 
 **Problem description:**
 I set the BaseIntermediateOutputPath to >$(OutputRootDirectory)obj\$(Configuration)\$(MSBuildProjectName)\
where OutputRootDirectory is a shared Output path of many projects
It is set via a build sdk that is included in the Directory.Build.props and Directory.Build.targets files in the root of my repository.

 **Actual behavior:** 
 When I build the project I get alot of build errors for xaml files, for example:
>D:\Git\Collaboration\Net\Framework\_out\obj\Debug\Siemens.Collaboration.Net.Presentation\netcoreapp3.0\Controls\AvalonDockApplicationControl.g.cs(54,18,54,62): error CS1504: Source file 'Controls\AvalonDockApplicationControl.xaml' could not be opened -- Could not find file.

It seems like the line information generated in to the g.cs files don't honor the the changed BaseIntermediateOutputPath setting.

 **Expected behavior:**
The build just runs fine

vatsan-madhavan:
Likely related to or duplicate of https://github.com/dotnet/wpf/issues/366

grubioe:
Closing as duplicate of #366

TFTomSun:
@grubioe I'm not sure whether this is a duplicate. It's more about how the compiler resolves line preprocessor statements with relative paths. For example into one of the generated cs files in the intermediate ouput is the following line statement generated:

>  #line 16 "Controls\AvalonDockApplicationControl.xaml"

I don't know how the compiler can find the path to the xaml file, because the xaml file is in my case not relatively located to the base intermediate output. I'm afraid that there's somewhere an assumption that the intermediate output is below the project directory.

TFTomSun:
I think I found the issue. The implementation of ParentFolderPath in the MarkupCompiler class of the PresentationBuildTasks assembly seems not to be clean:

```cs 
private string ParentFolderPrefix
    {
      get
      {
        string empty = string.Empty;
        if (this.TargetPath.StartsWith(this.SourceFileInfo.SourcePath, StringComparison.OrdinalIgnoreCase))
        {
[...]

Issue #366, opened 2019-02-20
Failure publishing WPF app with RID can fail if intermediate path is overridden


dsplaisted:
**Repro steps**

- `dotnet new wpf`
- Put the following in `Directory.Build.props` in project directory:

```xml
<Project>
    <PropertyGroup>
        <BaseOutputPath>$(MSBuildProjectDirectory)/out/$(MSBuildProjectName)/bin</BaseOutputPath>
        <BaseIntermediateOutputPath>$(MSBuildProjectDirectory)/out/$(MSBuildProjectName)/obj</BaseIntermediateOutputPath>
    </PropertyGroup>
</Project>
```

- `dotnet build`
- `dotnet publish -r win10-x64 --self-contained`

**Expected result**

Publish succeeds

** Actual result**

Errors such as:

> out\wpf\obj\Debug\netcoreapp3.0\wpf_oqgucux2_wpftmp.AssemblyInfo.cs(10,12): error CS0579: Duplicate 'System.Reflection.AssemblyCompanyAttribute' attribute [C:\git\repro\wpf\wpf_xfkwhwdg_wpftmp.csproj]
out\wpf\obj\Debug\netcoreapp3.0\wpf_oqgucux2_wpftmp.AssemblyInfo.cs(11,12): error CS0579: Duplicate 'System.Reflection.AssemblyConfigurationAttribute' attribute [C:\git\repro\wpf\wpf_xfkwhwdg_wpftmp.csproj]

It looks like these errors are due to the generated files being generated in two separate paths (one from the RID-less `dotnet build`, and the other from the `dotnet publish` with the RID specified), but both are being included in the WPF temporary generated project (likely because the Directory.Build.props intermediate path depends on the project name, which changes for the temporary project).

The best fix is probably to stop generating a temporary project, and instead call a target in the main project, if that's possible.

FYI @nguerrera 

Thanks to @mjrousos for reporting this

nguerrera:
@ryalanms

nguerrera:
We have had to do various workarounds for the wpf temp project. I agree it would be good to eliminate it.

See https://github.com/dotnet/arcade/search?q=IsWpfTempPRoject&unscoped_q=IsWpfTempProject

The idea I've had to do that would be to do something like multi-targeting where we build the same csproj, but change some global properties. I don't think this can just be a target in the same project evaluation as some of the targets need to run twice IIRC.

nguerrera:
[...]

PR #3120, opened 2020-06-10
.NET Core WPF Build error on custom BaseIntermediateOutputPath #1718

ryalanms:
Reviving the simplest of the proposed fixes for the BaseIntermediateOutputPath issue for comments.   (.NET standard's Path.GetRelativePath is one of the alternatives.)

Building WPF applications using .NET Framework and .NET Core with various custom BaseIntermediateOutputPaths on different drives works as expected.  

Note that PresentationBuildTasks.dll is built and consumed by both .NET Framework and .NET Core.


src/Microsoft.DotNet.Wpf/src/PresentationBuildTasks/MS/Internal/MarkupCompiler/NETFrameworkGetRelativePath.cs
```diff
@@ -0,0 +1,380 @@
...
+            if (EndsInDirectorySeparator(relativeTo.AsSpan()))
+                relativeToLength--;
+
+            bool pathEndsInSeparator = EndsInDirectorySeparator(path.AsSpan());
+            int pathLength = path.Length;
+            if (pathEndsInSeparator)
+                pathLength--;
+
+            // If we have effectively the same path, return "."
+            if (relativeToLength == pathLength && commonLength >= relativeToLength) return ".";
```
vatsan-madhavan, line 88: If "." is a constant with a special significance "current dir" for e.g., consider defining it as a const string with a name.

src/Microsoft.DotNet.Wpf/src/PresentationBuildTasks/MS/Internal/MarkupCompiler/NETFrameworkGetRelativePath.cs
```diff
@@ -0,0 +1,380 @@
...
+            //  C:\Foo C:\Foo\Bar L6, S0 -> Bar
+            //  C:\Foo\Bar C:\Bar\Bar L3, S2 -> ..\..\Bar\Bar
+            //  C:\Foo\Foo C:\Foo\Bar L7, S1 -> ..\Bar
+
+            StringBuilder sb = new StringBuilder();
+
+            // Add parent segments for segments past the common on the "from" path
+            if (commonLength < relativeToLength)
+            {
+                sb.Append("..");
```
vatsan-madhavan, line 105: Likewise, if ".." is "parent dir", define it as a const with a name.

src/Microsoft.DotNet.Wpf/src/PresentationBuildTasks/MS/Internal/MarkupCompiler/NETFrameworkGetRelativePath.cs
```diff
@@ -0,0 +1,380 @@
...
+        {
+            return c == DirectorySeparatorChar || c == AltDirectorySeparatorChar;
+        }
+
+        /// <summary>
+        /// Get the common path length from the start of the string.
+        /// </summary>
+        internal static int GetCommonPathLength(string first, string second, bool ignoreCase)
+        {
+            int commonChars = EqualStartingCharacterCount(first, second, ignoreCase: ignoreCase);
```
vatsan-madhavan, line 155: If possible, don't name the argument here since the variable is self-descriptive.

src/Microsoft.DotNet.Wpf/src/PresentationBuildTasks/MS/Internal/MarkupCompiler/NETFrameworkGetRelativePath.cs
```diff
@@ -0,0 +1,380 @@
...
+        }
+
+        /// <summary>
+        /// Returns true if the path is effectively empty for the current OS.
+        /// For unix, this is empty or null. For Windows, this is empty, null, or
+        /// just spaces ((char)32).
+        /// </summary>
+        internal static bool IsEffectivelyEmpty(ReadOnlySpan<char> path)
+        {
+            if (path.IsEmpty)
```
vatsan-madhavan, line 201: Why not something like this ?

```c#
if (RutimeInformation.IsOSPlatform(Windows))
{
    return string.IsNullOrWhiteSpace(path);
}
else
{
    return string.IsNullOrEmpty(path);
}
```

src/Microsoft.DotNet.Wpf/src/PresentationBuildTasks/MS/Internal/MarkupCompiler/NETFrameworkGetRelativePath.cs
```diff
@@ -0,0 +1,380 @@
...
+        public static bool EndsInDirectorySeparator(ReadOnlySpan<char> path)
+            => path.Length > 0 && PathInternal.IsDirectorySeparator(path[path.Length - 1]);
+
+        /// <summary>
+        /// Returns true if the path ends in a directory separator.
+        /// </summary>
+        public static bool EndsInDirectorySeparator(string path)
+              => path != null && path.Length > 0 && PathInternal.IsDirectorySeparator(path[path.Length - 1]);
+        
+        internal const char DirectorySeparatorChar = '\\';
```
vatsan-madhavan, line 369: Are you not leveraging System.IO.Path.DirectorySeparatorChar and IO.Path.AltDirectorySeparatorChar so that these values remain the same across all platforms?

src/Microsoft.DotNet.Wpf/src/PresentationBuildTasks/MS/Internal/MarkupCompiler/NETFrameworkGetRelativePath.cs
```diff
@@ -0,0 +1,380 @@
...
+                && path[0] == '\\'
+                && (path[1] == '\\' || path[1] == '?')
+                && path[2] == '?'
+                && path[3] == '\\';
+        }
+
+        /// <summary>
+        /// Returns true if the path ends in a directory separator.
+        /// </summary>
+        public static bool EndsInDirectorySeparator(ReadOnlySpan<char> path)
```
vatsan-madhavan, line 360: Consider naming this in a way that makes it clear that the result is a bool (i.e., it answers a question). As is, the name could be read either as "ends in directory separator" or "ends in directory separator?". Maybe DoesEndInDirectorySeparator, for e.g., as an alternative.

src/Microsoft.DotNet.Wpf/src/PresentationBuildTasks/MS/Internal/MarkupCompiler/NETFrameworkGetRelativePath.cs
```diff
@@ -0,0 +1,380 @@
...
+            // common Length and Segment count past the length.
+            //
+            // Some examples:
+            //
+            //  C:\Foo C:\Bar L3, S1 -> ..\Bar
+            //  C:\Foo C:\Foo\Bar L6, S0 -> Bar
+            //  C:\Foo\Bar C:\Bar\Bar L3, S2 -> ..\..\Bar\Bar
+            //  C:\Foo\Foo C:\Foo\Bar L7, S1 -> ..\Bar
+
+            StringBuilder sb = new StringBuilder();
```
vatsan-madhavan, line 100: nit: prefer `var`

src/Microsoft.DotNet.Wpf/src/PresentationBuildTasks/MS/Internal/MarkupCompiler/NETFrameworkGetRelativePath.cs
```diff
@@ -0,0 +1,380 @@
...
+
+            // Add parent segments for segments past the common on the "from" path
+            if (commonLength < relativeToLength)
```
vatsan-madhavan, line 103: What kinds of test-cases did you try this out on? If you have a unit-test for GetRelativePath, please do share. I know you can't (yet) merge those into the repo, but the test-data could help your reviewers give you some feedback (or offer additional test-data ideas).

src/Microsoft.DotNet.Wpf/src/PresentationBuildTasks/PresentationBuildTasks.csproj
```diff
@@ -279,6 +280,7 @@
     <Compile Include="MS\Internal\MarkupCompiler\CompilationUnit.cs" />
     <Compile Include="MS\Internal\MarkupCompiler\FileUnit.cs" />
     <Compile Include="MS\Internal\MarkupCompiler\MarkupCompiler.cs" />
+    <Compile Include="MS\Internal\MarkupCompiler\NETFrameworkGetRelativePath.cs" />
```
Nirmal4G, line 283: Why not use `GetRelativePath.NETFramework` as a filename instead?

vatsan-madhavan: Also consider PathInternal.cs.

src/Microsoft.DotNet.Wpf/src/PresentationBuildTasks/MS/Internal/MarkupCompiler/NETFrameworkGetRelativePath.cs
```diff
@@ -0,0 +1,380 @@
...
+{
+    internal sealed class PathInternal 
+    {
+        internal static string GetRelativePath(string relativeTo, string path, StringComparison comparisonType)
+        {
+            if (relativeTo == null)
+                throw new ArgumentNullException(nameof(relativeTo));
+
+            if (PathInternal.IsEffectivelyEmpty(relativeTo.AsSpan()))
+                throw new ArgumentException(nameof(relativeTo));
```
lindexi, line 54: The constructor of ArgumentException with one argument is a message not a parameter name

```
public ArgumentException (string message)
```

https://docs.microsoft.com/zh-cn/dotnet/api/system.argumentexception.-ctor?view=netcore-3.1#System_ArgumentException__ctor_System_String_

src/Microsoft.DotNet.Wpf/src/PresentationBuildTasks/MS/Internal/MarkupCompiler/NETFrameworkGetRelativePath.cs
```diff
@@ -0,0 +1,380 @@
...
+                    commonChars++;
+                    l++;
+                    r++;
+                }
+            }
+
+            return commonChars;
+        }
+
+            /// <summary>
```
lindexi, line 302: Format issues

src/Microsoft.DotNet.Wpf/src/PresentationBuildTasks/MS/Internal/MarkupCompiler/NETFrameworkGetRelativePath.cs
```diff
@@ -0,0 +1,380 @@
...
+        /// </summary>
+        public static bool EndsInDirectorySeparator(string path)
+              => path != null && path.Length > 0 && PathInternal.IsDirectorySeparator(path[path.Length - 1]);
+        
+        internal const char DirectorySeparatorChar = '\\';
+        internal const char AltDirectorySeparatorChar = '/';
+        internal const char VolumeSeparatorChar = ':';
+          // \\?\UNC\, \\.\UNC\
+        internal const int UncExtendedPrefixLength = 8;
+            // \\?\, \\.\, \??\
```
lindexi, line 374: format issues

src/Microsoft.DotNet.Wpf/src/PresentationBuildTasks/MS/Internal/MarkupCompiler/NETFrameworkGetRelativePath.cs
Nirmal4G: tip: If you have more than one review comment to offer, start a review instead of adding a comment. That way it doesn't pollute the PR with unnecessary breaks.

ryalanms:
@vatsan-madhavan and @lindexi:  Any thoughts on PresentationBuildTasks.dll taking a dependency on .NET Standard instead and avoiding this new code?

vatsan-madhavan:
> Any thoughts on PresentationBuildTasks.dll taking a dependency on .NET Standard instead and avoiding this new code?

I would be wary of doing this. This is an extensively used MSBuild task that must work robustly, and needs to load in VS, msbuild.exe and dotnet.exe reliably. The current way it builds (two outputs, net472, netcoreapp2.1) was carefully selected and tested, and the corresponding dependencies bundled with the SDK after much trial and error. 

The change you are proposing can be done, but perhaps not as a reflexive workaround for an otherwise tractable problem in a PR. This will be a big architectural change and must be done carefully, with extensive testing and enough time (in previews) for reacting to errors (which you don't really have in .NET 5 at this point). 

If/when you make the decision to enable cross-compilation on MacOS/Ubuntu etc., you'll have the opportunity to make this kind of change - at which time you'll also have the time to do it carefully keeping in mind  packaging dependencies that need to be updated, additional testing, allocating time for ecosystem feedback etc.

src/Microsoft.DotNet.Wpf/src/PresentationBuildTasks/MS/Internal/MarkupCompiler/NETFrameworkGetRelativePath.cs
```diff
@@ -0,0 +1,380 @@
...
+            if (relativeTo == null)
+                throw new ArgumentNullException(nameof(relativeTo));
+
+            if (PathInternal.IsEffectivelyEmpty(relativeTo.AsSpan()))
+                throw new ArgumentException(nameof(relativeTo));
+
+            if (path == null)
+                throw new ArgumentNullException(nameof(path));
+
+            if (PathInternal.IsEffectivelyEmpty(path.AsSpan()))
```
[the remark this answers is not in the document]
ryalanms: There are only two uses of each, but caching can be added when a third use is required.  Reactive if you disagree and I'll make the change.  Thanks.

src/Microsoft.DotNet.Wpf/src/PresentationBuildTasks/MS/Internal/MarkupCompiler/NETFrameworkGetRelativePath.cs
ryalanms: Only manual testing with the reported failing projects and the WPF samples (which don't exercise relative paths) have been tested.  I'll create a test application that wraps this and run through more extensive test cases.

src/Microsoft.DotNet.Wpf/src/PresentationBuildTasks/MS/Internal/MarkupCompiler/NETFrameworkGetRelativePath.cs
ryalanms: I think this was due to ReadOnlySpan not being convertible to string.  Please reactivate if you feel strongly.  Thanks.

ryalanms: This would require pulling in another assembly for .NET Framework PBT:

src/Microsoft.DotNet.Wpf/src/PresentationBuildTasks/MS/Internal/MarkupCompiler/NETFrameworkGetRelativePath.cs
ryalanms: Thanks, @lindexi.  This is standard convention for the .NET Core runtime.  Let me know if you disagree and I'll reactive.

lindexi: @ryalanms Sorry, I do not know it is standard convention for the .NET Core runtime. But I think if the parameter name is used as the message, then it may be difficult for us to get enough information from the message. We can only learn which parameter is wrong from the message, but we don’t know the reason for the error.

fabiant3 (approved):

src/Microsoft.DotNet.Wpf/src/PresentationBuildTasks/MS/Internal/MarkupCompiler/NETFrameworkGetRelativePath.cs
ryalanms: These issues are fixed for both .NET Framework and .NET Core: 

https://github.com/dotnet/wpf/issues/1718
https://github.com/dotnet/wpf/issues/366

vatsan-madhavan (approved):

src/Microsoft.DotNet.Wpf/src/PresentationBuildTasks/MS/Internal/MarkupCompiler/NETFrameworkGetRelativePath.cs
ryalanms: (The results of the internal unit testing were provided offline.)

PR #3207, opened 2020-06-30
Add the property for MicrosoftWindowsDesktopSdkImported

wli3:
So SDK side of the change will not be blocked by WPF side of change
More info in https://github.com/dotnet/sdk/issues/12108


wli3:
@ryalanms this is a simple marker to avoid double import. Could you approve that?

wli3:
cc @dsplaisted

vatsan-madhavan:
Will you have a similar marker in Sdk.targets as well?

wli3:
This property is enough, we won't import both props and targets if it has the mark

wli3:
And please help me merge it. I don't have the merge access

wli3:
@vatsan-madhavan @ryalanms could you help me merge this change?

vatsan-madhavan:
@ryalanms should take point here.

PR #3278, opened 2020-07-23
Use faster char based overload of String.IndexOf

davidwrighton:
- In dotnet/runtime#13339 I found that due to our changes to use icu by default there was a significant regression in Baml parsing time in .NET 5.0
- There is a plan to address most, if not all of that regression, but it won't be as good as simply using the char based IndexOf

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Markup/Baml2006/Baml2006Reader.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Markup/Baml2006/Baml2006Reader.cs
@@ -2065,7 +2065,7 @@ namespace System.Windows.Baml2006
 
         private string Logic_GetFullXmlns(string uriInput)
         {
-            int colonIdx = uriInput.IndexOf(":");
+            int colonIdx = uriInput.IndexOf(':');
             if (colonIdx != -1)
             {
                 string uriTypePrefix = uriInput.Substring(0, colonIdx);
@@ -2073,7 +2073,7 @@ namespace System.Windows.Baml2006
                 {
                     //We have a clr-namespace so do special processing
                     int clrNsStartIdx = colonIdx + 1;
-                    int semicolonIdx = uriInput.IndexOf(";");
+                    int semicolonIdx = uriInput.IndexOf(';');
                     if (-1 == semicolonIdx)
                     {
                         // We need to append local assembly
@@ -2085,7 +2085,7 @@ namespace System.Windows.Baml2006
                     else
                     {
                         int assemblyKeywordStartIdx = semicolonIdx + 1;
-                        int equalIdx = uriInput.IndexOf("=");
+                        int equalIdx = uriInput.IndexOf('=');
                         if (-1 == equalIdx)
                         {
                             throw new ArgumentException(SR.Get(SRID.MissingTagInNamespace, "=", uriInput));


src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Markup/Baml2006/Baml2006Reader.cs
```diff
@@ -2065,15 +2065,15 @@ private string Logic_GetFullyQualifiedNameForType(XamlType type)
 
         private string Logic_GetFullXmlns(string uriInput)
         {
-            int colonIdx = uriInput.IndexOf(":");
+            int colonIdx = uriInput.IndexOf(':');
             if (colonIdx != -1)
             {
                 string uriTypePrefix = uriInput.Substring(0, colonIdx);
                 if (String.Equals(uriTypePrefix, "clr-namespace"))
```
pentp, line 2072: This IndexOf+Substring+Equals is pretty inefficient, a simple StartsWith would be faster (optionally with a check for ':' at a const index(13) before to short-circuit).

src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Markup/Baml2006/Baml2006Reader.cs
```diff
@@ -2065,15 +2065,15 @@ private string Logic_GetFullyQualifiedNameForType(XamlType type)
...
                 {
                     //We have a clr-namespace so do special processing
                     int clrNsStartIdx = colonIdx + 1;
-                    int semicolonIdx = uriInput.IndexOf(";");
+                    int semicolonIdx = uriInput.IndexOf(';');
```
pentp, line 2076: Could add a startIndex:14 parameter

src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Markup/Baml2006/Baml2006Reader.cs
```diff
@@ -2085,7 +2085,7 @@ private string Logic_GetFullXmlns(string uriInput)
                     else
                     {
                         int assemblyKeywordStartIdx = semicolonIdx + 1;
-                        int equalIdx = uriInput.IndexOf("=");
+                        int equalIdx = uriInput.IndexOf('=');
```
pentp, line 2088: The `assembly=` check could be also done with a more efficient string.Compare check (and only search for the '=' in case of failure to get the proper error message) - no need for any substring allocation.

pentp: And probably don't need a substring allocation for the final empty assembly name check also.

src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Markup/Baml2006/Baml2006Reader.cs
```diff
@@ -2065,15 +2065,15 @@ private string Logic_GetFullyQualifiedNameForType(XamlType type)
...
                 string uriTypePrefix = uriInput.Substring(0, colonIdx);
                 if (String.Equals(uriTypePrefix, "clr-namespace"))
                 {
                     //We have a clr-namespace so do special processing
                     int clrNsStartIdx = colonIdx + 1;
-                    int semicolonIdx = uriInput.IndexOf(";");
+                    int semicolonIdx = uriInput.IndexOf(';');
                     if (-1 == semicolonIdx)
                     {
                         // We need to append local assembly
```
pentp, line 2079: The condition should be refactored to avoid one intermediate string allocation:
```
return (_settings.LocalAssembly != null)
  ? uriInput + ";assembly=" + GetAssemblyNameForNamespace(_settings.LocalAssembly)
  : uriInput;

davidwrighton:
@pentp This is on a somewhat hot startup path, but I don' t have the engineering time to spend to verify a more complex change as you propose before we lockdown to ship, and this change more than addresses the regression I saw. As you say, there is quite a bit of further improvements that could be had here. I believe your suggestions would do well in a new issue in the wpf repo.

fabiant3:
Agree with @davidwrighton, let's check this in and remove some of the heat. 
@pentp those are great suggestions, would you mind opening a separate issue and posting a PR? We'll try to pick up the change and test it internally next time we look at merging an external PR.

PR #3279, opened 2020-07-24
Use the new Marshal.IsTypeVisibleFromCom Api

lindexi:
See https://github.com/dotnet/coreclr/pull/19507

https://github.com/dotnet/runtime/issues/26497


fabiant3:
We'll take a look, thanks @lindexi. @ryalanms let's see if we can validate this though one of our test passes and merge in at some point.

ryalanms (approved):

PR #3302, opened 2020-08-03
Update to net5.0 template with backcompact

wli3:
Works for

```
 dotnet new wpf -n wpf1 -f netcoreapp3.1
 dotnet new wpf -n wpf2
 dotnet new wpf -n wpf3 -f netcoreapp3.1 --target-framework-override net461
 dotnet new wpf -n wpf3 -f net5.0 --target-framework-override net6.0
 dotnet new wpf -n wpf4 -f net5.0 --target-framework-override net6.0
```

### DIFF
--- a/packaging/Microsoft.Dotnet.Wpf.ProjectTemplates/content/WpfApplication-CSharp/Company.WpfApplication3x1.csproj
+++ b/packaging/Microsoft.Dotnet.Wpf.ProjectTemplates/content/WpfApplication-CSharp/Company.WpfApplication3x1.csproj
@@ -1,9 +1,9 @@
-﻿<Project Sdk="Microsoft.NET.Sdk">
+﻿<Project Sdk="Microsoft.NET.Sdk.WindowsDesktop">
 
   <PropertyGroup>
     <OutputType>WinExe</OutputType>
-    <TargetFramework Condition="'$(TargetFrameworkOverride)' == ''">FrameworkParameter-windows</TargetFramework>
-    <TargetFramework Condition="'$(TargetFrameworkOverride)' != ''">TargetFrameworkOverride-windows</TargetFramework>
+    <TargetFramework Condition="'$(TargetFrameworkOverride)' == ''">FrameworkParameter</TargetFramework>
+    <TargetFramework Condition="'$(TargetFrameworkOverride)' != ''">TargetFrameworkOverride</TargetFramework>
     <RootNamespace Condition="'$(name)' != '$(name{-VALUE-FORMS-}safe_namespace)'">Company.WpfApplication1</RootNamespace>
     <LangVersion Condition="'$(langVersion)' != ''">$(ProjectLanguageVersion)</LangVersion>
     <UseWPF>true</UseWPF>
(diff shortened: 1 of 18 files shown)


packaging/Microsoft.Dotnet.Wpf.ProjectTemplates/content/WpfApplication-CSharp/Company.WpfApplication3x1.csproj
```diff
@@ -0,0 +1,11 @@
+﻿<Project Sdk="Microsoft.NET.Sdk.WindowsDesktop">
+  <PropertyGroup>
+    <OutputType>WinExe</OutputType>
+    <TargetFramework Condition="'$(TargetFrameworkOverride)' == ''">net5.0</TargetFramework>
```
vatsan-madhavan, line 4: netcoreapp3.1

wli3: I actually just went down the rabbit hole on this property. The current authoring is a string place "net5.0" with the Framework input. It is odd, but I find the pattern is use in ASPNET and Winform too. I could change it to something like "FrameworkParameter" if you prefer breaking the pattern

```json
"Framework": {
      "type": "parameter",
      "description": "The target framework for the project.",
      "datatype": "choice",
      "choices": [
        {
          "choice": "netcoreapp3.0",
          "description": "Target netcoreapp3.0"
        },
        {
          "choice": "netcoreapp3.1",
          "description": "Target netcoreapp3.1"
        },
        {
          "choice": "net5.0",
          "description": "Target net5.0"
        }
      ],
      "replaces": "net5.0", <-----------------here
      "defaultValue": "net5.0"
    },
```

wli3: @vatsan-madhavan do you prefer to have "net5.0" and break the pattern or use "FrameworkParameter"?

vatsan-madhavan: I see. The usefulness of "net5.0"/'netcoreapp3.1' etc. in the csproj template is so that the template can be used as an actual csproj without modifications.  

So there is a tension between the two approaches. 

- FrameworkParameter/variable name makes it self-documenting and the 3x1 case becomes more accurate. At the same time, it forces the same change onto the net5 template as well. 
- net5.0 is a bit confusing, esp. to a reader looking at the 3x1 template. e.g. I left a comment asking to change it to netcoreapp3.1, which was obviously due to a misunderstanding of the *meaning* of this field in this particular context. OTOH, net5.0 looks just fine in the main template. 

I'm inclined to say that you should just change it to FrameworkParameter. The templating mechanism is very obscure and whatever makes it less so is better. Changing it improves the readers understanding of how things work, and leads them to the json Framework.replaces value. That has net higher value IMO than the templates mimicking a real csproj etc.

wli3:
Thank you @ryalanms and @vatsan-madhavan for helping me on this PR

PR #3318, opened 2020-08-08
Call SHA1.HashData static helper instead of instantiating SHA1CSP

GrabYourPitchforks:
We're trying to minimize usage of the `*CryptoServiceProvider` and `*Managed` types through our code bases so that we can eventually write analyzers to recommend moving off of them. See https://github.com/dotnet/runtime/issues/40169 for some related discussion w.r.t. `RNGCryptoServiceProvider`.

This PR changes the call site to use the static one-shot helper method so that you don't need to instantiate a hasher object. The `HashData` static method uses the best available implementation for the current OS.


reflectronic:
This code is actually copied from [C#/WinRT](https://github.com/microsoft/CsWinRT/blob/master/WinRT.Runtime/GuidGenerator.cs). It is supposed to eventually go away when the support for WinRT interop flows into the SDK in a way that WPF can consume. You may want to submit the fix there.

ryalanms (approved):

ryalanms:
Thanks, @GrabYourPitchforks.

Issue #2870, opened 2020-04-10
Unpin privatized C++ toolsets and ship using real-signed shipping tools


vatsan-madhavan:
https://github.com/dotnet/wpf/pull/2857 introduced use of privatized C++/CLI toolsets. This is a temporary fix and should be rolled back in favor of production tools. 

//build and RC/RTM products should not ship using these tools - they should be changed to production tools obtained from ambient tools coming from build pipeline machines. 

This would become possible as soon as Dev16.6p3 ships. 

**This should be treated as a release blocker for broad releases like //build, RC etc ** 

/cc @dotnet/wpf-developers

vatsan-madhavan:
You'll need to make 2 additional changes when fixing this one. 

- https://github.com/wpfcontrib/wpf/commit/11c24e3afb31cf7a40ee649e4e1f3022d5c9f9f3
- Disable LTCG for bilinearspan.lib. This causes [error C1047](https://docs.microsoft.com/en-us/cpp/error-messages/compiler-errors-1/fatal-error-c1047) when compilers change. Though this is only a transient problem, it can be a bother in dev-builds. 

Here is a preview of an upcoming build failure: 

> C:\Program Files (x86)\Microsoft Visual Studio\2019\Preview\VC\Tools\MSVC\14.28.29115\include\yvals_core.h(503,1): fatal error C1189: #error:  STL1001: Unexpected compiler version, expected MSVC 19.27 or newer.

ryalanms:
Thanks, @Vatsan.  I'll try to get this in soon.

vatsan-madhavan:
> Thanks, @vatsan. I'll try to get this in soon.

I might just be willing to lose by contrib history if I could get that username ;-)

ryalanms:
Thanks, @vatsan-madhavan.  This has been merged.

PR #3329, opened 2020-08-12
Remove VerifyWpfDllSet

### LINKED ISSUE #3322: Remove WpfDllVerifier checks from WPF
WPF has runtime checks to verify the correct versions of WPF assemblies are being loaded. It does so by validating that the DLLs are loaded from the same path. These checks were very useful when porting WPF to .NET Core, but WPF assembly load locations are varied for .NET 5 scenarios. The WpfDllVerifier class and uses of it should be removed for .NET 5.

    /// This class attempts to verify that core WPF binaries are all loaded from the same
    /// location.  Due to the fact that WPF can load binaries in different orders, this class
    /// needs to be called from all of the core managed binaries in static constructors of
    /// commonly used classes.
    /// 
    /// This can also be used to verify specific additional binaries as needed (say if they are
    /// optionally loaded).

    /// If the binaries fail to load from the same location an exception is thrown as we cannot 
    /// guarantee there won't be difficult to debug issues.

#11350
dotnet/runtime#38636

agocke:
Fixes #3322


vatsan-madhavan:
Your build failures are unrelated to your PR. You may have to be patient until the team gets around to dealing with #2870

agocke:
@vatsan-madhavan Is this ready to merge?

agocke:
ping @vatsan-madhavan

vatsan-madhavan:
/cc @rladuca

Recapping an internal email: I believe our discussion ended with the suggestion to retain the verifier them rather than remove it. Instead of checking consistency based on file-paths, the idea was to use a more flexible scheme, for e.g., a SHA, GUID or another unique ID, and factor in considerations like the presence of binaries from an other repo (dotnet-wpf-int) in the implementation, and also keep the implementation flexible enough to allow development in the inner-loop remain unconstrained.

I'll let the current team (already added as reviewers) take this PR forward how they see appropriate.  I'm happy to review additional changes as time permits (I'm on vacation for a few days - check my out-of-office message at work for details for those of you who have access 😉 ). This change looks like the right implementation if the goal were "removing the DLL verifier".

agocke:
>  the idea was to use a more flexible scheme, for e.g., a SHA, GUID or another unique ID

This makes sense, it's basically what Roslyn does to make sure that the compiler client and server are in-sync.

But I think we should take a weighted approach to problem solving here: without any change, there's a 100% probability that single-file publish will not work on .NET 5. Binaries getting out-of-sync is a problem, but more of a theoretical one at this point. In the mentioned thread I don't think we knew of any recent times customers had a problem with binary mismatch.

If we think the proposed solution could be implemented in time for .NET 5, I have no problem with that. But if there's a chance that it will miss, I would propose we remove the check for now, and add a separate verification scheme later.

hez2010:
Any updates on it?
I think unable to use WPF with PublishSingleFile is a critical problem.

ryalanms:
@vatsan-madhavan, the updated version of the Verifier with more capabilities will need to be added back to .NET 6.  We need to take this for RC1.  Thanks.

agocke:
@ryalanms It looks like this was merged into master. Is that the right branch for 5.0?

ryalanms:
Yes, the intention is to bring this in to release/5.0, but there is an approval process required.  I'll merge it in when we get permission.  Thanks.

agocke:
Any update on this? Is there an issue I can track for progress in Core Tactics?

Issue #3245, opened 2020-07-13
Add globbing support for WinUI 


stevenbrix:
The WinUI team needs support for globbing support so that Visual Studio will not edit the .csproj whenever you add a .xaml file to the project. See this bug as a reference: https://github.com/microsoft/microsoft-ui-xaml/issues/2498

In following the same pattern as WPF, we can add a `UseWinUI` property that will pull in the `Microsoft.NET.WindowsDesktop.Sdk` and enable the globbing support.

If both `UseWinUI` and `UseWpf` are true, then the SDK can assume all xaml files are for WPF. The WinUI nuget will respect the use of `UseWpf` and not invoke the winui xaml compiler if that is the case. If only `UseWinUI` is true, then the SDK should not invoke PBT. Below is a table describing the different combinations and what should happen:

| UseWpf | UseWinUI | Xaml Compiler invoked |
|      :-:       |        :-:         |             :-:           |
|  `True`    |  `True`           | PBT                    |
|  `True`    |  `False`          | PBT                    |
|  `False`    |  `True`          | WinUI Compiler |
|  `False`    |  `False`         | neither |

vatsan-madhavan:
Should all globbing support live in the base SDK instead of living in the WindowsDesktop SDK? This may not work well if WPF's targets are deciding which markup compiler to invoke, but perhaps WPF's targets should not be deciding that anyway. IIRC WPF's targets are making **local** decisions for itself, based on `UseWpf==true`. IMHO other compilers should do so based on their own rules. Why do these rules have to be all coordinated in the same place anyway? 

In .NET Framework/old-style projects, when there was no use of `UseWpf`, `UseWinUI` etc., `*.xaml` files were implicitly *inferred* to be WPF, UWP etc. These *inferences* required heuristic judgment. We did away with these heuristics and decided that the projects/developers had to explicitly state the intended XAML dialect being used by means these properties. This in turn disambiguates the XAML targets/compiler etc. Given this design, is there even a need for a central point-of-coordination based on a matrix of these property values? If both `UseWpf` and `UseWinUI` are set to `true` for the same XAML file, then both compilers would be invoked and things won't work out - but that's just a *"developer beware*" scenario that we can mitigate with warnings from the base SDK. 


https://github.com/microsoft/microsoft-ui-xaml/issues/2498#issuecomment-631781043
[...]

PR #3332, opened 2020-08-13
adding XamlRuntime msbuild property/metadata

stevenbrix:
This PR contains a scoped version of the changes described in #3245 

For now, we'll keep the globbing in WPF targets, and add the appropriate Item metadata and project property, so that vs tooling can distinguish WPF from WinUI in .NET core projects.

/cc @dsplaisted @ryalanms @chabiss


packaging/Microsoft.NET.Sdk.WindowsDesktop/targets/Microsoft.NET.Sdk.WindowsDesktop.targets
```diff
@@ -4,6 +4,7 @@
     <_MicrosoftNetSdkWindowsDesktop>true</_MicrosoftNetSdkWindowsDesktop>
     <EnableDefaultPageItems Condition="'$(EnableDefaultPageItems)' == ''">true</EnableDefaultPageItems>
     <EnableDefaultApplicationDefinition Condition="'$(EnableDefaultApplicationDefinition)' == ''">true</EnableDefaultApplicationDefinition>
+    <DefaultXamlRuntime Condition="'$(DefaultXamlRuntime)'==''">Wpf</DefaultXamlRuntime>
```
dsplaisted, line 7: Possibly this value should be capitalized: `WPF`

I'm not certain what the style guidelines are though.

stevenbrix: That's a fair point. `UseWPF` is caps, so it makes sense to be consistent

Issue #3317, opened 2020-08-07
Warning C4244 - implicit double -> float conversion with possible data-loss in BaseMatrix.cpp


vatsan-madhavan:
Here is the warning: 

> src\Microsoft.DotNet.Wpf\src\WpfGfx\core\common\BaseMatrix.cpp(1034,57): warning C4244: 'argument':
conversion from 'double' to 'float', possible loss of data

https://github.com/dotnet/wpf/blob/49cc5c6dcf0ead39a08b1917eeb1d472bcfbe739/src/Microsoft.DotNet.Wpf/src/WpfGfx/core/common/BaseMatrix.cpp#L1034

`CFloatFpu::Ceiling` takes a `float` param, but `std::pow` returns a `double`, which leads to the warning: 

https://github.com/dotnet/wpf/blob/49cc5c6dcf0ead39a08b1917eeb1d472bcfbe739/src/Microsoft.DotNet.Wpf/src/WpfGfx/common/shared/real.h#L628-L630

`std::pow` is defined to return `double` since C++11 ([reference](https://en.cppreference.com/w/cpp/numeric/math/pow))

>```C++
>Promoted    pow ( Arithmetic1 base, Arithmetic2 exp ); | (7) | (since C++11)
>```
>7) A set of overloads or a function template for all combinations of arguments of arithmetic type not covered by 1-3). If any argument has integral type, it is cast to `double`. If any argument is `long double`, then the return type Promoted is also `long double`, otherwise the return type is always `double`.

-- 

The result of `std::pow` should be safe to `static_cast` to `REAL` since `rScaleThreshold <= 1` (this is assert'ed early in the method). 

-- 

Why isn't this warning seen in CI/dev builds using build.cmd today? I can see it easily when building wpfgfx.vcxproj with `msbuild` directly.

ryalanms:
Thanks, @vatsan-madhavan.

PR #3333, opened 2020-08-13
Remove private Cpp tools in favor of using production tools

ryalanms:
Unpin privatized C++ toolsets and ship using real-signed shipping tools
> #2857 introduced use of privatized C++/CLI toolsets. This is a temporary fix and should be rolled back in favor of production tools.

Tracked by issue #2870. 

> //build and RC/RTM products should not ship using these tools - they should be changed to production tools obtained from ambient tools coming from build pipeline machines.

> This would become possible as soon as Dev16.6p3 ships.

> **This should be treated as a release blocker for broad releases like //build, RC etc **

/cc @dotnet/wpf-developers

Requires part 2 (see @vatsan-madhavan's comment): 

> You'll need to make 2 additional changes when fixing this one.

> wpfcontrib@11c24e3
> Disable LTCG for bilinearspan.lib. This causes error C1047 when compilers change. Though this is only a transient problem, it can be a bother in dev-builds.
> Here is a preview of an upcoming build failure:

> C:\Program Files (x86)\Microsoft Visual Studio\2019\Preview\VC\Tools\MSVC\14.28.29115\include\yvals_core.h(503,1): fatal error C1189: #error: STL1001: Unexpected compiler version, expected MSVC 19.27 or newer.

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/DirectWriteForwarder/CPP/DWriteWrapper/Factory.cpp
+++ b/src/Microsoft.DotNet.Wpf/src/DirectWriteForwarder/CPP/DWriteWrapper/Factory.cpp
@@ -92,7 +92,9 @@ namespace MS { namespace Internal { namespace Text { namespace TextInterface
         _pFactory = (IDWriteFactory*)factoryTemp;
     }
 
+    #pragma warning (disable : 4950) // The Constrained Execution Region (CER) feature is not supported.  
     [ReliabilityContract(Consistency::WillNotCorruptState, Cer::Success)]
+    #pragma warning (default : 4950) // The Constrained Execution Region (CER) feature is not supported.  
     __declspec(noinline) bool Factory::ReleaseHandle()
     {
         if (_wpfFontCollectionLoader != nullptr)
--- a/NuGet.config
+++ b/NuGet.config
@@ -9,4 +9,5 @@
     <add key="dotnet5-transport" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet5-transport/nuget/v3/index.json" />
     <add key="nuget.org" value="https://api.nuget.org/v3/index.json" />
   </packageSources>
+  <disabledPackageSources />
 </configuration>
--- a/eng/Version.Details.xml
+++ b/eng/Version.Details.xml
@@ -61,9 +61,9 @@
       <Uri>https://github.com/dotnet/runtime</Uri>
       <Sha>2d9dbbb538322a7cc05397f5e2e15d445c7d3ff0</Sha>
     </Dependency>
-    <Dependency Name="Microsoft.DotNet.Wpf.DncEng" Version="5.0.0-preview.8.20412.2">
+    <Dependency Name="Microsoft.DotNet.Wpf.DncEng" Version="5.0.0-rc.1.20414.3">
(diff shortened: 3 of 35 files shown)


ryalanms:
> src\Microsoft.DotNet.Wpf\src\WpfGfx\core\common\BaseMatrix.cpp(1034,57): error C4244: >> (NETCORE_ENGINEERING_TELEMETRY=Build) 'argument': conversion from 'double' to 'float', possible loss of data

vatsan-madhavan:
> > src\Microsoft.DotNet.Wpf\src\WpfGfx\core\common\BaseMatrix.cpp(1034,57): error C4244: >> (NETCORE_ENGINEERING_TELEMETRY=Build) 'argument': conversion from 'double' to 'float', possible loss of data

Follow the advice in https://github.com/dotnet/wpf/issues/3317 ;-)

src/Microsoft.DotNet.Wpf/src/DirectWriteForwarder/CPP/DWriteWrapper/Factory.cpp
```diff
@@ -92,7 +92,9 @@ namespace MS { namespace Internal { namespace Text { namespace TextInterface
         _pFactory = (IDWriteFactory*)factoryTemp;
     }
 
+    #pragma warning (disable : 4950) // The Constrained Execution Region (CER) feature is not supported.
```
vatsan-madhavan, line 95: If CER is not supported, then how do you plan on ensuring that this code works as intended? Simply suppressing the warning won't do that, would it?

/cc @SamBent

ryalanms: https://github.com/dotnet/designs/blob/master/accepted/2020/better-obsoletion/obsoletions-in-net5.md

The CER API calls need to be removed before GA; this is intended to unblock the builds.

ryalanms:
I've added the static_cast to REAL in https://github.com/dotnet/wpf/issues/3317.  This should be the final fix to get the builds going again.  wpf-int still requires the darc-update and undo of UsePrivateCppTools.  

>> src\Microsoft.DotNet.Wpf\src\WpfGfx\core\common\BaseMatrix.cpp(1034,57): error C4244: >> >> (NETCORE_ENGINEERING_TELEMETRY=Build) 'argument': conversion from 'double' to 'float', possible loss of data

> Follow the advice in #3317 ;-)

vatsan-madhavan:
I'm still puzzled at why this is working without https://github.com/wpfcontrib/wpf/commit/11c24e3. 

`$(VcrtLibDir)` shouldn't resolve to anything very useful because `VCToolsInstallDir` gets set much later. Eventually, `AdditionalLibraryDirectories` doesn't include all the paths that are usually needed to resolve msvcurt_netcore[d].lib and a few other libs from ambient toolset.

At least from what I've noticed, it doesn't get resolved every time, or there are times when the failure to set these paths correctly leads to failures. 

Have you checked the logs to see how these properties are evaluated and why the linker is able to find libraries correctly even though `AdditionalLibraryDirectories` isn't set quite correctly?

ryalanms:
Thanks, @vatsan-madhavan!

PR #3436, opened 2020-09-02
[release/5.0-rc2] Remove WpfDllVerifier checks from WPF 

ryalanms:
Remove WpfDllVerifier checks from WPF #3322 

>WPF has runtime checks to verify the correct versions of WPF assemblies are being loaded. It does so by validating that the DLLs are loaded from the same path. These checks were very useful when porting WPF from .NET Framework to .NET Core, but WPF assembly load locations are varied for .NET 5 scenarios. The WpfDllVerifier class and uses of it should be removed for .NET 5.

The brings @agocke's change (https://github.com/dotnet/wpf/pull/3329) to release/5.0 from master.

Issue: https://github.com/dotnet/wpf/issues/3322


ryalanms:
WPF received approval to take this for RC2.  I'll merge after the checks complete.  Thanks.

PR #3437, opened 2020-09-02
[release/5.0] Fix raw stylus data to support per-monitor DPI 

ryalanms:
### [release/5.0] Fix raw stylus data to support per-monitor DPI 

This brings @rladuca's change from master to release/5.0. 

Adjust the transform applied to the input per dpi scale, retrieved from the CompositionTarget of the input source. This allows the interemdiate points to be correct, if there more than one dpi scale (multiple monitors).
####  `Regression?`
No.  

*Then, why is this being brought up for servicing now?  What changed?*

This is a serious bug for applications that rely on touch, and we have two customers asking for a fix in 5.0.  
####  `Risk (see taxonomy)`
Low.  

The change has been in master since August 24, 2020, and test pass results showed no test failures in touch.  WPF has touch tests but none for this specific scenario (this was tested manually on the most recent release/5.0).  
##### Original Issue:

TouchMove intermediate points are not DPI aware (https://github.com/dotnet/wpf/issues/2072)

##### Original PR:

Fix raw stylus data to support per-monitor DPI (https://github.com/dotnet/wpf/pull/2891)

No.


vatsan-madhavan:
You're still building release/5.0 against a very old SDK 

https://github.com/dotnet/wpf/blob/be89bbfaded6ad3f0f4143bf1a8dd075a5789715/global.json#L3

This should be maintained regularly (each preview, when branch-snapping happens) and updated to the most recent preview-SDK. I suspect that you'll encounter problems/failures when moving up to a newer SDK that will require (pre-RTM) fixups. The repo cannot stay on an old SDK like this post-RTM during servicing. Arcade will sooner or later (often sooner) force this value to be updated to the RTM value of net5.0. Better updated this now and deal with the fallout.

dotMorten:
Any update on this? Just got pinged again by customers hitting this.

ryalanms:
@dotMorten:  We will bring this up for consideration again on Thursday.  Thanks.

ryalanms:
This has been approved.  We are waiting for the 5.0.x servicing branch to open for merges.

dotMorten:
Thank you! 🎉

PR #3505, opened 2020-09-14
Add BindingFailed event for diagnostics of binding failures

spadapet:
Visual Studio 16.8 will include a new set of XAML Binding Failures diagnostic improvements. The challenge with a binding failure is locating the source of the binding in XAML. We determined that to locate a binding failure in XAML, the source location must be provided by the platform. This will allow the user to quickly navigate to the binding failure.

The design is that WPF will trigger a new event any time debug traces are written for a binding failure. Debug traces alone aren’t enough to find the source of the failure. The new event will contain more information about the binding failure, such as the BindingExpression object, which helps locate the binding in the XAML source.

**Risk**

The risk for the runtime is very low because this feature doesn’t get activated for any application running. It only gets executed when the application is run under the debugger.

The new event has no effect on memory or performance unless certain conditions are met (these condition are the same as the existing ResourceDictionaryDiagnostics feature): 

1. VisualDiagnostics must be enabled 
2. The environment variable ENABLE_XAML_DIAGNOSTICS_SOURCE_INFO must be set so that source information is cached for XAML elements 
3. Trace output must be enabled for binding failures, if there’s no trace there’s no event 

**Event cache**

When Visual Studio injects a DLL into the user’s WPF app, it will attach to the new event BindingDiagnostics.BindingFailed. This happens well after the app starts running so it will miss initial failure events. To solve that, all BindingFailed events are cached until VS can add an event handler. There is a limit to the cache, but it will use memory until VS can add the handler. Then the cache is flushed. Again, this cache is only created if the diagnostics conditions are met (environment variable, etc).

Look in the new file BindingDiagnostics.cs for the cache (and most of the new code).

**Testing**

Tests are currently being written in the Visual Studio repository, which will fully test the new BindingFailed event. Tests will also be added to the internal dotnet-wpf-test repository, they have not been added yet but work is starting on that ASAP.

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Diagnostics/BindingDiagnostics.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Diagnostics/BindingDiagnostics.cs
@@ -22,23 +22,23 @@ namespace System.Windows.Diagnostics
     {
         internal static bool IsEnabled { get; private set; }
 
-        private static event EventHandler<BindingFailedEventArgs> bindingFailed;
-        private static List<BindingFailedEventArgs> pendingEvents;
-        private static object pendingEventsLock;
-        private const int maxPendingEvents = 2000;
+        private static event EventHandler<BindingFailedEventArgs> s_bindingFailed;
+        private static List<BindingFailedEventArgs> s_pendingEvents;
+        private static readonly object s_pendingEventsLock;
+        private const int MaxPendingEvents = 2000;
 
         static BindingDiagnostics()
         {
-            BindingDiagnostics.IsEnabled = VisualDiagnostics.IsEnabled && VisualDiagnostics.IsEnvironmentVariableSet(null, XamlSourceInfoHelper.XamlSourceInfoEnvironmentVariable);
+            IsEnabled = VisualDiagnostics.IsEnabled && VisualDiagnostics.IsEnvironmentVariableSet(null, XamlSourceInfoHelper.XamlSourceInfoEnvironmentVariable);
 
-            if (BindingDiagnostics.IsEnabled)
+            if (IsEnabled)
             {
                 // Listeners may miss the initial set of binding failures, so cache events until the first listener attaches.
                 // Normally there will only be one listener added soon after the process starts,
                 // and it will want to know about any binding failures that already happened.
 
-                BindingDiagnostics.pendingEvents = new List<BindingFailedEventArgs>();
-                BindingDiagnostics.pendingEventsLock = new object();
+                s_pendingEvents = new List<BindingFailedEventArgs>();
+                s_pendingEventsLock = new object();
             }
         }
 
@@ -49,16 +49,16 @@ namespace System.Windows.Diagnostics
         {
             add
             {
-                if (BindingDiagnostics.IsEnabled)
+                if (IsEnabled)
                 {
-                    BindingDiagnostics.bindingFailed += value;
-                    BindingDiagnostics.FlushPendingBindingFailedEvents();
+                    s_bindingFailed += value;
+                    FlushPendingBindingFailedEvents();
                 }
             }
 
             remove
             {
-                BindingDiagnostics.bindingFailed -= value;
+                s_bindingFailed -= value;
             }
         }
 
@@ -67,23 +67,23 @@ namespace System.Windows.Diagnostics
         /// </summary>
         private static void FlushPendingBindingFailedEvents()
         {
-            if (BindingDiagnostics.pendingEvents != null)
+            if (s_pendingEvents != null)
             {
-                BindingFailedEventArgs[] pendingEventsCopy = null;
+                List<BindingFailedEventArgs> pendingEvents = null;
 
-                lock (BindingDiagnostics.pendingEventsLock)
+                lock (s_pendingEventsLock)
                 {
-                    pendingEventsCopy = BindingDiagnostics.pendingEvents?.ToArray();
+                    pendingEvents = s_pendingEvents;
 
                     // Don't allow any more event caching
-                    BindingDiagnostics.pendingEvents = null;
+                    s_pendingEvents = null;
                 }
 
-                if (pendingEventsCopy != null)
+                if (pendingEvents != null)
                 {
-                    foreach (BindingFailedEventArgs args in pendingEventsCopy)
+                    foreach (BindingFailedEventArgs args in pendingEvents)
                     {
-                        BindingDiagnostics.bindingFailed?.Invoke(null, args);
+                        s_bindingFailed?.Invoke(null, args);
                     }
                 }
             }
@@ -94,21 +94,21 @@ namespace System.Windows.Diagnostics
         /// </summary>
         internal static void NotifyBindingFailed(BindingFailedEventArgs args)
         {
-            if (!BindingDiagnostics.IsEnabled)
+            if (!IsEnabled)
             {
                 return;
             }
 
-            if (BindingDiagnostics.pendingEvents != null)
+            if (s_pendingEvents != null)
             {
-                lock (BindingDiagnostics.pendingEventsLock)
+                lock (s_pendingEventsLock)
                 {
-                    if (BindingDiagnostics.pendingEvents != null)
+                    if (s_pendingEvents != null)
                     {
                         // Limit the pending event count so that memory doesn't grow unbounded if no event handler is ever added
-                        if (BindingDiagnostics.pendingEvents.Count < BindingDiagnostics.maxPendingEvents)
+                        if (s_pendingEvents.Count < MaxPendingEvents)
                         {
-                            BindingDiagnostics.pendingEvents.Add(args);
+                            s_pendingEvents.Add(args);
                         }
 
                         return;
@@ -116,7 +116,7 @@ namespace System.Windows.Diagnostics
                 }
             }
 
-            BindingDiagnostics.bindingFailed?.Invoke(null, args);
+            s_bindingFailed?.Invoke(null, args);
         }
     }
 }
--- a/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Diagnostics/BindingDiagnostics.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Diagnostics/BindingDiagnostics.cs
@@ -7,7 +7,6 @@
 //  Binding diagnostics API
 //
 
-using System.Collections.Concurrent;
 using System.Collections.Generic;
 
 namespace System.Windows.Diagnostics
@@ -21,7 +20,7 @@ namespace System.Windows.Diagnostics
     /// </remarks>
     public static class BindingDiagnostics
     {
-        internal static bool IsEnabled { get; }
+        internal static bool IsEnabled { get; private set; }
 
         private static event EventHandler<BindingFailedEventArgs> bindingFailed;
         private static List<BindingFailedEventArgs> pendingEvents;
@@ -43,6 +42,9 @@ namespace System.Windows.Diagnostics
             }
         }
 
+        /// <summary>
+        /// Handlers of this event should return control to WPF quickly, and not cache BindingFailedEventArgs for future use.
+        /// </summary>
         public static event EventHandler<BindingFailedEventArgs> BindingFailed
         {
             add
--- a/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Diagnostics/BindingDiagnostics.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Diagnostics/BindingDiagnostics.cs
@@ -1,0 +1,120 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+//
+// Description:
+//  Binding diagnostics API
+//
+
+using System.Collections.Concurrent;
+using System.Collections.Generic;
+
+namespace System.Windows.Diagnostics
+{
+    /// <summary>
+    /// Provides a notification infrastructure for listening to binding failure events.
+    /// </summary>
+    /// <remarks>
+    /// This type supports the .NET Framework infrastructure and is not intended to be used directly
+    /// from application code.
+    /// </remarks>
+    public static class BindingDiagnostics
+    {
+        internal static bool IsEnabled { get; }
+
+        private static event EventHandler<BindingFailedEventArgs> bindingFailed;
+        private static List<BindingFailedEventArgs> pendingEvents;
+        private static object pendingEventsLock;
+        private const int maxPendingEvents = 2000;
+
+        static BindingDiagnostics()
+        {
+            BindingDiagnostics.IsEnabled = VisualDiagnostics.IsEnabled && VisualDiagnostics.IsEnvironmentVariableSet(null, XamlSourceInfoHelper.XamlSourceInfoEnvironmentVariable);
+
+            if (BindingDiagnostics.IsEnabled)
+            {
+                // Listeners may miss the initial set of binding failures, so cache events until the first listener attaches.
+                // Normally there will only be one listener added soon after the process starts,
+                // and it will want to know about any binding failures that already happened.
+
+                BindingDiagnostics.pendingEvents = new List<BindingFailedEventArgs>();
+                BindingDiagnostics.pendingEventsLock = new object();
+            }
+        }
+
+        public static event EventHandler<BindingFailedEventArgs> BindingFailed
+        {
+            add
+            {
+                if (BindingDiagnostics.IsEnabled)
+                {
+                    BindingDiagnostics.bindingFailed += value;
+                    BindingDiagnostics.FlushPendingBindingFailedEvents();
+                }
+            }
+
+            remove
+            {
+                BindingDiagnostics.bindingFailed -= value;
+            }
+        }
+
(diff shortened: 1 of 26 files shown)


chabiss:
>             if (TraceData.IsEnabled)

What does control TraceData being enabled? There is a potentially an issue here where an external settings will prevent Binding Diagnostics from working? 

---
Refers to: src/Microsoft.DotNet.Wpf/src/PresentationFramework/MS/Internal/Helper.cs:325 in 6d1ec56. [](commit_id = 6d1ec561f9d259974ce9ae45f917309649181252, deletion_comment = False)

chabiss (approved):
:shipit:

spadapet:
>             if (TraceData.IsEnabled)

As far as I can tell, TraceData.IsEnabled is always true when debugging.

There's code that forces it enabled while debugging here:
[https://github.com/dotnet/wpf/blob/15f304663e0f9ed66f3b581032859230927b039b/src/Microsoft.DotNet.Wpf/src/PresentationFramework/MS/Internal/TraceData.cs#L70](https://github.com/dotnet/wpf/blob/15f304663e0f9ed66f3b581032859230927b039b/src/Microsoft.DotNet.Wpf/src/PresentationFramework/MS/Internal/TraceData.cs#L70)

---
In reply to: [692323046](https://github.com/dotnet/wpf/pull/3505#issuecomment-692323046) [](ancestors = 692323046)

---
Refers to: src/Microsoft.DotNet.Wpf/src/PresentationFramework/MS/Internal/Helper.cs:325 in 6d1ec56. [](commit_id = 6d1ec561f9d259974ce9ae45f917309649181252, deletion_comment = False)

src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Diagnostics/BindingDiagnostics.cs
```diff
@@ -0,0 +1,120 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+//
+// Description:
+//  Binding diagnostics API
+//
+
+using System.Collections.Concurrent;
```
SamBent, line 10: Unused, I think.  If so, delete it.

src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Diagnostics/BindingFailedEventArgs.cs
```diff
@@ -0,0 +1,53 @@
...
+
+        /// <summary>
+        /// This is the full message that is also written to debug output.
+        /// </summary>
+        public string Message { get; }
+
+        /// <summary>
+        /// Can be null for some failure codes that don't have a BindingExpressionBase context.
+        /// </summary>
+        public BindingExpressionBase Binding { get; }
```
SamBent, line 37: Lifetime problem:  A listener that retains a reference to the Binding beyond the scope of the event will affect the lifetime of the binding object and everything it refers to, which could encompass a lot - visual subtrees, etc..  This could change the behavior of the app, in obvious ways like perceived memory leaks, and in less obvious ways like bindings and their target elements remaining active after the app thinks they're gone.  [This applies to your "pending event" list as well.]

How is VS planning to use Binding?  In particular, will you need it after the event, say to help a user who asks about a binding error that happened long ago and whose binding wouldn't exist any more sans BindingDiagnostics?  If so, we can't fix the problem, but we (well, you - VS) should warn users that "running with BindingDiagnostics enabled can change the behavior of the app, take appropriate precautions".   If not, i.e. if your event listeners will harvest what they need from the Binding, is there some way to put the harvested data in the event args?

In any case, please add a comment here to warn other providers of diagnostic assistants of the issue.

SamBent (changes requested):

batzen:
Please do not make this new feature depend on the presence of a debugger and an environment variable.
Making it depend on those will prevent people from being able to use it in their own application and will prevent tools like snoop from being able to use it.

chabiss:
No... It is strictly forbidden for application to rely on this property. This follows the *exact* same guideline as we set for VisualDiagnostics.IsEnabled.     

---
In reply to: [692472986](https://github.com/dotnet/wpf/pull/3505#issuecomment-692472986) [](ancestors = 692472986)

batzen:
I was not talking about regular usage, but self diagnostic instead. It's, most of the time, not possible to use a debugger on customer machines, while using tools like snoop is possible.

Having a requirement for an attached debugger does not seem to have any benefit, besides from preventing other tools from using the diagnostics.

spadapet:
During the initial design review Sam said BindingDiagnostics.IsEnabled must follow VisualDiagnostics.IsEnabled plus check the source info environment variable. I can see how other tools (like snoop) may want to use this event, but in order to not affect perf it is only triggered for debug diagnostics scenarios.

---
In reply to: [692755627](https://github.com/dotnet/wpf/pull/3505#issuecomment-692755627) [](ancestors = 692755627,692472986)

src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Diagnostics/BindingDiagnostics.cs
spadapet: Good catch, it was my final change that removed the ConcurrentQueue and just went with a lock to protect pendingEvents.

---
In reply to: [488320841](https://github.com/dotnet/wpf/pull/3505#discussion_r488320841) [](ancestors = 488320841)

src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Diagnostics/BindingFailedEventArgs.cs
spadapet: Actually all VS does is harvest some strings and source info from the Binding, nothing is cached. The pending event list will hold onto Bindings until VS has a chance to flush the events. So that will prevent garbage collection. Seems like my options are:

1. Not cache the Binding, just cache everything I need from it. Hopefully VS doesn't need more info in the future
2. Use a WeakReference instead, and hope that Bindings aren't garbage collected before the pending events are flushed
3. Keep as is, but add more comments and documentation about how objects may have a long lifetime with diagnostics enabled

I'll see exactly what change I'd need for #1 (remove the Binding reference and cache data in here rather than VS code)

---
In reply to: [488330836](https://github.com/dotnet/wpf/pull/3505#discussion_r488330836) [](ancestors = 488330836)

SamBent:
It's not perf so much as risk.  We raise diagnostic events at unsafe times, when running arbitrary user code could break assumptions, cause re-entrancy, and do all kinds of damage that we're not prepared to support.  We can allow listeners whose intent is diagnostic, since any bugs they introduce are their fault _a _priori__ and not bugs in the app or in WPF itself;  they won't arise in a production environment.   We can't allow arbitrary listeners.

I'd be willing to allow Snoop (and similar tools) to use the diagnostic features, provided there was some reliable way they can identify themselves as a diagnostic add-on, and not part of the app itself.  The debugger/environment check is how we agreed to do this ("we" = WPF and VS).

batzen:
@SamBent Having a way to enable diagnostics and tell the runtime there's some diagnostics tool being attached would be very nice. As snoop is, currently, cut off from using anything related to `VisualDiagnostics` (currently BindingDiagnostics and ResourceDictionaryDiagnostics as far as i know).

vatsan-madhavan:
>The debugger/environment check is how we agreed to do this ("we" = WPF and VS).

It's more generic than that IMO. Although some product development in .NET world happens in symbiosis with VS product development, public API's and functionality like this are usually designed with everybody in mind. 

These diagnostics API's are dangerous for regular applications, but happen to be within the purview of one class of applications - namely **debuggers**. That's the class of apps for which they are already enabled by default. *Snoop* should implement `ICorDebug` interface and advertise itself as a managed debugger in order to consume this functionality safely IMHO. 

The recognition that some applications cannot (or do not want to) advertise themselves as debuggers is why the env-var alternative exists (as well as the fact that it aids in testing). 

If the real problem is that (a) apps like Snoop cannot advertise itself as a managed debugger *and* (b) cannot use the env-var alternative either, then it would make sense to think about the design here and look for a better approach.

batzen:
> The recognition that some applications cannot (or do not want to) advertise themselves as debuggers is why the env-var alternative exists (as well as the fact that it aids in testing).

The env-var is used in the cctor. If some code caused the cctor to run, for whatever reason, makes/would make it impossible to activate it later, and maybe even more important, deactivate it when snoop is detached.
I already thought about advertising snoop as a debugger, but decided to not do that as there can only be one debugger active in a process and i don't see snoop being a debugger but a diagnostics tool instead.

`ResourceDictionaryDiagnostics` has a private setter for `IsEnabled` so it can, in theory be turned on and off by snoop.
`BindingDiagnostics` does not have a private setter for `IsEnabled`. Could that be changed? Then snoop would have a chance to enable/disable it.

It would be very nice if you could keep tools like snoop in mind when adding future diagnostics. Many people, including me of course, prefer to use snoop as a diagnostics tool because it's much more lightweight and doesn't even require an installation.

And i have to say that i am very happy to be able to even have this kind of discussion/conversation now that WPF is OSS. 👍❤

miloush:
I think asking for `ICorDebug` being implemented is fair, but I also happen to attach snoop to an app being debugged in VS, so if this would hit a single debugger rule then it wouldn't help.

Am I correct expecting that any VS extension can take advantage of the API as it would be coming from a debugger process (VS)?

spadapet:
I can definitely implement BindingDiagnostics.IsEnabled the same way ResourceDictionaryDiagnostics.IsEnabled is implemented if that allows the value to be tweaked using reflection.

---
In reply to: [692897519](https://github.com/dotnet/wpf/pull/3505#issuecomment-692897519) [](ancestors = 692897519)

vatsan-madhavan:
I didn't remember the single-debugger rule - thanks for pointing that out. 

Maybe there should be a new interface that's specific to the land of visual diagnostics which is implemented by *visual debuggers*, and allows more than one *visual debugger* to be 'attached' to an app simultaneously. 

Put differently, looks like *'is a debugger?'* test isn't an apt one here and we should probably invent a better one.

vatsan-madhavan:
> allows the value to be tweaked using reflection.

My 2c - Other visual-diagnostics products should be able to opt-in without having to rely on undocumented or unsupported behavior.

miloush:
I was under the impression that the API itself is meant to be unsupported. If that itself is not enough and its usage should be actively blocked, then using reflection seems to be a reasonable expression intent on caller to do unsupported stuff.

If this is supported API then anyone should be able to call it.

batzen:
> I can definitely implement BindingDiagnostics.IsEnabled the same way ResourceDictionaryDiagnostics.IsEnabled is implemented if that allows the value to be tweaked using reflection.

That would be nice. 👍

batzen:
> My 2c - Other visual-diagnostics products should be able to opt-in without having to rely on undocumented or unsupported behavior.

That would also be very nice.
Being able to retrieve a list of supported diagnostics would be even better, but that should be part of a separate discussion.
If you (WPF team) ever want to talk/discuss about such things i would be happy to join. ;-)

vatsan-madhavan:
> I was under the impression that the API itself is meant to be unsupported. If that itself is not enough and its usage should be actively blocked, then using reflection seems to be a reasonable expression intent on caller to do unsupported stuff.
> 
> If this is supported API then anyone should be able to call it.

This is not a boolean choice situation, there are 3 choices really - (i) supported, (ii) unsupported, (iii) supported only for visual-diagnostics tools. 😉 

This falls under choice (iii). We seem to be *testing* for '*is a visual-diagnostics tool?*' by proxy by asking the question 'is this a debugger?', which works reasonably well (but as @batzen points out, not always). 

Short term, reflection-hackable BindingDiagnostics.IsEnabled is probably the quickest solution as @spadapet has offered. And before we add the next visual-diag feature, it would be good to switch to a new interface for identifying visual-diag tools and using that (in place of 'is a debugger?' test).

spadapet:
> Short term, reflection-hackable BindingDiagnostics.IsEnabled is probably the quickest solution as @spadapet has offered. And before the WPF team adds the next visual-diag feature, it would be good to switch to a new interface for identifying visual-diag tools and using that (in place of 'is a debugger?' test).

@vatsan-madhavan Sounds good to me. I'm not trying to unify how all visual diagnostics work here, that's another task.

src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Diagnostics/BindingFailedEventArgs.cs
spadapet: @SamBent I have the change locally to do all the binding harvesting in the event args. It seems good to me, and I've been working with data binding failures enough to be confident that the harvested data won't need to change (at least for Visual Studio's feature). I'm tempted to put a WeakReference to the binding just in case...

SamBent: @spadapet I'm actually fine with (3), given that you aren't actually caching the bindings.  It might be more flexible if we decide to expose this to more tools.  

A WeakReference doesn't help much, as nothing prevents the listener from caching its target.  And I'm not too worried about the pending events, given your claim that the listener will arrive as soon as it can to collect them.   All the diagnostic features rely on the good will of the listeners anyway, so adding to the "good will" list is reasonable.  Mostly I wanted to know that you weren't (inadvertently) breaking that good will, and get you to document what "good will" means for this event, so that other tools will know how to behave.

spadapet: @SamBent Thanks, I'll stick with (1) then, and add a comment here. The diagnostics features already allow the app to be broken by messing with property values, so maybe there is already a warning.

src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Diagnostics/BindingDiagnostics.cs
```diff
@@ -0,0 +1,122 @@
...
+    /// <remarks>
+    /// This type supports the .NET Framework infrastructure and is not intended to be used directly
+    /// from application code.
+    /// </remarks>
+    public static class BindingDiagnostics
+    {
+        internal static bool IsEnabled { get; private set; }
+
+        private static event EventHandler<BindingFailedEventArgs> bindingFailed;
+        private static List<BindingFailedEventArgs> pendingEvents;
```
lindexi, line 26: Naming conventions issue

See https://github.com/dotnet/corefx/blob/master/Documentation/coding-guidelines/coding-style.md

We use `_camelCase` for internal and private fields and use readonly where possible. Prefix internal and private instance fields with`_`, static fields with `s_` and thread static fields with `t_`.

src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Diagnostics/BindingDiagnostics.cs
```diff
@@ -0,0 +1,122 @@
...
+        private static object pendingEventsLock;
```
lindexi, line 27: Should we make it as readonly?

src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Diagnostics/BindingDiagnostics.cs
```diff
@@ -0,0 +1,122 @@
...
+        /// </summary>
+        private static void FlushPendingBindingFailedEvents()
+        {
+            if (BindingDiagnostics.pendingEvents != null)
+            {
+                BindingFailedEventArgs[] pendingEventsCopy = null;
+
+                lock (BindingDiagnostics.pendingEventsLock)
+                {
+                    pendingEventsCopy = BindingDiagnostics.pendingEvents?.ToArray();
```
lindexi, line 76: Can we not do call ToArray ?

Because the BindingDiagnostics.pendingEvents will never change

chabiss:
This feature is not a production feature for people to consume. It's only meant to be a diagnostics feature when the debugger is attached. We're following the same guideline as we set for  VisualDiagnostics.IsEnabled 

---
In reply to: [692772985](https://github.com/dotnet/wpf/pull/3505#issuecomment-692772985) [](ancestors = 692772985)

src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Diagnostics/BindingDiagnostics.cs
spadapet: Thanks

src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Diagnostics/BindingDiagnostics.cs
spadapet: Absolutely

src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Diagnostics/BindingDiagnostics.cs
spadapet: Makes sense, thanks

SamBent (approved):

fabiant3 (approved):

vatsan-madhavan (approved):

PR #3537, opened 2020-09-18
WindowsDesktop should not support netcoreapp3.0

### LINKED ISSUE #3532: Winform should not support netcoreapp3.0
---
**Issue moved from dotnet/sdk#12799**
- Please respond to @wli3.

---

_From @wli3 on Thursday, August 6, 2020 5:05:26 AM_

image

RussKie:
Closes #3532

/cc: @wli3 @dsplaisted @Shyam-Gupta

### DIFF
--- a/packaging/Microsoft.NET.Sdk.WindowsDesktop/targets/Microsoft.NET.Sdk.WindowsDesktop.props
+++ b/packaging/Microsoft.NET.Sdk.WindowsDesktop/targets/Microsoft.NET.Sdk.WindowsDesktop.props
@@ -133,6 +133,7 @@
     <_UnsupportedNETCoreAppTargetFramework Include=".NETCoreApp,Version=v2.0" />
     <_UnsupportedNETCoreAppTargetFramework Include=".NETCoreApp,Version=v2.1" />
     <_UnsupportedNETCoreAppTargetFramework Include=".NETCoreApp,Version=v2.2" />
+    <_UnsupportedNETCoreAppTargetFramework Include=".NETCoreApp,Version=v3.0" />
 
     <!--
         Windows Forms and WPF are not supported an any .NET Standard targets


wli3:
Does **WPF** also not not support netcoreapp3.0?

RussKie:
WPF depends on Windows Forms, so if Windows Forms doesn't support it, I can't see how WPF can.
Besides why doesn't Windows Forms support .NET Core 3.0?

wli3:
I don't remember who exactly, someone from Winform team, maybe @Shyam-Gupta, mentioned it should not support netcoreapp3.0. I remember it is due to you can build the project from commandline, but the designer in VS does not really support it.

Shyam-Gupta:
Yes, WinForms .NET Core designer does not support netcoreapp3.0.

vatsan-madhavan:
This is not a good approach to solving this problem IMO. WindowsDesktop on .NET Core was supported on 3.0 onwards and we should not change that historical baseline for reasons that have nothing to do with that support actually failing to work (and I'm not talking about lifecycle policies here. It's not our practice to prune away out-of-lifecycle TFM's from all build targets, for example). 

WinForms Designer not supporting something should not be the standard for deciding whether the SDK would drop support for a whole TFM. Compat considerations alone should preclude us from droppping netcoreapp3.0. 

The WinForms designer should find a more surgical way to avoid loading netcoreapp3.0 projects. Invent a new property or Item that keeps track of what TFM's are supported by WinForms designer and pivot off of that, for example. 

/cc @richlander @terrajobst

dsplaisted:
To clarify, the proposal (and what this PR would do) is merely to remove the value from the target framework dropdown in VS.  It won't stop existing projects from building, nor from retargeting to .NET Core 3.0 by editing the TargetFramework in the project file.

vatsan-madhavan:
I see. 🤔 

It's unfortunate that the project-system uses an item named `SupportedTargetFramework` to make this sort of decision about designer. Would have been nice if showing/not-showing the designer was decided separately from project-system support. Maybe this is a new situation - wherein supported TFM's and supported-for-designer TFM's are no longer coincident. 

Ideally the project-system should not rely on `SupportedTargetFramework` item for designer-level decisions. The fact that `SupportedTargetFramework` item is being filled-in here in a .NET SDK component means that in the future it will inevitably get used as its name semantically implies - to calculate what frameworks are supported by the *SDK* - and the decisions that follow are likely going to be in some error. 😟 

In general, since this is a public property, and I'm not really in favor of changing this even for the convenience of project system/designer. The natural meaning conveyed by the name of the property should hold more sway here IMO. Besides, https://github.com/dotnet/sdk/blob/561558916a4c9de8bd603adf22fd419fbdcaddcb/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.SupportedTargetFrameworks.props also suggests that `SupportedTargetFramework` means just that - *supported TFMs* - and not something less impactful like *supported TFM's in the VS/project system (but you know more TFM's may be supported by the SDK itself)* 😉  

💡 I would recommend providing the project system with a new item - something like `SupportedTargetFrameworkForProjectSystem`. 

As a design approach here, I would suggest changing the internal item `_UnsupportedNETCoreAppTargetFramework`'s name to something more apt like `_UnsupportedNETCoreAppTargetFrameworkForProjectSystem`, and adding a new public item `SupportedTargetFrameworkForProjectSystem` .

RussKie:
I should have marked it as draft from the get go, until there's a consensus on the direction.

terrajobst:
@Shyam-Gupta 

> Yes, WinForms .NET Core designer does not support netcoreapp3.0.

What specifically happens when someone opens a `netcoreapp3.0` with the latest Visual Studio and the .NET 5 SDK? Will it generally crash? Will the designer crash? Are certain controls not supported in designer (e.g. the ones we removed in 3.1)?

Shyam-Gupta:
> What specifically happens when someone opens a `netcoreapp3.0` with the latest Visual Studio and the .NET 5 SDK? Will it generally crash? Will the designer crash? Are certain controls not supported in designer (e.g. the ones we removed in 3.1)?

WinForms .NET Core designer uses an external netcore process called as `DesignToolsServer` to instantiate .NET Core types. Target framework for this process is set to `.netcoreapp3.1`. Due to this when user opens a `netcoreapp3.0` WinForms project in VS and tries to launch designer, the `DesignToolsServer` process fails to load 3.0 assemblies. In VS it manifests as an error dialog: "Timed out while connecting to named pipe.". But VS doesn't crash. The 3.0 project does build and run fine.
 
The reason for targeting `DesignToolsServer` to `netcoreapp3.1` was removal of some of the WinForms controls in 3.1.

/cc: @merriemcgaw @DustinCampbell @OliaG

dipeshmsft:
Closing this PR as 3.1 is out of support now and it is no more displayed in the project system dropdown list for Winforms projects in VS.

Issue #2490, opened 2020-01-31
WPF Tree View Freeze / Hang


pjchristmas:
* .NET Framework 4.8 with [latest patch](https://support.microsoft.com/en-us/help/4534130/kb4534130-cumulative-update-for-net-framework) that claimed to fix similar sounding bug.
* Windows version: (`winver`): Win10 1803

 **Problem description:**
 Identical to issue [1962](https://github.com/dotnet/wpf/issues/1962)
but not fixed by [update listed there](https://support.microsoft.com/en-us/help/4534130/kb4534130-cumulative-update-for-net-framework).
We have a virtualized Tree View with up to 30k items. Going to the bottom and scrolling up 51 times (or using scroll wheel) will cause it to freeze. If I leave it up for an hour it will usually come back(or maybe it's just logging out and back in that kicks it out of the layout loop)
We've tried changing the virtualization settings (ScrollUnit, VirtualizationCacheLength, VirtualizationCacheLengthUnit and VirtualizationMode.Recycling) but none had any effect.

 **Minimal repro:**
Attached app, just modified the one from previous posting and added reproducing data.
[HangTest.zip](https://github.com/dotnet/wpf/files/4154025/HangTest.zip)

Looking for a hero @SamBent

SamBent:
@pjchristmas  - Just saw this, and looked at your repro just enough to confirm that it hangs, but for a different root cause than in 1962.    I don't know the root cause yet.

Can you speak to the business case for fixing this in .NET servicing?   How often does it happen, when did it start happening, what fraction of the data sets hit it, is there a reliable workaround - that sort of thing.  (You've alluded to some of the already, I just want to fill in more blanks.)

BTW, I didn't get an email six days ago when you tagged me.  Perhaps GitHub only sends email when you tag me in a new comment, not when you add the tag in an edit of an existing comment?

pjchristmas:
**Business case:** This is important to key products/customers at National Instruments. We will be elevating through service channels as well.
**Reproducing:** It happens every time (you do upward scrolling) for certain data sets. The particulars of the dataset are customer driven, and we've seen it in a number of their files, but not all. Some combination of size and shape of tree seem to matter.
**Workaround:** We don't have one yet. If you have ideas, we'd love to try them. Our datasets get big enough that turning off virtualization is not an option.

Thanks for looking into this @SamBent .

BradenCalderMurphy:
Hi, have there been any updates with regards to this?

[...]

PR #3564, opened 2020-09-24
Vsp freeze

SamBent:
Addresses #2490 
This is a port of a servicing fix in .NET 4.7/4.8.

**Issue:** Scrolling a TreeView hangs.

**Discussion:**
The customer's data is highly non-uniform, in the sense that a given node's children govern subtrees whose sizes are quite different.  This means the estimates VSP computes can differ wildly from the truth, and are subject to frequent (and sometimes violent) revision as more of the tree de-virtualizes.  This is not necessarily bad, but it stresses the algorithms in ways that uniform trees don't.  This stress uncovers two problems:

1.  VSP stores "effective offsets", intended to support correct layout after bottom-up size changes (e.g. a deep node gets larger because of new content), which shouldn't move the visible items.  When VSP revises its estimates, it also changes its effective offsets.  These are intended to stay in effect until the user scrolls, moving the visible items.  But VSP used an indirect way to detect this (namely the local viewport offset).  The non-uniform churn exposed a flaw - the viewport offset after a scroll can be the same as it was before (by coincidence, if estimates in nearby nodes change in just the wrong way).  If so, the layout algorithm will choose the same node as before to be at the top of the viewport, and the "anchor" logic that checks whether the scroll ended up in the right place will ask for a remeasure, which merely repeats the same futile actions.  Infinite loop.

2. A violent enough change can cause the "anchor" logic to think that the anchor node's scroll offset is outside the extent of the full tree.  This is because the former is computed bottom-up, and can include revised estimates that have not yet propagated to the latter (computed top-down).  When this happens, the anchor logic simply accepts the current state, and the TreeView scrolls to the wrong place (sometimes by a lot - I've seen examples that were off by 1000s of nodes).

The fix for (1) is to keep a "scroll generation" counter, incremented each time the top-level scroll offset changes.  Effective offsets remember the generation at which they were created, and are only applied during layout requests in the same generation.

The fix for (2) is to detect the bad situation and ask for a remeasure.

The customer's real app (but sadly not the sample repro they provided) suffers from a third problem.  
3. When UseLayoutRounding=true, the anchor logic can fail because it uses unrounded size estimates of never-been-devirtualized items, inconsistent with the layout algorithm that rounds everything.  As usual, this kind of failure leads to an infinite-remeasure hang, depends sensitively on the size and shape of the data and the history of scrolling and virtualization, and seems to be more likely in large non-uniform datasets.

The fix for (3) is to apply layout rounding to the estimated sizes.

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Controls/Primitives/HierarchicalVirtualizationConstraints.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Controls/Primitives/HierarchicalVirtualizationConstraints.cs
@@ -21,6 +21,7 @@ namespace System.Windows.Controls
             _cacheLength = cacheLength;
             _cacheLengthUnit = cacheLengthUnit;
             _viewport = viewport;
+            _scrollGeneration = 0;  // internal field set separately by caller
         }
 
         #endregion
@@ -131,11 +132,22 @@ namespace System.Windows.Controls
 
         #endregion
 
+        #region Internal properties
+
+        internal long ScrollGeneration
+        {
+            get { return _scrollGeneration; }
+            set { _scrollGeneration = value; }
+        }
+
+        #endregion
+
         #region Data
 
         VirtualizationCacheLength _cacheLength;
         VirtualizationCacheLengthUnit _cacheLengthUnit;
         Rect _viewport;
+        long _scrollGeneration;
 
         #endregion
     }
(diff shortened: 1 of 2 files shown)


src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Controls/Primitives/HierarchicalVirtualizationConstraints.cs
```diff
@@ -131,11 +132,22 @@ public override int GetHashCode()
...
+        }
+
+        #endregion
+
         #region Data
 
         VirtualizationCacheLength _cacheLength;
         VirtualizationCacheLengthUnit _cacheLengthUnit;
         Rect _viewport;
+        long _scrollGeneration;
```
lindexi, line 150: LOL

> If you do 1000 scroll operations per second, that will happen in about 280 million years.

Maybe we can use ulong  😂

fabiant3 (approved):

Issue #3100, opened 2020-06-08
WPF App crashes due to COMException (UCEERR_RENDERTHREADFAILURE) from System.Windows.Media.Composition.DUCE.Channel.SyncFlush


SamBent:
Problem description:  HostVisual interacts with its rendering channel on the wrong thread.  This can lead to render thread failures, perhaps minutes or hours after the bad interaction.

[Port of servicing fix for .NET 4.7-4.8]

BrannonKing:
Is this a duplicate of #128?

jakeyeagley:
I'm also seeing this issue. Here is my stack trace.

System.Runtime.InteropServices.COMException (0x88980406): UCEERR_RENDERTHREADFAILURE (Exception from HRESULT: 0x88980406)
   at MS.Internal.HRESULT.Check(Int32 hr)
   at System.Windows.Media.Composition.DUCE.Channel.SyncFlush()
   at System.Windows.Media.MediaContext.CompleteRender()
   at System.Windows.Media.MediaContext.CreateChannels()
   at System.Windows.Media.MediaSystem.ConnectChannels(MediaContext mc)
   at System.Windows.Media.MediaContext..ctor(Dispatcher dispatcher)
   at System.Windows.Media.MediaContext.From(Dispatcher dispatcher)
   at System.Windows.Media.MediaContext.RegisterICompositionTarget(Dispatcher dispatcher, ICompositionTarget iv)
   at System.Windows.Media.VisualTarget..ctor(HostVisual hostVisual)

SamBent:
SyncFlush is a catch-all for all render thread failures.  The callstack just shows how the main thread learned that the render thread failed, without telling us anything at all about the failure.

This issue addresses a very specific failure, caused by the HostVisual element doing work on the wrong thread.  It is not a dupe of #128, and is unlikely to be @jakeyeagley's issue (unless maybe you're using HostVisual).

jakeyeagley:
I am using HostVisual.

BrannonKing:
There's something wrong with .NET Framework 4.8.4200.0 -- that version specifically that came out just recently and was applied via Windows Update over the last few days. That's causing complete breakage of HostVisual, throwing this COMException error. It's these two specific patches: KB4569751 and KB4569745

SamBent:
[...]

PR #3567, opened 2020-09-24
HostVisual threading

SamBent:
Addresses #3100
This is a port of a servicing fix in .NET 4.7-4.8

**Issue:** HostVisual can disconnect from its rendering channel on the wrong thread.  This can result in corruption that can crash the render thread, usually in some unrelated activity many minutes (or hours) later.

**Discussion:**
HostVisual sometimes adds commands to the graphics channel on the wrong thread.  And sometimes this leaves the channel in a corrupt state, where the primary handle table releases a slot but the secondary handle table doesn't release the corresponding slot.  This is a time-bomb - the next time the primary table allocates the slot, the secondary table allocation will fail, which will crash the app at the next SyncFlush.  (It could take minutes or hours for this to happen.)

The fix is for HostVisual to marshal the offending work back to the right thread.  I'm also adding checks to detect wrong-thread channel activity, available in private builds.

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Media/HostVisual.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Media/HostVisual.cs
@@ -125,9 +125,16 @@ namespace System.Windows.Media
 
             using (CompositionEngineLock.Acquire())
             {
-                DisconnectHostedVisual(
-                    channel,
-                    /* removeChannelFromCollection */ true);
+                // if there's a pending disconnect, do it now preemptively;
+                // otherwise do the disconnect the normal way.
+                // This ensures we do the disconnect before calling base,
+                // as required.
+                if (!DoPendingDisconnect(channel))
+                {
+                    DisconnectHostedVisual(
+                        channel,
+                        /* removeChannelFromCollection */ true);
+                }
             }
 
             base.FreeContent(channel);
@@ -252,7 +259,7 @@ namespace System.Windows.Media
             //
             if (!(channel.IsSynchronous)
                 && _target != null
-                && !_connectedChannels.Contains(channel))
+                && !_connectedChannels.ContainsKey(channel))
             {
                 Debug.Assert(IsOnChannel(channel));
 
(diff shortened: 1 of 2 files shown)


src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Media/HostVisual.cs
```diff
@@ -382,23 +398,42 @@ private void DisconnectHostedVisualOnAllChannels()
...
-                // the ref count for VisualTarget.
-                //
-
-                _target._contentRoot.ReleaseOnChannel(channel);
-
-                SetFlags(channel, false, VisualProxyFlags.IsContentNodeConnected);
+                // Adding commands to a channel is not thread-safe,
+                // we must do the actual work on the same dispatcher thread
+                // where the connection happened.
+                if (CoreAppContextSwitches.HostVisualDisconnectsOnWrongThread ||
```
vatsan-madhavan, line 407: Adding more AppContext switches into .NET Core codebase is not ideal. Consider removing these.

SamBent: I thought I had removed this.  I must have messed up GIT somehow.  Fixed now.

vatsan-madhavan (approved):

src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Media/HostVisual.cs
```diff
@@ -382,23 +398,41 @@ private void DisconnectHostedVisualOnAllChannels()
...
+                               _target._contentRoot.GetHandle(channel),
+                               _target._contentRoot);
+                }
+                else
+                {
+                    // marshal to the right thread
+                    if (channelDispatcher != null)
+                    {
+                        DispatcherOperation op = channelDispatcher.BeginInvoke(
+                            DispatcherPriority.Normal,
```
lindexi, line 421: Should we use Render priority? I think it's a rendering task.

fabiant3 (approved):

Issue #3114, opened 2020-06-09
WPF is creating and formatting stack traces when removing HwndSource from visual tree


SamBent:
Removing an HwndSource from the visual tree exercises logic that was recently augmented to help track down an unexplained crash.  The new logic creates a stack trace, but does even in the normal case, causing a non-trivial perf hit for no reason.

[Port of servicing fix for .NET 4.8.]

SamBent:
Fix ported to 3.1.19, 5.0.0, 6.0.

PR #3574, opened 2020-09-24
Move StackTrace to error branch

SamBent:
Addresses #3114 
This is a port of a servicing fix in .NET 4.8

**Issue:**  Expensive diagnostic code is running on the normal code-path.

**Discussion:**
A recent servicing fix added code to capture a stack trace whenever a certain anomalous condition occurs.  However it always creates the stack trace, and merely discards it in the normal case.

Fixed by moving the creation of the stack trace into the anomalous branch.

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Interop/HwndHost.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Interop/HwndHost.cs
@@ -983,7 +983,6 @@ namespace System.Windows.Interop
                     // window later when a new parent is available
                     var hwnd = SystemResources.GetDpiAwarenessCompatibleNotificationWindow(_hwnd);
                     Debug.Assert(hwnd != null);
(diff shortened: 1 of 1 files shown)


fabiant3 (approved):

src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Interop/HwndHost.cs
```diff
@@ -998,6 +997,10 @@ private void BuildOrReparentWindow()
                         // shutdown event after this HwndHost.
                         SystemResources.DelayHwndShutdown();
                     }
+                    else
+                    {
+                        Trace.WriteLineIf(hwnd == null, $"- Warning - Notification Window is null\n{new System.Diagnostics.StackTrace(true).ToString()}");
```
lindexi, line 1002: I do not think we should check `hwnd` is null twice. Because only when hwnd is null, will it enter this branch.

PR #3594, opened 2020-09-29
Fix the casing of 'Shared' in WpfSharedDir

FraGag:
The value was specified with a lower case 's', but the directory in Git is spelt with a capital 'S'. This discrepancy causes failures on source server lookups. For example, stepping into `MS.Utility.FrugalList` tries to load (using .NET 5.0 RC1):

https://raw.githubusercontent.com/dotnet/wpf/be89bbfaded6ad3f0f4143bf1a8dd075a5789715/src/Microsoft.DotNet.Wpf/src/shared/MS/Utility/FrugalList.cs

but the server returns a 404 response. The source is actually at:

https://raw.githubusercontent.com/dotnet/wpf/be89bbfaded6ad3f0f4143bf1a8dd075a5789715/src/Microsoft.DotNet.Wpf/src/Shared/MS/Utility/FrugalList.cs

### DIFF
--- a/eng/WpfArcadeSdk/tools/FolderPaths.targets
+++ b/eng/WpfArcadeSdk/tools/FolderPaths.targets
@@ -1,7 +1,7 @@
 <Project>
   <PropertyGroup>
     <!-- $(WpfSharedDir) is present in the current repository -->
-    <WpfSharedDir Condition="'$(WpfSharedDir)'=='' And Exists('$(WpfSourceDir)shared\')">$(WpfSourceDir)shared\</WpfSharedDir>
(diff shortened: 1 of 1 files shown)


vatsan-madhavan:
This is a benign change. Some of these properties obviously need to be case-sensitive since we rely on them to represent file names in projects etc. 

👍

vatsan-madhavan (approved):

vatsan-madhavan:
@dotnet/wpf-developers  to fix these PR build breaks, do the following - 

- Add `<RestoreProjectStyle>None</RestoreProjectStyle>` to `eng\wpfarcadesdk\tools\wpf.cpp.props`, somewhere around L28. (just after after `<BuildProjectReferences>true</BuildProjectReferences>`)

PR #3603, opened 2020-10-02
[PresentationCore] [Performance] [System.Windows.Media]

Mikolaytis:
Optimize memory allocations on Geometry parsing.
We are using a lot of geometries in our app that are located in the resource dictionary. I've found some unnecessary string allocations on parsing and fixed them. 

[Profiler Screenshot]
Screenshot 2020-10-02 at 12 31 50

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Media/ParsersCommon.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Media/ParsersCommon.cs
@@ -435,11 +435,11 @@ namespace MS.Internal.Markup
             }
             else
             {
-                string subString = _pathString.Substring(start, _curIndex - start);
+                ReadOnlySpan<char> slice = _pathString.AsSpan().Slice(start, _curIndex - start);
 
                 try
                 {
-                    return System.Convert.ToDouble(subString, _formatProvider);
+                    return double.Parse(slice, provider: _formatProvider);
                 }
                 catch (FormatException except)
                 {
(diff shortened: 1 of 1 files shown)


src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Media/ParsersCommon.cs
```diff
@@ -435,11 +435,11 @@ void SkipDigits(bool signAllowed)
             }
             else
             {
-                string subString = _pathString.Substring(start, _curIndex - start);
+                ReadOnlySpan<char> slice = _pathString.AsSpan().Slice(start, _curIndex - start);
```
gfoidl, line 438:
```suggestion
                ReadOnlySpan<char> slice = _pathString.AsSpan(start, _curIndex - start);
```

src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Media/ParsersCommon.cs
```diff
@@ -435,11 +435,11 @@ void SkipDigits(bool signAllowed)
             }
             else
             {
-                string subString = _pathString.Substring(start, _curIndex - start);
+                ReadOnlySpan<char> slice = _pathString.AsSpan(start, _curIndex - start);
 
                 try
                 {
-                    return System.Convert.ToDouble(subString, _formatProvider);
+                    return double.Parse(slice, provider: _formatProvider);
```
gfoidl, line 442: Not directly related to your change, but use `double.TryParse` and throw in the false-branch?
Thus avoiding the use of exceptions (has runtime cost) for control-flow?

Mikolaytis: I agree. It would be better. I've thought about this, but didn't do this because I want my PR to be as simple as possible.

gfoidl: 👍 makes sense. We can do this is a follow-up PR.

dipeshmsft:
@Mikolaytis, your changes in this PR were subsumed in another PR #6272. Hence, closing the PR.

PR #3637, opened 2020-10-12
[release/5.0] Revert path change to markup compiler

ryalanms:
## [release/5.0] Revert path change to markup compiler

AppCompat testing revealed a bug in the markup compiler due to the path change.  Revert the change.  (It can be reapplied later.)

#### Original PR:

https://github.com/dotnet/wpf/pull/3120


SamBent (approved):

clairernovotny:
Would that cause errors like this with the nightly:
```
D:\a\1\s\PackageExplorer\obj\Release\net5.0-windows10.0.19041\win-x64\AboutWindow.g.cs(45,18): error CS1504: Source file '..\..\..\..AboutWindow.xaml' could not be opened -- Could not find file. [D:\a\1\s\PackageExplorer\NuGetPackageExplorer.csproj]
```

ryalanms:
@clairernovotny:  Yes, that is due to this change.

clairernovotny:
@ryalanms given that was for the deterministic build work, when are we aiming for that up to be fixed?

ryalanms:
It would be in a servicing release: 5.01.  The PackageReference change needs to go in as well.  

I'm waiting to hear back on how to get that in to VS 16.9 Preview 1.

ryalanms:
@clairernovotny:  From the internal thread, this will be in VS 16.9 Preview 2.  I don't have an exact date but the estimate is a few weeks after 5.0.  Thanks.

PR #3659, opened 2020-10-16
Temporarily suppress obsolete API errors to get WPF master building

ryalanms:
WPF is using obsolete APIs that now cause compilation errors:

WebRequest.WebRequest()
WebRequest.Create(Uri)
Uri.EscapeUriString(string)

As discussed with @SamBent, temporarily disable the obsolete APIs to get WPF master building again.  These need to be replaced in 6.0.  

/cc @mmitche


ryalanms:
Obsolete API work tracked here: #3660.

Issue #811, opened 2019-06-03
When including any Page item via NuGet package props/targets files, all included Compile items will lost.


walterlv:
* .NET Core Version: 3.0 Preview5
* Windows version: 10.0.18362.145 (1903)
* Does the bug reproduce also in WPF for .NET Framework 4.8?: Yes

 **Problem description:**

1. Build a NuGet package with both XAML and C# files in.
1. Add Package.targets file into the NuGet package build folder and include the XAML files as `Page` and C# files as `Compile`.
1. When Installing this package, we'll get a compiling error.
 
 **Actual behavior:** 
 
1. If we check the compiling message of the `CoreCompile` target, we'll find that the `Compile` files we just included are not in the file list.
1. If we delete all the `Page` items, the `Compile` items will come back normal.

 **Expected behavior:**
 
No `Compile` file is lost if we include `Page` items.

 **Minimal repro:**

1st. Build a NuGet package with such a file structure:

```
- build
    - DemoPackage.targets
+ buildMultiTargeting
+ lib
- src
    - DemoView.xaml
    - DemoView.xaml.cs
```

In the DemoPackage.targets file, we have such the codes:

```xml
<Project>

  <PropertyGroup>
    <MSBuildAllProjects>$(MSBuildAllProjects);$(MSBuildThisFileFullPath)</MSBuildAllProjects>
  </PropertyGroup>

  <ItemGroup>
    <!-- I use the absolute path to avoid the slide effect of the property value evaluation. -->
    <Compile Include="C:\Users\lvyi\.nuget\packages\demopackage\0.1.0\src\DemoView.xaml.cs" />
    <Page Include="C:\Users\lvyi\.nuget\packages\demopackage\0.1.0\src\DemoView.xaml" />
  </ItemGroup>

</Project>
```

2nd. When we install this package into another empty WPF project just as the csproj below, we can't compile it correctly.

```xml
<Project Sdk="Microsoft.NET.Sdk.WindowsDesktop">
  
  <PropertyGroup>
    <OutputType>WinExe</OutputType>
    <TargetFramework>netcoreapp3.0</TargetFramework>
    <UseWPF>true</UseWPF>
  </PropertyGroup>

  <ItemGroup>
    <PackageReference Include="demopackage" Version="0.1.0" />
  </ItemGroup>

</Project>
```

3rd. Check the compiling message of the `CoreCompile` target.

[...]

PR #3585, opened 2020-09-25
Support PackageReferences in WPF projects

ryalanms:

PR #3846, opened 2020-11-23
[release/5.0] Support Source Generators in WPF projects

ryalanms:
### [release/5.0] Support Source Generators in WPF projects

*This feature requires applications to **opt-in** by explicitly setting a property in their project file.  Otherwise, the existing GenerateTemporaryTargetAssembly code is executed.*

This PR brings the changes from master to release/5.0: 

> https://github.com/dotnet/roslyn/blob/master/docs/features/source-generators.md

Original PR: 

> https://github.com/dotnet/wpf/pull/3585

> * Include generated NuGet props and targets from outer WPF parent project in generated temp project
> * Add condition and cleanup generated NuGet props and targets after compilation of temp project
> * Move BuildEngine global properties to project properties

The source generator samples now work in WPF applications:  

> https://github.com/dotnet/roslyn-sdk/tree/master/samples/CSharp/SourceGenerators

A single property turns on the new PackageReference support `(IncludePackageReferencesDuringMarkupCompilation = true)` and enables the new behavior.  By default, the unmodified GenerateTemporaryTargetAssembly code is called, the *old* path.

There was an incomplete change previously committed and reverted.  The new change includes:

> * Relative path fixes from previous commit: https://github.com/dotnet/wpf/pull/3120
> * Addition of `ResolveProjectReferences` target when `IncludePackageReferencesDuringMarkupCompilation` is not false
> * Replacement of BuildEngine global properties with project properties that caused build failures in referenced projects and various other scenarios (e.g. `AssemblyName`)


src/Microsoft.DotNet.Wpf/src/PresentationBuildTasks/MS/Internal/MarkupCompiler/PathInternal.cs
```diff
@@ -0,0 +1,382 @@
...
+
+using System.Threading;
+using MS.Internal.Markup;
+using MS.Internal.Tasks;
+using MS.Utility;   // for SR
+using Microsoft.Build.Utilities;
+using Microsoft.Build.Tasks.Windows;
+using System.Runtime.CompilerServices;
+
+namespace MS.Internal
```
fabiant3, line 44: This entire file is a port and not written from scratch, correct?

ryalanms: Yes, that's correct: the methods in PathInternal.cs are ported and modified as little as possible.  (The behavior in PresentationBuildTasks must be the same between .NET Framework and .NET Core.)

fabiant3 (approved):

src/Microsoft.DotNet.Wpf/src/PresentationBuildTasks/Microsoft.WinFX.targets
```diff
@@ -397,6 +404,59 @@
...
+      <Message Text="MSBuildProjectExtensionsPath: $(MSBuildProjectExtensionsPath)" />
+
+      <!-- Get the name of the outer WPF parent project -->
+      <PropertyGroup>
+          <CurrentProjectName>$([System.IO.Path]::GetFileName('$(MSBuildProjectFullPath)'))</CurrentProjectName>
+      </PropertyGroup>
+
+      <!-- Collect the generated NuGet files from the parent project required support PackageReferences -->
+      <ItemGroup>
+        <SourceGeneratedNuGetPropsAndTargets Include="$(MSBuildProjectExtensionsPath)\*.props"/>
```
lindexi, line 442: Should we only include the `*.nuget.g.props` files?

lindexi:
Can this fix https://github.com/dotnet/wpf/issues/811 ?

Looks like it can fix the source code nuget packages that package by SourceYard: https://github.com/dotnet-campus/SourceYard

SamBent (approved):

src/Microsoft.DotNet.Wpf/src/PresentationBuildTasks/Microsoft.WinFX.targets
```diff
@@ -377,18 +380,22 @@
...
+                     GenerateTemporaryTargetAssembly2;
                      MarkupCompilePass2;
                      AfterMarkupCompilePass2;
                      CleanupTemporaryTargetAssembly
        </MarkupCompilePass2ForMainAssemblyDependsOn>
 
-       <_CompileTargetNameForLocalType Condition="'$(_CompileTargetNameForLocalType)' == ''">_CompileTemporaryAssembly</_CompileTargetNameForLocalType>
+        <_CompileTargetNameForLocalType Condition="'$(_CompileTargetNameForLocalType)' == ''">_CompileTemporaryAssembly</_CompileTargetNameForLocalType>
 
+        <_ResolveProjectReferencesTargetName Condition="'$(ShouldWPFSupportPackageReferences)' == 'true'">ResolveProjectReferences</_ResolveProjectReferencesTargetName>
```
vatsan-madhavan, line 393: `ShouldWPFSupport...` reads like a C# variable name and not like an MSBuild Property. Consider renaming it to `IncludePackageReferencesDuringMarkupCompilation`

vatsan-madhavan (commented):
I seem to recall that this PR was merged, then reverted, and is now being brought back. If this recollection is correct, then perhaps an explanation about what went wrong, and how the problem has been remedied in this version (besides providing an opt-out) would be a helpful to incorporate in the description.

src/Microsoft.DotNet.Wpf/src/PresentationBuildTasks/Microsoft.WinFX.targets
ryalanms: Updated.  Thanks.

ryalanms:
The description has been updated to include the differences between this and the previous PR.  Thanks.

src/Microsoft.DotNet.Wpf/src/PresentationBuildTasks/Microsoft.WinFX.targets
```diff
@@ -377,18 +380,22 @@
...
+                     GenerateTemporaryTargetAssembly2;
                      MarkupCompilePass2;
                      AfterMarkupCompilePass2;
                      CleanupTemporaryTargetAssembly
        </MarkupCompilePass2ForMainAssemblyDependsOn>
 
-       <_CompileTargetNameForLocalType Condition="'$(_CompileTargetNameForLocalType)' == ''">_CompileTemporaryAssembly</_CompileTargetNameForLocalType>
+        <_CompileTargetNameForLocalType Condition="'$(_CompileTargetNameForLocalType)' == ''">_CompileTemporaryAssembly</_CompileTargetNameForLocalType>
 
+        <_ResolveProjectReferencesTargetName Condition="'$(IncludePackageReferencesDuringMarkupCompilation)' == 'true'">ResolveProjectReferences</_ResolveProjectReferencesTargetName>
```
vatsan-madhavan, line 393: I should have caught this the first time around... 

For public properties, strongly consider using negative tests like `"'$(IncludePackageReferencesDuringMarkupCompilation)' != 'false'"`

Inevitably, these properties have a tendency to be used somewhere else besides the props/targets file that initially declared it. Then you'll have to *initialize* it everywhere, which tends to be another common problem. 

The safest pattern for public properties, if you can get away with it is...

- No initialization
- Negative tests ( `!= false`) 

You don't really have to drop the initialization in this file if you feel really attached to it (it's somewhat benign, and has a mildly self-documenting value), but I don't see a strong value there.

src/Microsoft.DotNet.Wpf/src/PresentationBuildTasks/Microsoft.WinFX.targets
vatsan-madhavan: Good question - why would this not be limited to nuget specific props here?

src/Microsoft.DotNet.Wpf/src/PresentationBuildTasks/MS/Internal/MarkupCompiler/PathInternal.cs
```diff
@@ -0,0 +1,382 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+//---------------------------------------------------------------------------
+//
+// Description:
+//   Returns a relative path from one path to another.
```
vatsan-madhavan, line 8: Could you comment and add where this source came from (repo, sha etc.?) so that future maintenance would be easier?

vatsan-madhavan (commented):
The question about the glob-pattern looks interesting and should be resolved before merging - maybe it's nothing; and a few maintainability nits best taken care of now. otherwise looks good.

src/Microsoft.DotNet.Wpf/src/PresentationBuildTasks/Microsoft.WinFX.targets
ryalanms: No, it should only take the *.nuget.g.* files which should include the correct references to any other props/targets files for msbuild extensions.   

```xml
   <SourceGeneratedNuGetPropsAndTargets Include="$(MSBuildProjectExtensionsPath)\*.nuget.g.props"/>
   <SourceGeneratedNuGetPropsAndTargets Include="$(MSBuildProjectExtensionsPath)\*.nuget.g.targets"/>
```

```
<name>.csproj.nuget.dgspec.json
<name>.csproj.nuget.g.props
<name>.csproj.nuget.g.targets
```

It should only be picking up these generated files.

ryalanms: (I've made the update.)

src/Microsoft.DotNet.Wpf/src/PresentationBuildTasks/Microsoft.WinFX.targets
```diff
@@ -432,6 +492,46 @@
...
+                 >
+
+       </GenerateTemporaryTargetAssembly2>
+
+       <CreateItem Include="$(IntermediateOutputPath)$(TargetFileName)" >
+               <Output TaskParameter="Include" ItemName="AssemblyForLocalTypeReference" />
+       </CreateItem>
+
+       <Message Text="Deleting generated NuGet props and targets files from the temporary project: @(_DestinationGeneratedNuGetPropsAndTargets)"/>
+       <Delete Files="@(_DestinationGeneratedNuGetPropsAndTargets)"/>
```
rainersigwald, line 533:
```suggestion
       <Delete Files="@(_DestinationGeneratedNuGetPropsAndTargets)" Condition=" '$(GenerateTemporaryTargetAssemblyDebuggingInformation)' != 'true' " />
```

?

src/Microsoft.DotNet.Wpf/src/PresentationBuildTasks/Microsoft.WinFX.targets
```diff
@@ -377,18 +380,22 @@
   <PropertyGroup>
 
        <MarkupCompilePass2ForMainAssemblyDependsOn>
+                     CreateTemporaryTargetAssemblyProject;
                      GenerateTemporaryTargetAssembly;
+                     GenerateTemporaryTargetAssembly2;
```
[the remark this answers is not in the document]
vatsan-madhavan: Good catch. This came as a pre 5.0 change initially and rules like "don't introduce publics unless unavoidable" went into a blindspot for previous reviewers like me... fresh reviewers are great for these.

src/Microsoft.DotNet.Wpf/src/PresentationBuildTasks/Microsoft.WinFX.targets
```diff
@@ -432,6 +492,46 @@
                <Output TaskParameter="Include" ItemName="AssemblyForLocalTypeReference" />
        </CreateItem>
 
+  </Target>
+
+    <!--
+
+                ==========================================
+                       GenerateTemporaryTargetAssembly2
```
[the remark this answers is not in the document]
ryalanms: The target/task has been renamed to _CompileTemporaryTargetAssembly/CompileTemporaryTargetAssembly.

src/Microsoft.DotNet.Wpf/src/PresentationBuildTasks/MS/Internal/MarkupCompiler/PathInternal.cs
ryalanms: > //   Ported to .NET Framework from the GetRelativePath implementation in:
> //
> //   https://github.com/dotnet/corefx/blob/b123ba4b9107c73cbc02010dc1ee78eb8ffccb93/src/Common/src/CoreLib/System/IO/Path.cs
> //
> //   This ported method ('GetRelativePath') is only called by the .NET Framework 
> //   version of PresentationBuildTasks.dll that ships with the Microsoft.NET.Sdk.WindowsDesktop 
> //   SDK as part of .NET 5.0.

src/Microsoft.DotNet.Wpf/src/PresentationBuildTasks/Microsoft.WinFX.targets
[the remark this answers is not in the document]
ryalanms: Thanks.  One possible solution is to use the GenerateTemporaryTargetAssembly target as the entry point for both the old and new behavior, so applications hooking GenerateTemporaryTargetAssembly continue to work.  Any objections to that approach?  

For the consolidation in 6.0, there is already discussion of removing the temp project generation and compilation entirely and instead using source generators during the main application compilation.  But if we do keep the temp project, the idea would be to move the new behavior in to the GTTA target and GTTA task.  The GTTA task would need to create the temporary project, copy and rename the parent project's generate NuGet props/targets, and compile the temporary project.  WPF applications that call the public GTTA task directly should continue to work.  In other words, all the logic from the CreateTemporaryTargetASsemblyProject task and CompileTemporaryTargetAssemblyProject task would be moved in to the GTTA task.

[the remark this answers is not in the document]
ryalanms: I'll go ahead with moving everything under a common GenerateTemporaryTargetAssembly target.  We can roll it back if there are objections.  Thanks.

src/Microsoft.DotNet.Wpf/src/PresentationBuildTasks/Microsoft/Build/Tasks/Windows/CreateTemporaryTargetAssemblyProject.cs
```diff
@@ -0,0 +1,414 @@
...
+            if (xmlProjectDoc == null || properties == null )
+            {
+                // When the parameters are not valid, simply return it, instead of throwing exceptions.
+                return;
+            }
+
+            XmlNode root = xmlProjectDoc.DocumentElement;
+
+            // Create a new ItemGroup element
+            XmlNode nodeItemGroup = xmlProjectDoc.CreateElement("PropertyGroup", root.NamespaceURI);
```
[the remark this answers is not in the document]
ryalanms: Yes, this is required because the PackageReference NuGet props/targets can hook in to the temp project build and build other projects, etc. after the temp project has defined immutable global proeprties.  These globals interfere with other properties of the same name and cause unexpected behavior with the build.  One of them was 'AssemblyName': very likely to collide with other projects.

src/Microsoft.DotNet.Wpf/src/PresentationBuildTasks/Microsoft.WinFX.targets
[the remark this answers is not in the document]
ryalanms: As the two of you pointed out, there are scenarios that may break due to this change: 

1. Wpf application hooks the GTTA (GenerateTemporaryTargetAssembly) target
2. Wpf application calls the public GTTA target directly
3. Wpf applications calls the public GTTA task directly

To support 1), the GTTA target must always be called during the build; for 2), The GTTA target must exist with the old behavior; for 3), the GTTA task must remain as-is (old behavior).

These should continue to work when compiling existing applications without having to set properties explicitly.  

And for consolidation to a single task/target in .NET 6.0 (and to avoid breaking applications written between 5.0.2 and 6.0 by removing public targets/tasks) no new public targets/tasks should be created, if possible.   So the new behavior should placed directly in the the existing GTTA task or should be easily movable in to the existing GTTA task.

A possible change that supports the scenarios and allows for removal of the legacy behavior in .NET 6.0 is here: https://github.com/dotnet/wpf/commit/9e0a1f4d7f779f314b1c9acf46f2dd61c3c2b85e

The new tasks have been removed and everything has been moved in to GTTA.  If this seems reasonable I'll move it over to this PR.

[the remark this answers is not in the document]
ryalanms: This is resolved with the most recent change.  Thanks.

src/Microsoft.DotNet.Wpf/src/PresentationBuildTasks/Microsoft/Build/Tasks/Windows/CreateTemporaryTargetAssemblyProject.cs
```diff
@@ -0,0 +1,414 @@
...
+using MS.Internal.Tasks;
+
+// Since we disable PreSharp warnings in this file, PreSharp warning is unknown to C# compiler.
+// We first need to disable warnings about unknown message numbers and unknown pragmas.
+#pragma warning disable 1634, 1691
+
+namespace Microsoft.Build.Tasks.Windows
+{
+    /// <summary>
+    ///   This task is used to generate a temporary target assembly project.
```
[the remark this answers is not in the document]
ryalanms: Everything is consolidated in the most recent change.  Thanks.

ryalanms:
@rainersigwald:  Thank you for reviewing.  I believe all the design suggestions were accommodated here.  Everything has been consolidated in to a single task and target.

src/Microsoft.DotNet.Wpf/src/PresentationBuildTasks/Microsoft.WinFX.targets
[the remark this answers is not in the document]
ryalanms: ResolveProjectReferences resolves the refererences in the PackageReferences.  There doesn't appear to be a specific target for PackageReferences.

ryalanms:
This was approved by .NET Tactics today for 5.0.2.  Thanks.

PR #3926, opened 2020-12-24
Fix a wrong property name in a comment

ForNeVeR:
Current wording could create a false assumption that `UseWinForms` is a supported property name, while it really isn't.


ryalanms:
Thanks, @ForNeVeR.

PR #3932, opened 2020-12-27
Replace the Dictionary with HashSet in AppDomainShutdownMonitor

lindexi:
Benchmark:

|             Method |              argList |     Mean |    Error |   StdDev | Ratio | RatioSD |
|------------------- |--------------------- |---------:|---------:|---------:|------:|--------:|
| 'AddAndRemove Old' | Syste(...)ence] [55] | 679.7 us | 10.42 us |  8.70 us |  1.00 |    0.00 |
|                    |                      |          |          |          |       |         |
| 'AddAndRemove New' |  Syst(...)er]] [113] | 667.7 us | 11.07 us | 10.35 us |     ? |       ? |

|                Method |              argList |     Mean |     Error |    StdDev | Ratio | RatioSD |
|---------------------- |--------------------- |---------:|----------:|----------:|------:|--------:|
| 'AddBeforeRemove Old' | Syste(...)ence] [55] | 1.414 ms | 0.0174 ms | 0.0154 ms |  1.00 |    0.00 |
|                       |                      |          |           |           |       |         |
| 'AddBeforeRemove New' |  Syst(...)er]] [113] | 1.197 ms | 0.0097 ms | 0.0081 ms |     ? |       ? |

|        Method |     Mean |     Error |    StdDev |   Median | Ratio | RatioSD |
|-------------- |---------:|----------:|----------:|---------:|------:|--------:|
| 'AddOnly Old' | 2.886 ms | 0.0574 ms | 0.1160 ms | 2.821 ms |  1.00 |    0.00 |
| 'AddOnly New' | 2.880 ms | 0.0639 ms | 0.1865 ms | 2.827 ms |  1.00 |    0.09 |

Benchmark Code: https://github.com/lindexi/lindexi_gd/tree/b63ba44c2077bcc4136455f16b309fe74a288dcc/AppDomainShutdownMonitor

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/Shared/MS/Internal/AppDomainShutdownMonitor.cs
+++ b/src/Microsoft.DotNet.Wpf/src/Shared/MS/Internal/AppDomainShutdownMonitor.cs
@@ -30,57 +30,53 @@ namespace MS.Internal
         {
             AppDomain.CurrentDomain.DomainUnload += OnShutdown;
             AppDomain.CurrentDomain.ProcessExit += OnShutdown;
-
-            _dictionary = new Dictionary<WeakReference, WeakReference>();
         }
-        
-        public static void Add(WeakReference listener)
-        {
-            Debug.Assert(listener.Target != null);
-            Debug.Assert(listener.Target is IAppDomainShutdownListener);
 
-            lock (_dictionary)
+        public static void Add(WeakReference<IAppDomainShutdownListener> listener)
+        {
+            Debug.Assert(listener.TryGetTarget(out _));
+
+            lock (_hashSet)
             {
                 if (!_shuttingDown)
                 {
-                    _dictionary.Add(listener, listener);
+                    _hashSet.Add(listener);
                 }
             }
         }
 
-        public static void Remove(WeakReference listener)
+        public static void Remove(WeakReference<IAppDomainShutdownListener> listener)
         {
-            Debug.Assert(listener.Target == null || listener.Target is IAppDomainShutdownListener);
-
-            lock (_dictionary)
+            lock (_hashSet)
             {
                 if (!_shuttingDown)
                 {
-                    _dictionary.Remove(listener);
+                    _hashSet.Remove(listener);
                 }
             }
         }
 
         private static void OnShutdown(object sender, EventArgs e)
         {
-            lock (_dictionary)
-            {     
+            lock (_hashSet)
+            {
                 // Setting this to true prevents Add and Remove from modifying the list. This
                 // way we call out without holding a lock (which would be bad)
                 _shuttingDown = true;
             }
-            
-            foreach (WeakReference value in _dictionary.Values)
+
+            foreach (var weakReference in _hashSet)
             {
-                IAppDomainShutdownListener listener = value.Target as IAppDomainShutdownListener;
-                if (listener != null)
+                if (weakReference.TryGetTarget(out var listener))
                 {
                     listener.NotifyShutdown();
                 }
             }
         }
-        
-        private static Dictionary<WeakReference, WeakReference> _dictionary;
+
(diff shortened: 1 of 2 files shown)


src/Microsoft.DotNet.Wpf/src/Shared/MS/Internal/AppDomainShutdownMonitor.cs
```diff
@@ -30,57 +30,53 @@ static AppDomainShutdownMonitor()
...
+            lock (_hashSet)
+            {
                 // Setting this to true prevents Add and Remove from modifying the list. This
                 // way we call out without holding a lock (which would be bad)
                 _shuttingDown = true;
             }
-            
-            foreach (WeakReference value in _dictionary.Values)
+
+            foreach (var weakReference in _hashSet)
```
miloush, line 68: perhaps we should keep the coding style consistent i.e. not use `var` here

src/Microsoft.DotNet.Wpf/src/Shared/MS/Internal/AppDomainShutdownMonitor.cs
```diff
@@ -30,57 +30,53 @@ static AppDomainShutdownMonitor()
...
+                if (weakReference.TryGetTarget(out var listener))
                 {
                     listener.NotifyShutdown();
                 }
             }
         }
-        
-        private static Dictionary<WeakReference, WeakReference> _dictionary;
+
+        private static readonly HashSet<WeakReference<IAppDomainShutdownListener>> _hashSet =
```
miloush, line 77: this isn't (and wasn't) a very helpful name and does not convey the purpose of the field (demonstrated by the need to rename it in this case), maybe `_listeners` would be better?

src/Microsoft.DotNet.Wpf/src/Shared/MS/Internal/AppDomainShutdownMonitor.cs
```diff
@@ -30,57 +30,53 @@ static AppDomainShutdownMonitor()
...
-            _dictionary = new Dictionary<WeakReference, WeakReference>();
         }
-        
-        public static void Add(WeakReference listener)
+
+        public static void Add(WeakReference<IAppDomainShutdownListener> listener)
         {
-            Debug.Assert(listener.Target != null);
-            Debug.Assert(listener.Target is IAppDomainShutdownListener);
+            Debug.Assert(listener.TryGetTarget(out _));
```
miloush, line 37: I don't see why this check is necessary or even valid, as the target can become null any time after...

src/Microsoft.DotNet.Wpf/src/Shared/MS/Internal/AppDomainShutdownMonitor.cs
miloush: moving the initialization from constructor to initializer looks like a deliberate code style change, although I am less bothered by this one, especially as it is more correct to initialize the storage before subscribing to events

src/Microsoft.DotNet.Wpf/src/Shared/MS/Internal/AppDomainShutdownMonitor.cs
lindexi: @miloush Thank you and it is just a debug code

src/Microsoft.DotNet.Wpf/src/Shared/MS/Internal/AppDomainShutdownMonitor.cs
lindexi: @miloush Thank you and I fixed it.

ryalanms (approved):

ryalanms:
Thanks again, @lindexi.

PR #3967, opened 2021-01-09
Add OpenFolderDialog

### LINKED ISSUE #438: WPF alternative for WinForms FolderBrowserDialog

Related links:

Symbai:
Fixes #438

Usage:
```C#
var dialog = new OpenFolderDialog();
if (dialog.ShowDialog() != true) return;
MessageBox.Show(dialog.FileName);
```
```C#
var dialog = new OpenFolderDialog { Multiselect = true };
if (dialog.ShowDialog() != true) return;
foreach (var directory in dialog.FileNames)
{
    MessageBox.Show(directory);
}
```

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/PresentationFramework/Microsoft/Win32/OpenFolderDialog.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationFramework/Microsoft/Win32/OpenFolderDialog.cs
@@ -5,7 +5,7 @@
 //
 // 
 // Description:
-//              OpenFolderDialog is a sealed class derived from CommonDialog that
+//              OpenFolderDialog is a sealed class that
 //              implements Folder Open dialog-specific functions. 
 //
 
--- a/src/Microsoft.DotNet.Wpf/src/PresentationFramework/Microsoft/Win32/OpenFolderDialog.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationFramework/Microsoft/Win32/OpenFolderDialog.cs
@@ -1,0 +1,645 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+//
+// 
+// Description:
+//              OpenFolderDialog is a sealed class derived from CommonDialog that
+//              implements Folder Open dialog-specific functions. 
+//
+
+
+namespace PresentationFramework.Microsoft.Win32
+{
+    using System;
+    using System.Runtime.InteropServices;
+    using System.Windows.Interop;
+    using System.Windows;
+    using System.IO;
+    using System.Linq;
+    using System.Runtime.CompilerServices;
+
+    public sealed class OpenFolderDialog
+    {
+
+        //---------------------------------------------------
+        //
+        // Public methods
+        //
+        //---------------------------------------------------
+        #region Public methods
+        public Nullable<bool> ShowDialog() => ShowDialog(null);
+
+        public Nullable<bool> ShowDialog(Window window)
+        {
+            INativeFileOpenDialog openDialogCoClass = new INativeFileOpenDialog();
+            try
+            {
+                ApplyNativeSettings(openDialogCoClass);
+                window ??= Helpers.GetDefaultOwnerWindow();
+                int hresult = openDialogCoClass.Show(GetHandleFromWindow(window));
+                if (ErrorHelper.Matches(hresult, Win32ErrorCode.ERROR_CANCELLED))
+                    return false;
+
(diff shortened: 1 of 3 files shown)


miloush:
Sorry, this is just a bad copy&paste implementation that does not fit into the framework.

Symbai:
> Sorry, this is just a bad copy&paste implementation that does not fit into the framework.

This was requested since nearly 2 years. Nobody bothered to add a PR for it, including yourself. Instead of downvoting my PR and calling it "bad" without giving any argument, you should make it better then. At least I'm trying to solve the problem, what's your excuse?

ryalanms:
Thanks, @Symbai and @miloush.  We will review this during the next triage session.

src/Microsoft.DotNet.Wpf/src/PresentationFramework/Microsoft/Win32/OpenFolderDialog.cs
```diff
@@ -0,0 +1,645 @@
...
+
+            string directory = string.IsNullOrEmpty(InitialDirectory) ? null : Path.GetDirectoryName(InitialDirectory);
+            if (directory == null) return;
+            try
+            {
+                NativeMethods.SHCreateItemFromParsingName(InitialDirectory, IntPtr.Zero, new Guid(IIDGuid.IShellItem), out IShellItem folder);
+                if (folder != null)
+                    dialog.SetFolder(folder);
+            }
+            catch
```
lindexi, line 176: Should we catch all exception?

src/Microsoft.DotNet.Wpf/src/PresentationFramework/Microsoft/Win32/OpenFolderDialog.cs
```diff
@@ -0,0 +1,645 @@
...
+        /// Gets or sets a value that controls whether to show or hide the list of pinned  places that the user can choose.
+        /// Default value is true.
+        /// </summary>
+        public bool ShowPlacesList { get; set; } = true;
+
+        /// <summary>
+        /// Gets or sets a value that controls whether to show or hide the list of places where the user has recently opened or saved items.
+        /// Default value is true.
+        /// </summary>
+        public bool AddToMruList { get; set; } = true;
```
lindexi, line 131: May I know what "Mru" is?

miloush: most recently used

lindexi: Thank you. But I don’t think abbreviations are suitable here.

miloush:
@Symbai My excuse is quality expectations. I am not submitting a PR if I cannot test the code is even working.  If there is anything worse than a project not accepting PRs it is a project accepting anything. The folder dialog seems to be a heated topic and there is a pressure from the community to demonstrate commitment, and since it was flagged for review I wanted to note that this implementation does not meet the expectations of at least some of us.

That said, I could have been more constructive in the reasoning, so please accept my apologies, here are some of the reasons I am unhappy with this PR (in no particular order):

- It is a file dialog (and common dialog), but the class is not in the file dialog hierarchy.
- It is in different namespace than the other common dialogs, and you are introducing a new root namespace that is not used for anything else.
- Introducing enormous number of types that already exist in the framework (and worryingly with different signatures), have you seen e.g. [ShellProvider.cs](https://github.com/dotnet/wpf/blob/d49f8ddb889b5717437d03caa04d7c56819c16aa/src/Microsoft.DotNet.Wpf/src/PresentationFramework/MS/Internal/AppModel/ShellProvider.cs)?
- All types in single file
- Every property has an automatic backing field, taking unnecessary amount of memory for what is effectively a flag number.
- Inconsistent type accessibility (internal vs unspecified), and naming conventions differ from the existing code
- No support for custom places
- No hooks to process the dialog notifications
- No security checks, no threading checks, no error handling.

Here is an alternative implementation with minimal impact on the existing codebase: https://github.com/dotnet/wpf/compare/master...miloush:PickFolderFileDialog

src/Microsoft.DotNet.Wpf/src/PresentationFramework/Microsoft/Win32/OpenFolderDialog.cs
Symbai: Names were taken from WindowsAPICodePack

Symbai:
> 
> Here is an alternative implementation with minimal impact on the existing codebase: [master...miloush:PickFolderFileDialog](https://github.com/dotnet/wpf/compare/master...miloush:PickFolderFileDialog)

Great, go ahead and make a PR then.

PR #2843, opened 2020-04-05
Using array empty to replace create an empty array

lindexi:

PR #3983, opened 2021-01-12
Community PR test pass 1/18/2021

ryalanms:
**Community PR Test Pass 1/11/2021**

These PRs will be merged if no regressions are found.

**Community PRs included in this test branch:**

Remove custom copy of PtrToStringChars #3117
Using array empty to replace create an empty array #2843

**Test Matrix**

| OS | Architecture |
| --- | --- |
| Windows 7 SP1	Enterprise	| x86 |	
| Windows 7 SP1	Enterprise	| x64 |	
| Windows Blue	Enterprise	| x86 |	
| Windows Server 2012 R2 SP1	| x64 |
| Win10 RS1                	Enterprise	| x86 |
| Win10 RS3	Enterprise	| x64 |
| Win10 RS4	Enterprise	| x86 |
| Windows Server 2019		| x64  |
| Win10 RS6	Enterprise	| X86 |
| Windows Server 2016	 | x64 |


src/Microsoft.DotNet.Wpf/src/PresentationCore/MS/internal/Generated/AvTraceMessages.cs
```diff
@@ -82,7 +82,7 @@ static public void Trace( TraceEventType type, AvTraceDetails traceDetails, para
         /// <summary> These help delay allocation of object array </summary>
         static public void Trace( TraceEventType type, AvTraceDetails traceDetails )
         {
-            _avTrace.Trace( type, traceDetails.Id, traceDetails.Message, traceDetails.Labels, new object[0] );
+            _avTrace.Trace( type, traceDetails.Id, traceDetails.Message, traceDetails.Labels, Array.Empty<object>() );
```
SamBent, line 85: This is a generated file (see comment at the top of the file).  You need to make these changes in the template as well, or they'll get overwritten the next time the template changes. 

The right way to do this change is:
a. Change the template file AvTraceMessages.tt
b. Re-run the generator to propagate the changes to the 3 AvTraceMessages.cs files (for Base, Core, and FW).
c. Check in all 4 changed files.
I think step (b) happens automatically at the start of a full build, but you should verify that.

SamBent (changes requested):

src/Microsoft.DotNet.Wpf/src/PresentationCore/MS/internal/Generated/AvTraceMessages.cs
lindexi: @SamBent Thank you and I fix it in https://github.com/dotnet/wpf/pull/2843

ryalanms:
I will wait for Sam's sign-off before starting the test pass.  If any other PRs come in before then, we can add them to this test run.

ryalanms:
I am still waiting on sign-offs to start the test pass.  There are only two PRs, so this will have to wait until Monday.  Thanks.

PR #3991, opened 2021-01-13
Merge internal/release/5.0 to release/5.0 and update branding

ryalanms:
FYI, the internal feeds in nuget.config are intentional and benign.  Thanks.


ryalanms:
The non-shipping packages are still being pushed to the public transport feed.  This will fail until dependencies are available.

RussKie:
> The non-shipping packages are still being pushed to the public transport feed. This will fail until dependencies are available.

I had a quick chat with @mmitche, and he said it's been worked on.
/cc: @wtgodbe

ryalanms:
The CI build is passing now.  I'll merge when it completes.

PR #3998, opened 2021-01-13
Use char overload in System.Xaml

ThomasGoulet73:
Replace all method call that uses a single char string while there is another method that accepts a char.

Should yield tiny performance benefit on hot paths.

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/System.Xaml/System/Xaml/InfosetObjects/XamlObjectWriter.cs
+++ b/src/Microsoft.DotNet.Wpf/src/System.Xaml/System/Xaml/InfosetObjects/XamlObjectWriter.cs
@@ -1996,7 +1996,7 @@ namespace System.Xaml
                 if (value is NameFixupToken && parentProperty != XamlLanguage.Items)
                 {
                     NameFixupToken token = value as NameFixupToken;
-                    string names = String.Join(",", token.NeededNames.ToArray());
+                    string names = String.Join(',', token.NeededNames.ToArray());
                     string msg = SR.Get(SRID.ForwardRefDirectives, names);
                     throw ctx.WithLineInfo(new XamlObjectWriterException(msg));
                 }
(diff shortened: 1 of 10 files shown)


src/Microsoft.DotNet.Wpf/src/System.Xaml/System/Xaml/InfosetObjects/XamlObjectWriter.cs
```diff
@@ -1996,7 +1996,7 @@ private void Logic_DoAssignmentToParentProperty(ObjectWriterContext ctx)
                 if (value is NameFixupToken && parentProperty != XamlLanguage.Items)
                 {
                     NameFixupToken token = value as NameFixupToken;
-                    string names = String.Join(",", token.NeededNames.ToArray());
+                    string names = String.Join(',', token.NeededNames.ToArray());
```
sixlettervariables, line 1999: Curious, are these ToArray calls needed? It seems like `NeededNames` implements `IEnumerable<String>`.

ThomasGoulet73: It would definetly be worth benchmarking but List may have optimization that makes it quick to return a copy of the internal array so I don't know if it is faster than enumerating with IEnumerable. My guess would be that ToArray is faster but uses more memory (It may vary with the number of item).

miloush: `List` creates a new array and copies elements over, while enumeration uses in-place value-type enumerator. It's a one method call vs. method call every item during enumerating, but also don't forget that the `string.Join` implementation differs - the array uses unsafe, pre-allocated char buffer while the enumerable uses `StringBuilder` to avoid double enumeration, and another difference is that changing the collection while being enumerated is illegal and throws.

That said, it probably doesn't matter here as it's an exception path and the number of items in the list would typically be quite small.

ryalanms:
Thank you for your contribution, @ThomasGoulet73.

ryalanms (approved):

PR #4021, opened 2021-01-18
Inline VerifyAccess

ThomasGoulet73:
The method `VerifyAccess` is not being inlined because of the Exception. Not inlining the throw is enough to inline `VerifyAccess` because the method `CheckAccess` already qualifies to be inlined.

I used the following code to benchmark it :
https://gist.github.com/ThomasGoulet73/3dbc0cf5d24ca06b639ba6888c27fe02

Here are the results :
|               Method | count |      Mean |    Error |   StdDev |  Gen 0 | Gen 1 | Gen 2 | Allocated |
|--------------------- |------ |----------:|---------:|---------:|-------:|------:|------:|----------:|
|          CheckAccess |    10 |  95.25 ns | 1.469 ns | 1.374 ns | 0.0764 |     - |     - |     480 B |
|         VerifyAccess |    10 | 143.90 ns | 1.693 ns | 1.501 ns | 0.0763 |     - |     - |     480 B |
|  CheckAccess_Inlined |    10 |  94.92 ns | 1.222 ns | 1.020 ns | 0.0764 |     - |     - |     480 B |
| VerifyAccess_Inlined |    10 |  97.63 ns | 0.627 ns | 0.556 ns | 0.0764 |     - |     - |     480 B |

The perf gain is small but since `VerifyAccess` is used in a lot of place, any perf gain is good.

In the future, we could create a ThrowHelper like in the runtime repo. To prevent inlining throws which may prevent methods to otherwise be inlined.

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/WindowsBase/System/Windows/Threading/Dispatcher.cs
+++ b/src/Microsoft.DotNet.Wpf/src/WindowsBase/System/Windows/Threading/Dispatcher.cs
@@ -17,6 +17,8 @@ using MS.Internal.Interop;                   // WM
 using MS.Internal.WindowsBase;               // SecurityHelper
 using System.Threading;
 using System.ComponentModel;                 // EditorBrowsableAttribute, BrowsableAttribute
+using System.Diagnostics.CodeAnalysis;
+using System.Runtime.CompilerServices;
 
 // Disabling 1634 and 1691:
 // In order to avoid generating warnings about unknown message numbers and
@@ -222,8 +224,16 @@ namespace System.Windows.Threading
         {
             if(!CheckAccess())
             {
-                throw new InvalidOperationException(SR.VerifyAccess);
+                ThrowVerifyAccess();
             }
+        }
+
+        // Used to inline VerifyAccess.
+        [DoesNotReturn]
+        [MethodImpl(MethodImplOptions.NoInlining)]
+        private static void ThrowVerifyAccess()
+        {
+            throw new InvalidOperationException(SR.VerifyAccess);
         }
 
         /// <summary>
--- a/src/Microsoft.DotNet.Wpf/src/WindowsBase/System/Windows/Threading/Dispatcher.cs
+++ b/src/Microsoft.DotNet.Wpf/src/WindowsBase/System/Windows/Threading/Dispatcher.cs
@@ -224,16 +224,14 @@ namespace System.Windows.Threading
         {
             if(!CheckAccess())
             {
+                // Used to inline VerifyAccess.
+                [DoesNotReturn]
+                [MethodImpl(MethodImplOptions.NoInlining)]
+                static void ThrowVerifyAccess()
+                    => throw new InvalidOperationException(SR.VerifyAccess);
+
                 ThrowVerifyAccess();
             }
-        }
-
-        // Used to inline VerifyAccess.
-        [DoesNotReturn]
-        [MethodImpl(MethodImplOptions.NoInlining)]
-        private static void ThrowVerifyAccess()
-        {
-            throw new InvalidOperationException(SR.VerifyAccess);
         }
 
         /// <summary>


src/Microsoft.DotNet.Wpf/src/WindowsBase/System/Windows/Threading/Dispatcher.cs
```diff
@@ -222,10 +224,18 @@ public void VerifyAccess()
...
             if(!CheckAccess())
             {
-                throw new InvalidOperationException(SR.Get(SRID.VerifyAccess));
+                ThrowVerifyAccess();
             }
         }
 
+        // Used to inline VerifyAccess.
+        [DoesNotReturn]
+        [MethodImpl(MethodImplOptions.NoInlining)]
```
gfoidl, line 233: FYI: JIT won't inline method that do not return (like throwing only methods) by default, so this attribute isn't needed.

Althoug JIT x86 for .NET 4.8 won't have this optimization, only RyuJIT-based implementations (.NET 4.8 x64 and .NET Core (all supported versions)).

For style, as `ThrowVerifyAccess` is used only at one place, I'd make it a static local function.

gfoidl: There's a different epilogue of the methods (see where the `ret` is). Looks like the JIT applies a different pattern when it performs an implicit no inlining for never-return-methods...
It won't change anything relevant. On a micro processor view `VerifyAccess1` is better for the branch, as when the branch predictor has no history, it will assume that forward-branche is not taken (i.e. it will fall through). 
But this has really no relevance here.

Keep the attribute -- which makes the intent also clearer that it's cold-path.

src/Microsoft.DotNet.Wpf/src/WindowsBase/System/Windows/Threading/Dispatcher.cs
```diff
@@ -222,10 +224,18 @@ public void VerifyAccess()
...
+        private static void ThrowVerifyAccess()
```
lindexi, line 234: This will change the behavior, I think some automated analysis tools will not work

gfoidl: > some automated analysis tools will not work

They should respect the DoesNotReturn-Attribute. Or do you have something different in mind?

ThomasGoulet73: @lindexi I don't think I understand what would change behavior. I may be incorrect but analysis tools should only look at the method VerifyAccess. Having private static method here only changes the analysis that may occur in this specific file. It would be great if you could explain because I don't think I understand what you mean.

lindexi: > @lindexi I don't think I understand what would change behavior. I may be incorrect but analysis tools should only look at the method VerifyAccess. Having private static method here only changes the analysis that may occur in this specific file. It would be great if you could explain because I don't think I understand what you mean.

Such as the exception call stack. Thank you.

src/Microsoft.DotNet.Wpf/src/WindowsBase/System/Windows/Threading/Dispatcher.cs
ThomasGoulet73: I tried using a static local function but I kept getting this [error](https://docs.microsoft.com/dotnet/csharp/language-reference/compiler-messages/cs8400). Don't know if I did something wrong but for now I'll leave it as it is.

src/Microsoft.DotNet.Wpf/src/WindowsBase/System/Windows/Threading/Dispatcher.cs
miloush: You don't get VerifyAccess call at the same level of stack every time, so any tool would have to go through the stack to find it, so the position should be irrelevant as long as it is there. Moreover, I think if we can get time down by 30% for something as common as VerifyAccess, it might be worth considering having the developer to fix a tool rather than have users experience slower applications.

miloush: Sorry from the other end where exception happens it would be the same level. Still sounds like an implementation detail to me that shouldn't be relied upon.

ThomasGoulet73: I really don't think anything should rely on the stacktrace of a method. It would mean that we cannot modify the content of any method because it would break the stacktrace. At this point, any change would be breaking change.

lindexi: I re-write my tools that can support the new stacktrace.

And I think it is reasonable for automated analysis tools to rely on the stacktrace.

lindexi (approved):

src/Microsoft.DotNet.Wpf/src/WindowsBase/System/Windows/Threading/Dispatcher.cs
dipeshmsft: @ThomasGoulet73, I tried using it ThrowVerifyAccess as a static local function but did not see any build errors. Can you confirm this on your end once ??

dipeshmsft (approved):

src/Microsoft.DotNet.Wpf/src/WindowsBase/System/Windows/Threading/Dispatcher.cs
ThomasGoulet73: @dipeshmsft Static local function were not supported by the C# version of this repo at the time of creating this PR. I'll look into changing this PR to use static local function soon.

ThomasGoulet73: Done in e03a6e81e207d199121835b9fa40a5040ac3bf76

dipeshmsft:

PR #4027, opened 2021-01-20
Remove redundant static fields to reduce memory usage in XamlTypeInvoker

lindexi:


lindexi:
@pchaurasia14 It just remote a static field. And I think it hard to write the benchmarks code.

h3xds1nz:
@lindexi @pchaurasia14 The original code could be "faster" thanks to code-locality, the PR however saves few bytes in the class.  If code-locality doesn't play a factor, both shall be equal in terms of performance.

In my opinion the perfect answer doesn't exist, code-gen itself is gonna be identical, `Array.Empty<object>()` returns a singleton, the code explosion already happened for `object` in some runtime lib, and in my opinion @lindexi's PR improves the code quality.

It is basically always going to boil down to this for a method that takes an arr as a param (obviously in case of the particular method in question, that would be already `r9` register on `x86_64`)
```asm
mov rdx, 0x17664001eb8 ;immediate singleton ptr
mov rdx, [rdx] ;pass empty array ptr to func
call 0x00007ffbc4e90030 ;the function call
```
```csharp
private object[] EmptyMethod(object[] array)
{
    return suggestions;
}
```

h3xds1nz (approved):

siagupta0202 (approved):

dipeshmsft:
Thank you @lindexi

Issue #4036, opened 2021-01-20
Update WPF repo TFM from netcoreapp5.0 to net5.0


ryalanms:
WPF is still using the netcoreapp5.0 TFM and should updated to net5.0 before taking changes for ARM64 from the internal public repo mirror.

PR #4037, opened 2021-01-20
Update WPF repo TFM from netcoreapp5.0 to net5.0

ryalanms:
TFMs must be updated to 5.0 prior to committing WPF ARM64 changes.


ryalanms:
Fixes #4036.

Issue #4029, opened 2021-01-20
How to test API changes?


Symbai:
I've followed the [guide](https://github.com/dotnet/wpf/blob/master/Documentation/developer-guide.md#testing-api-changes) and added the lines to my csproj file. I've changed the Wpfrootpath and fixed the Reference paths.

```
	<PropertyGroup>
		<!-- Change this value based on where your local repo is located -->
		<WpfRepoRoot>C:\Users\abc\source\repos\wpf</WpfRepoRoot>
		<!-- Change based on which assemblies you build (Release/Debug) -->
		<WpfConfig>Debug</WpfConfig>
		<!-- Publishing a self-contained app ensures our binaries are used. -->
		<SelfContained>true</SelfContained>
		<!-- The runtime identifier needs to match the architecture you built WPF assemblies for. -->
		<RuntimeIdentifier>win-x86</RuntimeIdentifier>
	</PropertyGroup>
	<ItemGroup>
		<Reference Include="$(WpfRepoRoot)\artifacts\packaging\$(WpfConfig)\Microsoft.DotNet.Wpf.GitHub.Debug\lib\netcoreapp5.0\*.dll" />
		<ReferenceCopyLocalPaths Include="$(WpfRepoRoot)\artifacts\packaging\$(WpfConfig)\Microsoft.DotNet.Wpf.GitHub.Debug\lib\netcoreapp5.0\*.dll" />
	</ItemGroup>
```

My API changes are recognized by VS. However I cannot build as I get the following errors:

`Assembly XX  with identity XY uses ZZ  which has a higher version than referenced assembly ....`

```
Error	CS1705	Assembly 'PresentationFramework' with identity 'PresentationFramework, Version=42.42.42.42, Culture=neutral, PublicKeyToken=31bf3856ad364e35' uses 'System.Runtime, Version=6.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a' which has a higher version than referenced assembly 'System.Runtime' with identity 'System.Runtime, Version=5.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a'	
[...]

PR #4039, opened 2021-01-20
Add OpenFolderDialog v2

### LINKED ISSUE #438: WPF alternative for WinForms FolderBrowserDialog

Symbai:
Fixes #438, this is mostly a copy of @miloush's work based on the feedback on my previous attempt.

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/PresentationFramework/Microsoft/Win32/OpenFolderDialog.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationFramework/Microsoft/Win32/OpenFolderDialog.cs
@@ -1,0 +1,215 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+//
+// 
+// Description:
+//              OpenFolderDialog is a sealed class derived from FileDialog that
+//              sets the pick folders option. The class does not support the legacy
+//              SHBrowseForFolder dialog.
+
+namespace Microsoft.Win32
+{
+    using System;
+    using System.Collections.Generic;
+    using System.IO;
+    using System.Security;
+    using System.Windows;
+
+    using MS.Internal.AppModel;
+    using MS.Internal.Interop;
+    using MS.Internal.PresentationFramework;
+    using MS.Win32;
+
+    /// <summary>
+    ///  Represents a common dialog box that allows the user to open one or more file(s). 
+    ///  This class cannot be inherited.
+    /// </summary>
+    public sealed class OpenFolderDialog : FileDialog
+    {
+        //---------------------------------------------------
+        //
+        // Constructors
+        //
+        //---------------------------------------------------
+        #region Constructors
+
+        //   We add a constructor for our OpenFolderDialog since we have
+        //   additional initialization tasks to do in addition to the
+        //   ones that FileDialog's constructor takes care of.
+        /// <summary>
+        ///  Initializes a new instance of the OpenFolderDialog class.
+        /// </summary>
+        public OpenFolderDialog() : base()
+        {
+            Initialize();
+        }
+
+        #endregion Constructors
+
+        //---------------------------------------------------
+        //
+        // Public Methods
+        //
+        //---------------------------------------------------
+        #region Public Methods
+
+        //   We override the FileDialog implementation to set a default
+        //   for OFN_FILEMUSTEXIST in addition to the other option flags
+        //   defined in FileDialog.
+        /// <summary>
+        ///  Resets all properties to their default values.
+        /// </summary>
+        /// <Remarks>
+        ///     Callers must have FileIOPermission(PermissionState.Unrestricted) to call this API.
+        /// </Remarks>
+        public override void Reset()
+        {
+
+            // it is VERY important that the base.reset() call remain here
+            // and be located at the top of this function.
+            // Since most of the initialization for this class is actually
+            // done in the FileDialog class, we must call that implementation
+            // before we can finish with the Initialize() call specific to our
+            // derived class.
+            base.Reset();
+
+            Initialize();
+        }
+
+        #endregion Public Methods
+
+        //---------------------------------------------------
+        //
+        // Public Properties
+        //
+        //---------------------------------------------------
+        //#region Public Properties
+        //#endregion Public Properties
+
+        //---------------------------------------------------
+        //
+        // Public Events
+        //
+        //---------------------------------------------------
+        //#region Public Events
+        //#endregion Public Events
+
+        //---------------------------------------------------
+        //
+        // Protected Methods
+        //
+        //---------------------------------------------------
+        #region Protected Methods
+
+        /// <summary>
+        ///  Demands permissions appropriate to the dialog to be shown.
+        /// </summary>
+        protected override void CheckPermissionsToShowDialog()
+        {
+            base.CheckPermissionsToShowDialog();
+        }
+
+        #endregion Protected Methods
+
+        //---------------------------------------------------
+        //
+        // Internal Methods
+        //
+        //---------------------------------------------------
+        #region Internal Methods
+
+        internal override bool RunFileDialog(NativeMethods.OPENFILENAME_I ofn)
+        {
+            // Windows XP is not supported by .NET Core so this path should never execute.
+
+            // An alternative would be to override the RunDialog method and duplicate
+            // System.Windows.Forms.FolderBrowserDialog.RunDialogOld code when UseVista is false.
+
+            throw new PlatformNotSupportedException();
+        }
+
+        internal override void PrepareVistaDialog(IFileDialog dialog)
+        {
+            base.PrepareVistaDialog(dialog);
+
+            FOS options = dialog.GetOptions() | FOS.PICKFOLDERS;
+            dialog.SetOptions(options);
+        }
+
+        internal override string[] ProcessVistaFiles(IFileDialog dialog)
+        {
+            var openDialog = (IFileOpenDialog)dialog;
+            IShellItem item = openDialog.GetResult();
+            return new[] { item.GetDisplayName(SIGDN.DESKTOPABSOLUTEPARSING) };
+        }
+
+        internal override IFileDialog CreateVistaDialog()
+        {
+            return (IFileDialog)Activator.CreateInstance(Type.GetTypeFromCLSID(new Guid(CLSID.FileOpenDialog)));
+        }
+
+        #endregion Internal Methods
+
+        //---------------------------------------------------
+        //
+        // Internal Properties
+        //
+        //---------------------------------------------------
+        //#region Internal Properties
+        //#endregion Internal Properties
+
+        //---------------------------------------------------
+        //
+        // Internal Events
+        //
+        //---------------------------------------------------
+        //#region Internal Events
+        //#endregion Internal Events
+
+        //---------------------------------------------------
+        //
+        // Private Methods
+        //
+        //---------------------------------------------------
+        #region Private Methods
+
+        //  Provides the actual implementation of initialization tasks.
+        //  Initialize() is called from both the constructor and the
+        //  public Reset() function to set default values for member
+        //  variables and for the options bitmask.
+        //
+        //  We only perform OpenFolderDialog() specific reset tasks here;
+        //  it's the calling code's responsibility to ensure that the
+        //  base is initialized first.
+        //
+        private void Initialize()
+        {
+            // OFN_FILEMUSTEXIST
+            // Specifies that the user can type only names of existing files
+            // in the File Name entry field. If this flag is specified and 
+            // the user enters an invalid name, we display a warning in a 
+            // message box.   Implies OFN_PATHMUSTEXIST.
+            SetOption(NativeMethods.OFN_FILEMUSTEXIST, true);
+        }
+
+        #endregion Private Methods
+
+        //---------------------------------------------------
+        //
+        // Private Properties
+        //
+        //---------------------------------------------------
+        //#region Private Properties
+        //#endregion Private Properties
+
+        //---------------------------------------------------
+        //
+        // Private Fields
+        //
+        //---------------------------------------------------
+        //#region Private Fields
+        //#endregion Private Fields        
+    }
+}
--- a/src/Microsoft.DotNet.Wpf/cycle-breakers/PresentationFramework/PresentationFramework.cs
+++ b/src/Microsoft.DotNet.Wpf/cycle-breakers/PresentationFramework/PresentationFramework.cs
@@ -83,6 +83,12 @@ namespace Microsoft.Win32
         public System.IO.Stream[] OpenFiles() { throw null; }
         public override void Reset() { }
     }
+    public sealed partial class OpenFolderDialog : Microsoft.Win32.FileDialog
+    {
+        public OpenFolderDialog() { }
+        protected override void CheckPermissionsToShowDialog() { }
+        public override void Reset() { }
+    }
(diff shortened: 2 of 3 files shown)


src/Microsoft.DotNet.Wpf/src/PresentationFramework/Microsoft/Win32/OpenFolderDialog.cs
```diff
@@ -0,0 +1,215 @@
...
+        {
+            // Windows XP is not supported by .NET Core so this path should never execute.
+
+            // An alternative would be to override the RunDialog method and duplicate
+            // System.Windows.Forms.FolderBrowserDialog.RunDialogOld code when UseVista is false.
+
+            throw new PlatformNotSupportedException();
+        }
+
+        internal override void PrepareVistaDialog(IFileDialog dialog)
```
fabiant3, line 133: Have you tried running this code on a Vista machine?

Symbai: #4029

lindexi: > 
> 
> Have you tried running this code on a Vista machine?

Can we run dotnet core application in Vista?

ThomasGoulet73: According to the [doc](https://github.com/dotnet/core/blob/master/release-notes/5.0/5.0-supported-os.md#windows), Vista is not supported in .Net 5.

Symbai: It hasn't anything to do with Vista except that since Vista, there is a "modern" file/folder dialog instead of the legacy. So on any Windows >= Vista, including Windows 10, it calls PrepareVistaDialog, just like the existing FileDialog. This is just a wrapper based on FileDialog doing nothing else than setting a flag to make this being a folder dialog instead. **IF** it won't work on Vista machine, then FileDialog won't work either.

https://github.com/dotnet/wpf/blob/d49f8ddb889b5717437d03caa04d7c56819c16aa/src/Microsoft.DotNet.Wpf/src/PresentationFramework/Microsoft/Win32/FileDialog.cs#L817-L829
https://github.com/dotnet/wpf/blob/d49f8ddb889b5717437d03caa04d7c56819c16aa/src/Microsoft.DotNet.Wpf/src/PresentationFramework/Microsoft/Win32/FileDialog.cs#L1835-L1845

miloush: The reason I haven't submitted a PR myself is that I wasn't able to test the code. Did you manage to run it from an app?

It looks like the only thing you changed is the name - I used `PickFolderFileDialog` because it's a file dialog with a pick folder flag. I think an "open" operation on folder is a bit misleading, what would be the corresponding save folder dialog?

That said, I am not _that_ picky about the name, as long as we get the dialog!

miloush:
The reason I haven't submitted a PR myself is that I wasn't able to test the code. Did you manage to run it from an app?

It looks like the only thing you changed is the name - I used `PickFolderFileDialog` because it's a file dialog with a pick folder flag. I think an "open" operation on folder is a bit misleading, what would be the corresponding save folder dialog?

That said, I am not _that_ picky about the name, as long as we get the dialog!

src/Microsoft.DotNet.Wpf/src/PresentationFramework/Microsoft/Win32/OpenFolderDialog.cs
miloush: @fabiant3 if implementing and testing the legacy dialog would make it eligible for .NET FW I am happy to do that.

Symbai:
> I think an "open" operation on folder is a bit misleading

Tell this Microsoft, they have chosen "OpenFileDialog" for picking/selecting a file instead of "SelectFileDialog". Thus "OpenFolderDialog" matches the line of "OpenFileDialog". On Winforms its called "FolderBrowserDialog". And on the Windows API Code Pack they called it "CommonOpenFileDialog" allowing setting a property "IsFolderPicker" to true.

So if even Microsoft haven't followed any line, there is no ultimate solution for its name. I for myself think that (existing) "OpenFileDialog" and (new) "OpenFolderDialog" seems quite nice together, while calling one "OpenFileDialog" and the other "FolderBrowserDialog" is not matching at all.

Anyway, we have several ways of calling it and there is no good or bad one. They all make sense, its just a matter of taste. I can rename it if necessary.

src/Microsoft.DotNet.Wpf/src/PresentationFramework/Microsoft/Win32/OpenFolderDialog.cs
fabiant3: Yup, pretty much the testing portion is what the concern is. Thank you!

miloush:
@Symbai having open and save file dialogs separate instead a common select file one has some advantages, such as remembering different recent locations for the two tasks separately.

I am sure the team will pick a name they see best fit.

ryalanms:
Hi @Symbai.  We are still establishing an API review process for WPF.  @SamBent will likely be the final approver, but I will follow up when we have confirmation.  Thanks.

Symbai:
Thanks for the update, I appreciate that. I can wait, take your time.

ryalanms:
Adding @MikeHillberg for API review...

Symbai:
What's the status on this?

src/Microsoft.DotNet.Wpf/src/PresentationFramework/Microsoft/Win32/OpenFolderDialog.cs
```diff
@@ -0,0 +1,215 @@
...
+    using MS.Internal.AppModel;
+    using MS.Internal.Interop;
+    using MS.Internal.PresentationFramework;
+    using MS.Win32;
+
+    /// <summary>
+    ///  Represents a common dialog box that allows the user to open one or more file(s). 
+    ///  This class cannot be inherited.
+    /// </summary>
+    public sealed class OpenFolderDialog : FileDialog
```
MikeHillberg, line 29: Can this be a [CommonDialog](http://msdn.microsoft.com/library/Microsoft.Win32.CommonDialog) rather than a `FileDialog`? Making it a `FileDialog` gives it APIs that don't make sense for a folder, such as `CheckFileExists`, `DefaultExt`, `SafeFileName`, etc.

Symbai: CommonDialog has no fields and especially no validations, they're all in FileDialog. So either we copy everything except the few APIs you mentioned (which people have criticized on v1), or we just accept the fact that some APIs are irrelevant. Or we make a new base class for both, but that's a deeper change.

FYI: The Windows API Code Pack-Shell also has a FileDialog which becomes a Folder Dialog by setting a property to true. It also has SafeFileName (which will be the folder path then).

Symbai:
What's the status on this? Is this ever getting reviewed by someone either merging it or telling me what exactly I have it to change it to so it will be merged?

Symbai:
Closing as WPF team is unable to review for 5 months and refuses to tell whats the problem. If anyone else wants to do this to themselves, go ahead. I'm giving up on this repo and this team, sorry.

nzain:
What a pity. It seems the WPF team has very little resources or little interest. WPF feels dead to me :(

fabiant3:
Reopening issue, I left a comment for @Symbai. 
We are very much interested in the community PRs, we're just heads down making progress on open sourcing the test repo which is vital for accepting community PRs.

fabiant3:
Let us know if you'd like to resubmit, looks like your repo PR has been deleted.

miloush:
I am happy to resubmit if needed, but I think some design discussion should take place first. Is that also waiting for the testing infrastructure?

fabiant3:
Based on the previous PR I don't believe this change would require the test repo for finalizing it. I'd appreciate if you could bring back the design main points and give us and the community clear options. @MikeHillberg brings some good points, would be good if you could address them. Thanks @miloush!

miloush:
Right. So, from my point of view there are basically 3 things to decide:

### 1. Inheritance

The options were:
 a) no ancestor
 b) inheriting from [`CommonDialog`](https://docs.microsoft.com/en-us/dotnet/api/microsoft.win32.commondialog)
 c) inheriting from [`FileDialog`](https://docs.microsoft.com/en-us/dotnet/api/microsoft.win32.filedialog)

I am against a). It is a common dialog that everyone expects to be shown using `ShowDialog`, and that will have an owner window, which is basically all that `CommaonDialog` provides.

From these options, I prefer c). It is a file dialog after all, and there is _a lot_ of code in `FileDialog` that would need to be duplicated. This includes extracting and validating the paths, hook procedure, OK confirmation event, support for custom places, and some of the flags. As @MikeHillberg pointed out, some of the flags do not apply, and the legacy dialog would need to be overridden if needed. However, if we go for c), the new code is very lightweight and mirrors the Win32 implementation (which is also only a flag on file dialog).

There are other options too. We might not want to do a breaking change moving properties around, but there might be a possibility to split the inheritance chain, that is, `OpenFileDialog` > `FileDialog` > something new > `CommonDialog`. The something new could be for example `CommonFileDialog`, where most of the `FileDialog` code would be moved and that would not have the unnecessary properties. Then the new dialog would directly inherit this new one. If this is an option from compatibility point of view, I am all up for it.

### 2. Naming

The options were:
 a) `FolderBrowserDialog`. This is a name WinForms uses, based on the legacy shell API.
 b) `PickFolderFileDialog`. This was based on the fact that it is a `FileDialog` with a pick folder flag, for consistency with `OpenFileDialog` and `SaveFileDialog`. If we split the inheritance chain, `PickFolderDialog` would also make sense.
 c) `OpenFolderDialog`. This was to match `OpenFileDialog`.

I don't see any rationale for a) in a new WPF code. My worry about c) was that there is no "open" operation on folder involved and there would be no corresponding "save" operation/dialog, but as stated above I am quite easy between b) and c).

### 3. Legacy support

Windows XP which .NET Framework version of WPF supports does not have the common dialog APIs. The options are:

 a) throw `PlatformNotSupportedException`. If this is going to .NET Core only, then this is a valid and easy option.
 b) implement fallback using legacy `SHBrowseForFolder` API. Note that this has a non-trivial amount of native types to include.
 c) call the WinForms implementation


If anyone thinks I forgot something, feel free chime in.

ThomasGoulet73:
@miloush That's a great summary of the design discussion required for this PR and I think you've covered everything.

Here's my personal opinion on this discussion:

1. Inheritance
My favorite is inheriting from a new class that inherits from CommonDialog. Both FileDialog and the folder dialog class would inherit from it. I wouldn't move anything public from FileDialog to the new base class. I would only move the private/protected/internal members to the base class and duplicate the public members for compatibility.

If we need full compatiblity and zero API changes to FileDialog (No new base class), we could either use a static class with the shared code or an internal implementation class. By "internal implementation class", I mean a new internal class that contains the shared implementation code from FileDialog and the new folder dialog. The public API would forward their calls to the internal class.

I personally don't like option c). There is a bunch of API that would do nothing and I don't think that it's very developer-friendly.

2. Naming
I always disliked the name `FolderBrowserDialog` in WinForms because I don't think that it's very intellisense-friendly. Having `OpenFolderDialog` and `OpenFileDialog` would make it easier to see the two types of Dialog in intellisense.

3. Legacy support
I wouldn't add any new code to support pre-Vista dialogs. .Net 6 does not support Windows XP and I don't see the team adding it back in a future .Net version. I would just let it fail if with an exception that the Win32 API for Vista dialog does not exist if somehow someone was able to run a WPF on a pre-Vista Windows.

nzain:
@fabiant3 You wanted to re-open this PR, right? It looks closed to me :)

What @miloush and @ThomasGoulet73 posted sounds smart to me. On the other side, I'd be happy to have _any_ WPF option at all. This piece is missing for many years. Don't give up on this! 👍

miloush:
I tried to look more into the splitting inheritance option and I don't think it is that beneficial after all if we can't move around public members - they would have to be defined on both the `FileDialog` and the new intermediate. If we wanted to keep them non-public on the intermediate, then we would have a public class with no members, which is not great either.

The whole picking folder capability of the file dialog feels like stitched in as an after-thought, and pretending on the WPF side that there is a clean design is a fallacy. _I am now actually inclined to just add `PickFolders` property on the `FileDialog` and call it a day. That's what the underlying model is anyway._

* On Windows Vista, the save file dialog supports the pick folders flag, and it lets users to enter a non-existing path, i.e. a folder to be created (this support has been removed on later Windows versions)
* The `Multiselect` option kind of works for folders in the sense that the dialog allows you to select multiple things in the UI, however, you are not allowed to confirm such selection.
* The legacy browse for folder dialog has the ability to display files when selecting a folder. If this feature is added to the new dialog too, then the filter properties become relevant.


These are the current public properties on the `FileDialog`:

Applicable to folders:
* `CustomPlaces`
* `DereferenceLinks` - whether links to folders are followed or not allowed
* `FileName` - e.g. _C:\Users\Thomas_
* `FileNames` - would always have 1 item at the moment
* `InitialDirectory`
* `Options`
* `SafeFileName` - e.g. _Thomas_
* `SafeFileNames` - would always have 1 item at the moment
* `Tag`
* `Title`

Doing something, but wouldn't be necessary for folders:
* `AddExtension` - internal WPF flag to ensure the file/folder name ends with an extension
* `DefaultExt` - the extension to add by default
* `ValidateNames` - this one is a red flag, the WPF documentation says "indicating whether the dialog box accepts only valid Win32 file names." which is how the `OFN_NOVALIDATE` flag [works](https://docs.microsoft.com/en-us/windows/win32/api/commdlg/ns-commdlg-openfilenamea) in the legacy file dialogs. However, for the Vista dialogs, the `FOS_NOVALIDATE` flag [means](https://docs.microsoft.com/en-us/windows/win32/api/shobjidl_core/ne-shobjidl_core-_fileopendialogoptions) "check for situations that would prevent an application from opening the selected file, such as sharing violations or access denied errors."). In addition to the dialog behavior, WPF uses the flag for adding extensions and prompting for overwriting existing files.

Valid but currently without any effect:
* `CheckFileExists` - only existing folders are allowed nowadays
* `CheckPathExists` -  only existing folders are allowed nowadays
* `RestoreDirectory` - applicable to legacy dialogs only, [not used](http://web.archive.org/web/20131017051601/http://msdn.microsoft.com/en-us/library/bb761832(VS.85).aspx) in Vista dialogs

Invalid (the native call returns an error):
* `Filter`
* `FilterIndex`

So in fact @ThomasGoulet73, there isn't _that many_ properties that would do nothing, and all of them would have either worked in the past or might work in the future/on other platform.

As for legacy implementation, I am not quite sure what is the support requirements for running on a supported version of Windows but in a compatibility mode.

src/Microsoft.DotNet.Wpf/src/PresentationFramework/Microsoft/Win32/OpenFolderDialog.cs
MikeHillberg: We shouldn't have API that's known to never work though. We want great documentation, but at the same time we want to design easy APIs that don't require outside reading. Basic OOP design is that a subclass add functionality, not remove it.

What makes it difficult is that we also use subclassing for sharing implementations. If the `FileDialog` code is stable maybe a copy of the relevant parts of it into OpenFolderDialog is OK, and maybe as a separate PR when it's important enough. I think there are also places in code where common functionality has been put into static helpers to provide shared code for cases like this.

PR #4089, opened 2021-01-27
Make source generator and package reference support opt-out (on by default)

ryalanms:
.NET 6.0 Preview 1 has snapped.  Make source generator and package reference support opt-out (on by default) for Preview 2.

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/PresentationBuildTasks/Microsoft.WinFX.targets
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationBuildTasks/Microsoft.WinFX.targets
@@ -363,7 +363,7 @@
         <_CompileTargetNameForLocalType Condition="'$(_CompileTargetNameForLocalType)' == ''">_CompileTemporaryAssembly</_CompileTargetNameForLocalType>
 
         <!-- The updated .NET 5.0.2 GenerateTemporaryTargetAssembly behavior that supports source generators is off by default. -->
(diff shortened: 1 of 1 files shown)


src/Microsoft.DotNet.Wpf/src/PresentationBuildTasks/Microsoft.WinFX.targets
```diff
@@ -363,7 +363,7 @@
         <_CompileTargetNameForLocalType Condition="'$(_CompileTargetNameForLocalType)' == ''">_CompileTemporaryAssembly</_CompileTargetNameForLocalType>
 
         <!-- The updated .NET 5.0.2 GenerateTemporaryTargetAssembly behavior that supports source generators is off by default. -->
-        <IncludePackageReferencesDuringMarkupCompilation Condition="'$(IncludePackageReferencesDuringMarkupCompilation)' == ''">false</IncludePackageReferencesDuringMarkupCompilation>
+        <IncludePackageReferencesDuringMarkupCompilation Condition="'$(IncludePackageReferencesDuringMarkupCompilation)' == ''">true</IncludePackageReferencesDuringMarkupCompilation>
```
lindexi, line 366: It looks like we can delete this code line, and the default value provided by the SDK will be used by default. See https://github.com/dotnet/sdk/pull/15465

ryalanms: Thanks, @lindexi.  I wasn't told this was being turned on in the SDK.

ryalanms: We will remove this after the SDK change is committed to master.

PR #4053, opened 2021-01-21
Fix incremental build for WPF projects with dotnet.exe

MichaeIDietrich:
This PR fixes incremental builds for WPF projects if `dotnet.exe` is used for building.

The `GetHashCode()` method of strings can produce different hash values between subsequent program runs in .NET Core and cannot be used to persist the `MarkupCompile.cache` file.

A copy of the `GetNonRandomizedHashCode()` method from `String.Comparison.cs` is used instead.

The bug was found as part of #3876.

Issue #3876, opened 2020-12-02
Building WPF projects is not deterministic with dotnet.exe


MichaeIDietrich:
* .NET Core version: 5.0.100
* Windows version: 19042.630
* MSBuild .NET Framework version: 16.8.2.56705
 
 **Problem description:**

 Building a **WPF** .NET Core project with `dotnet.exe` is not deterministic, but building it with `msbuild.exe` is deterministic as expected.
 This issue also breaks incremental builds.
 
 **Actual behavior:** 

 Each execution of `dotnet.exe build` produces assemblies that are binary different.
 
 **Expected behavior:**

 Building a .NET Core project should result in identical output as long as the input did not change.
 SDK style projects define `<Deterministic>true</Deterministic>` by default, which is meant 
 to produce identical output each build.
 
 **Minimal repro:**

1. `dotnet new wpf --name deterministic_bug`
2. `cd deterministic_bug`
3. `dotnet build -p:OutputPath=bin\dotnet_1\ -p:IntermediateOutputPath=obj\dotnet_1\`
4. `dotnet build -p:OutputPath=bin\dotnet_2\ -p:IntermediateOutputPath=obj\dotnet_2\`
5. `msbuild /restore -p:OutputPath=bin\msbuild_1\ -p:IntermediateOutputPath=obj\msbuild_1\`
6. `msbuild /restore -p:OutputPath=bin\msbuild_2\ -p:IntermediateOutputPath=obj\msbuild_2\`

Compare `bin\dotnet_1\deterministic_bug.dll` to `bin\dotnet_2\deterministic_bug.dll`
>_both are different_

Compare `bin\msbuild_1\deterministic_bug.dll` to `bin\msbuild_2\deterministic_bug.dll`
>_both are identical_

**Deeper look**

It seems to be a problem with the XAML markup compiler. 
If you check `obj\dotnet_1\deterministic_bug_MarkupCompile.cache` and `obj\dotnet_2\deterministic_bug_MarkupCompile.cache`, you can see that they differ from line 12 to 15. Those lines represent hashes over `<Page>`-items, references, `<Content>`-files and `cs`-files. Check out `GenerateCacheForFileList` from [CompilerState.cs](https://github.com/dotnet/wpf/blob/master/src/Microsoft.DotNet.Wpf/src/PresentationBuildTasks/MS/Internal/Tasks/CompilerState.cs).
The hash generation seems to be straightforward, but for some reason the result is not stable and differs everytime in .NET Core.

[...]

PR #4098, opened 2021-01-28
[release/5.0] Fix incremental build issue in WPF's markup compiler 

ryalanms:
### [release/5.0] Fix incremental build issue in WPF's markup compiler 
This moves @MichaeIDietrich's original PR from master. 

*WPF rebuilds its intermediate assembly containing local types when no inputs have changed.*

As part of the WPF's incremental build logic, WPF applies a hash to the list of reference filenames during compilation.  The hash value is written to a cache file and compared against the hash value of the new list of reference filenames next compile.  This comparison always fails due to the hash value for the same list of strings differing between runs and causes WPF to always build its intermediate project, even when inputs have not changed.  .NET's GetHashCode is not stable between separate application runs as it was in .NET Framework.
####  `Customer Impact`
One large internal Microsoft customer and one large external customer have both asked for this to be fixed as soon as possible.
 
The WPF team has been working with both customers to reduce their build times.  Customers tested instrumented builds of our markup compiler with earlier versions of this fix to identify causes of unnecessary temporary target assembly builds.  Replacing GetHashCode fixed the majority of the incremental build issues for both customers.  (Accidental file changes during the build was another non-WPF cause.)
####  `Regression?`
Not from 3.0, but from .NET Framework.  

####  `Risk (see taxonomy)`
Low.  The replacement hash function is a port of existing .NET code.  This is only called during markup compilation.

##### Original Issue:
Building WPF projects is not deterministic with dotnet.exe #3876
##### Original PR:
Fix incremental build for WPF projects with dotnet.exe #4053

No.

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/PresentationBuildTasks/MS/Internal/Tasks/CompilerState.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationBuildTasks/MS/Internal/Tasks/CompilerState.cs
@@ -215,7 +215,7 @@ namespace MS.Internal.Tasks
 
                 for (int i = 0; i < iCount; i++)
                 {
-                    iHashCode += fileItemList[i].ItemSpec.GetHashCode();
+                    iHashCode += GetNonRandomizedHashCode(fileItemList[i].ItemSpec);
                 }
 
                 StringBuilder sb = new StringBuilder();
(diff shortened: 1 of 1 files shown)


ryalanms:
The request from Tactics was to verify that updating the .NET 5.0 version with this fix always forces a rebuild of existing projects.  The concern was that the cache file from a previous version without the fix may prevent the rebuild from occurring.

ryalanms:
I've confirmed that building a project with release/5.0, and updating to the release/5.0 branch with the fix, forced the project to rebuild. The detection happened here: 

https://github.com/dotnet/wpf/blob/master/src/Microsoft.DotNet.Wpf/src/PresentationBuildTasks/MS/Internal/Tasks/IncrementalCompileAnalyzer.cs#L88

```cs
                // if PresenationBuildTasks.dll is changed last build, rebuild the entire project for sure.

                if (IsFileChanged(Assembly.GetExecutingAssembly().Location) ||
                    IsFileListChanged(_mcPass1.ExtraBuildControlFiles))
                {
                    _analyzeResult = RecompileCategory.All;
                }
```

PR #4106, opened 2021-01-29
In WispLogic.cs, improve performance by reducing memory copy of arrays

lindexi:
We haven't modified the contents of the array, so we don't need to call the RawStylusInputReport.GetRawPacketData method to create a copy of the array. We can use RawStylusInputReport.Data to return the original array


ryalanms:
Hi, @lindexi.  Have you done any sanity testing with a touchscreen or tablet?  Thanks.

lindexi:
@ryalanms Yeah, bcause the software I develop now supports touch, I have many touch devices. But I don’t think my code has been tested enough

Issue #4062, opened 2021-01-23
WPF binary names with _cor3 suffix should be updated 


ryalanms:
PenImc_cor3.dll, wpfgfx_cor3.dll, and PresentationNative_cor3.dll were named during the port from .NET Framework to .NET Core and need to be updated.

lindexi:
How about PenImc_cor5.dll and wpfgfx_cor5.dll

ryalanms:
These have been updated to _net6 from _cor3.  Thanks.

PR #4132, opened 2021-02-03
Update WPF binary names with _cor3 suffix to _net6

ryalanms:
PenImc_cor3.dll, wpfgfx_cor3.dll, and PresentationNative_cor3.dll were named during the port from .NET Framework to .NET Core and need to be updated.

These have been updated to _net6 from _cor3.  Fixes https://github.com/dotnet/wpf/issues/4062.  

/cc @SamBent

### DIFF
--- a/eng/WpfArcadeSdk/tools/WpfProjectReference.targets
+++ b/eng/WpfArcadeSdk/tools/WpfProjectReference.targets
@@ -54,19 +54,19 @@
     <WpfProjectPath Include="UIAutomationClientSideProviders" ProjectPath="$(WpfSourceDir)UIAutomation\UIAutomationClientSideProviders\UIAutomationClientSideProviders.csproj" />
 
 
-    <WpfProjectPath Include="PresentationNative_cor3" ProjectPath="$(WpfSourceDir)PresentationNative\dll\PresentationNative.vcxproj">
+    <WpfProjectPath Include="PresentationNative_net6" ProjectPath="$(WpfSourceDir)PresentationNative\dll\PresentationNative.vcxproj">
       <ReferenceOutputAssembly>false</ReferenceOutputAssembly>
       <OutputItemType>Content</OutputItemType>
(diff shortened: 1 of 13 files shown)


eng/WpfArcadeSdk/tools/WpfProjectReference.targets
```diff
@@ -54,19 +54,19 @@
     <WpfProjectPath Include="UIAutomationClientSideProviders" ProjectPath="$(WpfSourceDir)UIAutomation\UIAutomationClientSideProviders\UIAutomationClientSideProviders.csproj" />
 
 
-    <WpfProjectPath Include="PresentationNative_cor3" ProjectPath="$(WpfSourceDir)PresentationNative\dll\PresentationNative.vcxproj">
+    <WpfProjectPath Include="PresentationNative_net6" ProjectPath="$(WpfSourceDir)PresentationNative\dll\PresentationNative.vcxproj">
```
batzen, line 57: Shouldn't these includes use the suffix property to build the name?

ryalanms:
This is an update to the Wpf Arcade SDK which will require breaking the build.  wpf-int must consume the new SDK and push a package before public can build successfully.

ThomasGoulet73:
Shouldn't you also update this file: https://github.com/dotnet/wpf/blob/d49f8ddb889b5717437d03caa04d7c56819c16aa/src/Microsoft.DotNet.Wpf/src/Shared/MS/Win32/ExternDll.cs#L36 ? It seems to be used in code for the DllImportAttribute.

eng/WpfArcadeSdk/tools/WpfProjectReference.targets
ryalanms: Thanks, @batzen.

Nirmal4G:
Do we really need to specify a version or even a suffix at all?

dipeshmsft:
Closing this PR as stale.

batzen:
Stale? The reviews from the WPF team didn't even happen.
There was no further feedback the author could have acted on.

PR #4155, opened 2021-02-09
Don't use PreserveSig = false for DwmIsCompositionEnabled.

### LINKED ISSUE #4166: A WPF application running on Wine (on Linux) via wine-mono will crash when calling DwmIsCompositionEnabled due to PreserveSig = false.
The native DwmIsCompositionEnabled dllimport function currently uses PreserveSig = false to automatically "convert" the signature.

The actual native Windows API signature in C is this:

HRESULT DwmIsCompositionEnabled(BOOL *enabled);

That is, it returns a HRESULT, not a BOOL, and the actual BOOL is returned as an output parameter (given via a pointer). However, because of PreserveSig = false, this is now specified as:

BOOL DwmIsCompositionEnabled(void);

But mono doesn't support PreserveSig = false for dllimports, and so wine-mono (which also relies on wpf) doesn't either. When this native function is called, the PreserveSig is ignored, so it calls the second version, which is wrong.

For a 64-bit application (such as the one described below), since no parameter was supplied (it's empty / void), it will have whatever value rcx was had when it was invoked, due to the MS x64 ABI. In my particular case, it had the value rcx=0x00000000FFFFFFFF.

This was used as the "BOOL *enabled" pointer, which is an invalid address. wine's DwmIsCompositionEnabled then tried to write a value of FALSE to this address, which causes a crash. (it's not wine's fault here, it was simply supplied an invalid pointer due to wrong signature, even though the function always returns FALSE, it shouldn't be a problem).

A simple fix would be to do the same as was done for the imported native DwmGetColorizationColor API—use the correct signature with PreserveSig = true, and call the API with its correct signature.

Background: This happened while trying to get the launcher working for the game Divinity: Original Sin II Definitive Edition with wine-mono. Note that there are several issues with it right now, and just this patch isn't enough, but to get to this bug, the following patches (PRs) are also needed on mono first:

https://github.com/mono/mono/pull/20831
https://github.com/mono/mono/pull/20832

There's another fix for wine-mono required before this bug can be reproduced, but a workaround is to simply set the environment variable before launching it: MONO_THREADS_SUSPEND=preemptive

After this, it will arrive at this bug, i.e. crash in native DwmIsCompositionEnabled due to the wrong pointer supplied.

g-insn:
# Description

mono doesn't support it, which causes Divinity: Original Sin II's launcher to crash with wine-mono.

This fixes #4166.

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Standard/NativeMethods.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Standard/NativeMethods.cs
@@ -2558,9 +2558,9 @@ namespace Standard
         public static extern void DwmExtendFrameIntoClientArea(IntPtr hwnd, ref MARGINS pMarInset);
 
         [SuppressMessage("Microsoft.Performance", "CA1811:AvoidUncalledPrivateCode")]
-        [DllImport("dwmapi.dll", EntryPoint = "DwmIsCompositionEnabled", PreserveSig = false)]
+        [DllImport("dwmapi.dll", EntryPoint = "DwmIsCompositionEnabled", PreserveSig = true)]
         [return: MarshalAs(UnmanagedType.Bool)]
-        private static extern bool _DwmIsCompositionEnabled();
+        private static extern HRESULT _DwmIsCompositionEnabled([Out, MarshalAs(UnmanagedType.Bool)] out bool enabled);
 
         [SuppressMessage("Microsoft.Performance", "CA1811:AvoidUncalledPrivateCode")]
         [DllImport("dwmapi.dll", EntryPoint = "DwmGetColorizationColor", PreserveSig = true)]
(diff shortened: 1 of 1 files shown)


ryalanms:
Thanks, @g-insn.  Could you give more context in to the problem?  WPF doesn't run on Mono.

ryalanms:
Thanks, @g-insn.  To confirm the issue:

*A WPF application running on Wine (on Linux) will crash due to a failed pinvoke call to DwmIsCompositionEnabled.  DwmIsCompositionEnabled is missing on Wine.*

Your fix changes the pinvoke signature and accommodates handling of the missing function at runtime.  

Could you open an issue with a description and link it to the PR?  I will get this in to our community test pass.  Thank you for the fix.

g-insn:
Hi, I've created an issue #4166 and linked it. The "simple fix" mentioned in the issue is what this PR does; it uses the actual signature of the *imported* API and calls it with PreserveSig = true, so there are no implicit conversions (from HRESULT).

ryalanms:
@g-insn:  Thank you for analysis and fix.  This is great.  

We will review this in the WPF triage meeting on Thursday.

madewokherd:
In wine-mono, we don't merge in updates from wpf, so I just applied this to our fork. If a program using .NET Core runs under Wine, it should already work correctly because .NET Core supports PreserveSig=false. So, from Wine's perspective, it doesn't matter whether this change is merged here.

The only practical difference for .NET Core is that if DwmIsCompositionEnabled returns a failing HRESULT for some reason, it would be handled the same as returning false, rather than throwing an exception.

If for some reason you wanted to use the mono runtime with wpf on Windows, you would have to work around many more uses of PreserveSig=false, or add support for PreserveSig=false to the mono runtime.

ryalanms:
@g-insn: It looks like @madewokherd took this change, and we don't need it in the main WPF repo.  If so, can this be closed?  Thanks.

g-insn:
Sure, since it's fixed in wine-mono, Closing.

ryalanms:
Thank you, @g-insn and @madewokherd.

Issue #2492, opened 2020-01-31
Compilation error MC3064 when importing internal user control from signed class library


vsfeedback:
_This issue has been moved from [a ticket on Developer Community](https://developercommunity.visualstudio.com/content/problem/887201/compilation-error-mc3064-when-importing-internal-u.html)._

---
The Wpf project fails to compile with the  error MC3064 when importing signed user control if its visibility set to internal.

Note: The imported library has appropriate directive [InternalsVisibleTo()].

Same control compiles without problem if library is not strongly signed (check attached project to reproduce problem).

The attached project WpfClientApp will not compile with error:
1&gt;------ Build started: Project: WpfClientApp, Configuration: Debug Any CPU ------
1&gt;C:\Users\aza\source\repos\WpfClientApp\MainWindow.xaml(12,10): error MC3064: Only public or internal classes can be used within markup. 'ButtonControl' type is not public or internal.

If you remove signing of the assembly and the application (Project properties-&gt;Signing tab-&gt;uncheck  &quot;Sign the assembly&quot; and remove PublicKey parameter from the ControlsLibrary\Properties\AssemblyInfo.cs), project will compile and run successfully.

This makes it impossible to distincts controls for public and internal use in the community controls library.

---
#### Visual Studio Feedback System on 1/16/2020, 07:04 PM: 

<p>We have directed your feedback to the appropriate engineering team for further evaluation. The team will review the feedback and notify you about the next steps.</p>

#### azatvornitskiy on 1/19/2020, 11:35 PM: 

Thank you, will be witing.

---
### Original Solutions
(no solutions)

adamhewitt627:
[DotnetWpf2492.zip](https://github.com/dotnet/wpf/files/5969852/DotnetWpf2492.zip)

I've also run into this, but only with SDK-style projects. I've attached a reproduction project.

ryalanms:
Thanks, @gomathip02 and @adamhewitt627.  

WPF does type validation during both markup compilation passes.  As part of this, there is a check to allow non-visible types if the containing assembly for that type has the custom attribute for InternalsVisibleTo and the parameter for that attribute matches the target assembly.  The comparison of assembly names only happens if the custom property retrieved from the assembly containing the referenced type is of typeof(InternalsVisibleToAttribute).  This initial check was testing the attribute type against the default load context, not the reflection-only load context.  These are incompatible.  

[...]

PR #4192, opened 2021-02-19
Use reflection-only context for InternalsVisibleToAttribute type comparison

ryalanms:
## Description

Use reflection-only context for InternalsVisibleToAttribute type comparison.

Fixes Issue #2492: Compilation error MC3064 when importing internal user control from signed class library.  The wrong load context was used for a type comparison that always failed.  

## Customer Impact

Internal types contained in assemblies with InternalsVisibleTo attributes that are referenced in markup cause a markup compiler error:  *error MC3064: Only public or internal classes can be used within markup. 'InternalClass' type is not public or internal.*

## Regression

This is a regression from .NET Framework.  

## Testing

Repro'd the issue and tested the fix against .NET Core targeting .NET Framework 4.72/4.8 + .NET Core 3.1/6.0 and .NET Framework targeting .NET Framework 4.72/4.8 + .NET Core 3.1/6.0.  Referencing internal types in markup that are contained in an assembly with InternalsVisibleTo (which grants access to the target assembly) now works as expected.

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/Shared/System/Windows/Markup/ReflectionHelper.cs
+++ b/src/Microsoft.DotNet.Wpf/src/Shared/System/Windows/Markup/ReflectionHelper.cs
@@ -505,7 +505,7 @@ namespace System.Xaml
 
             for (int j = 0; j < list.Count; j++)
             {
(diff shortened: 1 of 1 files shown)


fabiant3 (approved):

ryalanms:
Thanks, @fabiant3.

Issue #4172, opened 2021-02-10
Tooltips show on keyboard focus


batzen:
* .NET Core Version: 5.0
* Windows version: 20H2
* Does the bug reproduce also in WPF for .NET Framework 4.8?: Yes

 **Problem description:**
Tooltips started to show on keyboard focus with no way to disable this, without also disabling other accessibility features.
`ToolTip` only has `internal virtual bool ShouldShowOnKeyboardFocus` and not, as one would expect, an override able dependency property.
This is a massive breaking change in behavior and also affects .NET 4.8. 
I say "massive breaking" because:
1. Tooltips may be expensive to show/render
2. Tooltips may cover other content

IMHO: I think it's a sad decision from Microsoft to introduce new behavior and then only allow internal consumers to change that behavior while affecting every app which uses tooltips. As the internal usage shows you fully understood that this behavior is not always the desired behavior, but still decided to make it impossible for everyone else to opt out of it.
 
 **Actual behavior:**
Tooltips show on keyboard focus.
 
 **Expected behavior:**
Tooltips should not show on keyboard focus by default, IMHO.
The behavior must be controllable through a dependency property.
 
 **Minimal repro:**
1. Add a button to a window and set it tooltip to some string
2. Select the button via keyboard
3. Tooltip shows up

elyoh:
Is it not possible to get the earlier behaviour using the AppContextSwitch Switch.UseLegacyToolTipDisplay?

batzen:
This would also disable other accessibility features related to tooltips.

ryalanms:
Thank you for the report, @batzen.

SamBent:
@batzen - The change was driven by an internal team responsible for accessibility issues across all Microsoft products, who are in turn driven by various consortiums and organizations such as [WCAG](https://www.w3.org/WAI/standards-guidelines/wcag/) that set accessibility standards and guidelines for all software producers.  In other words, it wasn't WPF's or even Microsoft's decision, and we can't do what you suggest without violating internal rules and putting WPF apps at risk of failing to be eligible for use by agencies with strong accessibility requirements (e.g. governments).  

I understand your concerns, but you should raise them with WCAG et al.

batzen:
@SamBent @ryalanms That's a very strange and concerning answer.
My main request was "The behavior must be controllable through a dependency property.".

It was Microsofts decision to make this non override-able for everyone except internal users.

[...]

PR #4195, opened 2021-02-21
Adding DependencyProperty for ShouldShowOnKeyboardFocus

batzen:
Fixes Issue #4172

# Description

Converts the virtual internal property for ShouldShowOnKeyboardFocus to a public property and also adds a DependencyProperty for it.

# Customer Impact

Tooltips always show on keyboard focus which is not the desired behavior in all customer applications.

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Controls/tooltip.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Controls/tooltip.cs
@@ -87,11 +87,30 @@ namespace System.Windows.Controls
             }
         }
 
-        internal virtual bool ShouldShowOnKeyboardFocus
+        /// <summary>
+        /// The DependencyProperty for the ShouldShowOnKeyboardFocus property.
+        /// Default: true
+        /// </summary>
+        public static readonly DependencyProperty ShouldShowOnKeyboardFocusProperty =
+            DependencyProperty.Register(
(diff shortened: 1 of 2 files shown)


batzen:
ApiCompat will complain about the changes i made.
I don't know what has to be done to stop ApiCompat from complaining.

batzen:
IMHO, this should also be backported to .NET 4.8.

SamBent:
@batzen - There is another round of tooltip accessibility changes on the way.  We will include a proper solution to the extensibility issue you raised in #4172.  (Your solution here isn't quite flexible enough.)  So I'm closing this PR.

Changes to the public API surface cannot be backported to 4.8.  Unfortunately that rules out any solution that introduces a new property or method - yours or mine.  We'll add the extensibility point in .NET 6.0, but older versions are stuck with the current behavior.

batzen:
@SamBent Could you be a bit more specific on why my solution isn't flexible enough?

Is there a chance that Microsoft could be more open/transparent about such upcoming changes?
Currently it feels like Microsoft alone decides what and how additions to WPF happen, and that doesn't feel very good to be honest.

Just to make sure i understand this correctly:
Microsoft changes the behavior in .NET 4.8, making it the new current, and says "now you are stuck with that as it's current"?
If behavior in .NET 4.8 can't be changed, why was the behavior changed in .NET 4.8?

SamBent:
See https://github.com/dotnet/wpf/pull/4242

PR #4205, opened 2021-02-22
Fix publishing in VS when runtime identifier in publishing profile differs from project rid

ryalanms:
## Description

1: In Visual Studio, publishing with a publishing profile that contains a runtime identifier which differs from the runtime identifier in the project results in a publish failure.  

In this scenario, VS overrides 'ProjectToOverrideProjectExtensionsPath'.  The parent WPF project must set the `MSBuildProjectExtensionsPath` of the child project.  (See Sdk.props:35).  

2: A `Directory.Build.Props` that uses `MSBuildProjectName` to set `IntermediateOutputPath` will cause the temporary local type assembly to not be found.   

The parent targets assumes the assembly will be placed in the `IntermediateOutputPath` of the parent, not the `IntermediateOutputPath` of the child project.  Return the assembly location from `_CompileTemporaryAssembly` so these paths no longer need to match.  

## Customer Impact

Breaks some publishing profiles in VS and a fairly common usage of directory.build.props for solutions with a large number of projects.

## Regression

Yes.  This is a regression when PackageReference support in WPF is enabled. 

## Testing

Repro'd the publish failure in VS, applied the change, and verified publishing was working.  

Repro'd the missing assembly, applied the change, and verified the project compiled.


src/Microsoft.DotNet.Wpf/src/PresentationBuildTasks/Microsoft/Build/Tasks/Windows/GenerateTemporaryTargetAssembly.cs
```diff
@@ -179,7 +179,13 @@ private bool ExecuteLegacyGenerateTemporaryTargetAssembly()
                 globalProperties[assemblyNamePropertyName] = AssemblyName;
                 globalProperties[targetAssemblyProjectNamePropertyName] = currentProjectName;
 
-                retValue = BuildEngine.BuildProjectFile(tempProj, new string[] { CompileTargetName }, globalProperties, null);
+                Dictionary<string, ITaskItem[]> targetOutputs = new Dictionary<string, ITaskItem[]>();
+                retValue = BuildEngine.BuildProjectFile(TemporaryTargetAssemblyProjectName, new string[] { CompileTargetName }, globalProperties, targetOutputs);
+
+                if (retValue)
+                {
+                    TemporaryAssemblyForLocalTypeReference = targetOutputs[CompileTargetName][0].ItemSpec;
```
ryalanms, line 187: Some additional checks need to be added here and below.

fabiant3 (approved):

Issue #3452, opened 2020-09-04
CommandParameter evaluates CanExecute


vsfeedback:
_This issue has been moved from [a ticket on Developer Community](https://developercommunity.visualstudio.com/content/idea/361782/commandparameter-evaluates-canexecute.html)._

---
<p>Behavior concerning ButtonBase-Class:</p><p>Currently, only the ValueChangedCallback from the CommandProperty invokes the CanExecute-Method of the given Command. It would be highly useful, if also the ValueChangedCallback of the CommandParameterProperty would invoke CanExecute on the Command (if not null). </p>

---
#### Fiona Niu[MSFT] on 10/19/2018, 05:51 PM: 

<p>Thank you for taking the time to provide your suggestion. We will do some preliminary checks to make sure we can proceed further.&nbsp;You will hear from us in about a week on our next steps. </p>

peterduniho:
Please note that this issue exists also with the MenuItem class. I haven't checked, but this probably also applies to the InputBinding class, and possibly any other class that has a pair of Command and CommandParameter dependency properties (e.g. ThumbButtonInfo, Hyperlink, etc.)

SetTrend:
When implementing a solution to this issue, may I humbly suggest to [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself) and implement all command related code into a separate `CommandHandler` class.

olejorgensen:
I hate this issue so much. It has been broken forever

pchaurasia14:
Closing as the completed since the corresponding PR has been merged and released.

Issue #4078, opened 2021-01-26
MenuItem: CommandParameters are ignored


SetTrend:
- .NET Core Version: 5
- Windows version: 20H2

## Problem description:

I created an MVVM sample repository. In this repository I'm loading [`MenuItem`](https://docs.microsoft.com/dotnet/api/system.windows.controls.menuitem)s and [`Button`](https://docs.microsoft.com/dotnet/api/system.windows.controls.button)s with commands.

While the `Button` elements call `ICommand::Execute(object? parameter)` with the given `CommandParameter`, `MenuItem` elements don't.

Here's a code snipped showing both declarations:

```xml
<MenuItem Header="Film_liste" ItemsSource="{Binding Path=Commands}">
    <MenuItem.ItemTemplate>
      <DataTemplate>
        <MenuItem Command="{Binding}" CommandParameter="{Binding ElementName=Filmliste, Path=SelectedItem}" Header="{Binding Path=Text}" />
      </DataTemplate>
    </MenuItem.ItemTemplate>
  </MenuItem>

  ...

  <ItemsControl.ItemTemplate>
    <DataTemplate>
      <Button Command="{Binding}" CommandParameter="{Binding ElementName=Filmliste, Path=SelectedItem}" Content="{Binding Path=Text}" Margin="5" Padding="10 3"></Button>
    </DataTemplate>
  </ItemsControl.ItemTemplate>
```
 
## Actual behavior:

Commands triggered by `MenuItem` elements are always passed `null` as command parameter.
 
## Expected behavior:

Commands triggered by `MenuItem` elements should be passed the value provided as command parameter.
 
## Minimal repro:

[MVVM Test repository](/SetTrend/MVVM-Test)

## Steps to reproduce:

https://user-images.githubusercontent.com/9283914/105822923-da8b8880-5fbc-11eb-9c90-e665297a710a.mp4

miloush:
Your `MenuItem.ItemTemplate` is not doing what you think it is doing - you are creating a `MenuItem` inside a `MenuItem`. The binding for the inner one works as expected, however, there is no binding on the outer one so it gets disabled which applies to its children too.

Try this instead:
```
<MenuItem Header="Film_liste" ItemsSource="{Binding Path=Commands}">
    <MenuItem.ItemContainerStyle>
        <Style TargetType="MenuItem">
            <Setter Property="Command" Value="{Binding}" />
[...]

PR #4217, opened 2021-02-24
CommandParameter invalidates CanExecute

miloush:
Fixes Issue #3452 #4078 (also discussed in #316)

In summary, the issue is that 
```xml
<MenuItem CommandParameter="P" Command="C" />
```
and
```xml
<MenuItem Command="C" CommandParameter="P" />
```
behave differently. In the latter case, `C.CanExecute` is called to determine whether the element should be enabled, but since the parameter has not been parsed yet, `CanExecute` is called with a `null` argument. After the parameter is parsed, the status of the command is not re-evaluated and therefore the element stays disabled when it shouldn't be.

A common workaround is to use the former syntax.

# Description

Adds property changed callback to `CommandParameterProperty` wherever applicable (`MenuItem`, `ButtonBase`, `Hyperlink`, `RibbonGallery`, `RibbonTextBox`) to call `UpdateCanExecute`.

This way, changing command parameter makes it re-evaluate command's `CanExecute` value. Most notably, this will make the order of `Command` and `CommandParameter` properties in XAML irrelevant.

Note that `ThumbButtonInfo` already implements this behaviour.

# Customer Impact

Not taking this fix retains the existing XAML attributes order significance (and the inconsistency with `ThumbButtonInfo` behaivour).

# Testing

*This has not been tested as the branch does not build at the time of submitting this PR*. Happy to test if further instructions provided.

# Risk

No public API changes, 5 private static methods and callback instances in DP properties introduced. All changes in the  _PresentationFramework_ assembly.

The new behaviour will cause the `CanExecute` method of commands to be called more often (with every parameter change), and specifically twice during loading of the element (which could be avoided by deferring `CanExecute` to `EndInit` phase).

If user changes both parameter and command and changes the command parameter first, with the new behaviour the existing command will potentially be called with a command parameter of an unexpected type. Since the API accepts `object` though, the implementations of `CanExecute` should be able to handle both `null` and any object type.

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Controls/MenuItem.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Controls/MenuItem.cs
@@ -486,7 +486,9 @@ namespace System.Windows.Controls
         public static readonly DependencyProperty CommandParameterProperty =
                 ButtonBase.CommandParameterProperty.AddOwner(
                         typeof(MenuItem),
-                        new FrameworkPropertyMetadata((object) null));
+                        new FrameworkPropertyMetadata(
+                                (object)null,
+                                new PropertyChangedCallback(OnCommandParameterChanged)));
 
         /// <summary>
         ///     The parameter to pass to MenuItem's Command.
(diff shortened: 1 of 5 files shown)


ryalanms:
@miloush:  As soon as Sam has bandwidth, he will review this.  Thanks.

lindexi:
A year has passed..

PR #4218, opened 2021-02-25
Remove WPF's temporary project generation during markup compilation.  Compile the local type assembly directly by calling csc/vbc. 

ryalanms:
## Description

Remove WPF's temporary project generation during markup compilation.  Compile the local type assembly directly by calling csc/vbc. 

This can be used for further testing.  

## Testing

* Built the WPF repository using the new PresentationBuilldTasks.  
* Built all the samples in WPF's sample repository: https://github.com/Microsoft/WPF-Samples
* Tested several of the sample projects that use local type references. 
* Built and tested a VB project that uses local type references.


rladuca:
@ryalanms Will this cause any issues with something like [Fody](https://github.com/Fody/Fody) that might hang off of CoreCompile since you're no longer running MSBuild targets?

Also, @dsplaisted @rainersigwald does calling CSC directly break any sort of promises that MSBuild relies on?  I recall once a long time ago that @nguerrera said doing a restore again would be bad, would anything like this pose a problem?

vatsan-madhavan:
I worry that this is going to be fragile. 

A project may receive sources from many places - from its own Compile elements, and from the *targets* it runs during build - these targets may come from NuGet packages. In addition, these dynamically generated sources may also come from generators like C#'s source generator, or other custom generators. 

It's MSBuild's job to keep all of this book-keeping straight, run various targets at the right opportunity, insert new compile items into the Compile Items at the time points. 

Building a temp project achieves this cleanly. 

It should be possible to do what you hope do to - namely infer compile items and pass it to CL etc. tasks. But this seems like a difficult-to-get-right undertaking. 

If the ideal solution is not going to be funded any time soon (rearchitect markup compilation to use Roslyn and do away with needing a separate compilation), then use the *inner build* paradigm instead of the current design you're pursuing. 

Rearrange properties in the WPF build in such a way that you can recursively call a build of the project (w/o getting into infinite recursion). You should be able to achieve this by calling build on self (i.e., the project being compiled) with some build-time property changed so that MSBuild sees it as a distinct build. Then arrange it so that the "inner build" doesn't run off into the XAML build path - instead, set it up with properties (that are toggled at build-launch) so that all the markup-compilation targets are disabled, and XAML sources are ignored - this ought to simply give you a straight C#-like compilation. 

Once the "inner build" completes, the "outer build" can resume and use the artifacts (dll's) produced by the inner-build to bootstrap its own XAML build (it will eventually produce artifacts that will overwrite the ones produced by the inner build).

dsplaisted:
> Also, @dsplaisted @rainersigwald does calling CSC directly break any sort of promises that MSBuild relies on? I recall once a long time ago that @nguerrera said doing a restore again would be bad, would anything like this pose a problem?

I think that using the CSC task certainly doesn't have the same issues as trying to do a restore would have.  There might be other issues like Vatsan points out, or from the fact that if there are changes to how CSC is called then you'll need to keep the WPF code in sync.

However, if I understood correctly then UWP is already using CSC this way so it seems like that would mitigate some of the risk.

I also lean towards Vatsan's view that an inner build that uses the normal compile targets would be a better solution.  However, some other people seemed to feel that would be more risky, so I don't think there's a clear-cut answer.

vatsan-madhavan:
I don't view the two paradigms (inner-build vs. CSC) as equivalent in this context even if they seem work equally well on surface (esp. based on evidence from other projects). 

In general, WPF's markup compilation uses an inner-build like paradigm today - it just so happens that the paradigm was implemented a long time ago and the implementation is not very sophisticated 😉. It turns out that the old project format was not easy to work with wrt doing things like inner-build (unlike the he SDK style project lends itself naturally for these sorts of things). So instead of leveraging the "self" project file to call an inner-build, WPF constructs a (stripped) clone of the project file to call the "inner-build" upon. 

Put simply, my suggestion above to use inner-build is not some new insight - its merely an observation about how markup compilation already works in WPF, and the suggestion is to upgrade it with few changes to use better msbuild design patterns that fit the SDK-style project world aptly.

If the CSC-paradigm has been proven to work well, that's certainly nice to hear (and that means if WPF ultimately chooses this design, it gives us that much more confidence 👍) . That said, using CSC in an implementation-approach would still amount to a paradigm switch, so here it would be good to understand what's gained when that sort of switch is made in a widely used codebase like this one. 

What's missing here is a good argument about why the CSC approach is beneficial to the extent that a breakaway from the current paradigm is necessary. And a design proposal would have been a good place to start, I think. 

Re testing: For a wholesale change like this, building a bunch of samples and projects seems like a very modest replacement for intentional unit-testing and functional testing. Customer/field testing will give you a lot of feedback, but this may be a good opportunity to build some tests for this area as well.

nguerrera:
Hi all, long time no see! I hope you're all having fun in .NET land. I don't remember what I said about restoring twice exactly, but indeed, that doesn't seem like a good idea. I also don't see it being proposed in any of the alternatives being discussed so it seems moot. Did I miss something?

An even longer time ago, before WPF on .NET Core, I thought the inner build approach @vatsan-madhavan is talking about would have been the easiest tactical fix as well. It would basically preserve the whole process and just not use a temporary project to do it. I do think it's worth considering. However, getting multi-targeting inner builds working correctly was not easy at all. The biggest pitfall there is global properties infecting project references. The solution to that was quite messy.

I think it would be good to chat with the Blazor folks on this. They had a one-off csc invocation and it was indeed fragile with lots of breaks happening as csc task changed. There were multiple fire drills getting coherent bits together from different teams into SDK in response to such changes. At that time I made a tactical msbuild feature proposal to make invoking csc yourself less fragile, but that was before roslyn generators were a thing, and I don't think it would be worth pushing on anymore.

I believe the roslyn generator feature should be the north star and I think the motivation for it included exactly these Blazor and Xaml scenarios so hopefully it can be adopted at some point. No matter how you compile twice, there are dragons, and just avoiding it altogether would be the biggest win for the long term. :)  I'll leave it to you to think about what the best option is in the meantime and I hope some of the above helps to think about that. Cheers!

dipeshmsft:
Closing this PR as stale as of now.

PR #4242, opened 2021-03-04
Spec for ToolTip accessibility changes (for WCAG 2.1)

SamBent:


src/Microsoft.DotNet.Wpf/specs/tooltip.md
```diff
@@ -0,0 +1,305 @@
...
+
+# Requirements 
+This spec cites the following sources (the [brackets] contain abbreviations used elsewhere in this document):
+
+1. WCAG 2.1, section 1.4.13: [Content on hover or focus](https://www.w3.org/TR/WCAG21/#content-on-hover-or-focus) [WCAG] 
+2. WCAG 2.1:  [Understanding 1.4.13](https://www.w3.org/WAI/WCAG21/Understanding/content-on-hover-or-focus.html)   [WCAG]
+3. MAS 1.4.13:  [Content on hover or focus](https://microsoft.sharepoint.com/:w:/r/sites/accessibility/_layouts/15/Doc.aspx?sourcedoc=%7B56729308-859e-4b32-ab27-04d4364b3003%7D&action=view&wdLOR=c9C9640FB-F04E-401A-BF46-A79887F322D3&wdAccPdf=0&wdparaid=38854747) [MAS]
+4. The [Figma diagram](https://www.figma.com/file/8LeT0Eq7a3ylAyM19c0FJp/ASC%3A-Tooltip?node-id=110%3A0) (as of February 2021) [FD]
+5. Email thread (MS internal) [ET] 
+6. WPF [Tooltip Overview](https://docs.microsoft.com/en-us/dotnet/desktop/wpf/controls/tooltip-overview?view=netframeworkdesktop-4.8) (and related links) [TO]
```
MikeHillberg, line 31: What does "[TO]" mean?

src/Microsoft.DotNet.Wpf/specs/tooltip.md
```diff
@@ -0,0 +1,305 @@
...
+
+1. Hover the mouse over the parent.
+2. Move keyboard focus to the parent.
+3. Type a keyboard shortcut.
+4. Open the tooltip programmatically.
+
+Unless mentioned otherwise below, the tooltip is shown immediately and its size and position are governed by the placement properties described in [TO].
+
+### Hover
+A hover occurs when the mouse enters the parent element and stays there for a certain time, without any other activity.
```
MikeHillberg, line 73: Or touch?

src/Microsoft.DotNet.Wpf/specs/tooltip.md
```diff
@@ -0,0 +1,305 @@
...
+# Conceptual pages (How To)
+
+## Declaring a tooltip
+You can declare a tooltip in XAML or code as described in [TO].
+Starting in .NET 6.0, your declarations can also use the properties
+`ToolTip.ShowsToolTipOnKeyboardFocus`, and
+`ToolTipService.ShowsToolTipOnKeyboardFocus`.
+
+## Opening a tooltip
+When a parent element declares a tooltip, there are four ways to open (show) the tooltip:
```
MikeHillberg, line 63: What does parent mean? Does this mean the element to which the attached ToolTip property was attached?

src/Microsoft.DotNet.Wpf/specs/tooltip.md
```diff
@@ -0,0 +1,305 @@
...
+4. Open the tooltip programmatically.
+
+Unless mentioned otherwise below, the tooltip is shown immediately and its size and position are governed by the placement properties described in [TO].
+
+### Hover
+A hover occurs when the mouse enters the parent element and stays there for a certain time, without any other activity.
+The tooltip is opened after the hover time, whose value is a function of the `InitialShowDelay` and `BetweenShowDelay` properties, and whether another tooltip is already showing [TO].
+
+### Keyboard focus
+When you use keyboard navigation to move focus to the parent element, the tooltip opens.
```
MikeHillberg, line 77: This is only if the new ShowsToolTipOnKeyboardFocus property is set? But the other behaviors listed above are with/without that property?

src/Microsoft.DotNet.Wpf/specs/tooltip.md
```diff
@@ -0,0 +1,305 @@
...
+
+|           | False | null  | True  |
+|-----------| ----- | ----  | ----  |
+| **False** | False | False | False |
+| **null**  | False | True  | True  |
+| **True**  | True  | True  | True  |
+
+If there is no ToolTip element, the 'null' column applies.
+
+In other words, honor an explicit value (True or False); if both the parent element and the ToolTip element have explicit values, the parent's value takes precedence;  if neither has an explicit value, open the tooltip.  The latter case is the default.
```
MikeHillberg, line 141: This is great and succinct, I would leave out the table.

src/Microsoft.DotNet.Wpf/specs/tooltip.md
```diff
@@ -0,0 +1,305 @@
...
+### Programmatically
+[TODO: Check if this is possible.  If so, describe it]
+
+## Closing a tooltip
+A tooltip closes for one of the following reasons.
+
+- **Timeout**. When the tooltip's `ShowDuration` time expires, the tooltip closes.
+- **Hover ends**.  When the tooltip opened due to hover, moving the mouse outside a "safe region" closes the tooltip.
+The safe region is the union of the parent element, the tooltip, and the mouse buffer zone connecting the two (this is explained further in the next section).
+- **Keyboard Shortcut**.  When the parent has keyboard focus, typing Ctrl+Shift+F10 or simply Ctrl (without any other keys) closes the tooltip.
```
MikeHillberg, line 95: If I tab to the parent and it opens, then press Control, does it close?

Editing note: I don't know that the docs.microsoft.com page for APIs generally goes into full detail about the specific user experience. From the app developer's POV the framework interacts with the end user how it sees best, maybe differently across releases, but it's nothing the app needs to write code fore. 

It's great it be written down here though. So I usually put user experience and implementation details in the appendix. But here is fine too.

src/Microsoft.DotNet.Wpf/specs/tooltip.md
```diff
@@ -0,0 +1,305 @@
...
+
+### Keyboard focus
+When you use keyboard navigation to move focus to the parent element, the tooltip opens.
+(Using the mouse to move focus does not show the tooltip.)
+You can override this behavior using the `ShowsToolTipOnKeyboardFocus` properties.
+
+### Keyboard shortcut
+When the parent has keyboard focus, typing Ctrl+Shift+F10 opens the tooltip.
+
+When the tooltip opens due to either of the keyboard actions, a [PlacementMode](https://docs.microsoft.com/en-us/dotnet/api/system.windows.controls.primitives.placementmode?view=net-5.0) value of `Mouse` or `MousePoint` is replaced by `Bottom`.
```
MikeHillberg, line 84: It behaves this way, or literally sets the property. Setting the property would be bad for the app (confusing when read/write properties are magically changed underneath them.)

src/Microsoft.DotNet.Wpf/specs/tooltip.md
```diff
@@ -0,0 +1,305 @@
...
+
+## ToolTip.ShowsToolTipOnKeyboardFocus property
+Gets or sets whether the ToolTip is shown when its parent element acquires keyboard focus.
+
+```
+    public bool? ShowsToolTipOnKeyboardFocus { get; set; }
+```
+
+### Remarks
+Setting this property on ToolTip element helps control whether to open the tooltip when its parent element acquires focus by keyboard navigation.
```
MikeHillberg, line 151: Is this the same description as the Service version? If so, please say "see ...", so I don't have to diff these two to find the differences :). When we copy this to docs.microsoft.com we can paste twice.

src/Microsoft.DotNet.Wpf/specs/tooltip.md
```diff
@@ -0,0 +1,305 @@
...
+2. Moving the mouse within the buffer zone does not close the tooltip.
+3. (6.0 only) Properties `ToolTip.ShowsToolTipOnKeyboardFocus` and `TooltipService.ShowsToolTipOnKeyboardFocus` control whether acquiring keyboard focus shows the tooltip.
+4. Ctrl closes the tooltip.
+
+Also, some changes were previously made in .NET 4.8 (and appear in .NET Core 3.0, 3.1, and .NET 5.0), without documentation:
+
+1. Keyboard focus opens the tooltip.  (Except for RibbonToolTip.)
+2. Ctrl-Shift-F10 opens or closes the tooltip.
+
+# API Details
```
MikeHillberg, line 290: Could you put this above the implementation section?

src/Microsoft.DotNet.Wpf/specs/tooltip.md
```diff
@@ -0,0 +1,305 @@
...
+
+```c#
+class ToolTipService
+{
```
MikeHillberg, line 294: Could you add a comment about "current members elided" or some such?

src/Microsoft.DotNet.Wpf/specs/tooltip.md
```diff
@@ -0,0 +1,305 @@
...
+All the changes described here are required to appear in .NET 6.0.
+
+The behavior changes will also appear in servicing updates for earlier versions as follows:
+- .NET 4.6-4.7.2, .NET 4.8:  the next biannual accessibility update (expected summer 2021)
+- .NET Core 3.1, .NET 5.0:  same time frame as above, patch version TBD
+
+The API changes (and behavior that depends on them) are only in .NET 6.0;
+servicing updates cannot include changes to public API.
+
+In the servicing updates, the new behavior is opt-in by setting
```
ThomasGoulet73, line 49: If I understood correctly, it's supposed to be use the new behavior by default but developers can opt-out of the new behavior by setting `Switch.UseLegacyAccessibilityFeatures.5` to false.

I would suggest changing it, unless I misunderstood the spec.
```suggestion
In the servicing updates, the new behavior is enabled by default but developers can opt-out by setting
```

src/Microsoft.DotNet.Wpf/specs/tooltip.md
```diff
@@ -0,0 +1,305 @@
...
+A hover occurs when the mouse enters the parent element and stays there for a certain time, without any other activity.
+The tooltip is opened after the hover time, whose value is a function of the `InitialShowDelay` and `BetweenShowDelay` properties, and whether another tooltip is already showing [TO].
+
+### Keyboard focus
+When you use keyboard navigation to move focus to the parent element, the tooltip opens.
+(Using the mouse to move focus does not show the tooltip.)
+You can override this behavior using the `ShowsToolTipOnKeyboardFocus` properties.
+
+### Keyboard shortcut
+When the parent has keyboard focus, typing Ctrl+Shift+F10 opens the tooltip.
```
fabiant3, line 82: What happens if the app or the system has registered this key combo? Does the notification still reaches the app or this will be filtered out by the WPF framework? Is the app or the framework handling it first?
Also I assume we want this key combo to work for any app, written in whatever framework and language, so this hot key need to be identical across Microsoft products. Let's make sure we mention this to CELA.

src/Microsoft.DotNet.Wpf/specs/tooltip.md
SamBent: It is the "abbreviation used elsewhere in this document" (see line 24) to refer to this link.
The letters TO stand for Tooltip Overview.

src/Microsoft.DotNet.Wpf/specs/tooltip.md
SamBent: Yes.

src/Microsoft.DotNet.Wpf/specs/tooltip.md
SamBent: Yes.  This is the term used in several of the cited docs.

Well, your question is a little garbled.  Parent means "the element that owns the tooltip".  In WPF terms, it's the element whose `ToolTip` property is non-null (that property is available on `FrameworkElement` and `FrameworkContentElement`, it's a normal property, not an attached property).  I assume that's what you were trying to say.

src/Microsoft.DotNet.Wpf/specs/tooltip.md
YuliKl: CELA defined this key combo, based on the implementation already in Office apps. So Office suggested this combo, CELA approved the suggestion. This key combination is coming from the overall platform-agnostic guidance doc.

src/Microsoft.DotNet.Wpf/specs/tooltip.md
YuliKl: In Xaml, tooltips show with touch while the parent item is pressed. As soon as the user lifts up their finger, the tooltip closes. Similar behavior to mouse hover but requires more effort by the user :) Feels like the correct touch behavior for WPF to use as well.

src/Microsoft.DotNet.Wpf/specs/tooltip.md
SamBent: In the real doc pages, perhaps.  Although I bet there are people who prefer the table to the text.  

In the spec, the table is more precise and definitive, less likely to be misinterpreted by the implementer, tester, or compliance reviewer.

src/Microsoft.DotNet.Wpf/specs/tooltip.md
SamBent: >If I tab to the parent and it opens, then press Control, does it close?

Well yes.  Tab to the parent puts focus on the parent, so pressing Ctrl closes the tooltip.  (Well, pressing and releasing Ctrl without touching any other keys, to be super precise.)

The API reference pages don't have this kind of detail, but the overview pages do.  And the audience for this spec includes the folks who want us to comply with [MAS] and [FD] - shunting this off to an appendix sends them the wrong message (that compliance is an afterthought).  After all, this spec is primarily about user experience.

src/Microsoft.DotNet.Wpf/specs/tooltip.md
SamBent: It behaves this way.  (This is already in 4.8 et al.)

src/Microsoft.DotNet.Wpf/specs/tooltip.md
SamBent: They're the same.

src/Microsoft.DotNet.Wpf/specs/tooltip.md
SamBent: I'd rather not.  This is a summary section, repeating material that appears earlier.  Summaries belong at the end, or at least after the things they summarize.

src/Microsoft.DotNet.Wpf/specs/tooltip.md
SamBent: OK.

src/Microsoft.DotNet.Wpf/specs/tooltip.md
SamBent: It's correct as written.  Setting the switch to true means "use the old behavior".   Runtimes that know about the switch (i.e. 6.0 and patched versions of older runtimes) use the target version of app to set the default value for the switch, in this case defaulting to false if the app targets 6.0 and true otherwise.  Thus apps that target 6.0 get the new behavior, apps that target older versions get the old behavior.  An older app that wants the new behavior has to opt-in by setting the switch to false.  (More literally, the app opts out of using the old behavior.)

This confuses everyone the first time you see it.  But eventually you realize that it's the right design (it took me a few weeks).   Your assertion
>it's supposed to be use the new behavior by default 

isn't quite right.  It's true for apps that target the new runtime, but older apps should use the old behavior by default (for compat).

src/Microsoft.DotNet.Wpf/specs/tooltip.md
SamBent: It comes from [FD], the spec that's driving "identical across MS products".    CELA is probably the source.

src/Microsoft.DotNet.Wpf/specs/tooltip.md
ThomasGoulet73: You're right, I read the Applicability section again and you're comment and I understand what you mean. It's for apps targeting older frameworks than .Net 6 to be able to have the accessibility behavior of .Net 6. So, apps that target .Net 6 uses the new behavior by default but developers can opt-out and apps that target the frameworks getting a servicing update can opt-in.

Thank you!

fabiant3 (approved):

MikeHillberg (approved):

PR #4250, opened 2021-03-07
Fix loc add a space after the colon

lindexi:
## Description

Fix loc by add a space after the colon.

You can see that the other translated content of this document is to add a space after the colon.


lindexi:
See https://github.com/dotnet/wpf/pull/4327

PR #2976, opened 2020-05-10
Don't double import WinFX targets

Nirmal4G:
Don't import .NET Framework's WinFX targets, When the SDK's implementation is being used and vice versa

**Reference**: microsoft/msbuild#4948

PR #4254, opened 2021-03-07
[release/3.1] WinFX casing and double-import fixes

### LINKED ISSUE #4253: Backport WinFX changes to "v3.1" and "v5.0" SDKs
Can we backport WinFX changes from PRs (#2975, #2976, #4629, #4630) into `v3.1` and `v5.0` SDKs?

These changes (_casing fix and double-import fix_) shouldn't be high-risk change when all they do is fix some issues that were most common and frustrating to work around (_at least in our case_). I have checked with both patched SDKs from our side and `v6.0` preview SDKs. They don't cause no major issues.

I already have the patches available targeting both `v3.1` and `v5.0` branches. It would be very helpful if the team could look this at priority. We thought we could wait until `v6.0` releases but our build requires these two changes immediately. Having to apply the workaround on our CI while build is really a recipe for disaster.

Nirmal4G:
Closes #4253

Master PRs: #2975, #2976, #4630

## Description

Casing of `WinFX` is wrong in file name of the props/targets leading to file not found errors in case-sensitive file systems, leading to build break.
When using the Desktop SDK, the build double imports both Full framework and Core `WinFX` targets, leading to build break.
Also, there's no straight forward way to switch between either `WinFX` targets as needed. This patch provides such mechanism through `ImportFrameworkWinFXTargets`.

_Although, there are workarounds for the above issues, they require modifying the installation of the .NET SDKs. So, it's preferable to have them fixed at least for LTS (`v3.1`) .NET SDK._

## Customer Impact

 - dotnet/sdk#11108
 - novotnyllc/MSBuildSdkExtras#224
 - novotnyllc/MSBuildSdkExtras#207

## Regression

Yes, Desktop SDK and WinFX targets were introduced since v3.0!

## Testing

Manual testing and mostly with our WPF projects builds against both .NET `Core` and `Full` framework CLRs.

## Risk

There should be minimal to no risk to most when taking this patch.
But, if you're building both Full framework and Core projects with heavily modified build like ours, depending on the exact casing and behaviour of the double import, then Yes, the build will break.
However, if you already have your workarounds placed, then there should be no problem. You can also remove those workarounds when updating to the SDK with this patch as it fixes those issues.


Nirmal4G:
Can someone/anyone (_please_) take a look at this patch at priority? 🧐

Been asking for months. Any feedback or response would be nice… 😔

Nirmal4G:
This doesn't look good on you, M$FT! Not even taking a look at a patch and then closing it when it becomes irrelevant. Not a good open-source practice! Very disappointed here!!

pchaurasia14:
@Nirmal4G - This PR got closed as a side effect of deleting the `release/3.1` branch from the repository.  Also, we ask all PRs to be submitted against `main` branch instead of `release/*` branches.

If the problem addressed by this PR is still relevant, you may want to open a PR against `main` branch and we will review it in upcoming CTP(s).

We apologize for the inconvenience.

PR #4255, opened 2021-03-07
[release/5.0] Switch between either WinFX targets

### LINKED ISSUE #4253: Backport WinFX changes to "v3.1" and "v5.0" SDKs

Nirmal4G:
Closes #4253

Master PRs: #2976, #4629, #4630

## Description

When using the Desktop SDK, the build double imports both Full framework and Core `WinFX` targets, leading to build break.
Also, there's no straight forward way to switch between either `WinFX` targets as needed. This patch provides such mechanism through `ImportFrameworkWinFXTargets`.

_Although there is a workaround for the above issue, they require modifying the installation of the .NET SDKs. So, it's preferable to have them fixed at least for Current (`v5.0`) .NET SDK._

## Customer Impact

We (_and possibly many others like me_) have a large set of NETCLR WPF projects that can't move to CoreCLR right now. I have recently converted them to sdk-style projects with the `WindowsDesktop` SDK which includes custom tasks that are meant for Full framework projects. We also have some silverlight projects to maintain and I recently converted them to sdk-style too. Seems they also need WinFX targets.

So, to use the official `WindowsDesktop`, we want to switch between either `WinFX` targets for our custom tasks to take effect. This patch is just allowing that. We're migrating our projects and tools to `.NET` CoreCLR and it'll take some time. Until then, we need this ability to maintain our projects.

## Regression

Yes, Desktop SDK and WinFX targets were introduced since v3.0!

## Testing

Manual testing and mostly with our WPF projects builds against both .NET `Core` and `Full` framework CLRs.

## Risk

There should be minimal to no risk to most when taking this patch.
But, if you're building both Full framework and Core projects with heavily modified build like ours, depending on the behaviour of the double import, then Yes, the build will break.
However, if you already have your workarounds placed, then there should be no problem. You can also remove those workarounds when updating to the SDK with this patch as it fixes the issue.

### DIFF
--- a/packaging/Microsoft.NET.Sdk.WindowsDesktop/Sdk/Sdk.props
+++ b/packaging/Microsoft.NET.Sdk.WindowsDesktop/Sdk/Sdk.props
@@ -15,10 +15,20 @@ Copyright (c) .NET Foundation. All rights reserved.
     <_MicrosoftWindowsDesktopSdkImported>true</_MicrosoftWindowsDesktopSdkImported>
   </PropertyGroup>
 
+  <!--
+    Workaround: https://github.com/microsoft/msbuild/issues/4948
+    Disable .NET Framework's inbox WinFX targets when using the SDK, since, we really don't use it's build logic
+    and is superseded by 'WindowsDesktop' SDK that provides it's own WinFX for both NETFX and CoreCLR targets.
+  -->
+  <PropertyGroup>
+    <ImportFrameworkWinFXTargets>false</ImportFrameworkWinFXTargets>
+  </PropertyGroup>
+
   <Import Sdk="Microsoft.NET.Sdk" Project="Sdk.props" />
 
-  <!-- Set ImportWindowsDesktopTargets to force import WindowsDesktop Targets to enable pre 5.0 behavior.
-      Post 5.0, Microsoft.NET.Sdk.WindowsDesktop.props will always be imported by SDK
+  <!--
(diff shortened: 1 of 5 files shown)


Nirmal4G:
Can someone/anyone (_please_) take a look at this patch at priority? 🧐

Been asking for months. Any feedback or response would be nice… 😔

Nirmal4G:
This doesn't look good on you, M$FT! Not even taking a look at a patch and then closing it when it becomes irrelevant. Not a good open-source practice! Very disappointed here!!

pchaurasia14:
@Nirmal4G - Since `release/3.1` and `release/5.0` branches are deleted from the repository, this PR got closed as a consequence. 

Also, we recommend all PRs to be made against `main` branch (instead of any `release/*` branches.

We apologize for the inconvenience you have experienced. If the PR addresses an issue that is still relevant, you may open the same against `main` branch, we will review it for our upcoming CTP(s).

Issue #4120, opened 2021-02-02
WPF's default branch will be renamed to 'main' from 'master'


ryalanms:
Rename WPF's default branch to 'main' from 'master'.

ClosetBugSlayer:
Because...?

ryalanms:
.NET is renaming all repositories to 'main'.  The GitHub default for new repositories is also now 'main'.

ClosetBugSlayer:
Why?

lindexi:
@ClosetBugSlayer  See: https://github.com/github/renaming  https://www.zdnet.com/article/ibm-microsoft-staff-rally-to-remove-racially-insensitive-language-from-products/

ryalanms:
The deadline for .NET repos is 3/19.

ryalanms:
The default branch in the WPF repo will be renamed this week.  Thanks.

PR #4289, opened 2021-03-17
Change WPF's default branch to 'main':  Update pipeline triggers to include 'main'

ryalanms:
Please see https://github.com/dotnet/wpf/issues/4120.  This change updates the triggers to include the 'main' branch.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -37,6 +37,7 @@ trigger:
   branches:
     include: 
     - master
+    - main 
     - release/3.*
     - release/5.*
     - internal/release/5.*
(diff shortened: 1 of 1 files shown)


azure-pipelines.yml
```diff
@@ -37,6 +37,7 @@ trigger:
   branches:
     include: 
     - master
```
lindexi, line 39: Should we remove the master branch?

ryalanms: This is updated to remove master in a later step after changing the default branch and recreating the DARC subscriptions.

ryalanms: Actually, I'm going to disable the PR builds entirely so all PRs retargeted to main don't queue up PR builds simultaneously.

Issue #4338, opened 2021-03-28
Replace DoubleUtil.IsNaN with double.IsNaN 


omariom:
BCL's double.IsNaN  got faster long ago. In .NET Core it is a single instruction.

The usage of [DoubleUti.IsNaN](https://github.com/dotnet/wpf/blob/7e118beef92050c247fac7ad24ce0816e42e77b6/src/Microsoft.DotNet.Wpf/src/Shared/MS/Internal/DoubleUtil.cs#L303) should be replaced by [double.IsNaN](https://github.com/dotnet/runtime/blob/4f8be1992cc79807d377afce640f219d4caffb5b/src/libraries/System.Private.CoreLib/src/System/Double.cs#L91).

[...]

PR #4339, opened 2021-03-28
Replace DoubleUtil.IsNaN with double.IsNaN 

omariom:
Fixes Issue https://github.com/dotnet/wpf/issues/4338

## Description

Replaces DoubleUtil.IsNaN with double.IsNaN which is times faster.

## Testing

I haven't managed to build the project.
## Risk

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/UIElement.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/UIElement.cs
@@ -585,7 +585,7 @@ namespace System.Windows
                 using (Dispatcher.DisableProcessing())
                 {
                     //enforce that Measure can not receive NaN size .
-                    if (DoubleUtil.IsNaN(availableSize.Width) || DoubleUtil.IsNaN(availableSize.Height))
+                    if (double.IsNaN(availableSize.Width) || double.IsNaN(availableSize.Height))
                         throw new InvalidOperationException(SR.Get(SRID.UIElement_Layout_NaNMeasure));
 
                     bool neverMeasured = NeverMeasured;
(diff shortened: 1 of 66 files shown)


src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/UIElement.cs
```diff
@@ -1403,8 +1403,8 @@ private static void RenderTransform_Changed(DependencyObject d, DependencyProper
         private static bool IsRenderTransformOriginValid(object value)
         {
             Point v = (Point)value;
-            return (    (!DoubleUtil.IsNaN(v.X) && !Double.IsPositiveInfinity(v.X) && !Double.IsNegativeInfinity(v.X))
-                     && (!DoubleUtil.IsNaN(v.Y) && !Double.IsPositiveInfinity(v.Y) && !Double.IsNegativeInfinity(v.Y)));
+            return (    (!double.IsNaN(v.X) && !Double.IsPositiveInfinity(v.X) && !Double.IsNegativeInfinity(v.X))
```
gfoidl, line 1406: These checks could use `double.IsNormal` instead of IsNan, IsPositiveInfinity, and IsNegativeInfinity.
Same on other places.

But this API isn't available on .NET Full, only starting at .NET Core 2.1 and .NET Standard 2.1.

Instead of IsPositiveInfinity and IsNegativeInfinity use just `double.IsInfinity`.

omariom: I would implement it in a separate PR. 
This one just mechanically replaces one call with another.

gfoidl: Makes sense.

ayushverma-ms (approved):

Issue #4316, opened 2021-03-23
After adding a splashscreen Application.Current.Activated event is no longer fired


Symbai:
* .NET Core Version: 5.0.4
* Windows version: Win10
* Does the bug reproduce also in WPF for .NET Framework 4.8?: Yes
 
 **Problem description:**
After [adding a splashscreen](https://docs.microsoft.com/en-us/dotnet/desktop/wpf/app-development/how-to-add-a-splash-screen-to-a-wpf-application?view=netframeworkdesktop-4.8), the [Application.Current.Activated](https://docs.microsoft.com/en-us/dotnet/api/system.windows.application.activated?view=netframework-4.7.2) event is no longer fired
 
 **Actual behavior:** 
Activated event does not fire
 
 **Expected behavior:**
Activated event fires, like it does without splashscreen
 
 **Minimal repro:**
1. Create a blank WPF project, open App.xaml.cs and copy this into it
```C#
    public partial class App : Application
    {
        public App()
        {
            Current.Activated += Current_Activated;
        }

        private void Current_Activated(object sender, EventArgs e)
        {
            MessageBox.Show("Hi");
        }
    }
```
2. Run the project and you see the message box appears
3. Add a splash screen: https://docs.microsoft.com/en-us/dotnet/desktop/wpf/app-development/how-to-add-a-splash-screen-to-a-wpf-application?view=netframeworkdesktop-4.8
4. Re-run the project, the splash screen no longer appears

lindexi:
@Symbai Why the Application.Current.Activated event is no longer fired?

Because Activated is fired in the WmActivateApp method of the Application class. 
The WmActivateApp method is fired by Windows messages. See [WM_ACTIVATEAPP](https://docs.microsoft.com/zh-cn/windows/win32/winmsg/wm-activateapp?WT.mc_id=WD-MVP-5003260 )


```csharp
        private void EnsureHwndSource()
        {
            if (_parkingHwnd == null)
            {
                // _appFilterHook needs to be member variable otherwise
                // it is GC'ed and we don't get messages from HwndWrapper
                // (HwndWrapper keeps a WeakReference to the hook)

                _appFilterHook = new HwndWrapperHook(AppFilterMessage);
                HwndWrapperHook[] wrapperHooks = {_appFilterHook};

                _parkingHwnd = new HwndWrapper(
                                0,
                                0,
                                0,
                                0,
                                0,
                                0,
                                0,
                                "",
                                IntPtr.Zero,
[...]

PR #4340, opened 2021-03-29
Try to create application before show SplashScreen

lindexi:
Fixes Issue https://github.com/dotnet/wpf/issues/4316

## Description

Because Activated is fired in the WmActivateApp method of the Application class. The WmActivateApp method is fired by Windows messages. See [WM_ACTIVATEAPP](https://docs.microsoft.com/zh-cn/windows/win32/winmsg/wm-activateapp?WT.mc_id=WD-MVP-5003260 )

```csharp
        private void EnsureHwndSource()
        {
            if (_parkingHwnd == null)
            {
                // _appFilterHook needs to be member variable otherwise
                // it is GC'ed and we don't get messages from HwndWrapper
                // (HwndWrapper keeps a WeakReference to the hook)

                _appFilterHook = new HwndWrapperHook(AppFilterMessage);
                HwndWrapperHook[] wrapperHooks = {_appFilterHook};

                _parkingHwnd = new HwndWrapper(
                                0,
                                0,
                                0,
                                0,
                                0,
                                0,
                                0,
                                "",
                                IntPtr.Zero,
                                wrapperHooks);
            }
        }

        private IntPtr AppFilterMessage(IntPtr hwnd, int msg, IntPtr wParam, IntPtr lParam, ref bool handled)
        {
            IntPtr retInt = IntPtr.Zero;
            switch ((WindowMessage)msg)
            {
                case WindowMessage.WM_ACTIVATEAPP:
                    handled = WmActivateApp(NativeMethods.IntPtrToInt32(wParam));
                    break;
                case WindowMessage.WM_QUERYENDSESSION :
                    handled = WmQueryEndSession(lParam, ref retInt);
                    break;
                default:
                    handled = false;
                    break;
            }
            return retInt;
        }

        private bool WmActivateApp(Int32 wParam)
        {
            int temp = wParam;
            bool isActivated = (temp == 0? false : true);

            // Event handler exception continuality: if exception occurs in Activate/Deactivate event handlers, our state would not
            // be corrupted because no internal state are affected by Activate/Deactivate. Please check Event handler exception continuality
            // if a state depending on those events is added.
            if (isActivated == true)
            {
                OnActivated(EventArgs.Empty);
            }
            else
            {
                OnDeactivated(EventArgs.Empty);
            }
            return false;
        }
```

But after setting SplashScreen, WPF will create the code in the generated App.g.cs file.

```csharp
        [System.STAThreadAttribute()]
        [System.Diagnostics.DebuggerNonUserCodeAttribute()]
        [System.CodeDom.Compiler.GeneratedCodeAttribute("PresentationBuildTasks", "5.0.1.0")]
        public static void Main() 
        {
            SplashScreen splashScreen = new SplashScreen("SplashScreen.png");
            splashScreen.Show(true);
            App app = new App();
            app.InitializeComponent();
            app.Run();
        }
```

As you can see, in the Main function, SplashScreen will be called first and then the App object will be created.

And SplashScreen will create windows very fast. After the SplashScreen creates the window, the system will send the WM_ACTIVATEAPP message to the application. But at this time the App object has not been created yet. 

Therefore, before EnsureHwndSource method in Application is called, the system has already sent the WM_ACTIVATEAPP message to the application.
So AppFilterMessage method in Application will not receive WM_ACTIVATEAPP message, and will not trigger the Activated event 

I fix it by create application before show SplashScreen

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/PresentationBuildTasks/MS/Internal/MarkupCompiler/MarkupCompiler.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationBuildTasks/MS/Internal/MarkupCompiler/MarkupCompiler.cs
@@ -3207,15 +3207,17 @@ namespace MS.Internal
 
                 if (cmmMain != null)
                 {
+                    //   MyApplication app = new MyApplication();
+                    //
+                    CodeVariableReferenceExpression cvreApp = GenerateAppInstance(cmmMain);
+
+                    //   SplashScreen splashScreen = new SplashScreen("SplashScreen.png");
+                    //
                     CodeVariableReferenceExpression cvreSplashScreen = null;
                     if (!string.IsNullOrEmpty(_splashImage) && !HostInBrowser)
                     {
                         cvreSplashScreen = GenerateSplashScreenInstance(cmmMain);
                     }
-
-                    //   MyApplication app = new MyApplication();
-                    //
-                    CodeVariableReferenceExpression cvreApp = GenerateAppInstance(cmmMain);
 
                     if (_ccRoot.InitializeComponentFn != null)
                     {


ThomasGoulet73:
I think this will cause the OnStartup event to be called before the splashcreen which I don't think is desired. Could you test it ? Because if it fixes an event just to break another one, I think it should stay as it is.

lindexi:
> 
> 
> I think this will cause the OnStartup event to be called before the splashcreen which I don't think is desired. Could you test it ? Because if it fixes an event just to break another one, I think it should stay as it is.

@ThomasGoulet73 I changed the behavior, so I think you are right

Symbai:
On Splashscreen (FYI: can even be a messagebox) both (OnStartup / Activated) is **not** called. I have not tested lindexi's fix here but the original issue is bigger than fixing just **one** event.

miloush:
Yeah I agree that moving the splash screen after creating the `Application` instance defeats the point of the splash screen. Any static members of the `App` class would be executed before splash and these can potentially take a long time.

PR #4424, opened 2021-04-18
Use pattern matching in TaskExtensions

lindexi:
## Description

Use pattern matching in TaskExtensions.cs. It will be clearer.

## Testing

Just CI.

## Risk

Low.


anjalisheel-wpf (approved):
There should be some description for the PR requests, it should not be empty.

Issue #4481, opened 2021-05-05
System.Windows.Documents.Serialization.SerializerProvider default constructor fails creating System.Uri


ryalanms:
System.Windows.Documents.Serialization.SerializerProvider creation always fails while creating System.Uri. 

```cs
SerializerProvider sp = new SerializerProvider();
```   
> at System.Uri.CreateThis(String uri, Boolean dontEscape, UriKind uriKind)
   at System.Uri..ctor(String uriString)
   at System.Windows.Xps.Serialization.XpsSerializerFactory.get_ManufacturerWebsite()
   at System.Windows.Documents.Serialization.SerializerDescriptor.CreateFromFactoryInstance(ISerializerFactory factoryInstance)
   at System.Windows.Documents.Serialization.SerializerProvider..ctor()
   at TestSerializerProvider.MainWindow..ctor() in D:\PluginSerializer\TestSerializerProvider\MainWindow.xaml.cs:line 30

Issue #4479, opened 2021-05-05
SerializerDescriptor relies on unsupported reflection API and will always throw


ryalanms:
SerializerDescriptor needs to be updated to use System.Reflection.MetadataLoadContext.   

(This is covered by internal test Avalon.Test.D2.Print.PluginSerializer.PluginSrzTestApp.)

ryalanms:
After porting PluginSerializer.PluginSrzTestApp to .NET Core, the test is crashing while creating a SerializerProvider using its default constructor.  The test doesn't make it to the SerializerDescriptor.  

> D:\PluginSerializer\PluginSrzTest\bin\Debug\net6.0-windows>PluginSrzTest.exe
Beginning Test *********************************************************
+Start Test: PluginSrzTestApp
  Fail: EXCEPTION System.UriFormatException during test. Message: Invalid URI: The format of the URI could not be determined.. StackTrace:
  Fail:    at System.Uri.CreateThis(String uri, Boolean dontEscape, UriKind uriKind) in System.Private.Uri.dll:token 0x6000096+0x69
   at System.Uri..ctor(String uriString) in System.Private.Uri.dll:token 0x600003d+0x14
   at System.Windows.Xps.Serialization.XpsSerializerFactory.get_ManufacturerWebsite() in ReachFramework.dll:token 0x6000d86+0x0
   at System.Windows.Documents.Serialization.SerializerDescriptor.CreateFromFactoryInstance(ISerializerFactory factoryInstance) in PresentationFramework.dll:token 0x6005d00+0x3e
   at System.Windows.Documents.Serialization.SerializerProvider..ctor() in PresentationFramework.dll:token 0x6005d10+0xe
   at Avalon.Test.D2.Print.PluginSerializer.PluginSrzTestApp.Run(CmdLineParser parser) in D:\PluginSerializer\PluginSrzTest\PluginSrzTest.cs:line 116
-End Test: PluginSrzTestApp Result: Fail

PR #4482, opened 2021-05-06
Fix ReachFramework's resource string resolution

ryalanms:
## Description

System.Windows.Documents.Serialization.SerializerProvider creation always fails while creating System.Uri. 

```cs
SerializerProvider sp = new SerializerProvider();
```   
> at System.Uri.CreateThis(String uri, Boolean dontEscape, UriKind uriKind)
   at System.Uri..ctor(String uriString)
   at System.Windows.Xps.Serialization.XpsSerializerFactory.get_ManufacturerWebsite()
   at System.Windows.Documents.Serialization.SerializerDescriptor.CreateFromFactoryInstance(ISerializerFactory factoryInstance)
   at System.Windows.Documents.Serialization.SerializerProvider..ctor()
   at TestSerializerProvider.MainWindow..ctor() in D:\PluginSerializer\TestSerializerProvider\MainWindow.xaml.cs:line 30

The issue is that resource lookups for the ReachFramework were not actually resolving the string, and instead were returning the key.  When the key `Xps_ManufacturersWebsite` was passed in to the URI constructor, URI creation failed.  Fixing the resource lookup path fixed `SerializerProvider`.  Fixes https://github.com/dotnet/wpf/issues/4481.  

This bug is blocking testing of the fix for another bug that is preventing WPF from taking the latest WinForms update: https://github.com/dotnet/wpf/issues/4479.  

## Regression

No.  This is a longstanding bug from 3.0.  

## Testing

Repro'd the failure, fixed the resource string lookup, and verified SerializerProvider is working.

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/ReachFramework/SR.cs
+++ b/src/Microsoft.DotNet.Wpf/src/ReachFramework/SR.cs
@@ -8,14 +8,14 @@ namespace System.Windows.Xps
 {
     internal static partial class SR
     {
-        public static string Get(string id)
+        public static string Get(string name)
         {
-             return SR.Format(id);
+            return GetResourceString(name, null);
         }
-        
-        public static string Get(string id, params object[] args)
+
+        public static string Get(string name, params object[] args)
         {
-             return SR.Format(id, args);
+            return Format(GetResourceString(name, null), args);
         }
     }
 }


SamBent (approved):
I trust you've checked all the other SR.cs files for the same problem?

ryalanms:
> I trust you've checked all the other SR.cs files for the same problem?

Thanks, Sam.  Yes, the other eleven SR.cs files are correct.  ReachFramework is the only component that was not calling GetResourceString before Format.

PR #4486, opened 2021-05-07
In FamilyCollection.cs replace Dictionary ContainsKey-Add with TryAdd

lindexi:


lindexi:
Thank you @dipeshmsft

PR #4497, opened 2021-05-10
Update 5.0 branding version to 5.0.7

ryalanms:


wtgodbe:
> LINK : fatal error C1047: The object or library file 'D:\workspace\_work\1\s\.packages\runtime.win-x86.microsoft.dotnet.wpf.dnceng\5.0.6-servicing.21206.6\runtimes\win-x86\native\bilinearspan.lib' was created by a different version of the compiler than other objects like 'D:\workspace\_work\1\s\artifacts\obj\wpfgfx\Win32\Release\precomp.obj'; rebuild all objects and libraries with the same compiler [D:\workspace\_work\1\s\src\Microsoft.DotNet.Wpf\src\WpfGfx\core\dll\wpfgfx.vcxproj]

@ryalanms any ideas what is causing this?

ryalanms:
If the libs in the transport package haven't been built recently, the compiler versions can differ.  I've already started a wpf-int build to get a newer version of the libs.

PR #4502, opened 2021-05-11
Add IntPtr overloads to MessageBox

0x90d:
From all existing overloads it takes the window handle (IntPtr) and call its core functionality in `ShowCore` method anyway. But it has no public overload for IntPtr yet. But one reason why this could be useful is that it allows us to do something like:

```C#
MessageBox.Show(Application.Current?.MainWindow == null ?
    IntPtr.Zero :
    new WindowInteropHelper(Application.Current.MainWindow).Handle, "Hello world");
```

Using the window overload the messagebox will be shown centered at the application window. However sometimes `Application.Current.MainWindow` can be null for example when called before the window is shown. And in this case it will throw an exception, rather than just fallback to IntPtr.Zero in which case the messagebox appears centered at the desktop. This is because it creates a new instance of `WindowInteropHelper` on the existing window overloads to obtain the window handle. But `WindowInteropHelper` does not allow NULL on its `ctor`.

Intially I wanted to just add nullify support of the window overloads but I don't know if the WPF project supports it, it seems like it does not. This would also be problematically since it modifies existing code. Adding IntPtr overloads was the safest solution as it has zero risk by not modifying existing code.

### DIFF
--- a/src/Microsoft.DotNet.Wpf/cycle-breakers/PresentationFramework/PresentationFramework.cs
+++ b/src/Microsoft.DotNet.Wpf/cycle-breakers/PresentationFramework/PresentationFramework.cs
@@ -943,12 +943,18 @@ namespace System.Windows
         public static System.Windows.MessageBoxResult Show(string messageBoxText, string caption, System.Windows.MessageBoxButton button, System.Windows.MessageBoxImage icon) { throw null; }
         public static System.Windows.MessageBoxResult Show(string messageBoxText, string caption, System.Windows.MessageBoxButton button, System.Windows.MessageBoxImage icon, System.Windows.MessageBoxResult defaultResult) { throw null; }
         public static System.Windows.MessageBoxResult Show(string messageBoxText, string caption, System.Windows.MessageBoxButton button, System.Windows.MessageBoxImage icon, System.Windows.MessageBoxResult defaultResult, System.Windows.MessageBoxOptions options) { throw null; }
+        public static System.Windows.MessageBoxResult Show(IntPtr owner, string messageBoxText, string caption, System.Windows.MessageBoxButton button) { throw null; }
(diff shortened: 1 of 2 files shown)


lindexi:
Maybe we can introduce optional parameter for the method `Show`

src/Microsoft.DotNet.Wpf/cycle-breakers/PresentationFramework/PresentationFramework.cs
```diff
@@ -943,12 +943,18 @@ public sealed partial class MessageBox
...
         public static System.Windows.MessageBoxResult Show(System.Windows.Window owner, string messageBoxText, string caption, System.Windows.MessageBoxButton button, System.Windows.MessageBoxImage icon) { throw null; }
         public static System.Windows.MessageBoxResult Show(System.Windows.Window owner, string messageBoxText, string caption, System.Windows.MessageBoxButton button, System.Windows.MessageBoxImage icon, System.Windows.MessageBoxResult defaultResult) { throw null; }
         public static System.Windows.MessageBoxResult Show(System.Windows.Window owner, string messageBoxText, string caption, System.Windows.MessageBoxButton button, System.Windows.MessageBoxImage icon, System.Windows.MessageBoxResult defaultResult, System.Windows.MessageBoxOptions options) { throw null; }
+        public static System.Windows.MessageBoxResult Show(System.Windows.Window owner, string messageBoxText) { throw null; }
+        public static System.Windows.MessageBoxResult Show(System.Windows.Window owner, string messageBoxText, string caption) { throw null; }
```
miloush, line 957: How are these lines different to 950-951?

miloush:
> in which case the messagebox appears centered at the desktop

In case of the overloads without `Window` argument, the active window is used as parent rather than desktop I believe.

I am not very keen on adding public `IntPtr` overloads, allowing null `Window` sounds like a better solution, without affecting any existing application or changing public API.

PR #4519, opened 2021-05-16
enable nullable for new projects

jmarolf:
resolves https://github.com/dotnet/templates/issues/485


jmarolf:
@dotnet/wpf-developers ping

jmarolf:
As winforms and the main .NET templates repos have merged in this change I am going to do so here for consistency. Please feel free to revert this if that is what the ream decides

PR #4559, opened 2021-05-24
Update ComboBoxAutomationPeer.cs

kronic:
Fix Invalid cast exception. ComboBoxAutomationPeer is not implemented IScrollProvider.


ayushverma-ms:
@kronic Combobox is the grandchild of ItemControl. ItemControl has a variable of ScrollViewer class, which can provide an implementation of IScrollProvider. (Basically, it is a ScrollViewerAutomationPeer instance).

PR #4605, opened 2021-06-01
Ingest and import Windows Forms analyzer props

RussKie:
## Description

This work relates to [Streamline Windows Forms application configuration and bootstrap design proposal](https://github.com/dotnet/designs/pull/223), and acts as a workaround until [Proposal to add Framework specific inbox source generators design proposal](https://github.com/dotnet/designs/pull/181) is approved and implemented.

Provide a mechanism to ingest Windows Forms specific analyzer props into Microsoft.NET.Sdk.WindowsDesktop from Windows Forms transport package, and copy the props file into `targets` folder of the SDK, so that the props file is resolved and imported when a developer builds a Windows Forms app.

In the end the new props file will end up in a location similar to this: `C:\Program Files\dotnet\sdk\6.0.<version>\Sdks\Microsoft.NET.Sdk.WindowsDesktop\targets\`.

The build process looks something like this:
* import props and targets
    image
    image
* execute the task that resolves and references Windows Forms analyzers:
    image

## Testing

A lot of manual testing

1. Create a Windows Forms transport package, e.g.: `winforms> .\build.cmd -pack`
2. Copy the build transport package from .\winforms\artifacts\packages\Debug\NonShipping over the currently referenced by WPF microsoft.private.winforms package in C:\Users\\<user\>\\.nuget\packages\microsoft.private.winforms
3. Build and pack WPF: `wpf> .\build -pack`
4. Inspect .\wpf\artifacts\packaging\Debug\Microsoft.NET.Sdk.WindowsDesktop.Debug\targets folder and see System.Windows.Forms.Analyzers.props and System.Windows.Forms.Analyzers.targets present

### DIFF
--- a/packaging/Microsoft.NET.Sdk.WindowsDesktop/Microsoft.NET.Sdk.WindowsDesktop.ArchNeutral.csproj
+++ b/packaging/Microsoft.NET.Sdk.WindowsDesktop/Microsoft.NET.Sdk.WindowsDesktop.ArchNeutral.csproj
@@ -1,4 +1,4 @@
-﻿<Project Sdk="Microsoft.NET.Sdk">
+<Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
     <ProjectGuid>{440d06b8-e3de-4c0d-ad25-cd4f43d836e1}</ProjectGuid>
     <TargetFramework>net6.0</TargetFramework>
@@ -25,4 +25,30 @@
     <PackagingContent Include="targets\*" SubFolder="root\targets" />
     <PackagingContent Include="useSharedDesignerContext.txt" SubFolder="root" />
   </ItemGroup>
+
+  <!-- Windows Forms specific -->
+  <PropertyGroup>
+    <_WinFormsNuGetPath>$(PkgMicrosoft_Private_Winforms)</_WinFormsNuGetPath>
+    <_WinFormsContentPath>$(_WinFormsNuGetPath)\sdk\dotnet-wpf\*</_WinFormsContentPath>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackagingContent Include="$(_WinFormsContentPath)" SubFolder="root\targets" />
+  </ItemGroup>
+
+  <!-- 
+    This target *only* validates the content of Microsoft.Private.Winforms NuGet package
+    to ensure we correctly import and reference props and targets.
+   -->
+  <Target Name="_EnsureWindowsFormsPackagingContent" BeforeTargets="IdentifyPackageAssets">
+    <Error Text="Unable to resolve path to Microsoft.Private.Winforms NuGet package. Is %24(PkgMicrosoft_Private_Winforms) defined?"
+           Condition="'$(_WinFormsNuGetPath)' == ''"/>
+
+    <ItemGroup>
+      <_WinFormsContent Include="$(_WinFormsContentPath)" />
+    </ItemGroup>
+
+    <Error Text="Microsoft.Private.Winforms NuGet package contains no content. Is this expected?"
+           Condition="@(_WinFormsContent->Count()) == 0"/>
+  </Target>
 </Project>
--- a/packaging/Microsoft.NET.Sdk.WindowsDesktop/Microsoft.NET.Sdk.WindowsDesktop.ArchNeutral.csproj
+++ b/packaging/Microsoft.NET.Sdk.WindowsDesktop/Microsoft.NET.Sdk.WindowsDesktop.ArchNeutral.csproj
@@ -1,4 +1,5 @@
-<Project Sdk="Microsoft.NET.Sdk">
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
   <PropertyGroup>
     <ProjectGuid>{440d06b8-e3de-4c0d-ad25-cd4f43d836e1}</ProjectGuid>
     <TargetFramework>net6.0</TargetFramework>
@@ -27,6 +28,8 @@
   </ItemGroup>
 
   <!-- Windows Forms specific -->
+  <Import Project="$(RepositoryEngineeringDir)WindowsForms.targets" />
+
   <PropertyGroup>
     <_WinFormsNuGetPath>$(PkgMicrosoft_Private_Winforms)</_WinFormsNuGetPath>
     <_WinFormsContentPath>$(_WinFormsNuGetPath)\sdk\dotnet-wpf\*</_WinFormsContentPath>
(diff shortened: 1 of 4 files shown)


packaging/Microsoft.NET.Sdk.WindowsDesktop/Microsoft.NET.Sdk.WindowsDesktop.ArchNeutral.csproj
```diff
@@ -24,5 +24,8 @@
     <PackagingContent Include="Sdk\*" SubFolder="root\Sdk" />
     <PackagingContent Include="targets\*" SubFolder="root\targets" />
     <PackagingContent Include="useSharedDesignerContext.txt" SubFolder="root" />
+
+    <!-- Windows Forms specific -->
+    <PackagingContent Include="$(PkgMicrosoft_Private_Winforms)\Wpf.Sdk.WindowsDesktop\*" SubFolder="root\targets" />
```
vatsan-madhavan, line 29: Agree that its supposed to be as simple as this. IIRC WPF repo uses  `GeneratePathProperty` with most Nuget packages and its going to be true for the WinForms transport package as well. 

Side note, I've always been sad that WinForms didn't get the casing for the casing for the transport package name like `WinForms` and it's like `Winforms` - so it's pretty confusing to type. 

Also, don't bundle your targets inside a subfolder named "Wpf.Sdk.WindowsDesktop" - bundle it inside a subfolder named "build\" or something canonical that NuGet understands well. See https://docs.microsoft.com/en-us/nuget/create-packages/creating-a-package. (Not a blocking comment - what you have here ought to work just fine). 

Overall your approach looks fine to me.

vatsan-madhavan: Outside these props, ensure that when `PkgMicrosoft_Private_Winforms` is undefined, the build fails. Good chance that's already the case (or should be) but it's very important this happens from here onwards. 

Likewise, make sure that the build fails if `S.W.F.Analyser.[props|targets]` aren't found under the expected locations under `$(PkgMicrosoft_Private_Winforms)`. 

The way `PackagingContent` works right now, if those files aren't found, then packaging will just proceed along without them - so you're responsible for validating.

RussKie: > IIRC WPF repo uses `GeneratePathProperty` with most NuGet packages and its going to be true for the WinForms transport package as well.

It is here:
https://github.com/dotnet/wpf/blob/89d172db0b7a192de720c6cfba5e28a1e7d46123/eng/WpfArcadeSdk/tools/SdkReferences.targets#L3-L10


> Also, don't bundle your targets inside a subfolder named "Wpf.Sdk.WindowsDesktop" - bundle it inside a subfolder named "build" or something canonical that NuGet understands well. See [docs.microsoft.com/en-us/nuget/create-packages/creating-a-package](https://docs.microsoft.com/en-us/nuget/create-packages/creating-a-package). (Not a blocking comment - what you have here ought to work just fine).

We have some props/targets that go to WPF and some that go to WindowsDesktop. This is an attempt to provide some indication of the destination (as well as some docs in our repo).
Also I tried putting these under "build", and that results in [NuGet Warning NU5129](https://docs.microsoft.com/en-us/nuget/reference/errors-and-warnings/nu5129).



> Outside these props, ensure that when `PkgMicrosoft_Private_Winforms` is undefined, the build fails.....

Good idea 👍 

Thank you

RussKie: Added a target to verify
image

vatsan-madhavan: NU5129 seems like a common enough problem and I've had to suppress it elsewhere for similar reasons when creating NuGet packages with build\ subfolder.

packaging/Microsoft.NET.Sdk.WindowsDesktop/Microsoft.NET.Sdk.WindowsDesktop.ArchNeutral.csproj
```diff
@@ -20,9 +20,30 @@
...
   <ItemGroup>
     <PackagingContent Include="Sdk\*" SubFolder="root\Sdk" />
     <PackagingContent Include="targets\*" SubFolder="root\targets" />
     <PackagingContent Include="useSharedDesignerContext.txt" SubFolder="root" />
+
+    <!-- Windows Forms specific -->
+    <PackagingContent Include="$(_WinFormsContentPath)" SubFolder="root\targets" />
   </ItemGroup>
+
+  <Target Name="_EnsureWindowsFormsPackagingContent" BeforeTargets="IdentifyPackageAssets">
```
vatsan-madhavan, line 37: This is an antipattern. Foo.*ArchNeutral.csproj, Foo.csproj etc. have certain conventions, and they weren't intended to be spots for additional domain-specific targets. Our domain-specific targets exist under a certain hierarchy under eng\ that I've already alluded to.

vatsan-madhavan: I haven't read the docs you've shared, but my sense/guess here is that analyzer related props/targets is a new category of payload that's being introduced. If this is right... then consider introducing a new `Item` name, something like `AnalyzerPackagingContent` that's used to track these items. These would become agnostic of WPF/Winforms etc. far as build-logic goes.

Then in `Packaging.targets`, add logic specific to validation of these. For e.g., anything that's been added to `AnalyzerPackagingContent` - ensure that they exist, make sure they're added to `PackagingContent` etc.. Then all you'd do in `*.WindowsDesktop.ArchNeutral.csproj` is adding your props/targets to the newly introduced `AnalyzerPackagingContent` `Item` and let the backend targets do its work as before.

packaging/Microsoft.NET.Sdk.WindowsDesktop/Microsoft.NET.Sdk.WindowsDesktop.ArchNeutral.csproj
RussKie: Our transport package has the following structure:
image

packaging/Microsoft.NET.Sdk.WindowsDesktop/Microsoft.NET.Sdk.WindowsDesktop.ArchNeutral.csproj
RussKie: My understanding that once https://github.com/dotnet/designs/pull/181 is approved and implemented none of this will be required. @ericstj am I reading your proposal correctly?

RussKie: I added this target purely to perform verifcations as you suggested earlier.
All work is done by these targets https://github.com/RussKie/winforms/blob/project_configuration/src/System.Windows.Forms.Analyzers/src/System.Windows.Forms.Analyzers.targets

RussKie: The reason I added this target in this project and not in Directory.Build.targets at other levels is these validations is only relevant to this project. I'm happy to move it out to a separate targets file under eng\ folder and reference it in this project.

vatsan-madhavan: A different architecture is fine, but this project is not the right place for adding targets, imports etc. - if you need to extend or hook the build system, it has to be done elsewhere.

RussKie: Does 9c0757e7519cff48333b376866d417cd0e1da3b5 look better?

RussKie:
Tested using an [experimental pipeline](https://dev.azure.com/dnceng/internal/_build/results?buildId=1175975&view=results):
image

Issue #1014, opened 2019-06-19
Improve WPF Tooltips Hover for Accessibility 


grubioe:
Copied from ADO ID: 917236

Repro Steps:

1. Launch affected app referenced in VSO ID
2. Navigate to various actionable UI in the toolbar.
3. Check if the tooltip displayed is satisfying the below conditions:​
         Hoverable​
         Persistent

-affected app has accessibility updates enabled in config file.
-affected app just uses ToolTip property on button or RibbonButton to add tooltip.  Changing app to use tooltip object via <Button.Tooltip><Tooltip/></Button.Tooltip> does not affect behavior.

Expected Result:

Hoverable: Should be able to hover on the additional content(tooltip)​
Persistent: The additional content(tooltip) should be visible until the hover or focus trigger is removed, the user dismisses it, or its information is no longer valid

Actual Result:

The tooltip is not Hoverable. Tooltip is dismissed on hovering to it.​  
The tooltip is not persistent. Tooltip is dismissed after some time without removing the focus or hover.

Initial Discussion from team research:
I don't believe we have any logic in place to detect hover on the tooltip and keep it alive, when the mouse moves out of the tooltip's parent element the tooltip is automatically dismissed.

arpitmathur:
This feature request is currently out of 3.0. We should evaluate this issue for 3.1 or 5.0.

jlukosky:
Shouldn't this be tagged _Accessibility MAS_ as two of the three requirements of MAS 1.4.13: Content on Hover or Focus? Removing the label may certainly be appropriate but I'm not sure of the reasoning.

SamBent:
Repurposing this issue to track the WCAG 2.1 ToolTip work (which subsumes the work for which this issue was opened).  See the [spec](https://github.com/dotnet/wpf/blob/main/src/Microsoft.DotNet.Wpf/specs/tooltip.md).

ryalanms:
This is fixed by @SamBent's change.

Issue #4612, opened 2021-06-02
ToolTip issues prior to .NET 6.0


SamBent:
This issue contains references to a number of ToolTip issues.  It is intended as a convenience for search, and a collection point for common information.  Please limit discussion to topics common to the collection as a whole;  comments about the individual bugs should go into the appropriate linked issue.

The issues collected here are
- Issues present in .NET 4.8 and above (i.e. .NET Core 3.x, .NET 5.x, previews of .NET 6.0)
- Actual bugs, as opposed to design changes
- Contrary to, or blocking of, the new ToolTip behavior mandated by the WCAG 2.1 compliance work.  See the [spec](https://github.com/dotnet/wpf/blob/main/src/Microsoft.DotNet.Wpf/specs/tooltip.md)
- Fixed by that work (in .NET 6.0)
- Not planned to be fixed in any earlier versions

Here is a sample app that illustrates most of the issues (see individual bugs for repro steps):
[WpfApp3.zip](https://github.com/dotnet/wpf/files/6650174/WpfApp3.zip)
The app artificially sets all tooltips to delay for 5 seconds before opening, and to close after displaying for 15 seconds.  This makes it easier to see the behavior, but does require a little patience.

SamBent:
Here are the issues (not all of which appear in the mentions you probably see above):

- **ToolTip shows when owner acquires focus via non-keyboard #4613**.   Moving between apps, minimizing/restoring an app, and other non-keyboard ways of changing focus, should not show the tooltip.
- **Moving focus to an element with no ToolTip opens an ancestor's ToolTip #4614**.   Moving focus should not open a different element's tooltip.
- **Ctrl+Shift+F10 waits before opening the ToolTip #4615**.  Ctrl+Shift+F10 should toggle the tooltip immediately.
- **Mouse motion closes a ToolTip shown by keyboard #4616**.   Unrelated mouse motion should not close the tooltip.
- **Moving keyboard focus while a ToolTip is open shows the new ToolTip only after a delay #4617**.  Keyboard should work the same as mouse - a new tooltip shows immediately if another tooltip was already showing.
- **ToolTip toggles on Ctrl+Shift+Alt+F10 #4618**.  Extra modifier keys shouldn't toggle the tooltip.
- **BetweenShowDelay=0 does not work #4619**.  Setting BetweenShowDelay to 0 should force the next tooltip to delay before opening.
[...]

PR #4623, opened 2021-06-02
Tooltip

SamBent:
Fixes Issue #1014

Also fixes _en passant_ the issues collected in #4612.

## Description

ToolTip behavior to comply with WCAG 2.1 and MAS 1.4.13.  See detailed [spec](https://github.com/dotnet/wpf/blob/main/src/Microsoft.DotNet.Wpf/specs/tooltip.md).

## Customer Impact

Enables compliance with WCAG 2.1 and MAS 1.4.13.

## Regression

No.

## Testing

(in progress)

## Risk

Required by WCAG, MAS.

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Controls/PopupControlService.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Controls/PopupControlService.cs
@@ -538,11 +538,15 @@ namespace System.Windows.Controls
 
                     // begin the BetweenShowDelay interval, during which another tooltip
                     // can open without the usual delay
-                    _quickShow = true;
-                    CurrentToolTipTimer = new DispatcherTimer(DispatcherPriority.Normal);
-                    CurrentToolTipTimer.Interval = TimeSpan.FromMilliseconds(ToolTipService.GetBetweenShowDelay(owner));
-                    CurrentToolTipTimer.Tick += new EventHandler(OnBetweenShowDelay);
-                    CurrentToolTipTimer.Start();
+                    int betweenShowDelay = ToolTipService.GetBetweenShowDelay(owner);
+                    _quickShow = (betweenShowDelay > 0);
+                    if (_quickShow)
+                    {
+                        CurrentToolTipTimer = new DispatcherTimer(DispatcherPriority.Normal);
+                        CurrentToolTipTimer.Interval = TimeSpan.FromMilliseconds(betweenShowDelay);
+                        CurrentToolTipTimer.Tick += new EventHandler(OnBetweenShowDelay);
+                        CurrentToolTipTimer.Start();
+                    }
                 }
                 else
                 {
--- a/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Controls/PopupControlService.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Controls/PopupControlService.cs
@@ -190,13 +190,24 @@ namespace System.Windows.Controls
         {
             if (!e.Handled)
             {
-                if ((e.SystemKey == Key.F10) && ((Keyboard.Modifiers & ModifierKeys.Shift) == ModifierKeys.Shift) && ((Keyboard.Modifiers & ModifierKeys.Control) == ModifierKeys.Control))
+                const ModifierKeys ModifierMask = ModifierKeys.Alt | ModifierKeys.Control | ModifierKeys.Shift | ModifierKeys.Windows;
+                ModifierKeys modifierKeys = Keyboard.Modifiers & ModifierMask;
+
+                if ((e.SystemKey == Key.F10) && (modifierKeys == (ModifierKeys.Control | ModifierKeys.Shift)))
                 {
                     e.Handled = OpenOrCloseToolTipViaShortcut();
                 }
-                else if ((e.SystemKey == Key.F10) && ((Keyboard.Modifiers & ModifierKeys.Shift) == ModifierKeys.Shift))
+                else if ((e.SystemKey == Key.F10) && (modifierKeys == ModifierKeys.Shift))
                 {
                     RaiseContextMenuOpeningEvent(e);
+                }
+
+                // track the last key-down, to detect Ctrl-KeyUp trigger
+                _lastCtrlKeyDown = Key.None;
+                if ((CurrentToolTip?.FromKeyboard ?? false) && (modifierKeys == ModifierKeys.Control) &&
+                        (e.Key == Key.LeftCtrl || e.Key == Key.RightCtrl))
+                {
+                    _lastCtrlKeyDown = e.Key;
                 }
             }
         }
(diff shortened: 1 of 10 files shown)


src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Controls/PopupControlService.cs
```diff
@@ -236,557 +221,557 @@ private void ProcessKeyUp(object sender, KeyEventArgs e)
...
-                    {
-                        return o;
-                    }
-                    else
+                    // get owner and tooltip rectangles, in owner window's client coords
+                    Rect rectElement = new Rect(new Point(0, 0), owner.RenderSize);
+                    Rect rectRoot = PointUtil.ElementToRoot(rectElement, owner, presentationSource);
+                    Rect ownerRect = PointUtil.RootToClient(rectRoot, presentationSource);
+
+                    Rect screenRect = tooltip.GetScreenRect();
```
lindexi, line 742: Can it work well when we have two Moniter with difference DPI and the owner displayed in the Moniter one and the tooltip displayed in the Moniter two?

src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Controls/PopupControlService.cs
```diff
@@ -1173,19 +1137,400 @@ internal KeyboardFocusChangedEventHandler FocusChangedEventHandler
...
+            {
+                _source = source;
+            }
+
+            // sort by y (and by x among equal y's)
+            private void SortPoints(PointList points)
+            {
+                // insertion sort is good enough.  We're dealing with a small
+                // set of points that are nearly in the right order already.
+                for (int i=0, N=points.Count; i<N; ++i)
```
lindexi, line 1227: Can we start with 1?

```suggestion
                for (int i=1, N=points.Count; i<N; ++i)
```

src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Controls/PopupControlService.cs
```diff
@@ -1173,19 +1137,400 @@ internal KeyboardFocusChangedEventHandler FocusChangedEventHandler
...
+                points.RemoveRange(hullCount, N - hullCount);
+                _points = points.ToArray();
+                SetDirections();
+            }
+
+            // set the Direction field on each point.  This enables optimizations during
+            // ContainsPoint.
+            private void SetDirections()
+            {
+                for (int i=0, N=_points.Length; i<N; ++i)
```
lindexi, line 1386: Should we set the `_points[0].Direction = Direction.Skew;` when the `_points.Length` is one?

src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Controls/PopupControlService.cs
```diff
@@ -236,557 +221,557 @@ private void ProcessKeyUp(object sender, KeyEventArgs e)
...
                 }
+
+                _currentToolTip.SetValue(OwnerProperty, o);
+                _currentToolTip.Opened += OnToolTipOpened;
+                _currentToolTip.Closed += OnToolTipClosed;
+                _currentToolTip.FromKeyboard = fromKeyboard;
+                _currentToolTip.IsOpen = true;
+
+                CurrentToolTipTimer = new DispatcherTimer(DispatcherPriority.Normal);
+                CurrentToolTipTimer.Interval = TimeSpan.FromMilliseconds(ToolTipService.GetShowDuration(o));
```
lindexi, line 400: We change the  ShowDuration default value from 5000 to int.MaxValue in ToolTipService.cs . And I think we can check the  ToolTipService.GetShowDuration result first, because we will do nothing when it is int.MaxValue

lindexi (commented):
Thank you @SamBent

src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Controls/PopupControlService.cs
```diff
@@ -236,557 +221,557 @@ private void ProcessKeyUp(object sender, KeyEventArgs e)
...
+                    // Bind the content of the tooltip to the ToolTip attached property
+                    Binding binding = new Binding();
+                    binding.Path = new PropertyPath(ToolTipService.ToolTipProperty);
+                    binding.Mode = BindingMode.OneWay;
+                    binding.Source = o;
+                    _currentToolTip.SetBinding(ToolTip.ContentProperty, binding);
                 }
-                else if (args.KeepCurrentActive)
+
+                if (!_currentToolTip.StaysOpen)
```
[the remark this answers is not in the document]
SamBent: There's no harm in checking StaysOpen here, even though it will always be true in the "new ToolTip" case.  The old code did it this way, and it's good defensive programming in case someone changes the default for ToolTip.StaysOpen.

src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Controls/PopupControlService.cs
SamBent: This is defensive programming, always starting the timer even when it won't tick for 25 days.  Code doesn't have to check for special cases (including whether CurrentToolTipTimer is null).

src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Controls/PopupControlService.cs
```diff
@@ -236,557 +221,557 @@ private void ProcessKeyUp(object sender, KeyEventArgs e)
...
-                else if (_lastChecked == null)
-                {
-                    _lastChecked = new WeakReference(value);
-                }
-                else
-                {
-                    _lastChecked.Target = value;
-                }
-            }
+        private bool MouseHasLeftSafeArea(RawMouseInputReport mouseReport)
```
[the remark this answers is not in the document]
SamBent: Reading aloud "if (MouseHasLeftSafeArea)" sounds like English;  reading "if (IsMouseHasLeftSafeArea)" doesn't.   The advice to start the name of a boolean property with "Is" is meant to improve readability;  it's good advice much of the time ("IsEnabled", "IsMouseDirectlyOver"), but needn't be followed slavishly if it makes readability worse.

src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Controls/PopupControlService.cs
SamBent: Yes.

src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Controls/PopupControlService.cs
SamBent: That can't happen in our use-case - the convex hull of a non-zero number of non-empty rectangles will have at least 4 points.  In general, the convex hull of any set of points will have at least 3 points unless it's 1-dimensional (all points are collinear, hull has 2 points) or 0-dimensional (the set is a single point, hull has 1 point).  

Our ContainsPoint implementation only works when the hull has 2-d area (which is the only case we'll see).  Changing the direction to Skew in the 1-point case doesn't fix that - it will still give the wrong answer, just for a different reason.

fabiant3 (approved):
Awesome to see this work!

ryalanms (approved):

src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Controls/PopupControlService.cs
SamBent: Tooltips are always displayed on the same monitor as the owner.

PR #4630, opened 2021-06-04
Disable NETFX WinFX targets only when the SDK is used directly

### LINKED ISSUE #4515: SDK-style WPF Projects targeting net472 are broken because ImportFrameworkWinFXTargets is false
This PR has regressed building WPF projects that use the SDK-style:
https://github.com/dotnet/wpf/pull/2976

By default ImportFrameworkWinFXTargets was `true`, but now changes to `false`, resulting in MarkupCompile not running.

FYI @Nirmal4G

Nirmal4G:
Partially Fixes #4515
Partially Fixes dotnet/sdk#18062

## Description

Disable .NET Framework's WinFX targets import only when the Windows Desktop SDK is used directly instead!

Example:
```xml
<Project Sdk="Microsoft.NET.Sdk.WindowsDesktop" />
```

Here, we're simply moving the logic to the `Sdk.props`. This still does not fix the root of the problem
where the SDK provided WinFX targets should work properly when targeting .NET Framework only projects.

## Customer Impact

As per the issues mentioned above, the .NET Framework projects that depends on older inbox WinFX targets no longer used those targets and so, the build fails.

## Regression

Yes, it was a regression of a Desktop SDK's PBT bug that revealed itself through #2976.

## Testing

Works when tested locally.

## Risk

Low, since the .NET 6 is still in preview.

It only fixes when the SDK used is .NET SDK instead of Windows Desktop SDK. Since .NET SDK in post 5.0 imports
`WindowsDesktop.props` and so, we simply move the logic to `Sdk.props`. So, for those who depends on this behavior
should use Windows Desktop SDK instead of the Main SDK, as shown above.

### DIFF
--- a/packaging/Microsoft.NET.Sdk.WindowsDesktop/Sdk/Sdk.props
+++ b/packaging/Microsoft.NET.Sdk.WindowsDesktop/Sdk/Sdk.props
@@ -15,10 +15,21 @@ Copyright (c) .NET Foundation. All rights reserved.
     <_MicrosoftWindowsDesktopSdkImported>true</_MicrosoftWindowsDesktopSdkImported>
   </PropertyGroup>
 
+  <!--
+    Workaround: https://github.com/microsoft/msbuild/issues/4948
+    Disable .NET Framework's inbox WinFX targets when using the SDK, since, we really don't use it's build logic
+    and is superseded by 'WindowsDesktop' SDK that provides it's own WinFX for both NETFX and CoreCLR targets.
+    Make it opt-out, just in case, if something fails or we don't want to use 'WindowsDesktop' SDK's version.
+  -->
+  <PropertyGroup>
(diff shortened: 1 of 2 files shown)


Nirmal4G:
This and its parent PRs related to WinFX targets have been backported to `v3.1` and `v5.0`. Please look into those too!

Also, this is just a workaround to unblock working builds which were previously using inbox WinFX targets. My #2976 introduced a change where it forces to use the SDK's WinFX targets instead. Anyhow, this should not have failed since the SDK's WinFX targets also have `net472` `PresentationBuildTasks` assembly. I should look into this more... BTW #4218 looks interesting as it could fix the root of the problem.

dsplaisted (approved):

ryalanms:
Thanks, @Nirmal4G.

ryalanms (approved):

PR #4632, opened 2021-06-05
Fix incorrect marshalling spec

kant2002:
DeactivateActCtx do not have HRESULT as returning type, so PreserveSig == false is no op.
SHCreateItemFromParsingName will always throw in case of error,
and error handling in
https://github.com/dotnet/wpf/blob/d26eaaca20aa2f5de768e4053593f50a30776380/src/Microsoft.DotNet.Wpf/src/PresentationFramework/MS/Internal/AppModel/ShellProvider.cs#L901-L909
would never works.

Discovered as part of attempt to compile WPF in Native AOT in https://github.com/dotnet/runtimelab/issues/1183
Also apologize for style correction, seems to be editorconfig picked up.


jkotas:
> so PreserveSig == false is no op.

`PreserveSig == false` is plain wrong. It is mismatched PInvoke signature that can lead to unpredictable behavior at runtime.

jkotas (approved):

ryalanms (approved):

PR #4717, opened 2021-06-23
Avoid new delegate allocation on each DispatcherTimer tick

stephentoub:
## Description

Every time DispatcherTimer.Restart is called (which may happen on every tick), it's allocating a new delegate to point to the instance's FireTick method.  We can instead make it a non-capturing delegate, such that the compiler caches it on first use and never has to allocate it again, passing in `this` via the object state.

## Customer Impact

Less garbage allocated for the GC to clean up.

## Regression

No

## Testing

Only what's done by CI.

## Risk

Minimal.  The change just modifies how `this` is passed to the callback.

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/WindowsBase/System/Windows/Threading/DispatcherTimer.cs
+++ b/src/Microsoft.DotNet.Wpf/src/WindowsBase/System/Windows/Threading/DispatcherTimer.cs
@@ -267,8 +267,8 @@ namespace System.Windows.Threading
                 // BeginInvoke a new operation.
                 _operation = _dispatcher.BeginInvoke(
                     DispatcherPriority.Inactive,
-                    new DispatcherOperationCallback(FireTick),
-                    null);
+                    (DispatcherOperationCallback)(state => ((DispatcherTimer)state).FireTick()),
+                    this);
 
                 
                 _dueTimeInTicks = Environment.TickCount + (int) _interval.TotalMilliseconds;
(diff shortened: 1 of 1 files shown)


lindexi:
Awesome!

src/Microsoft.DotNet.Wpf/src/WindowsBase/System/Windows/Threading/DispatcherTimer.cs
```diff
@@ -267,8 +267,8 @@ private void Restart()
                 // BeginInvoke a new operation.
                 _operation = _dispatcher.BeginInvoke(
                     DispatcherPriority.Inactive,
-                    new DispatcherOperationCallback(FireTick),
-                    null);
+                    (DispatcherOperationCallback)(state => ((DispatcherTimer)state).FireTick()),
```
kronic, line 270: (static state => ((DispatcherTimer)state).FireTick())

stephentoub: The project is being built with an older LangVersion that doesn't support using static lambdas.

lindexi: @stephentoub   Can the compiler cache it ?

stephentoub: Yes,  that's the purpose of this PR. The static keyword here would have zero impact on code gen; it would purely be there to trigger errors if the code were changed to try to capture something.

ryalanms (approved):

PR #4220, opened 2021-02-25
avoid boxing

i-kostikov:
Fixes Issue [4114](https://github.com/dotnet/wpf/issues/4114)

Master PR 

## Description
#4114 
Using the EqualityComparer<T>.Default instead the ReferenceEquals method to avoid the boxing for value types.

## Risk

No risks

PR #4724, opened 2021-06-23
Avoid boxing in FrugalList

stephentoub:
## Description

All comparisons are being performed with object.Equals(object).  If T is a value type, every comparison ends up boxing the argument.

## Customer Impact

Less boxing means less garbage generated and less time spent on collections.

## Regression

No

## Testing

Just CI

## Risk

Minimal. `EqualityComparer<T>.Default.Equals` will use `object.Equals` if the type doesn't provide an `IEquatable<T>` implementation.

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/Shared/MS/Utility/FrugalList.cs
+++ b/src/Microsoft.DotNet.Wpf/src/Shared/MS/Utility/FrugalList.cs
@@ -265,12 +265,12 @@ namespace MS.Utility
 
         public override bool Contains(T value)
         {
-            return _loneEntry.Equals(value);
+            return EqualityComparer<T>.Default.Equals(_loneEntry, value);
         }
 
         public override int IndexOf(T value)
         {
-            if (_loneEntry.Equals(value))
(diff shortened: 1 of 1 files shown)


lindexi:
The other PR: https://github.com/dotnet/wpf/pull/4220

ThomasGoulet73:
Would it be worth to do something like Dictionary to devirtualize when it's a value type and cache when it's a reference type ?
https://github.com/dotnet/runtime/blob/main/src/libraries/System.Private.CoreLib/src/System/Collections/Generic/Dictionary.cs#L274-L298

stephentoub:
> Would it be worth to do something like Dictionary to devirtualize when it's a value type and cache when it's a reference type ?

I don't think so.  We don't do so in the vast majority of use cases.  The difference is tiny and only relevant in the hottest of hot paths when the rest of the overheads are tiny.  It's also the kind of thing that is temporarily an optimization but could easily become a deoptimization as the JIT improves, e.g. it should become unnecessary in dictionary once https://github.com/dotnet/runtime/pull/50446 is merged.

ryalanms (approved):

PR #4736, opened 2021-06-24
Avoid calling GetInvocationList on hot paths

stephentoub:
## Description

GetInvocationList on one or more delegates is being called in response to most any input operation.  We can instead call GetInvocationList only when the set of delegates changes, and use that list over and over.

## Customer Impact

Reduced Delegate[] allocation leading to less pressure on the GC.  You can see the impact of this with a simple sample WPF app; in this trace, I just open the app and interact with it for a few seconds.  These are the top allocations:
image

## Regression

No

## Testing

Just CI

## Risk

I'm making an assumption that it's much more common for these events to be invoked than it is for them to be changed (delegates added or removed from the events).  If it were more common for them to be changed than for them to be invoked, this would be a deoptimization.


src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Input/InputManager.cs
```diff
@@ -974,14 +951,14 @@ private void RequestContinueProcessingStagingArea()
...
-        private event NotifyInputEventHandler _preNotifyInput;
-
-        private event NotifyInputEventHandler _postNotifyInput;
-
-        private event ProcessInputEventHandler _postProcessInput;
+        // These four events introduced for security purposes. Rather than just store the multicast delegate
+        // and have to do GetInvocationList on each invocation (in order to invoke delegates in reverse-registered
+        // order), we get the invocation list when the events are updated, and then use that cached list
+        // on every invocation.
+        private Tuple<PreProcessInputEventHandler, Delegate[]> _preProcessInput;
```
lindexi, line 958: How about ValueTuple?

stephentoub: That can't be swapped atomically with an Interlocked.  I don't actually know if thread-safety is necessary here, but `event` provides that automatically, so I brought the behavior forward.

lindexi: @stephentoub  Yeah, thread safety is necessary. Such as I will add handle in my `Stylus Input` thread.

src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/EventHelper.cs
```diff
@@ -0,0 +1,55 @@
...
+
+        /// <summary>Removes a delegate from an existing delegate.</summary>
+        /// <param name="field">A tuple of a delegate and its invocation list already in array form.</param>
+        /// <param name="value">The delegate to remove.</param>
+        public static void RemoveHandler<T>(ref Tuple<T, Delegate[]> field, T value) where T : Delegate
+        {
+            while (true)
+            {
+                Tuple<T, Delegate[]> oldTuple = field;
+                T delegateWithRemoval = (T)Delegate.Remove(oldTuple?.Item1, value);
```
lindexi, line 46: Should we do anything when the `oldTuple` is null?

stephentoub: It's doing `oldTuple?.Item1`, so if `oldTuple` is null, it'll pass null as the first argument to Remove, and Remove will return null.

lindexi: @stephentoub Thank you. Is it necessary to return fast when the `field` is null? I find some code which will remove the handle before add the handle in my application.


```csharp
            InputManager.Current.PreNotifyInput -= InputManager_PreNotifyInput; // Often this PreNotifyInput event is null
            InputManager.Current.PreNotifyInput += InputManager_PreNotifyInput;
```

ryalanms (approved):

Issue #3885, opened 2020-12-06
Use C# for PresentationCore module initializer instead of the existing IL injection


LakshanF:
Trimming a default WPF application causes issues for the linker and I'm following up on those. One issue is related with the use of IL injection for the module cctor since earlier version of the C# compiler didn't have this support. 

My colleague, @vitek-karas, has a change that addresses [this](https://github.com/dotnet/wpf/compare/master...vitek-karas:UseModuleInit) that can be leveraged for this purpose. The existing IL injection at [Arcade](https://github.com/dotnet/wpf/tree/master/eng/WpfArcadeSdk/tools/InjectModuleInitializer) might require some cleanup which could be done better by a WPF domain expert.

The other [issues](https://github.com/dotnet/sdk/issues/14261) related to WPF trimming are changes that we need to make on the linker side, at least to peel away to uncover any other deeper issues related to WPF heavy use of C++/CLI assemblies.

madewokherd:
FWIW, in Wine Mono I was able to replace DirectWriteForwarder with a C# assembly, except for the custom loader interfaces where I encountered GC issues and moved the implementation to a native dll as a light wrapper around a managed callback.

ryalanms:
Thanks, @LakshanF.

PR #4748, opened 2021-06-24
Update LangVersion to C# 9

ThomasGoulet73:
## Description
Update LangVersion to C# 9. WPF seems to be the only repo that still use C# 8.

Here's a list of popular dotnet repos and their version:
aspnetcore: [9.0](https://github.com/dotnet/aspnetcore/blob/main/eng/targets/CSharp.Common.props#L4)
efcore: [9.0](https://github.com/dotnet/efcore/blob/main/Directory.Build.props#L24)
runtime: [preview](https://github.com/dotnet/runtime/blob/main/Directory.Build.props#L282)
winforms: [preview](https://github.com/dotnet/winforms/blob/main/Directory.Build.props#L16)

This change allows us to use newer feature (E.g. Static anonymous functions: https://github.com/dotnet/wpf/pull/4717#discussion_r656966634)

## Customer Impact
None

## Regression
No

## Testing
I tested by doing a build locally.

## Risk
There shouldn't be any risk, C# 9 is already the default version for .Net 6.

### DIFF
--- a/eng/WpfArcadeSdk/Sdk/Sdk.props
+++ b/eng/WpfArcadeSdk/Sdk/Sdk.props
@@ -14,7 +14,7 @@
 
   <PropertyGroup>
     <GenXmlStringTable>$(WpfArcadeSdkToolsDir)GenXmlStringTable.pl</GenXmlStringTable>
-    <LangVersion Condition="'$(LangVersion)'==''">8.0</LangVersion>
+    <LangVersion Condition="'$(LangVersion)'==''">9.0</LangVersion>
     <CLSCompliant Condition="'$(CLSCompliant)'==''">true</CLSCompliant>
     <IncludeDllSafeSearchPathAttribute Condition="'$(IncludeDllSafeSearchPathAttribute )'==''">true</IncludeDllSafeSearchPathAttribute>
 


RussKie:
Why not `preview`?

vatsan-madhavan:
Has WPF adopted C# 9's module initializer support yet for PresentationCore, and stopped using ILDAsm/ILAsm round-tripping ? If not, then do that first before doing this here.

ThomasGoulet73:
@RussKie, I didn't update to `preview` or `latest` because they might change depending on the .Net version and I assumed the version was pinned in the first place because of this. Doing this might require a bigger discussion than this PR requires, which just changes the pinned version to be the default version of .Net 6. I didn't want to extend the already "slow" PR acceptance delay.

@vatsan-madhavan As far as I know, WPF didn't migrate to C# 9 module initializer and the issue about this is sill opened (#3885). I'm not sure how this would block this PR though. Maybe you were talking about @RussKie's comment ?

vatsan-madhavan:
ILDasm-ILAsm roundtripping, from what I'd heard at that time, doesn't always keep pace with language improvements. I chose to fix the `LangVersion` property to a known constant that was known to work well for PresentationCore - instead of `preview` or `latest` - so that ILDasm/ILAsm roundtripping of PresentationCore would operate on a well-understood set of language constructs. I especially didn't want PresentationCore to end up in a situation in which the roundtripping seemed to happen without problems (i.e., *sans* build/CI failures), but introduced subtle errors. 

If this situation has improved - i.e., if ILDasm/ILAsm round-tripping has become more robust over the past ~year, then changing `LangVersion` to `preview` or `latest` ought to be ok. You can also do things more piecemeal - like increasing `LangVersion` globally and then setting it to 8 for PresentationCore but that sounds rather kludgy. 

That said, #3885 is not a hard problem; I view it as an ~1 day effort item. 

/cc @rladuca, @danmoseley

RussKie:
We need to update to C#10 for a number of reasons.

1. In .NET 6 GA (or really from Preview 6 and on) targeting net6.0 in an application will also target C# 10 by default. That will enable all C# 10 features including global usings, file scoped namespaces, etc...
2. To enable global usings support in WPF (https://github.com/dotnet/wpf/issues/4774)

lindexi:
See https://github.com/dotnet/wpf/pull/4889

ryalanms:
@vatsan-madhavan and @SamBent:  I've replaced the IL injection with System.Runtime.CompilerServices.ModuleInitializerAttribute, but diff'ing the IL from each shows significant differences beyond the module initializer.  My expectation was that the IL injection would only modify what is necessary to simulate the module initializer, yet there are all sorts of minor changes unrelated to the initializer in the IL. 

Anyway, the module initializer without IL injection is working as expected: 

>>	PresentationCore.dll!ModuleInitializer.Initialize() Line 21	C#
 	PresentationCore.dll!<Module>.<Module>()	Unknown
 	[Native to Managed Transition]

rladuca:
> @vatsan-madhavan and @SamBent: I've replaced the IL injection with System.Runtime.CompilerServices.ModuleInitializerAttribute, but diff'ing the IL from each shows significant differences beyond the module initializer. My expectation was that the IL injection would only modify what is necessary to simulate the module initializer, yet there are all sorts of minor changes unrelated to the initializer in the IL.
> 
> Anyway, the module initializer without IL injection is working as expected:
> 
> > > PresentationCore.dll!ModuleInitializer.Initialize() Line 21	C#
> > > PresentationCore.dll!.()	Unknown
> > > [Native to Managed Transition]

I think what might be happening is that we were round-tripping IL.  The tasks would disassemble the entire `PresentationCore` binary, inject the initialize, then reassemble it.  So there are going to be many small round-tripping artifacts.  I'm not sure of the differences you are seeing, but I would imagine these issues would account for them.  With your, much cleaner, method you should now see "pristine" IL right from the compiler.

ryalanms (approved):

ryalanms:
Thanks, everyone, for the guidance.

PR #4754, opened 2021-06-25
Avoid POINT allocations

stephentoub:
## Description

There are a bunch of copies around the tree of POINT interop types: some are structs (as they should be), some are classes.  The latter are resulting in tons of allocations.  This standardizes on using structs for all of them.

In a simple WPF app just moving my mouse around for a few seconds, these are the resulting top allocations:
image
With this PR, the highlighted line goes away.

## Customer Impact

Less allocation, less GC, less pause time.

## Regression

No

## Testing

CI, plus validation that the allocations went away in the example scenarios I was looking at.

## Risk

There is some risk here.  The change itself is sound. However, the interop definitions throughout the repo are in a bit of disarray, with many different copies of the same interop types and functions, all defined subtly differently.  I believe I caught all of the relevant paths interacting with these types, but it's definitely possible something could have slipped through.

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/InterOp/HwndMouseInputProvider.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/InterOp/HwndMouseInputProvider.cs
@@ -152,15 +152,15 @@ namespace System.Windows.Interop
             // WORKAROUND for the fact that WM_MOUSELEAVE not sent when capture changes
             if (success && !_active)
             {
-                NativeMethods.POINT ptCursor = new NativeMethods.POINT();
+                NativeMethods.POINT ptCursor = default;
 
-                success = UnsafeNativeMethods.TryGetCursorPos(ptCursor);
+                success = UnsafeNativeMethods.TryGetCursorPos(ref ptCursor);
 
                 if(success)
                 {
                     try
                     {
-                        SafeNativeMethods.ScreenToClient(new HandleRef(this, _source.Value.CriticalHandle), ptCursor);
+                        SafeNativeMethods.ScreenToClient(new HandleRef(this, _source.Value.CriticalHandle), ref ptCursor);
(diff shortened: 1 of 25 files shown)


lindexi:
Can we use the same `Point` struct type in WPF?

stephentoub:
> Can we use the same Point struct type in WPF?

Do you mean use it for the interop? No, it has a different layout/size,  with double rather than int fields.

lindexi:
@stephentoub 

Such as:

https://github.com/dotnet/wpf/blob/192a38ece06826f2a0dd43bb83180490f8fa5048/src/Microsoft.DotNet.Wpf/src/Shared/MS/Win32/UnsafeNativeMethodsTextServices.cs#L750-L766

https://github.com/dotnet/wpf/blob/192a38ece06826f2a0dd43bb83180490f8fa5048/src/Microsoft.DotNet.Wpf/src/UIAutomation/UIAutomationClient/MS/Win32/UnsafeNativeMethods.cs#L159-L169

https://github.com/dotnet/wpf/blob/192a38ece06826f2a0dd43bb83180490f8fa5048/src/Microsoft.DotNet.Wpf/src/UIAutomation/UIAutomationClientSideProviders/MS/Win32/UnsafeNativeMethods.cs#L300-L310

And the POINT in `UIAutomation/UIAutomationClient/MS/Win32/UnsafeNativeMethods.cs` and `src/Microsoft.DotNet.Wpf/src/UIAutomation/UIAutomationClientSideProviders/MS/Win32/UnsafeNativeMethods.cs` do not add StructLayout attribute.

stephentoub:
Thanks. This is what I meant by "There are a bunch of copies around the tree of POINT interop types"; in fact, tons of interop is duplicated around the tree.  It could certainly be cleaned up / deduped / etc., but I'm not going to do that here.

RussKie (approved):

ryalanms (approved):

PR #4766, opened 2021-06-28
Remove lots of unnecessary string allocations

stephentoub:
## Description

There are many places we're allocating temporary strings, e.g. from Substring, where we can now use spans.

## Customer Impact

Less allocation/copying.

## Regression

No

## Testing

Just CI

## Risk

Minimal.  Changes are all local and can be easily reviewed for correctness.

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/PresentationCore/MS/internal/FontCache/FontCacheUtil.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationCore/MS/internal/FontCache/FontCacheUtil.cs
@@ -661,7 +661,7 @@ namespace MS.Internal.FontCache
             {
                 // No fragment separator. The entire string is a family name so convert to uppercase
                 // and add a fragment separator at the beginning.
-                return "#" + fontFamilyReference.Substring(startIndex, length).ToUpperInvariant();
+                return string.Concat("#", fontFamilyReference.AsSpan(startIndex, length)).ToUpperInvariant();
             }
             else if (fragmentIndex + 1 == startIndex + length)
             {
--- a/src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Input/InputScopeConverter.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Input/InputScopeConverter.cs
@@ -109,14 +109,18 @@ namespace System.Windows.Input
 
             if (null != stringSource)
             {
-                stringSource = stringSource.Trim();
+                ReadOnlySpan<char> spanSource = stringSource;
+                spanSource = spanSource.Trim();
 
-                if (-1 != stringSource.LastIndexOf('.'))
-                    stringSource = stringSource.Substring(stringSource.LastIndexOf('.')+1);
-                    
-                if (!stringSource.Equals(String.Empty))
+                int periodPos = spanSource.LastIndexOf('.');
+                if (periodPos != -1)
                 {
-                    sn = (InputScopeNameValue)Enum.Parse(typeof(InputScopeNameValue), stringSource);
+                    spanSource = spanSource.Slice(periodPos + 1);
+                }
+                
+                if (!spanSource.IsEmpty)
+                {
+                    sn = Enum.Parse<InputScopeNameValue>(spanSource);
                 }
             }
             
--- a/src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Media/Animation/KeyTimeConverter.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Media/Animation/KeyTimeConverter.cs
@@ -195,7 +195,7 @@ namespace System.Windows
                                     keyTime.Percent * 100.0,
                                     destinationType);
 
-                            return returnValue + _percentCharacter[0].ToString();
+                            return string.Concat(returnValue, (ReadOnlySpan<char>)_percentCharacter);
 
                         case KeyTimeType.TimeSpan:
 
--- a/src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Media/Animation/KeyTimeConverter.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Media/Animation/KeyTimeConverter.cs
@@ -19,12 +19,6 @@ namespace System.Windows
     /// </summary>
     public class KeyTimeConverter : TypeConverter
     {
-        #region Data
-
-        private static char[] _percentCharacter = new char[] { '%' };
-
-        #endregion
-
         /// <summary>
         /// Returns whether or not this class can convert from a given type
         /// to an instance of a KeyTime.
@@ -88,9 +82,9 @@ namespace System.Windows
                 {
                     return KeyTime.Paced;
                 }
-                else if (stringValue[stringValue.Length - 1] == _percentCharacter[0])
+                else if (stringValue[stringValue.Length - 1] == '%')
                 {
-                    stringValue = stringValue.TrimEnd(_percentCharacter);
+                    stringValue = stringValue.TrimEnd('%');
 
                     double doubleValue = (double)TypeDescriptor.GetConverter(
                         typeof(double)).ConvertFrom(
@@ -195,7 +189,7 @@ namespace System.Windows
                                     keyTime.Percent * 100.0,
                                     destinationType);
 
-                            return string.Concat(returnValue, (ReadOnlySpan<char>)_percentCharacter);
+                            return string.Concat(returnValue, (ReadOnlySpan<char>)stackalloc char[] { '%' });
 
                         case KeyTimeType.TimeSpan:
 
--- a/src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Media/Imaging/PropVariant.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Media/Imaging/PropVariant.cs
@@ -364,7 +364,7 @@ namespace System.Windows.Media.Imaging
                 else if (value is char)
                 {
                     varType = (ushort)VarEnum.VT_LPSTR;
-                    pszVal = Marshal.StringToCoTaskMemAnsi(new String(new char[] { (char)value }));
+                    pszVal = Marshal.StringToCoTaskMemAnsi(new String(stackalloc char[] { (char)value }));
                 }
                 else if (type == typeof(short))
                 {
--- a/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Controls/VirtualizingStackPanel.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Controls/VirtualizingStackPanel.cs
@@ -12217,11 +12217,11 @@ namespace System.Windows.Controls
                         sb.Append("/");
                     }
 
-                    string name = t.ToString();
+                    ReadOnlySpan<char> name = t.ToString();
                     isWPFControl = name.StartsWith("System.Windows.Controls.");
                     if (isWPFControl)
                     {
-                        name = name.Substring(24);  // 24 == length of "s.w.c."
+                        name = name.Slice(24); // 24 == length of "s.w.c."
                     }
 
                     sb.Append(name);
--- a/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Documents/FixedSchema.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Documents/FixedSchema.cs
@@ -441,15 +441,14 @@ namespace System.Windows.Documents
                 attrName.Equals("Fill", StringComparison.Ordinal) ||
                 attrName.Equals("Stroke", StringComparison.Ordinal))
             {
-                attrValue = attrValue.Trim();
-                if (attrValue.StartsWith(_contextColor, StringComparison.Ordinal))
+                ReadOnlySpan<char> attrValueSpan = attrValue.AsSpan().Trim();
+                if (attrValueSpan.StartsWith(_contextColor, StringComparison.Ordinal))
                 {
-                    attrValue = attrValue.Substring(_contextColor.Length);
-                    attrValue = attrValue.Trim();
-                    string[] tokens = attrValue.Split(new char[] { ' ' });
-                    if (tokens.GetLength(0) >= 1)
+                    attrValueSpan = attrValueSpan.Slice(_contextColor.Length).Trim();
+                    int spacePos = attrValueSpan.IndexOf(' ');
+                    if (spacePos >= 0)
                     {
-                        return new string[] { tokens[0] };
+                        return new string[] { attrValueSpan.Slice(0, spacePos).ToString() };
                     }
                 }
             }
--- a/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Documents/XamlToRtfWriter.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Documents/XamlToRtfWriter.cs
@@ -2737,14 +2737,11 @@ namespace System.Windows.Documents
                     xamlToRtfError = XamlToRtfError.InvalidFormat;
                     if (name.Length >= 5)
                     {
-                        string num = name.Substring(3, name.Length - 4);
                         int i = 0;
-                        bool ret = Converters.HexStringToInt(num, ref i);
+                        bool ret = Converters.HexStringToInt(name.AsSpan(3, name.Length - 4), ref i);
                         if (i >= 0 && i <= 0xFFFF)
                         {
-                            char[] ac = new char[1];
-                            ac[0] = (char)i;
-                            string s = new string(ac);
+                            string s = new string(stackalloc char[1] { (char)i });
                             return ((IXamlContentHandler)this).Characters(s);
                         }
                     }
@@ -2753,14 +2750,12 @@ namespace System.Windows.Documents
                 {
                     if (name.Length >= 4)
                     {
-                        string num = name.Substring(2, name.Length - 3);
+                        ReadOnlySpan<char> num = name.Substring(2, name.Length - 3);
                         int i = 0;
                         bool ret = Converters.StringToInt(num, ref i);
                         if (i >= 0 && i <= 0xFFFF)
                         {
-                            char[] ac = new char[1];
-                            ac[0] = (char)i;
-                            string s = new string(ac);
+                            string s = new string(stackalloc char[1] { (char)i });
                             return ((IXamlContentHandler)this).Characters(s);
                         }
                     }
@@ -3547,7 +3542,7 @@ namespace System.Windows.Documents
                 return true;
             }
 
-            internal static bool ConvertToFontSize(ConverterState converterState, string s, ref double d)
+            internal static bool ConvertToFontSize(ConverterState converterState, ReadOnlySpan<char> s, ref double d)
             {
                 if (s.Length == 0)
                 {
@@ -3561,11 +3556,11 @@ namespace System.Windows.Documents
                     n--;
                 }
 
-                string units = null;
+                ReadOnlySpan<char> units = default;
                 if (n < s.Length - 1)
                 {
-                    units = s.Substring(n + 1);
-                    s = s.Substring(0, n + 1);
+                    units = s.Slice(n + 1);
+                    s = s.Slice(0, n + 1);
                 }
 
                 // Now convert number part
@@ -3574,7 +3569,7 @@ namespace System.Windows.Documents
                 if (ret)
                 {
                     // No units mean pixels
-                    if (units == null || units.Length == 0)
+                    if (units.IsEmpty)
                     {
                         d = Converters.PxToPt(d);
                     }
@@ -3813,7 +3808,7 @@ namespace System.Windows.Documents
                 return ret;
             }
 
-            internal static bool ConvertToThickness(ConverterState converterState, string thickness, ref XamlThickness xthickness)
+            internal static bool ConvertToThickness(ConverterState converterState, ReadOnlySpan<char> thickness, ref XamlThickness xthickness)
             {
                 int numints = 0;
                 int s = 0;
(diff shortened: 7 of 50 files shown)


src/Microsoft.DotNet.Wpf/src/PresentationCore/MS/internal/FontCache/FontCacheUtil.cs
```diff
@@ -661,7 +661,7 @@ private static string NormalizeFontFamilyReference(string fontFamilyReference, i
             {
                 // No fragment separator. The entire string is a family name so convert to uppercase
                 // and add a fragment separator at the beginning.
-                return "#" + fontFamilyReference.Substring(startIndex, length).ToUpperInvariant();
+                return string.Concat("#", fontFamilyReference.AsSpan(startIndex, length)).ToUpperInvariant();
```
ThomasGoulet73, line 664: Could you call .ToUpperInvariant() on the span instead of on the string ?

It would save an allocation of string from .ToUpperInvariant(). I don't think we need to call .ToUpperInvariant() on "#", since it's a symbol.

src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Input/InputScopeConverter.cs
```diff
@@ -109,14 +109,15 @@ public override object ConvertFrom(ITypeDescriptorContext context, CultureInfo c
...
             if (null != stringSource)
             {
-                stringSource = stringSource.Trim();
+                ReadOnlySpan<char> spanSource = stringSource;
+                spanSource = spanSource.Trim();
 
-                if (-1 != stringSource.LastIndexOf('.'))
-                    stringSource = stringSource.Substring(stringSource.LastIndexOf('.')+1);
+                if (-1 != spanSource.LastIndexOf('.'))
+                    spanSource = spanSource.Slice(stringSource.LastIndexOf('.')+1);
```
ThomasGoulet73, line 116: Replace stringSource.LastIndexOf('.') with spanSource.LastIndexOf('.') because at this point, if there are whitespace in stringSource, stringSource != spanSource, because of spanSource.Trim(). You could also store the result of spanSource.LastIndexOf('.') in a local variable if you want.

src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Media/Animation/KeyTimeConverter.cs
```diff
@@ -195,7 +195,7 @@ public class KeyTimeConverter : TypeConverter
                                     keyTime.Percent * 100.0,
                                     destinationType);
 
-                            return returnValue + _percentCharacter[0].ToString();
+                            return string.Concat(returnValue, (ReadOnlySpan<char>)_percentCharacter);
```
ThomasGoulet73, line 198: Would it be faster to change _percentCharacter to be a single char instead of an array of char ? I think the only reason that _percentCharacter was an array was because it's used in TrimEnd(System.Char) which was introduced in .Net 5.

src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Media/Imaging/PropVariant.cs
```diff
@@ -364,7 +364,7 @@ internal void Init(object value)
                 else if (value is char)
                 {
                     varType = (ushort)VarEnum.VT_LPSTR;
-                    pszVal = Marshal.StringToCoTaskMemAnsi(new String(new char[] { (char)value }));
+                    pszVal = Marshal.StringToCoTaskMemAnsi(new String(stackalloc char[] { (char)value }));
```
ThomasGoulet73, line 367: Does ((char)value) already have this kind of optimization ? If yes, I think we should use it, if not I think its fine to use stackalloc.

src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Documents/glyphs.cs
```diff
@@ -406,14 +406,14 @@ private static bool IsEmpty(string s)
         /// <param name="glyphIndex"></param>
         /// <returns>true if glyph index is present, false if glyph index is not present.</returns>
         private bool ReadGlyphIndex(
-            string      valueSpec,
+            ReadOnlySpan<char> valueSpec,
```
ThomasGoulet73, line 409: I think we should keep the previous format (Argument name in a column to keep the same format in the file).

src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Markup/Primitives/ExtensionSimplifierMarkupObject.cs
```diff
@@ -355,16 +355,16 @@ private string ConvertMarkupItemToString(MarkupObject item)
                     Debug.Assert(!propertyWritten, "An argument was returned after a property was set. All arguments must be returned first and in order");
                 }
                 
-                string value = property.StringValue;
+                ReadOnlySpan<char> value = property.StringValue;
                 
-                if (value != null && value.Length > 0)
+                if (value.Length > 0)
```
ThomasGoulet73, line 360: In some other files, you used `span.IsEmpty` by there you used `value.Length > 0`. I think we should choose one or the other (I personally prefer IsEmpty, if there are no perf implications).

src/Microsoft.DotNet.Wpf/src/ReachFramework/Packaging/XpsFixedPageReaderWriter.cs
```diff
@@ -1866,26 +1866,24 @@ Uri         resourceUri
             Uri imageUri
             )
         {
-            //Extract file extension
+            //Extract file extension without '.'
             String path = imageUri.OriginalString;
-            String extension = Path.GetExtension(path).ToLower(CultureInfo.InvariantCulture);
-            //remove .
-            extension = extension.Substring(1);
+            ReadOnlySpan<char> extension = Path.GetExtension(path).ToLower(CultureInfo.InvariantCulture).AsSpan(1);
```
ThomasGoulet73, line 1871: Could you call `.ToLower(CultureInfo.InvariantCulture)` on the span instead of the string to avoid allocating a new string in ToLower.

src/Microsoft.DotNet.Wpf/src/PresentationCore/MS/internal/FontCache/FontCacheUtil.cs
stephentoub: There isn't a `ToUpperInvariant` that produces a string from a span.  There is one that operates in-place, but that requires a `Span<char>` not a `ReadOnlySpan<char>`.

The only downside here is, if upper-casing is required, we allocate two additional bytes unnecessarily.  In contrast, the original would end up allocating an entire additional string.

src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Input/InputScopeConverter.cs
stephentoub: Thanks, that was an oversight.  Fixed.

src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Media/Animation/KeyTimeConverter.cs
stephentoub: Not really faster, but it'd be a bit cleaner.

src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Media/Imaging/PropVariant.cs
stephentoub: I switched it to use `char.ToString((char)value)`.

src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Documents/glyphs.cs
stephentoub: Done

src/Microsoft.DotNet.Wpf/src/ReachFramework/Packaging/XpsFixedPageReaderWriter.cs
stephentoub: No, an in-place ToLower requires a mutable span.

src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Markup/Primitives/ExtensionSimplifierMarkupObject.cs
stephentoub: I'd just kept the original code here, but I don't mind changing it to IsEmpty.

src/Microsoft.DotNet.Wpf/src/PresentationCore/MS/internal/FontCache/FontCacheUtil.cs
ThomasGoulet73: Sorry, I checked the API doc a bit to quick. I didn't notice that it required a writable span as an argument. Thanks

src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Documents/FixedSchema.cs
```diff
@@ -441,15 +441,14 @@ public override string[] ExtractUriFromAttr(string attrName, string attrValue)
...
-                    attrValue = attrValue.Substring(_contextColor.Length);
-                    attrValue = attrValue.Trim();
-                    string[] tokens = attrValue.Split(new char[] { ' ' });
-                    if (tokens.GetLength(0) >= 1)
+                    attrValueSpan = attrValueSpan.Slice(_contextColor.Length).Trim();
+                    int spacePos = attrValueSpan.IndexOf(' ');
+                    if (spacePos >= 0)
                     {
-                        return new string[] { tokens[0] };
+                        return new string[] { attrValueSpan.Slice(0, spacePos).ToString() };
```
lindexi, line 451: Seems it changed the behavior

stephentoub: How so?

src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Controls/VirtualizingStackPanel.cs
```diff
@@ -12217,11 +12217,11 @@ private static string DisplayType(object o)
                         sb.Append("/");
                     }
 
-                    string name = t.ToString();
+                    ReadOnlySpan<char> name = t.ToString();
                     isWPFControl = name.StartsWith("System.Windows.Controls.");
                     if (isWPFControl)
                     {
-                        name = name.Substring(24);  // 24 == length of "s.w.c."
+                        sb.Append(name.Slice(24)); // 24 == length of "s.w.c."
```
halgab, line 12224: Isn't this a change in behaviour?

stephentoub: How so?

stephentoub: Oh, yes, typo. Will fix.

src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Documents/XamlToRtfWriter.cs
```diff
@@ -2753,14 +2750,12 @@ XamlToRtfError IXamlContentHandler.SkippedEntity(string name)
                 {
                     if (name.Length >= 4)
                     {
-                        string num = name.Substring(2, name.Length - 3);
+                        ReadOnlySpan<char> num = name.Substring(2, name.Length - 3);
```
halgab, line 2753: Did you mean `ReadOnlySpan<char> num = name.AsSpan(2, name.Length - 3);`?

stephentoub: I did, yes, thanks. That will avoid another string.

src/Microsoft.DotNet.Wpf/src/System.Xaml/System/Xaml/MS/Impl/KnownStrings.cs
```diff
@@ -62,6 +62,11 @@ public static bool Eq(string a, string b)
             return string.Equals(a, b, StringComparison.Ordinal);
         }
 
+        public static bool Eq(ReadOnlySpan<char> a, ReadOnlySpan<char> b)
```
lindexi, line 65: Should we add the `in` key word?

```suggestion
        public static bool Eq(in ReadOnlySpan<char> a, in ReadOnlySpan<char> b)
```

src/Microsoft.DotNet.Wpf/src/ReachFramework/Packaging/XpsFixedPageReaderWriter.cs
lindexi: @stephentoub Is it faster than before when we do not use ToLower and check string Equals with `StringComparison.OrdinalIgnoreCase`?

src/Microsoft.DotNet.Wpf/src/System.Xaml/System/Xaml/MS/Impl/KnownStrings.cs
stephentoub: No. Why?

lindexi: @stephentoub  Sorry. I read some documents and found that it is not necessary to add the `in` key word.

src/Microsoft.DotNet.Wpf/src/ReachFramework/Packaging/XpsFixedPageReaderWriter.cs
stephentoub: Sorry,  I don't understand the question.  Can you elaborate?

src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Documents/FixedSchema.cs
lindexi: @stephentoub Thank you. I wrote some test code, and found that the behavior has not changed.

src/Microsoft.DotNet.Wpf/src/ReachFramework/Packaging/XpsFixedPageReaderWriter.cs
lindexi: @stephentoub  How about this code:

```
            //Extract file extension without '.'
            ReadOnlySpan<char> path = imageUri.OriginalString.AsSpan();
            ReadOnlySpan<char> extension = Path.GetExtension(path).Slice(1);
            ContentType contentType;
            if (extension.Equals(XpsS0Markup.JpgExtension, StringComparison.OrdinalIgnoreCase))
            {
                contentType = XpsS0Markup.JpgContentType;
            }

            // And so on

```

Is it faster?

stephentoub: Ah, I understand the question now. The answer is, it depends.  It will save an allocation / copy, but it will need to do the casing operations for each comparison, and depending on how many of the comparisons it had to do, at some point that could become more expensive.  It would need to be measured, and if there's a cross-over point, weighed based on known use cases.

lindexi: @stephentoub Thank you. I wrote the test demo code and I found an improvement in performance.

|                                 Method |      Mean |    Error |    StdDev |  Gen 0 | Gen 1 | Gen 2 | Allocated |
|--------------------------------------- |----------:|---------:|----------:|-------:|------:|------:|----------:|
|           LookupContentTypeForImageUri | 114.17 ns | 3.096 ns |  9.130 ns | 0.0153 |     - |     - |      64 B |
| LookupContentTypeForImageUriIgnoreCase |  40.84 ns | 0.975 ns |  2.874 ns |      - |     - |     - |         - |
|           LookupContentTypeForImageUri | 115.00 ns | 3.858 ns | 11.375 ns | 0.0153 |     - |     - |      64 B |
| LookupContentTypeForImageUriIgnoreCase |  55.19 ns | 1.913 ns |  5.642 ns |      - |     - |     - |         - |
|           LookupContentTypeForImageUri |  84.97 ns | 1.773 ns |  3.198 ns | 0.0153 |     - |     - |      64 B |
| LookupContentTypeForImageUriIgnoreCase |  40.82 ns | 0.869 ns |  1.099 ns |      - |     - |     - |         - |
|           LookupContentTypeForImageUri |  91.29 ns | 1.693 ns |  1.322 ns | 0.0153 |     - |     - |      64 B |
| LookupContentTypeForImageUriIgnoreCase |  47.45 ns | 0.961 ns |  0.852 ns |      - |     - |     - |         - |
|           LookupContentTypeForImageUri | 103.35 ns | 1.876 ns |  2.234 ns | 0.0153 |     - |     - |      64 B |
| LookupContentTypeForImageUriIgnoreCase |  58.83 ns | 1.231 ns |  1.643 ns |      - |     - |     - |         - |
|           LookupContentTypeForImageUri |  52.88 ns | 0.973 ns |  0.911 ns | 0.0076 |     - |     - |      32 B |
| LookupContentTypeForImageUriIgnoreCase |  31.78 ns | 0.701 ns |  1.264 ns |      - |     - |     - |         - |
|           LookupContentTypeForImageUri |  65.51 ns | 1.325 ns |  1.361 ns | 0.0076 |     - |     - |      32 B |
| LookupContentTypeForImageUriIgnoreCase |  44.33 ns | 0.949 ns |  0.841 ns |      - |     - |     - |         - |
|           LookupContentTypeForImageUri |  66.49 ns | 1.383 ns |  2.386 ns | 0.0076 |     - |     - |      32 B |
| LookupContentTypeForImageUriIgnoreCase |  40.33 ns | 0.688 ns |  0.643 ns |      - |     - |     - |         - |
|           LookupContentTypeForImageUri |  76.48 ns | 1.560 ns |  1.669 ns | 0.0076 |     - |     - |      32 B |
| LookupContentTypeForImageUriIgnoreCase |  47.74 ns | 1.015 ns |  1.640 ns |      - |     - |     - |         - |
|           LookupContentTypeForImageUri |  83.05 ns | 1.536 ns |  1.769 ns | 0.0076 |     - |     - |      32 B |
| LookupContentTypeForImageUriIgnoreCase |  58.10 ns | 1.177 ns |  1.156 ns |      - |     - |     - |         - |


This is the benchmark code:


```csharp
    [MemoryDiagnoser]
    public class Program
    {
        {
            BenchmarkRunner.Run<Program>();
        }

        [Benchmark]
        [ArgumentsSource(nameof(GetArguments))]
        public object LookupContentTypeForImageUri(Uri imageUri)
        {
            //Extract file extension without '.'
            String path = imageUri.OriginalString;
            ReadOnlySpan<char> extension = Path.GetExtension(path).ToLower(CultureInfo.InvariantCulture).AsSpan(1);
            ContentType contentType;
            if (extension.Equals(XpsS0Markup.JpgExtension, StringComparison.Ordinal))
            {
                contentType = XpsS0Markup.JpgContentType;
            }
            else if (extension.Equals(XpsS0Markup.PngExtension, StringComparison.Ordinal))
            {
                contentType = XpsS0Markup.PngContentType;
            }
            else if (extension.Equals(XpsS0Markup.TifExtension, StringComparison.Ordinal))
            {
                contentType = XpsS0Markup.TifContentType;
            }
            else if (extension.Equals(XpsS0Markup.WdpExtension, StringComparison.Ordinal))
            {
                contentType = XpsS0Markup.WdpContentType;
            }
            else
            {
                //default to PNG
                contentType = XpsS0Markup.PngContentType;
            }
            return contentType;
        }

        [Benchmark]
        [ArgumentsSource(nameof(GetArguments))]
        public object LookupContentTypeForImageUriIgnoreCase(Uri imageUri)
        {
            //Extract file extension without '.'
            ReadOnlySpan<char> path = imageUri.OriginalString.AsSpan();
            ReadOnlySpan<char> extension = Path.GetExtension(path).Slice(1);
            ContentType contentType;
            if (extension.Equals(XpsS0Markup.JpgExtension, StringComparison.OrdinalIgnoreCase))
            {
                contentType = XpsS0Markup.JpgContentType;
            }
            else if (extension.Equals(XpsS0Markup.PngExtension, StringComparison.OrdinalIgnoreCase))
            {
                contentType = XpsS0Markup.PngContentType;
            }
            else if (extension.Equals(XpsS0Markup.TifExtension, StringComparison.OrdinalIgnoreCase))
            {
                contentType = XpsS0Markup.TifContentType;
            }
            else if (extension.Equals(XpsS0Markup.WdpExtension, StringComparison.OrdinalIgnoreCase))
            {
                contentType = XpsS0Markup.WdpContentType;
            }
            else
            {
                //default to PNG
                contentType = XpsS0Markup.PngContentType;
            }
            return contentType;
        }

        public IEnumerable<Uri> GetArguments()
        {
            yield return  new Uri(@"C:\foo.jpg") ;
            yield return  new Uri(@"C:\foo.JPG") ;
            yield return  new Uri(@"C:\foo.png") ;
            yield return  new Uri(@"C:\foo.PNG") ;
            yield return  new Uri(@"C:\foo.tif") ;
            yield return  new Uri(@"C:\foo.TIF") ;
            yield return  new Uri(@"C:\foo.wdp") ;
            yield return  new Uri(@"C:\foo.WDP") ;
            yield return  new Uri(@"C:\foo.none");
            yield return  new Uri(@"C:\foo.NONE");
        }
    }
```

You can find all the code in github

```
git init
git remote add origin https://github.com/lindexi/lindexi_gd.git
git pull origin 983c668ea7cd308a510c043b5d2e5de74da8efe7
cd ColalejayJearbearlerlekere
```

RussKie (approved):

ryalanms (approved):

PR #4771, opened 2021-06-29
Stop allocating new delegates in WriterDelegate.Dispose

stephentoub:
## Description

Every call to Dispose is allocating one or two delegates.  They're not unique to this instance; we can let the compiler cache them for us.

## Customer Impact

Less allocation, less GC impact.

## Regression

No

## Testing

Just CI

## Risk

Minimal.  It's just changing one form of delegate creation to another.

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/System.Xaml/System/Xaml/Primitives/WriterDelegate.cs
+++ b/src/Microsoft.DotNet.Wpf/src/System.Xaml/System/Xaml/Primitives/WriterDelegate.cs
@@ -72,10 +72,11 @@ namespace System.Xaml
                 if (disposing && !IsDisposed)
                 {
                     _addDelegate(XamlNodeType.None, XamlNode.InternalNodeType.EndOfStream);
-                    _addDelegate = new XamlNodeAddDelegate(ThrowBecauseWriterIsClosed);
-                    _addLineInfoDelegate = (_addLineInfoDelegate != null)
-                                ? new XamlLineInfoAddDelegate(ThrowBecauseWriterIsClosed2)
-                                : null;
+                    _addDelegate = delegate { throw new XamlException(SR.Get(SRID.WriterIsClosed)); };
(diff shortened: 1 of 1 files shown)


src/Microsoft.DotNet.Wpf/src/System.Xaml/System/Xaml/Primitives/WriterDelegate.cs
```diff
@@ -72,10 +72,11 @@ protected override void Dispose(bool disposing)
                 if (disposing && !IsDisposed)
                 {
                     _addDelegate(XamlNodeType.None, XamlNode.InternalNodeType.EndOfStream);
-                    _addDelegate = new XamlNodeAddDelegate(ThrowBecauseWriterIsClosed);
-                    _addLineInfoDelegate = (_addLineInfoDelegate != null)
-                                ? new XamlLineInfoAddDelegate(ThrowBecauseWriterIsClosed2)
-                                : null;
+                    _addDelegate = delegate { throw new XamlException(SR.Get(SRID.WriterIsClosed)); };
```
ThomasGoulet73, line 75: Do we even need to set `_addDelegate` and `_addLineInfoDelegate` ? ThrowIsDisposed should already be called before the use of `_addDelegate` or `_addLineInfoDelegate`.

ThomasGoulet73: I still think this PR is valid because it's faster than the existing code, which was already executed even if the new delegate is not invoked. A follow-up PR could be open to remove this "dead code" with more thorough testing to make sure that the delegates are never executed after dispose. I only said this over a quick glance over the code but they might be invoked.

stephentoub: > I still think this PR is valid

It is :smile:

ryalanms (approved):

PR #4777, opened 2021-06-30
Fix issue causing constant rebuilds

wjk:
## Description

In my experience, MSBuild marked WpfGfx.dll and all of its dependencies for rebuild every time because various `<WriteLinesToFile>` calls lacked `WriteOnlyWhenDifferent=true`. This made my inner-loop significantly slower.

## Customer Impact

None; this is a build-time fix only.


RussKie (approved):

ryalanms:
Thanks, @wjk.

PR #4779, opened 2021-07-01
Import Windows Forms related props and targets into Windows Desktop

RussKie:
## Proposed Changes:

* Establish a mechanism to import Windows Forms related functionality into Windows Desktop SDK via the Windows Forms transport package.
* Add a basic validation of the content of the shipping package in .\eng\WindowsFormsImports.targets.

Enables https://github.com/dotnet/winforms/pull/5183

## Screenshots 

image

## Test methodology 

- Manually locally modifying the installed SDK
- Building winforms -> wpf -> windowsdesktop

### DIFF
--- a/eng/WindowsFormsImports.targets
+++ b/eng/WindowsFormsImports.targets
@@ -1,0 +1,51 @@
+<Project>
+
+  <PropertyGroup>
+    <_WindowsFormsNuGetPath>$(PkgMicrosoft_Private_Winforms)</_WindowsFormsNuGetPath>
+  </PropertyGroup>
+
+  <!--
+    ============================================================
+                      _ValidateWindowsFormsPackagingContent
+    Validates the content of Microsoft.Private.Winforms NuGet package
+    to ensure we correctly import and reference props and targets.
+    ============================================================
+   -->
+  <Target Name="_ValidateWindowsFormsPackagingContent">
+    <PropertyGroup>
+      <_WindowsFormsContentPath>$(_WindowsFormsNuGetPath)\sdk\dotnet-wpf\*</_WindowsFormsContentPath>
+      <_WindowsFormsRequiredFileName>Microsoft.NET.SDk.WindowsDesktop.WindowsForms</_WindowsFormsRequiredFileName>
+    </PropertyGroup>
+
+    <Error Text="Unable to resolve path to Microsoft.Private.Winforms NuGet package. Is %24(PkgMicrosoft_Private_Winforms) defined?"
+           Condition="'$(_WindowsFormsNuGetPath)' == ''"/>
+
+    <ItemGroup>
+      <!-- Enumerate all transported files -->
+      <_WindowsFormsContent Include="$(_WindowsFormsContentPath)" />
+      <!-- ...and verify Microsoft.NET.SDk.WindowsDesktop.WindowsForms.props/.targets are present -->
+      <_WindowsFormsContentFiles Include="@(_WindowsFormsContent->'%(FileName)%(Extension)')"
+          Condition=" '%(FileName)%(Extension)' == '$(_WindowsFormsRequiredFileName).props'
+                   or '%(FileName)%(Extension)' == '$(_WindowsFormsRequiredFileName).targets' "/>
+    </ItemGroup>
+
+    <!-- Fail if the required files are missing -->
+    <Error Text="Microsoft.Private.Winforms NuGet package does not contain $(_WindowsFormsRequiredFileName).props or $(_WindowsFormsRequiredFileName).targets"
+           Condition="@(_WindowsFormsContentFiles->Count()) != 2"/>
+  </Target>
+
+
+  <!--
+    ============================================================
+                      _IdentifyWindowsFormsPackageAssets
+    Add props/targets shipped by Windows Forms to the collection of files that get packaged
+    in to Microsoft.NET.Sdk.WindowsDesktop.<configuration>.<version>.nupkg.
+    ============================================================
+  -->
+  <Target Name="_IdentifyWindowsFormsPackageAssets" BeforeTargets="IdentifyPackageAssets" DependsOnTargets="_ValidateWindowsFormsPackagingContent">
+    <ItemGroup>
+      <PackagingContent Include="$(_WindowsFormsContentPath)" SubFolder="root\targets" />
+    </ItemGroup>
+  </Target>
+
+</Project>
--- a/packaging/Microsoft.NET.Sdk.WindowsDesktop/targets/Microsoft.NET.Sdk.WindowsDesktop.props
+++ b/packaging/Microsoft.NET.Sdk.WindowsDesktop/targets/Microsoft.NET.Sdk.WindowsDesktop.props
@@ -161,4 +161,11 @@
     <SupportedTargetFramework Remove="@(_UnsupportedNETCoreAppTargetFramework);@(_UnsupportedNETStandardTargetFramework);@(_UnsupportedNETFrameworkTargetFramework)" />
   </ItemGroup>
 
+  <!--
+    Import Windows Forms props.
+    These come via the Windows Forms transport package, that can be found under
+    https://github.com/dotnet/winforms/tree/main/pkg/Microsoft.Private.Winforms/sdk
+  -->
+  <Import Project="Microsoft.NET.SDk.WindowsDesktop.WindowsForms.props" />
+
 </Project>
--- a/packaging/Microsoft.NET.Sdk.WindowsDesktop/targets/Microsoft.NET.Sdk.WindowsDesktop.props
+++ b/packaging/Microsoft.NET.Sdk.WindowsDesktop/targets/Microsoft.NET.Sdk.WindowsDesktop.props
@@ -166,6 +166,6 @@
     These come via the Windows Forms transport package, that can be found under
     https://github.com/dotnet/winforms/tree/main/pkg/Microsoft.Private.Winforms/sdk
   -->
-  <Import Project="Microsoft.NET.SDk.WindowsDesktop.WindowsForms.props" />
+  <Import Project="Microsoft.NET.Sdk.WindowsDesktop.WindowsForms.props" />
 
 </Project>
--- a/packaging/Microsoft.NET.Sdk.WindowsDesktop/targets/Microsoft.NET.Sdk.WindowsDesktop.targets
+++ b/packaging/Microsoft.NET.Sdk.WindowsDesktop/targets/Microsoft.NET.Sdk.WindowsDesktop.targets
@@ -160,6 +160,6 @@
     These come via the Windows Forms transport package, that can be found under
     https://github.com/dotnet/winforms/tree/main/pkg/Microsoft.Private.Winforms/sdk
   -->
-  <Import Project="Microsoft.NET.SDk.WindowsDesktop.WindowsForms.targets" Condition="'$(UseWindowsForms)' == 'true'"/>
+  <Import Project="Microsoft.NET.Sdk.WindowsDesktop.WindowsForms.targets" Condition="'$(UseWindowsForms)' == 'true'"/>
 
 </Project>
--- a/packaging/Microsoft.NET.Sdk.WindowsDesktop/targets/Microsoft.NET.Sdk.WindowsDesktop.targets
+++ b/packaging/Microsoft.NET.Sdk.WindowsDesktop/targets/Microsoft.NET.Sdk.WindowsDesktop.targets
@@ -155,4 +155,11 @@
   <!-- Import WPF Build logic only when we don't import NETFX's WinFX targets -->
   <Import Project="Microsoft.WinFX.targets" Condition="'$(ImportFrameworkWinFXTargets)' != 'true'"/>
 
+  <!--
+    Import Windows Forms targets.
+    These come via the Windows Forms transport package, that can be found under
+    https://github.com/dotnet/winforms/tree/main/pkg/Microsoft.Private.Winforms/sdk
+  -->
+  <Import Project="Microsoft.NET.SDk.WindowsDesktop.WindowsForms.targets" Condition="'$(UseWindowsForms)' == 'true'"/>
+
 </Project>
--- a/packaging/Microsoft.NET.Sdk.WindowsDesktop/Microsoft.NET.Sdk.WindowsDesktop.ArchNeutral.csproj
+++ b/packaging/Microsoft.NET.Sdk.WindowsDesktop/Microsoft.NET.Sdk.WindowsDesktop.ArchNeutral.csproj
@@ -27,7 +27,7 @@
     <PackagingContent Include="useSharedDesignerContext.txt" SubFolder="root" />
   </ItemGroup>
 
-  <!-- Windows Forms specific -->
-  <Import Project="$(RepositoryEngineeringDir)WindowsFormsImports.targets" />
+  <!-- Windows Forms validation and packaging -->
+  <Import Project="Microsoft.NET.Sdk.WindowsDesktop.WinForms.Packaging.targets" />
 
 </Project>
--- a/packaging/Microsoft.NET.Sdk.WindowsDesktop/Microsoft.NET.Sdk.WindowsDesktop.ArchNeutral.csproj
+++ b/packaging/Microsoft.NET.Sdk.WindowsDesktop/Microsoft.NET.Sdk.WindowsDesktop.ArchNeutral.csproj
@@ -1,4 +1,5 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
+
   <PropertyGroup>
     <ProjectGuid>{440d06b8-e3de-4c0d-ad25-cd4f43d836e1}</ProjectGuid>
     <TargetFramework>net6.0</TargetFramework>
@@ -25,4 +26,8 @@
     <PackagingContent Include="targets\*" SubFolder="root\targets" />
     <PackagingContent Include="useSharedDesignerContext.txt" SubFolder="root" />
   </ItemGroup>
+
+  <!-- Windows Forms specific -->
+  <Import Project="$(RepositoryEngineeringDir)WindowsFormsImports.targets" />
+
 </Project>
--- a/packaging/Microsoft.NET.Sdk.WindowsDesktop/Microsoft.NET.Sdk.WindowsDesktop.WinForms.Packaging.targets
+++ b/packaging/Microsoft.NET.Sdk.WindowsDesktop/Microsoft.NET.Sdk.WindowsDesktop.WinForms.Packaging.targets
@@ -14,7 +14,7 @@
   <Target Name="_ValidateWindowsFormsPackagingContent">
     <PropertyGroup>
       <_WindowsFormsContentPath>$(_WindowsFormsNuGetPath)\sdk\dotnet-wpf\*</_WindowsFormsContentPath>
-      <_WindowsFormsRequiredFileName>Microsoft.NET.SDk.WindowsDesktop.WindowsForms</_WindowsFormsRequiredFileName>
+      <_WindowsFormsRequiredFileName>Microsoft.NET.Sdk.WindowsDesktop.WindowsForms</_WindowsFormsRequiredFileName>
     </PropertyGroup>
 
     <Error Text="Unable to resolve path to Microsoft.Private.Winforms NuGet package. Is %24(PkgMicrosoft_Private_Winforms) defined?"
@@ -23,7 +23,7 @@
     <ItemGroup>
       <!-- Enumerate all transported files -->
       <_WindowsFormsContent Include="$(_WindowsFormsContentPath)" />
-      <!-- ...and verify Microsoft.NET.SDk.WindowsDesktop.WindowsForms.props/.targets are present -->
+      <!-- ...and verify Microsoft.NET.Sdk.WindowsDesktop.WindowsForms.props/.targets are present -->
       <_WindowsFormsContentFiles Include="@(_WindowsFormsContent->'%(FileName)%(Extension)')"
           Condition=" '%(FileName)%(Extension)' == '$(_WindowsFormsRequiredFileName).props'
                    or '%(FileName)%(Extension)' == '$(_WindowsFormsRequiredFileName).targets' "/>


RussKie:
Blocked by https://github.com/dotnet/wpf/pull/4823

vatsan-madhavan:
The props/targets changes (*.GlobalUsings.props/targets) are already merged (into dotnet/winforms) coming into the SDK via this PR is organized as a NuGet/transport-package integration. From what I can tell, those props/targets files are not part of this PR directly. 

IMO reviewers curating the WindowsDesktop SDK should be evaluating functional changes made to it by paying attention to the actual changes (like the contents of the new *.GlobalUsings.props/targets files), and looking for both correctness and functional/integration problems in those changes - tasks that are made difficult to perform in this format (i.e., a format that doesn't include the functional changes to the SDK directly/inline in the PR)

I'd suggest putting the changes in *GlobalUsings.props/targets directly into WinFx.targets instead - which is already a shared WPF/WinForms file. If you foresee a lot of WinForms-specific build-logic arising in the near-future, then perhaps creating a new targets file for WinForms next to WinFx.targets (for e.g., System.Windows.Forms.targets) might do the trick. 

The transport-package based approach will also lead to readability/maintainability degradation. Today, its easy to look up the sources of the WindowsDesktop SDK in one place, but with this type of forking of content, it will become much harder to look at its sources and understand how it works.

RussKie:
Thank you for your suggestions.

> The transport-package based approach will also lead to readability/maintainability degradation. Today, its easy to look up the sources of the WindowsDesktop SDK in one place, but with this type of forking of content, it will become much harder to look at its sources and understand how it works.

Equally it will also make it difficult to understand and appreciate Windows Forms related concerns if they are split across multiple repositories. It will also sound very odd "_If you need to see changes for Windows Forms in Windows Desktop go and make a change in WPF_".

RussKie:
> I'd suggest putting the changes in *GlobalUsings.props/targets directly into WinFx.targets instead - which is already a shared WPF/WinForms file. If you foresee a lot of WinForms-specific build-logic arising in the near-future, then perhaps creating a new targets file for WinForms next to WinFx.targets (for e.g., System.Windows.Forms.targets) might do the trick.

IIUIC, WinFx.targets is shipped with VS, which ties us to VS release cadence, and makes our functionality VS version specific. So I don't believe this is a good way forward either.

RussKie:
We plan to add more Windows Forms specific functionality (e.g. https://github.com/dotnet/wpf/pull/4605). I will refactor and create more generic System.Windows.Forms.props/targets and replace System.Windows.Forms.GlobalUsings.props/targets when the work on #4605 continues.
This work on the critical path, so I'm merging it as is.

vatsan-madhavan:
If WinFx.targets is included by VS, then it cannot possibly be carried in isolation. Either VS carries the .NET Framework version of WinFX.targets and thus it has no bearing on the contents of WindowsDesktop SDK, or VS carries the .NET Core version and it will necessarily have to carry all the contents of WindowsDesktop SDK no matter where its produced. Either way, worrying about VS seems moot. 

>Equally it will also make it difficult to understand and appreciate Windows Forms related concerns if they are split across multiple repositories. It will also sound very odd "If you need to see changes for Windows Forms in Windows Desktop go and make a change in WPF".

That's fair. On balance I'm more interested in ensuring a high-quality product and SDK's like this are used commonly and inadvertent mistakes in these can cause regressions easily. 

In the medium term, the appropriate solution would be to refactor the SDK into a separate repo (along with System.Xaml, PresentationBuildTasks) that builds prior to either of dotnet/winforms, dotnet/wpf, and then take a flow from it.

In the short term,  there is no perfect solution - either option is somewhat imperfect and I'm inclined to suggest that you optimize towards overall quality and favor easier reviews, overall maintainability etc.

RussKie:
> > Equally it will also make it difficult to understand and appreciate Windows Forms related concerns if they are split across multiple repositories. It will also sound very odd "If you need to see changes for Windows Forms in Windows Desktop go and make a change in WPF".
> 
> That's fair. On balance I'm more interested in ensuring a high-quality product and SDK's like this are used commonly and inadvertent mistakes in these can cause regressions easily.

If we have a well defined integration point (e.g. System.Windows.Forms.props/targets) then any regressions would be isolated to Windows Forms. We do routinely test builds (when they flow through, but that's a separate issue), so regression will be caught (sufficiently early). IMO dealing with such regressions would be easier if they are in a single repo.
We need to increase our test coverage in Windows Desktop, where we test both Windows Forms and WPF scenarios (e.g. "can create a blank app with top level statements", "can create a blank app with top level statements idsabled", etc.) similar to what the SDK has. This is probably our biggest gap to date.

> In the medium term, the appropriate solution would be to refactor the SDK into a separate repo (along with System.Xaml, PresentationBuildTasks) that builds prior to either of dotnet/winforms, dotnet/wpf, and then take a flow from it.

That'd be a great initiative. Though with the current load on both teams I can't see this work happening in the near-med term. 
@merriemcgaw / @Pilchie to prioritise.

merriemcgaw:
> That'd be a great initiative. Though with the current load on both teams I can't see this work happening in the near-med term.
@merriemcgaw / @Pilchie to prioritise.
Unfortunately at this stage I'd have to agree.

vatsan-madhavan:
> If we have a well defined integration point (e.g. System.Windows.Forms.props/targets) then any regressions would be isolated to Windows Forms. We do routinely test builds (when they flow through, but that's a separate issue), so regression will be caught (sufficiently early). IMO dealing with such regressions would be easier if they are in a single repo.

These sorts of changes are direct additions into the SDK - they're not additions to the winforms portion of the runtime. What matters here is the maintainability of the WindowsDesktop SDK. I'm interested in preventing WindowsDesktop SDK from generally breaking users during builds of either WinForms or WPF builds, or other generalized scenarios like multitargeting builds, .NET Core builds etc.  - and to that end maintainability and review-ability are important.

Take for instance the fact that I tried to review your change and failed to coherently build a mental picture to my satisfaction because this PR doesn't have that overall picture in it. And in order to gain that overall view, I'd have to build your topic branch, get a NuGet package out of it, and then inspect it. If I wanted to look at a diff, then I'd have to diff that locally built NuGet package against a previous version (for e.g, one built out of main) and only *then* I could review this PR with any level of confidence. I'm simply saying that's not the sort of of reviewability/maintainability loss that seems reasonable.

Nirmal4G:
Yeah, not adding the props/targets/tasks into the Windows Desktop SDK directly will **definitely** cause lot more pain in the future. Just as Vatsan Madhavan suggested, we can move the Desktop SDK and the `PresentationBuildTasks` into [.NET SDK](/dotnet/sdk) repo. Currently PBT depends on sources of `PresentationCore`, `PresentationFramework`, `WindowsBase` and a whole lot of shared sources! So, to move the Windows Desktop SDK into .NET SDK repo, we could reference them as assemblies and either link and trim them similar to what `NuGet.Build.Tasks.Pack` does or require Windows Desktop runtime to be installed to use PBT.

IMHO, I prefer link and trim solution than having to have Desktop runtime requirement!

Nirmal4G:
Yeah, this is baaaaaad!!! not adding the props/targets/tasks into the Windows Desktop SDK directly will **definitely** cause lot more pain in the future.

But as Vatsan Madhavan suggested, we can move the Desktop SDK and the `PresentationBuildTasks` into [.NET SDK](/dotnet/sdk) repo. I see that PBT currently depends on sources of `PresentationCore`, `PresentationFramework`, `WindowsBase` and a whole lot of shared sources! So, to move the Windows Desktop SDK into .NET SDK repo, we could reference them as assemblies and either link and trim them similar to what `NuGet.Build.Tasks.Pack` does or require Windows Desktop runtime to be installed to use PBT.

IMHO, I prefer link and trim solution than having to have Desktop runtime requirement!

RussKie:
@Nirmal4G are you volunteering to do the work and see it through?

packaging/Microsoft.NET.Sdk.WindowsDesktop/targets/Microsoft.NET.Sdk.WindowsDesktop.targets
```diff
@@ -155,4 +155,11 @@
   <!-- Import WPF Build logic only when we don't import NETFX's WinFX targets -->
   <Import Project="Microsoft.WinFX.targets" Condition="'$(ImportFrameworkWinFXTargets)' != 'true'"/>
 
+  <!--
+    Import Windows Forms targets.
+    These come via the Windows Forms transport package, that can be found under
+    https://github.com/dotnet/winforms/tree/main/pkg/Microsoft.Private.Winforms/sdk
+  -->
+  <Import Project="System.Windows.Forms.targets" />
```
Nirmal4G, line 163: Always make sure the WinFX targets is imported at last.

RussKie: Is there a specific reason for this?

Nirmal4G: Just a convention. I saw the content of the targets; it doesn't have much except a property to disable implicit references. For simple targets without any (_you know_) targets (_couldn't help myself_😂), placing it before any targets (_even the complex ones in WinFX_) is just easier to review, reason about and understand.

This is also why I moved the WinFX targets to bottom!

Nirmal4G:
> are you volunteering to do the work and see it through?

I could try but I don't know how the SDK, WPF and WinForms repo's build system work. Also, from what I understand, this needs changes across all three repos. Even if it's okay with the teams (_all three of them_), I might need a lot of help to understand all the three repos and its build mechanism to make these changes. Is that right?

So, is it okay for me to take this?

packaging/Microsoft.NET.Sdk.WindowsDesktop/targets/Microsoft.NET.Sdk.WindowsDesktop.props
```diff
@@ -161,4 +161,11 @@
     <SupportedTargetFramework Remove="@(_UnsupportedNETCoreAppTargetFramework);@(_UnsupportedNETStandardTargetFramework);@(_UnsupportedNETFrameworkTargetFramework)" />
   </ItemGroup>
 
+  <!--
+    Import Windows Forms props.
+    These come via the Windows Forms transport package, that can be found under
+    https://github.com/dotnet/winforms/tree/main/pkg/Microsoft.Private.Winforms/sdk
+  -->
+  <Import Project="System.Windows.Forms.props" />
```
Nirmal4G, line 169: It might be too late (_or not_) How about `Microsoft.NET.Desktop.WindowsForms` for WinForms specific props/targets and `Microsoft.NET.Desktop.WPF` for WPF specific props/targets?

We can also have `Microsoft.NET.WindowsDesktop.WindowsForms` or `Microsoft.NET.WindowsDesktop.Forms` and `Microsoft.NET.WindowsDesktop.WPF`, similar naming to `Microsoft.NET.WindowsDesktop.App.<WindowsForms|WPF>` packages.

RussKie:
> I could try but I don't know how the SDK, WPF and WinForms repo's build system work. Also, from what I understand, this needs changes across all three repos. Even if it's okay with the teams (_all three of them_), I might need a lot of help to understand all the three repos and its build mechanism to make these changes. Is that right?
> 
> So, is it okay for me to take this?

The Windows Desktop SDK is jointly owned by Windows Forms team (I'm from it) and the WPF team. The teams have different priorities, engagement models, workloads, etc. As noted above, at this stage the Windows Forms team doesn't any capacity to participate in the restructure of the Windows Desktop SDK; and I suspect the WPF is in the same boat.

Whilst we all appreciate help with a restructure I don't think I'd be able to provide necessary amount of attention or guidance at least until after we'll have shipped .NET 6.0.

packaging/Microsoft.NET.Sdk.WindowsDesktop/targets/Microsoft.NET.Sdk.WindowsDesktop.props
RussKie: We're having an offline discussion with @ryalanms and @SamBent about this.

Nirmal4G:
I thought so too. That's why, I've been looking at how the ASP.NET team moved their SDKs into the .NET SDK repo. I think I can do this without any help, since, the template on how to do it, is already there.

If it's okay with the teams, I can open a Draft PR soon.

Do we want to preserve history or just moving the latest HEAD is enough?

RussKie:
> Do we want to preserve history or just moving the latest HEAD is enough?

I don't have a strong preference either way.



> I thought so too. That's why, I've been looking at how the ASP.NET team moved their SDKs into the .NET SDK repo. I think I can do this without any help, since, the template on how to do it, is already there.
> 
> If it's okay with the teams, I can open a Draft PR soon.

Sure thing. IIRC there was a push to break the SDK into separate optional workloads, e.g. Windows Desktop, ASP.NET Core, etc., so that developer would download and install only relevant workloads.
We can use your draft as a way to start a discussion.

Nirmal4G:
For the Draft PR, I'll start with the latest HEAD. Then, if we need history preserved, I'll update it with the filtered commits.

> there was a push to break the SDK into separate optional workloads…

This is great. But the SDKs all have to be there inorder for this to come to fruition. So, moving the Windows Desktop SDK would definitely push that.

> We can use your draft as a way to start a discussion.

Sure.

Pilchie:
Note that before doing the work to move everything to the SDK repo, it would be good to open an issue there and discuss the proposal with that team.

packaging/Microsoft.NET.Sdk.WindowsDesktop/targets/Microsoft.NET.Sdk.WindowsDesktop.props
ryalanms: Based on the discussion in the WinForms/WPF teams meeting, WinForms has additional upcoming work that will require other WinForms-specific props/targets.  This is only the first of several changes.  Taking on the additional complexity of moving files in the WinForms transport package to be included in the WindowsDesktop SDK (in the WPF repo) seems reasonable; these \*WinForms\*.props/targets will only be imported when UseWinForms is true.   WPF will also scope our files to WPF-only.

ryalanms: Condition="'$(UseWindowsForms)' == 'true'"

packaging/Microsoft.NET.Sdk.WindowsDesktop/targets/Microsoft.NET.Sdk.WindowsDesktop.targets
ryalanms: These should only be imported if Condition="'$(UseWindowsForms)' == 'true'.

packaging/Microsoft.NET.Sdk.WindowsDesktop/targets/Microsoft.NET.Sdk.WindowsDesktop.targets
```diff
@@ -155,4 +155,11 @@
   <!-- Import WPF Build logic only when we don't import NETFX's WinFX targets -->
   <Import Project="Microsoft.WinFX.targets" Condition="'$(ImportFrameworkWinFXTargets)' != 'true'"/>
```
ryalanms, line 156: As you pointed out in the meeting, because the SDK is not referenced explicitly in the project element, the WindowsDesktop SDK is only imported if one or both of UseWPF and UseWindowsForms is true.  Microsoft.WinFX.targets can be scoped to WPF only.

ryalanms:
> IIUIC, WinFx.targets is shipped with VS, which ties us to VS release cadence, and makes our functionality VS version specific. So I don't believe this is a good way forward either.

There are two WinFX.targets, one that is shipped with VS for .NET Framework and one that is part of .NET Core.  WPF will import one or the other depending on the target framework and `UseWPF` value.

If targeting .NET Framework, the .NET Framework WinFX.targets is imported unless `UseWPF` is set.  If targeting .NET Core, the new WinFX.targets from the WindowsDesktop SDK is always imported.

ryalanms (changes requested):

packaging/Microsoft.NET.Sdk.WindowsDesktop/targets/Microsoft.NET.Sdk.WindowsDesktop.props
RussKie: No, we can't conditionally import as `UseWindowsForms` isn't defined yet
image

packaging/Microsoft.NET.Sdk.WindowsDesktop/targets/Microsoft.NET.Sdk.WindowsDesktop.targets
RussKie: Yes, we can conditionally import the targets
image

RussKie:
@ryalanms I have made changes as per our discussion. The PR description is updated to reflect that.

packaging/Microsoft.NET.Sdk.WindowsDesktop/targets/Microsoft.NET.Sdk.WindowsDesktop.props
```diff
@@ -161,4 +161,11 @@
     <SupportedTargetFramework Remove="@(_UnsupportedNETCoreAppTargetFramework);@(_UnsupportedNETStandardTargetFramework);@(_UnsupportedNETFrameworkTargetFramework)" />
   </ItemGroup>
 
+  <!--
+    Import Windows Forms props.
+    These come via the Windows Forms transport package, that can be found under
+    https://github.com/dotnet/winforms/tree/main/pkg/Microsoft.Private.Winforms/sdk
+  -->
+  <Import Project="Microsoft.NET.SDk.WindowsDesktop.WindowsForms.props" />
```
Nirmal4G, line 169:
```suggestion
  <Import Project="Microsoft.NET.Sdk.WindowsDesktop.WindowsForms.props"/>
```

Case fix

packaging/Microsoft.NET.Sdk.WindowsDesktop/targets/Microsoft.NET.Sdk.WindowsDesktop.targets
```diff
@@ -155,4 +155,11 @@
...
 
+  <!--
+    Import Windows Forms targets.
+    These come via the Windows Forms transport package, that can be found under
+    https://github.com/dotnet/winforms/tree/main/pkg/Microsoft.Private.Winforms/sdk
+  -->
+  <Import Project="Microsoft.NET.SDk.WindowsDesktop.WindowsForms.targets" Condition="'$(UseWindowsForms)' == 'true'"/>
```
Nirmal4G, line 163:
```suggestion
  <Import Project="Microsoft.NET.Sdk.WindowsDesktop.WindowsForms.targets" Condition="'$(UseWindowsForms)' == 'true'"/>
```

Case fix. Also, please add this before the `WinFX` targets.

Nirmal4G (changes requested):
Please add the `WinFX` targets at last.

eng/WindowsFormsImports.targets
```diff
@@ -0,0 +1,51 @@
+<Project>
```
ryalanms, line 1: Since this file will only be used when packaging the WinForms props/targets into the WindowsDekstop SDK, could it be moved to the WindowsDesktop packaging folder and renamed?

eng/WindowsFormsImports.target -> packaging/Microsoft.NET.Sdk.WindowsDesktop/Microsoft.NET.Sdk.WindowsDesktop.WinForms.Packaging.targets?

ryalanms (changes requested):

eng/WindowsFormsImports.targets
ryalanms: Actually, this could also go in eng\WpfArcadeSdk\tools alongside the other packaging targets.

RussKie: Putting Windows Forms specific targets in eng\WpfArcadeSdk\tools feels wrong. `packaging/Microsoft.NET.Sdk.WindowsDesktop/Microsoft.NET.Sdk.WindowsDesktop.WinForms.Packaging.targets` is probably a better suggestion.

packaging/Microsoft.NET.Sdk.WindowsDesktop/targets/Microsoft.NET.Sdk.WindowsDesktop.targets
RussKie: Thank you

packaging/Microsoft.NET.Sdk.WindowsDesktop/targets/Microsoft.NET.Sdk.WindowsDesktop.props
RussKie: Thank you

RussKie:
> Please add the `WinFX` targets at last.

Why? It is WPF specific, and I'm adding Windows Forms specific targets. They have no relation, and I can't a compelling reason to inject the new functionality in the middle.

packaging/Microsoft.NET.Sdk.WindowsDesktop/targets/Microsoft.NET.Sdk.WindowsDesktop.props
RussKie: Getting it fixed in our transport package too https://github.com/dotnet/winforms/pull/5332

packaging/Microsoft.NET.Sdk.WindowsDesktop/targets/Microsoft.NET.Sdk.WindowsDesktop.targets
RussKie: Yes, the SDK is imported. This is an existing behaviour. However to be able to build apps a developer must opt-in to Windows Forms and/or WPF.
Also our targets must be conditional to only execute when appropriate.

image

image

ryalanms (approved):

Nirmal4G:
[As I said before](4779#discussion_r672215140), it's just a matter of convention.

Also, we ensure that we don't accidently override something from the `WinFX` targets, if we import it before. This isn't 100% foolproof but not having the `WindowsForms` targets here helps a bit.

If I were to do this, I would take the opportunity to refactor this into WPF-specific logic into its own file.

May be `Microsoft.NET.Sdk.WindowsDesktop.WPF.<props/targets>`!?

RussKie:
> If I were to do this, I would take the opportunity to refactor this into WPF-specific logic into its own file.
> 
> May be `Microsoft.NET.Sdk.WindowsDesktop.WPF.<props/targets>`!?

It is something we discussed offline with @ryalanms. It is his call though.

Nirmal4G:
We don't need to do it now but planning ahead is not bad though! 🤓🧐

PR #4790, opened 2021-07-02
refresh API change process

SamBent:
* Add API spec template
* Cosmetic changes to ToolTip spec, to bring it closer to the new template, clarify some wording, and fix one glaring typo.
* Revise instructions for API-compat and GenAPI
* Disable api-compat checks during a gen-api build.  This has to be done anyway, so it should be built in.

This PR does not change any product code.

### DIFF
--- a/eng/WpfArcadeSdk/tools/ApiCompat.targets
+++ b/eng/WpfArcadeSdk/tools/ApiCompat.targets
@@ -10,8 +10,9 @@
                                           and !($(MSBuildProjectName.Contains('_wpftmp')))">true</RunNetFrameworkApiCompat>
   </PropertyGroup>
 
-  <!-- Only enable Ref API Compat for the lib projects using hand-crafted reference assemblies -->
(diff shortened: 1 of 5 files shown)


eng/WpfArcadeSdk/tools/ApiCompat.targets
```diff
@@ -10,8 +10,9 @@
                                           and !($(MSBuildProjectName.Contains('_wpftmp')))">true</RunNetFrameworkApiCompat>
   </PropertyGroup>
 
-  <!-- Only enable Ref API Compat for the lib projects using hand-crafted reference assemblies -->
-  <PropertyGroup Condition="'$(RunRefApiCompat)'==''">
+  <!-- Only enable Ref API Compat for the lib projects using hand-crafted reference assemblies,
+  but not if we're generating the reference assemblies -->
```
ryalanms, line 14: nit: $(GenerateReferenceAssemblySource)'!='true' for when it is explicitly false.

ryalanms (approved):

Issue #4781, opened 2021-07-01
Application.GetContentStream() may fail if the app is in a path that contains a "#".


neelabo:
* .NET Core Version:  .NET 5.0
* Windows version: 21H1
* Does the bug reproduce also in WPF for .NET Framework 4.8?: Yes

 **Problem description:**

Application.GetContentStream() will fail if you place the executable in a path that contains "#".
Apparently, the paths after the "#" have disappeared. 
 
 **Minimal repro:**

For example, if you place the executable file and content file in "C:\FooBar\Path#Name\\",
you will get the exception `FileNotFoundException: Could not find file 'C:\FooBar\test.png'.`.

``` cs
var uri = new Uri("/test.png", UriKind.Relative); // test.png is content file
var streamResourceInfo = Application.GetContentStream(uri); // <- Exception!!
```

lindexi:
I wrote a demo in GitHub:  https://github.com/lindexi/lindexi_gd/tree/e1022aededd1871017e4c43f228ad1fe26c99b57/  and I found that it seems is the Uri's bug, not the WPF

dipeshmsft:
Closing this issue as the corresponding fix has been merged in main branch

PR #4799, opened 2021-07-05
Fix GetEntryAssemblyLocation Uri

lindexi:
Fixes Issue #4781

## Description

We do not escape the path string in GetEntryAssemblyLocation method. And it will break when the application in the special path, such as the path contains a `#`. See https://github.com/dotnet/wpf/issues/4781

## Customer Impact

All the GetContentStream

## Regression

No.

## Testing

Just CI


miloush:
Is similar fix needed for when `#` is in the assembly file name rather than the directory path?

lindexi:
> 
> 
> Is similar fix needed for when `#` is in the assembly file name rather than the directory path?

Sorry, it seems hard to create the assembly with the name contains `#`? Could you tell me how to create it?

anjalisheel-wpf (approved):

lindexi:
@anjalisheel-wpf  Thank you. And any news?

singhashish-wpf (approved):

PR #4800, opened 2021-07-06
Update FxCop analyzers

ThomasGoulet73:
## Description
Update Microsoft.CodeAnalysis.FxCopAnalyzers to 3.3.1 and activate valid rules.

## Customer Impact
None, there are no behavior changes. The last version before the deprecated warning. I plan on doing a follow-up PR on chaning FxCopAnalyzers for NetAnalyzers.

## Regression
None.

## Testing
Local build.

## Risk
None.

### DIFF
--- a/eng/WpfArcadeSdk/tools/CodeAnalysis.targets
+++ b/eng/WpfArcadeSdk/tools/CodeAnalysis.targets
@@ -6,8 +6,7 @@
   <ItemGroup Condition="'$(IsTestProject)'!='true' AND '$(EnableAnalyzers)'=='true'">
     <!-- Managed Code Reference analyzers -->
     <PackageReference Include="Microsoft.DotNet.CodeAnalysis" Version="$(MicrosoftDotNetCodeAnalysisPackageVersion)" />
-    <PackageReference Include="Microsoft.CodeAnalysis.Common" Version="2.9.0" />
-    <PackageReference Include="Microsoft.CodeAnalysis.FxCopAnalyzers" Version="2.6.3" />
+    <PackageReference Include="Microsoft.CodeAnalysis.FxCopAnalyzers" Version="3.3.1" />
     <PackageReference Include="System.Runtime.Analyzers" Version="1.1.0" />
     <PackageReference Include="System.Runtime.InteropServices.Analyzers" Version="1.1.0" />
   </ItemGroup>
--- a/eng/WpfArcadeSdk/tools/CodeAnalysis/WpfCodeAnalysis.ruleset
+++ b/eng/WpfArcadeSdk/tools/CodeAnalysis/WpfCodeAnalysis.ruleset
@@ -21,7 +21,6 @@
   </Rules>
   <Rules AnalyzerId="Microsoft.CodeAnalysis.FxCopAnalyzers" RuleNamespace="Microsoft.CodeAnalysis.FxCopAnalyzers">
     <Rule Id="CA1062" Action="None" />
-    <Rule Id="CA2000" Action="None" />
     <Rule Id="CA2002" Action="None" />
   </Rules>
 </RuleSet>
--- a/eng/WpfArcadeSdk/tools/CodeAnalysis/WpfCodeAnalysis.ruleset
+++ b/eng/WpfArcadeSdk/tools/CodeAnalysis/WpfCodeAnalysis.ruleset
@@ -21,7 +21,6 @@
   </Rules>
   <Rules AnalyzerId="Microsoft.CodeAnalysis.FxCopAnalyzers" RuleNamespace="Microsoft.CodeAnalysis.FxCopAnalyzers">
     <Rule Id="CA1062" Action="None" />
-    <Rule Id="CA1805" Action="None" />
     <Rule Id="CA2000" Action="None" />
     <Rule Id="CA2002" Action="None" />
   </Rules>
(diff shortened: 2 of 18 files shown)


src/Microsoft.DotNet.Wpf/src/Shared/MS/Internal/Xaml/Parser/SpecialBracketCharacters.cs
```diff
@@ -106,7 +106,7 @@ internal bool EndsEscapeSequence(char ch)
 
         internal bool Match(char start, char end)
         {
-            return _endChars.IndexOf(end.ToString()) == _startChars.IndexOf(start.ToString());
+            return _endChars.IndexOf(end.ToString(), StringComparison.Ordinal) == _startChars.IndexOf(start.ToString(), StringComparison.Ordinal);
```
ThomasGoulet73, line 109: This is the only behavior change to use Ordinal by instead of the default current culture.

It is more inline with the usage of Contains, which defaults to Ordinal comparison here:
https://github.com/dotnet/wpf/blob/78e93de6c2450b4f7b1e3adb5a8b357ab8823d41/src/Microsoft.DotNet.Wpf/src/Shared/MS/Internal/Xaml/Parser/SpecialBracketCharacters.cs#L94
and here:
https://github.com/dotnet/wpf/blob/78e93de6c2450b4f7b1e3adb5a8b357ab8823d41/src/Microsoft.DotNet.Wpf/src/Shared/MS/Internal/Xaml/Parser/SpecialBracketCharacters.cs#L99

eng/WpfArcadeSdk/tools/CodeAnalysis/WpfCodeAnalysis.ruleset
```diff
@@ -19,4 +19,7 @@
   <Rules AnalyzerId="System.Runtime.CSharp.Analyzers" RuleNamespace="System.Runtime.CSharp.Analyzers">
     <Rule Id="CA2213" Action="None" />
   </Rules>
+  <Rules AnalyzerId="Microsoft.CodeAnalysis.FxCopAnalyzers" RuleNamespace="Microsoft.CodeAnalysis.FxCopAnalyzers">
+    <Rule Id="CA1062" Action="None" />
```
ThomasGoulet73, line 23: Disable [CA1062](https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1062) because there are false positives, when a public method is invoking another method that already does the null check.

eng/WpfArcadeSdk/tools/CodeAnalysis.targets
```diff
@@ -6,8 +6,7 @@
   <ItemGroup Condition="'$(IsTestProject)'!='true' AND '$(EnableAnalyzers)'=='true'">
     <!-- Managed Code Reference analyzers -->
     <PackageReference Include="Microsoft.DotNet.CodeAnalysis" Version="$(MicrosoftDotNetCodeAnalysisPackageVersion)" />
-    <PackageReference Include="Microsoft.CodeAnalysis.Common" Version="2.9.0" />
```
ThomasGoulet73, line 9: This shouldn't be needed and it conflicts with other PackageReference when enabling analyzers in any other project than System.Xaml.

eng/WpfArcadeSdk/tools/CodeAnalysis/WpfCodeAnalysis.ruleset
dipeshmsft: @ThomasGoulet73  There are only two such cases where this warning is raised. Shall we suppress only for those two locations, instead of globally suppressing it?

ThomasGoulet73: @dipeshmsft You're right, I thought about it more and I think we should enable it to make sure that we do null checks in the public method, even if there is a null check in a private method. This helps make sure that the argument name in the exception is the same as the public one, in case the argument name in the private method is not the same as the one in the public method. This additional check probably won't even be noticeable performance-wise.

I'll enable it in a separate commit and I'll leave the decision to the reviewers.

src/Microsoft.DotNet.Wpf/src/System.Xaml/System/Xaml/XamlSchemaContext.cs
```diff
@@ -542,6 +542,9 @@ private string GetCompatibleNamespace(string oldNs)
 
         public virtual XamlType GetXamlType(Type type)
         {
+            if (type == null)
```
ThomasGoulet73, line 545: This one would succeed in XamlLanguage.TypeAlias(type) but throw in GetXamlType, By adding an explicit throw here, we make sure to throw early, with the right stacktrace and the right parameter name.

src/Microsoft.DotNet.Wpf/src/System.Xaml/System/Windows/Markup/ValueSerializer.cs
```diff
@@ -244,6 +244,9 @@ public static ValueSerializer GetSerializerFor(PropertyDescriptor descriptor, IV
         /// </summary>
         protected Exception GetConvertToException(object value, Type destinationType)
         {
+            if (destinationType == null)
```
ThomasGoulet73, line 247: This one would already throw a NullReferenceException destinationType.FullName. By adding an explicit throw here, we make sure to throw early, with the right stacktrace and the right parameter name.

**This one is technically a breaking change, the type of exception returned is not the same as before. Let me know if you would prefer to revert this change.**

RussKie: Whilst this is a breaking change indeed, these types of changes have been generally considered acceptable across the .NET ecosystem.
https://github.com/dotnet/runtime/blob/main/docs/coding-guidelines/breaking-changes.md#bucket-2-reasonable-grey-area

dipeshmsft (approved):

PR #4868, opened 2021-07-17
Only enable `DisableImplicitNamespaceImports` when `UseWPF` is set

ryalanms:
WinForms project also import Microsoft.WinFX.targets, and WinForms now relies on implicit imports.  Scope disabling implicit imports to WPF.  This broke WinForms VB projects.

> But it wasn't @JunTaoLuo's change(s). It's actually caused by DisableImplicitNamespaceImports here: dotnet/wpf@ff1be8e#diff-681ca681001fbc20488df7fdbb46a8c40ac2df9dc2210df5ad869aee6a56df16R23
This disables these VB imports:
https://github.com/dotnet/sdk/blob/55195f3c4ba3f582aa91ec824bd24a7ba3e73528/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.DefaultItems.props#L118-L129

/cc @dotnet/wpf-developers


ericstj:
FWIW implicit namespace imports have been around for VB much longer than C#.  I’d be surprised if they broke WPF, I suspect it’s had them for VB even on .NETFramework.

Issue #4848, opened 2021-07-15
Undo workaround for inboxed analyzers after updating to Preview 7


ryalanms:
Undo workaround for inboxed analyzers after updating to Preview 7.

WPF, WinForms, and WindowsDesktop required this Arcade change.  After updating to Preview 7, remove Exclude inboxed analyzers from VerifyClosure task dotnet/arcade#7624.

ericstj:
Moving to WPF as workaround will be moved here.

ryalanms:
Reverted here:  https://github.com/dotnet/arcade/commit/7b88b24068ecb4f6cf1da8395dbc2dee25a700bd.  

Thanks, @ericstj.

ericstj:
I moved this here to track undoing https://github.com/dotnet/wpf/pull/4849

ericstj:
I believe you should be able to undo that now, since you're on an RC sdk.

ericstj:
Let me see if I can do this now.

PR #4849, opened 2021-07-15
Remove analyzers from framework references

ericstj:
These are incorrectly raised as references in pre-6.0-preview7 SDKs.

Fixes broken build dotnet/windowsdesktop#1803 by removing these references from WPF's C++/CLI assemblies.

#4848 tracks undoing this.

PR #4885, opened 2021-07-20
Revert "Remove analyzers from framework references"

ericstj:
Reverts dotnet/wpf#4849

This should no longer needed now that WPF is consuming an RC SDK.  Don't merge this until we manually validate the C++/CLI binaries don't have references to the JSON source generator (since that would only fail after ingestion upstack).

### DIFF
--- a/Directory.Build.targets
+++ b/Directory.Build.targets
@@ -1,16 +1,9 @@
 <Project>
+
   <Import Project="$(WpfArcadeSdkTargets)"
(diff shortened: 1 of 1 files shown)


ericstj:
Fixes https://github.com/dotnet/wpf/issues/4848

ericstj:
I have reviewed the binlogs and the issue no longer exists, so this should be good to go.  Can I get a review @ryalanms?

ryalanms (approved):

ryalanms:
Thank you, @ericstj!

PR #4889, opened 2021-07-20
Use preview language version for WPF

ryalanms:
Closes #4748

Update language version to support implicit global namespace feature in .NET 6.0 Preview 7.

(Waiting on test pass results...)


RussKie (approved):

ryalanms:
The IL injection used to create a PresentationCore needs to be replaced in this change.

ryalanms:
This was already updated with Replace and remove IL injection in PresentationCore #4911.

PR #4911, opened 2021-07-23
Replace and remove IL injection in PresentationCore

ryalanms:
Module initializers are now supported with a single attribute.  As a prerequisite to updating the WPF repo to build against C# 10, this change adds the single ModuleInitializer attribute to PresentationCore and removes the IL injection logic from the WPF repo. 

The module initializer without IL injection is working as expected:

 >PresentationCore.dll!ModuleInitializer.Initialize() Line 21 C#
PresentationCore.dll!.() Unknown
[Native to Managed Transition]

Prior to C# 9, the PresentationCore assembly was built, disassembled to IL, combined with additional IL containing a module initializer, and reassembled to a final PresentationCore assembly.  

Related:
https://github.com/dotnet/wpf/issues/3885
https://github.com/dotnet/wpf/pull/4748
https://github.com/dotnet/wpf/pull/4889

/cc @rladuca @vatsan-madhavan @SamBent


ryalanms:
global.json and versions.props still need to be updated.

eng/WpfArcadeSdk/Sdk/Sdk.props
```diff
@@ -14,7 +14,7 @@
 
   <PropertyGroup>
     <GenXmlStringTable>$(WpfArcadeSdkToolsDir)GenXmlStringTable.pl</GenXmlStringTable>
-    <LangVersion Condition="'$(LangVersion)'==''">8.0</LangVersion>
+    <LangVersion Condition="'$(LangVersion)'==''">preview</LangVersion>
```
rladuca, line 17: When you snap to a release branch I imagine that you're now going to have to set this to a true release or else your lang version is going to march forward in stuff like LTS branches.  You probably want to document that process somewhere.

ryalanms: It's an unusual situation because of implicit global namespace feature: 

> error CS8652: The feature 'global using directive' is currently in Preview and *unsupported*. To use Preview features, use the 'preview' language version.

rladuca: Got it, so I imagine the language features will stabilize before this becomes a release branch.  Just making sure that "preview" doesn't make its way to a release accidentally.

ryalanms: I've added a tracking issue here:  https://github.com/dotnet/wpf/issues/4913.

SamBent (approved):

eng/WpfArcadeSdk/Sdk/Sdk.props
vatsan-madhavan: I recommend sticking to a particular LangVersion here rather than preview. Alternatively, detect your branch name or use another breadcrumb like version string has words like 'preview' and use that to set this value to preview, and when your detect you're on a release/servicing branch, set it to a well-known stable value.

vatsan-madhavan:
lgtm. Have you checked IL-diff's before vs after that everthing looks good?

eng/WpfArcadeSdk/Sdk/Sdk.props
ryalanms: A condition has been added to set langversion to preview for "rc" and "preview" but not release.

vatsan-madhavan (approved):
Looked at IL-diff with @ryalanms over Teams - changes look good 👍

Issue #3811, opened 2020-11-13
Improve linkability for WPF


agocke:
At the top-level, WPF apps should be able to use trimming with predictable results.

## Goals

- [ ] WPF template produces zero/minimal trim warnings
- [ ] WPF APIs which are not linker-friendly are correctly annotated

### Information gathering

- [ ] Generate list of warnings produced when linking a representative WPF app
- [ ] Categorize warnings, sorting to netcore warnings and WPF-specific warnings (see https://github.com/dotnet/runtime/issues/43078)
- [ ] Identify the cause of the warnings and propose fixes

### Code fixes

- [ ] Fix warnings or annotate APIs as necessary
- [ ] Identify linker-unsafe constructs and propose replacements for functionality

samsp-msft:
The goal for .NET 6 should be to be able to the trim the framework libraries for WPF apps, but not the user code or 3rd party references.

mateoatr:
Here is the list of all the linker warnings that I get when publishing a wpf app with `-p:PublishTrimmed=true` and disabling the default `--no-warn`s: https://gist.github.com/mateoatr/596ba43e079a138ac79b907d2974089d

LakshanF:
WPF specific warnings can be seen in the file below,
[usinglinkerwrn3.txt](https://github.com/dotnet/wpf/files/5660731/usinglinkerwrn3.txt)


Warnings | Count
-- | --
IL2075 | 89
IL2070 | 59
IL2026 | 37
IL2072 | 25
IL2080 | 22
IL2094 | 11
IL2057 | 10
IL2067 | 8
IL2077 | 6
IL2060 | 5
IL2055 | 5
IL2090 | 4
IL2087 | 2
IL2065 | 1
IL2091 | 1
IL2058 | 1
IL2059 | 1

Symbai:
> The goal for .NET 6 should be to be able to the trim the framework libraries for WPF apps, but not the user code or 3rd party references.

What's the status on this? As of now WPF does not support any trimming in .NET6 and there are only a few months left before .NET6 releases. Is this still targeted for .NET6 or can someone say for sure that it will take longer?

agocke:
It seems very unlikely this will be done for 6.0. 

I'll leave it up to @ryalanms to comment on WPF resourcing/scheduling.

mansellan:
Is there an update on this? Might it make it into .Net 7.0 ?

rgwood:
[...]

PR #4917, opened 2021-07-23
Use generic Marshal.PtrToStructure

ThomasGoulet73:
## Description
Use generic Marshal.PtrToStructure instead of the non-generic. The generic one is slightly faster than the non-generic. Also, I removed the customs UnsafeNativeMethods.PtrToStructure which justs forwards to Marshal.PtrToStructure because it causes linker warnings because Marshal.PtrToStructure is annotated while UnsafeNativeMethods.PtrToStructure isn't.

Contributes to #3811

## Customer Impact
None, same behavior as the non-generic.

## Regression
No

## Testing
I manually tested a few of the generic and non-generic and they gave the same result.

## Risk
None, same behavior as the non-generic.

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/DpiChangedEventArgs.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/DpiChangedEventArgs.cs
@@ -10,6 +10,7 @@
 //              different DPI, or the dpi of the current monitor changes.
 
 using System.ComponentModel;
+using System.Runtime.InteropServices;
 using System.Security;
 using System.Threading;
 using System.Windows.Interop;
@@ -48,7 +49,7 @@ namespace System.Windows
         {
             OldDpi = new DpiScale(oldDpiX / DpiUtil.DefaultPixelsPerInch, oldDpiY / DpiUtil.DefaultPixelsPerInch);
             NewDpi = new DpiScale(newDpiX / DpiUtil.DefaultPixelsPerInch, newDpiY / DpiUtil.DefaultPixelsPerInch);
-            NativeMethods.RECT suggestedRect = (NativeMethods.RECT)UnsafeNativeMethods.PtrToStructure(lParam, typeof(NativeMethods.RECT));
+            NativeMethods.RECT suggestedRect = Marshal.PtrToStructure<NativeMethods.RECT>(lParam);
             this.SuggestedRect = new Rect((double)suggestedRect.left, (double)suggestedRect.top, (double)suggestedRect.Width, (double)suggestedRect.Height);
         }
 
--- a/src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Input/InputProcessorProfiles.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Input/InputProcessorProfiles.cs
@@ -160,7 +160,7 @@ namespace System.Windows.Input
                  for (int i = 0; i < nCount; i++)
                  {
                      // Unmarshal each langid from short array.
-                     short langid = (short)Marshal.PtrToStructure((IntPtr)((Int64)langids + sizeOfShort * i), typeof(short));
+                     short langid = Marshal.PtrToStructure<short>((IntPtr)((Int64)langids + sizeOfShort * i));
                      arrayLang.Add(new CultureInfo(langid));
                  }
 
--- a/src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/InterOp/HwndSource.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/InterOp/HwndSource.cs
@@ -1368,7 +1368,7 @@ namespace System.Windows.Interop
 
                 // Get WINDOWPOS structure data from lParam; it contains information about the window's
                 // new size and position.
-                NativeMethods.WINDOWPOS windowPos = (NativeMethods.WINDOWPOS)UnsafeNativeMethods.PtrToStructure(lParam, typeof(NativeMethods.WINDOWPOS));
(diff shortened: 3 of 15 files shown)


vishalmsft (approved):

src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Window.cs
```diff
@@ -4710,11 +4710,7 @@ internal virtual void WmMoveChangedHelper()
 
         private bool WmGetMinMaxInfo( IntPtr lParam )
         {
-            NativeMethods.MINMAXINFO mmi;
```
ThomasGoulet73, line 4713: Hey @stephentoub, could I have your opinion here please?

In one of your PR, you changed Marshal.PtrToStructure for this unsafe code: https://github.com/dotnet/wpf/pull/4754/files#diff-30ce94d35c72b94385aa439cafead5bd0677f7b4037f0fefcfc7a03d3730a681L4702-R4705

What is best, the unsafe code or generic Marshal.PtrToStructure ?

If it's the unsafe code, when should we use it ? Should I change my PR to convert Marshal.PtrToStructure to this unsafe code ?

Thank you!

stephentoub: Both are equally "unsafe", so it's purely a performance / functionality discussion.  If the type isn't blittable, use `Marshal.PtrToStructure`, as it appropriately handles all the necessary marshaling.  If the type is blittable, it's _way_ cheaper to just cast rather than using `PtrToStructure`.

|         Method |       Mean | Allocated |
|--------------- |-----------:|----------:|
|           Cast |  0.0821 ns |         - |
| PtrToStructure | 25.5936 ns |      24 B |

```C#
using BenchmarkDotNet.Attributes;
using BenchmarkDotNet.Running;
using System;
using System.Runtime.InteropServices;

[MemoryDiagnoser]
public class Program
{
    public static void Main(string[] args) => BenchmarkSwitcher.FromAssembly(typeof(Program).Assembly).Run(args);

    private IntPtr _ptr;

    [GlobalSetup]
    public unsafe void Setup() => _ptr = Marshal.AllocHGlobal(sizeof(MyPoint));

    [GlobalCleanup]
    public void Cleanup() => Marshal.FreeHGlobal(_ptr);

    [Benchmark]
    public unsafe MyPoint Cast() => *(MyPoint*)_ptr;

    [Benchmark]
    public MyPoint PtrToStructure() => Marshal.PtrToStructure<MyPoint>(_ptr);
}

[StructLayout(LayoutKind.Sequential)]
public struct MyPoint
{
    public int X;
    public int Y;
}
```

ThomasGoulet73: Thank you!

> Both are equally "unsafe", so it's purely a performance / functionality discussion
>
I didn't mean "unsafe" as "this code is less safe than this one", I just didn't know how to name your code and I used "unsafe" because it used unsafe blocks.

I'll revert this change and keep the usage of generic PtrToStructure everywhere else in this PR since the generic version is slightly faster than the non-generic.

I'll open an issue to use your code to cast IntPtr to blittable structs (Maybe also add a shared method to avoid having unsafe blocks everywhere to improve readability).

lindexi: @stephentoub Can I ask you why the PtrToStructureGeneric is faster than PtrToStructureNonGeneric?

See:

https://github.com/dotnet/runtime/blob/52c6dcb45cc1e8e98d267c07f10fe42d89c71656/src/libraries/System.Private.CoreLib/src/System/Runtime/InteropServices/Marshal.cs#L575-L600

https://github.com/dotnet/runtime/blob/52c6dcb45cc1e8e98d267c07f10fe42d89c71656/src/libraries/System.Private.CoreLib/src/System/Runtime/InteropServices/Marshal.cs#L615-L634

Thank you.

stephentoub: For the generic, and a value type, the JIT generates a specialization of the method for that value type, such that the PtrToStructure method knows what the type is, which means it can do at compile time things the non-generic needs to do at runtime.

lindexi: @stephentoub  Thank you very much.

src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/DpiChangedEventArgs.cs
```diff
@@ -48,7 +49,7 @@ internal HwndDpiChangedEventArgs(double oldDpiX, double oldDpiY, double newDpiX,
         {
             OldDpi = new DpiScale(oldDpiX / DpiUtil.DefaultPixelsPerInch, oldDpiY / DpiUtil.DefaultPixelsPerInch);
             NewDpi = new DpiScale(newDpiX / DpiUtil.DefaultPixelsPerInch, newDpiY / DpiUtil.DefaultPixelsPerInch);
-            NativeMethods.RECT suggestedRect = (NativeMethods.RECT)UnsafeNativeMethods.PtrToStructure(lParam, typeof(NativeMethods.RECT));
+            NativeMethods.RECT suggestedRect = Marshal.PtrToStructure<NativeMethods.RECT>(lParam);
```
RussKie, line 52: In general you want to avoid using Marshal class, and replace this with blittable structs and pointers

ThomasGoulet73: Yes, this is definitely something that I'm interested in doing in a separate PR. I would still keep this PR only on using the generic version since it's easier to review (Handles blittable and non-blittable), slightly faster than the non-generic overload and trim-friendly (Removes `UnsafeNativeMethods.PtrToStructure` which is not annotated for the linker).

Thank you!

RussKie (approved):

PR #4953, opened 2021-07-29
Fixing deterministic build bug with dotnet.exe

guimafelipe:
Fixes Issue #3876

Main PR 

## Description

Building a WPF .NET Core project with dotnet.exe is not deterministic, but building it with msbuild.exe is deterministic as expected.

This PR is based on [this](https://github.com/MichaeIDietrich/wpf/commit/cff984aaf851ba126a822190bf815d0bf5c4d40e) fix from @MichaeIDietrich.

## Customer Impact

This issue may break incremental builds.

## Regression

No regression.

## Testing

The fix was testing building some projects using the `dotnet build` command, and comparing the generated files using a diff program. Without the fix, the files are almost always different.


ryalanms:
LGTM.  I've included @MichaeIDietrich's commit as well so it's tracked in GitHub.

ryalanms (approved):

ryalanms:
This will need to be backported to 5.0 and 3.1.

Issue #4936, opened 2021-07-27
Replace helper functions through patternmatching


IAmTheCShark:
UIElement has the following helper functions
![image](https://user-images.githubusercontent.com/20905599/127213529-7a089ea9-e817-4b56-b842-d0c870ba5944.png)
that are used in a lot of places where they can easily be replaced by `is` pattern matching.

For example `CommandManager.FindCommandBinding` which is under heavy load in large applications should benefit from this.

![image](https://user-images.githubusercontent.com/20905599/127214395-1f99da4b-8174-42ed-90bc-71fe96da955e.png)

A quick test i threw together shoes how much faster this would be.

<html>
<body>

Method | o | Mean | Error | StdDev
-- | -- | -- | -- | --
IsInstanceOfType | ? | 2.6260 ns | 0.0040 ns | 0.0033 ns
IsExpression | ? | 0.2219 ns | 0.0038 ns | 0.0034 ns
IsInstanceOfType |   | 4.0874 ns | 0.0051 ns | 0.0042 ns
IsExpression |   | 0.3285 ns | 0.0010 ns | 0.0010 ns
IsInstanceOfType | 1 | 8.4281 ns | 0.0335 ns | 0.0297 ns
IsExpression | 1 | 0.3286 ns | 0.0013 ns | 0.0011 ns

</body>
</html>

lindexi:
@IAmTheCShark I hope to see this optimization.

dipeshmsft:
Closing this issue as the PR for the same is merged.

PR #4964, opened 2021-07-31
Replacing expensive type checks with "is" checks

batzen:
Fixes Issue 
#4936

## Description

Replacing expensive type checks with "is" checks.
I added a few todos to the code as i am not sure if we should throw explicit exceptions in those places. Before my changes those place would have thrown an InvalidCastException or a NullReferenceException.

## Customer Impact

Slower performance.

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Input/Command/CommandManager.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Input/Command/CommandManager.cs
@@ -352,7 +352,6 @@ namespace System.Windows.Input
 
             // Determine UIElement/ContentElement/Neither type
             DependencyObject targetElementAsDO = targetElement as DependencyObject;
-            // todo patternmatching: Would an as cast and null checks be faster here instead of casting multiple times?
             bool isUIElement = targetElementAsDO is UIElement;
             bool isContentElement = !isUIElement && targetElementAsDO is ContentElement;
             bool isUIElement3D = !isUIElement && !isContentElement && targetElementAsDO is UIElement3D;
--- a/src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Input/Command/CommandManager.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Input/Command/CommandManager.cs
@@ -352,9 +352,10 @@ namespace System.Windows.Input
 
             // Determine UIElement/ContentElement/Neither type
             DependencyObject targetElementAsDO = targetElement as DependencyObject;
-            bool isUIElement = InputElement.IsUIElement(targetElementAsDO);
-            bool isContentElement = !isUIElement && InputElement.IsContentElement(targetElementAsDO);
-            bool isUIElement3D = !isUIElement && !isContentElement && InputElement.IsUIElement3D(targetElementAsDO);
+            // todo patternmatching: Would an as cast and null checks be faster here instead of casting multiple times?
+            bool isUIElement = targetElementAsDO is UIElement;
+            bool isContentElement = !isUIElement && targetElementAsDO is ContentElement;
+            bool isUIElement3D = !isUIElement && !isContentElement && targetElementAsDO is UIElement3D;
 
             // Step 1: Check local input bindings
             InputBindingCollection localInputBindings = null;
@@ -370,6 +371,7 @@ namespace System.Windows.Input
             {
                 localInputBindings = ((UIElement3D)targetElement).InputBindingsInternal;
             }
+
             if (localInputBindings != null)
             {
                 InputBinding inputBinding = localInputBindings.FindMatch(targetElement, inputEventArgs);
@@ -423,6 +425,7 @@ namespace System.Windows.Input
                 {
                     localCommandBindings = ((UIElement3D)targetElement).CommandBindingsInternal;
                 }
+
                 if (localCommandBindings != null)
                 {
                     command = localCommandBindings.FindMatch(targetElement, inputEventArgs);
--- a/src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Input/MouseDevice.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Input/MouseDevice.cs
@@ -739,17 +739,17 @@ namespace System.Windows.Input
             // First, check things like IsEnabled, IsVisible, etc. on a
             // UIElement vs. ContentElement basis.
             //
-            if (dependencyObject is UIElement)
+            if (dependencyObject is UIElement uie)
             {
-                killCapture = !ValidateUIElementForCapture((UIElement)_mouseCapture);
+                killCapture = !ValidateUIElementForCapture(uie);
             }
-            else if (dependencyObject is ContentElement)
+            else if (dependencyObject is ContentElement ce)
             {
-                killCapture = !ValidateContentElementForCapture((ContentElement)_mouseCapture);
+                killCapture = !ValidateContentElementForCapture(ce);
             }
-            else if (dependencyObject is UIElement3D)
+            else if (dependencyObject is UIElement3D uie3D)
             {
-                killCapture = !ValidateUIElement3DForCapture((UIElement3D)_mouseCapture);
+                killCapture = !ValidateUIElement3DForCapture(uie3D);
             }
 
             //
--- a/src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Input/MouseDevice.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Input/MouseDevice.cs
@@ -739,15 +739,15 @@ namespace System.Windows.Input
             // First, check things like IsEnabled, IsVisible, etc. on a
             // UIElement vs. ContentElement basis.
             //
-            if (InputElement.IsUIElement(dependencyObject))
+            if (dependencyObject is UIElement)
             {
                 killCapture = !ValidateUIElementForCapture((UIElement)_mouseCapture);
             }
-            else if (InputElement.IsContentElement(dependencyObject))
+            else if (dependencyObject is ContentElement)
             {
                 killCapture = !ValidateContentElementForCapture((ContentElement)_mouseCapture);
             }
-            else if (InputElement.IsUIElement3D(dependencyObject))
+            else if (dependencyObject is UIElement3D)
             {
                 killCapture = !ValidateUIElement3DForCapture((UIElement3D)_mouseCapture);
             }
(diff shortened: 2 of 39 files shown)


src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Input/Command/CommandManager.cs
```diff
@@ -352,9 +352,10 @@ internal static void TranslateInput(IInputElement targetElement, InputEventArgs
 
             // Determine UIElement/ContentElement/Neither type
             DependencyObject targetElementAsDO = targetElement as DependencyObject;
-            bool isUIElement = InputElement.IsUIElement(targetElementAsDO);
-            bool isContentElement = !isUIElement && InputElement.IsContentElement(targetElementAsDO);
-            bool isUIElement3D = !isUIElement && !isContentElement && InputElement.IsUIElement3D(targetElementAsDO);
+            // todo patternmatching: Would an as cast and null checks be faster here instead of casting multiple times?
```
vishalmsft, line 355: @batzen If you are suggesting, something like?
localInputBindings = ((targetElement as UIElement) ?? (targetElement as ContentElement) ?? (targetElement 
 as UIElement3D))?.InputBindingsInternal;

"is" actually casts the object. Above line will just save redundant cast.

batzen: I was meant for the casts later inside the ifs. Getting the InputBindings here would be too early.
Maybe it's also ok as it is, as any change to the flags would make the code more difficult to read.
Just wanted to leave so todos for things like this.
If no one thinks we should have a further look at this place i can just delete the todo. ;-)

vishalmsft: Agreed. It won't work, unless inherited from the same interface. It will need more refactoring, than just converting that to one line.

src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Input/MouseDevice.cs
```diff
@@ -739,15 +739,15 @@ private object ReevaluateCaptureAsync(object arg)
             // First, check things like IsEnabled, IsVisible, etc. on a
             // UIElement vs. ContentElement basis.
             //
-            if (InputElement.IsUIElement(dependencyObject))
+            if (dependencyObject is UIElement)
             {
                 killCapture = !ValidateUIElementForCapture((UIElement)_mouseCapture);
             }
```
vishalmsft, line 745:
```suggestion
            if (dependencyObject is UIElement uie)
            {
                killCapture = !ValidateUIElementForCapture(uie);
            }
```

Please note, _mouseCapture is the dependencyObject at line#736

src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Input/MouseDevice.cs
```diff
@@ -739,15 +739,15 @@ private object ReevaluateCaptureAsync(object arg)
             // First, check things like IsEnabled, IsVisible, etc. on a
             // UIElement vs. ContentElement basis.
             //
-            if (InputElement.IsUIElement(dependencyObject))
+            if (dependencyObject is UIElement)
```
vishalmsft, line 742:
```suggestion
            if (dependencyObject is UIElement uie)
            {
                killCapture = !ValidateUIElementForCapture(uie);
            }
```
Please note, _mouseCapture is the dependencyObject at line#736
Simlar changes can be applied for below three cases.

src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Input/MouseDevice.cs
```diff
@@ -1527,9 +1527,13 @@ private void PreNotifyInput(object sender, NotifyInputEventArgs e)
...
                                             // find the containing element.
                                             if (!InputElement.IsValid(mouseOver))
+                                            {
                                                 mouseOver = InputElement.GetContainingInputElement(mouseOver as DependencyObject);
+                                            }
                                             if ((rawMouseOver != null) && !InputElement.IsValid(rawMouseOver))
+                                            {
                                                 rawMouseOver = InputElement.GetContainingInputElement(rawMouseOver as DependencyObject);
+                                            }
                                         }
```
vishalmsft, line 1537: Can you provide issue or test method, that cover this piece of code.

src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Input/Stylus/Pointer/PointerLogic.cs
```diff
@@ -774,18 +770,19 @@ private object ReevaluateCaptureAsync(object arg)
             // First, check things like IsEnabled, IsVisible, etc. on a
             // UIElement vs. ContentElement basis.
             //
-            if (InputElement.IsUIElement(dependencyObject))
+            if (dependencyObject is UIElement)
```
vishalmsft, line 773:
```suggestion
             if (dependencyObject is UIElement uie)
            {
                killCapture = !ValidateUIElementForCapture(uie);
            }
```

src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Input/Stylus/Wisp/WispLogic.cs
```diff
@@ -2409,18 +2413,19 @@ private object ReevaluateCaptureAsync(object arg)
             // First, check things like IsEnabled, IsVisible, etc. on a
             // UIElement vs. ContentElement basis.
             //
-            if (InputElement.IsUIElement(dependencyObject))
+            if (dependencyObject is UIElement)
```
vishalmsft, line 2416:
```suggestion
          if (dependencyObject is UIElement uie)
            {
                killCapture = !ValidateUIElementForCapture(uie);
            }
```

src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Input/MouseDevice.cs
batzen: I don't understand that request, as i only changed the casts.

src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/DescendentsWalker.cs
```diff
@@ -114,18 +112,13 @@ private void IterateChildren(DependencyObject d)
...
-                else
+                else if (d is Visual3D v3D)
                 {
-                    Visual3D v3D = d as Visual3D;
-                    if (v3D != null)
-                    {
-                        WalkVisualChildren(v3D);
-                    }
+                    WalkVisualChildren(v3D);
                 }
```
dipeshmsft, line 122: Is this change fine ?

src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Generated/FrameworkContentElement.cs
```diff
@@ -1227,6 +1227,7 @@ internal bool PotentiallyHasMentees
         #region Internal Fields
 
         // Optimization, to avoid calling FromSystemType too often
+        // todo BSC: type
         internal new static DependencyObjectType DType = DependencyObjectType.FromSystemTypeInternal(typeof(FrameworkContentElement));
```
dipeshmsft, line 1231: What does this mean ??

src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Input/Stylus/Wisp/WispStylusDevice.cs
```diff
@@ -418,52 +418,54 @@ private void UpdateOverProperty(IInputElement oldOver, IInputElement newOver)
...
+                else if (o is UIElement3D uie3D)
                 {
-                    ((UIElement3D)o).IsEnabledChanged += _overIsEnabledChangedEventHandler;
-                    ((UIElement3D)o).IsVisibleChanged += _overIsVisibleChangedEventHandler;
-                    ((UIElement3D)o).IsHitTestVisibleChanged += _overIsHitTestVisibleChangedEventHandler;
+                    uie3D.IsEnabledChanged += _overIsEnabledChangedEventHandler;
+                    uie3D.IsVisibleChanged += _overIsVisibleChangedEventHandler;
+                    uie3D.IsHitTestVisibleChanged += _overIsHitTestVisibleChangedEventHandler;
                 }
+                // todo patternmatching: Should we throw here?
```
dipeshmsft, line 468: What  to do about this ??

src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Generated/FrameworkContentElement.cs
batzen: BSC is short for "Bastian Schmidt".
I left the todo there because i couldn't find any usages of the `DType` fields.
Should it be removed, or is it used through reflection somewhere?

src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/DescendentsWalker.cs
batzen: It should be, as it doesn't change the original logic.

src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Input/Stylus/Wisp/WispStylusDevice.cs
dipeshmsft: Yeah, @batzen, sorry for the open ended comment. It was actually for the team. We have had a discussion regarding this and it would be better to throw an error here. But I am still not sure what error message should be thrown here. Previously, we would have got a Cast Error Exception , should we throw the same error now ??

batzen: Had a second look and it wouldn't even always be an InvalidCastException but a NullRef one.
The message in the Wisp* and PointerLogic classes would be slightly different to the one in PresentationSource and KeyboardDevice though, as the code in Wisp* and PointerLogic would, previously, also throw if the object wasn't a DependencyObject.

I guess it would be better to add some helper to throw an exception with a dedicated message for those places.
Something that states which types would have been support and which type was received might be the most helpful.

ThomasGoulet73: My suggestion would be to throw an exception. To me, it looks like an InvalidCastException could happen if oldOver is DependencyObject and is not UIElement, ContentElement or UIElement3D while a NullReferenceException could happen if oldOver is not DependencyObject. Since an exception would already be thrown, throwing a proper exception is probably the right thing to do.

Also, adding an else that should never happen should not impact perf here (The method is probably not inlined already because of its size). You could use copy this throw here:
https://github.com/dotnet/wpf/blob/6e053f99e31de32162f8e6a97f42a04742ed821e/src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Input/Stylus/Wisp/WispStylusDevice.cs#L280

batzen: That exception is meant to be thrown when InputElement.IsValid returns false and the input element is not null, so it means something else than the new cases have to express.

ThomasGoulet73: InputElement.IsValid does the same validation as the pattern matching code here but yes, you could add another throw when the input element is not a DependencyObject.

src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Generated/FrameworkContentElement.cs
dipeshmsft: With the changes in your PR, this type is no more used. So I think, we can go ahead and remove the `DType` field.

src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Input/Stylus/Wisp/WispStylusDevice.cs
dipeshmsft: In order to unblock this PR, we can go ahead with the `SRID.Invalid_IInputElement` exception for now and create a new issue to introduce a new exception type ( public API change, would require a complete review process ).

The reason for throwing an error is to ensure that we don't swallow exceptions which would have been the case if we don't throw any exception here.

dipeshmsft:
@batzen, once the above things are resolved, we can queue this PR for test pass.

batzen:
@dipeshmsft Everything should be addressed now.

dipeshmsft (approved):

dipeshmsft:

PR #4968, opened 2021-08-02
Remove Obsolete APIs

singhashish-wpf:
## Description

WPF is using obsolete APIs. These need to be replaced for 6.0.

WebRequest.WebRequest()
WebRequest.Create(Uri)
Uri.EscapeUriString(string)

## Customer Impact

Warnings related to various Obsolete APIs

## Regression

No

## Testing
Build
CI/CD


src/Microsoft.DotNet.Wpf/src/PresentationUI/MS/Internal/documents/DocumentSignatureManager.cs
```diff
@@ -506,9 +506,7 @@ internal void ShowSigningDialog(IntPtr parentWindow)
                                             // Get[Algorithm]PrivateKey methods would always have returned the private key if the PrivateKey property would
                                             // But Get[Algorithm]PrivateKey methods never throw but returns null in case of error during cryptographic operations
                                             // But we want exception to be thrown when an error occurs during a cryptographic operation so that we can redisplay the certificate picker
-                                            #pragma warning disable SYSLIB0028
-                                            AsymmetricAlgorithm testKey = x509Certificate2.PrivateKey;
-                                            #pragma warning restore SYSLIB0028
+                                            AsymmetricAlgorithm testKey = x509Certificate2.GetRSAPrivateKey();
```
vishalmsft, line 509: @singhashish-wpf As per comments above, the caller is expecting exception, if cryptographic operations fails. GetRSAPrivateKey won't do that.
Please cross check the consequence of not having the exceptions.

src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Markup/XamlTypeMapper.cs
```diff
@@ -245,7 +247,7 @@ public XamlTypeMapper(string[] assemblyNames)
 
 #endregion Methods
 
-#region Properties
+        #region Properties
```
vishalmsft, line 250: @singhashish-wpf  seems formatting moved.

src/Microsoft.DotNet.Wpf/src/PresentationFramework/MS/Internal/AppModel/AssemblyFilter.cs
```diff
@@ -47,28 +47,29 @@ internal void FilterCallback(Object sender, AssemblyLoadEventArgs args)
                 // other wise also we cannot do any processing on a reflection only assembly aside from reflection based actions
                 if (!a.ReflectionOnly)
```
vishalmsft, line 48: @singhashish-wpf complete block inside "if" is commented out. Shall this "if" go away? even everything inside "lock" may not be needed, if following statement do not perform an action.
Assembly a = args.LoadedAssembly;

singhashish-wpf: Updaetd in commit https://github.com/dotnet/wpf/pull/4968/commits/1e55e8532fffbff1dcb526e91fbbca6dd324c5ab

src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Markup/XamlTypeMapper.cs
singhashish-wpf: updated in commit https://github.com/dotnet/wpf/pull/4968/commits/1e55e8532fffbff1dcb526e91fbbca6dd324c5ab

src/Microsoft.DotNet.Wpf/src/PresentationUI/MS/Internal/documents/DocumentSignatureManager.cs
singhashish-wpf: Throwing exception if the privatekey is null.
Commit : https://github.com/dotnet/wpf/pull/4968/commits/c43cd8d2b5f9e1f73c8122a06a074138cff430d0

ThomasGoulet73:
There are still some usage of ReliabilityContractAttribute in C++/CLI assemblies: https://github.com/dotnet/wpf/search?l=C%2B%2B&q=4950. I think those should be removed as well if they are removed in C# code. They were probably not found at first because they do not use the same warning code. C++/CLI raise the warning with the code 4950 for usage of APIs with ObsoleteAttribute.

ryalanms (approved):

src/Microsoft.DotNet.Wpf/src/System.Printing/CPP/inc/GDIExporter/precomp.hpp
```diff
@@ -3,7 +3,6 @@
 // See the LICENSE file in the project root for more information.
 
 // Turn off CER warnings: The Constrained Execution Region (CER) feature is not supported.
```
SamBent, line 5: Delete the comment.  It makes no sense without the code you're deleting.

ryalanms:
Please do not merge this until our testing is completed later today.

src/Microsoft.DotNet.Wpf/src/PresentationFramework/MS/Internal/AppModel/AssemblyFilter.cs
```diff
@@ -47,30 +48,31 @@ internal void FilterCallback(Object sender, AssemblyLoadEventArgs args)
...
-                                System.Environment.Exit(-1);
-                            }
-                            finally
-                            {
-                                Debug.Fail("Environment.Exit() failed.");
-                            }
-                        }
-                    }
+
+                     * // check if it is in the Gac , this ensures that we eliminate any non GAC assembly which are of no risk
```
ryalanms, line 52: nit: This could be removed.

src/Microsoft.DotNet.Wpf/src/WindowsBase/MS/Internal/IO/Packaging/XmlDigitalSignatureProcessor.cs
```diff
@@ -874,9 +874,7 @@ private static AsymmetricAlgorithm GetPrivateKey(X509Certificate2 cert)
             // Get[Algorithm]PrivateKey methods would always have returned the private key if the PrivateKey property would
             // But Get[Algorithm]PrivateKey methods never throw but returns null in case of error during cryptographic operations
             // But we want exception to be thrown when an error occurs during a cryptographic operation so that we can revert the changes
-            #pragma warning disable SYSLIB0028
-            return cert.PrivateKey;
-            #pragma warning restore SYSLIB0028
+            throw new CryptographicException("PrivateKey is not retrievable.");
```
SamBent, line 877: You are now throwing an exception if all three of the Get*Key methods return null.
1. Does `cert.PrivateKey` also throw in that case?   I couldn't find a doc that guaranteed that.
2. Does it throw the same exception, with the same message?   If not, this is a breaking change.
3. Does the message need to be localized?   If it's visible to users (i.e. not caught by WPF code), then yes.
4. The comment at the head of the method says the keys returned by PrivateKey are different from Get*Key.  If answer to (1) is 'no', any code that acts on that difference is now obsolete.  Is there such code?  (Not visible in this review.)

src/Microsoft.DotNet.Wpf/src/PresentationUI/MS/Internal/documents/DocumentSignatureManager.cs
```diff
@@ -506,9 +506,7 @@ internal void ShowSigningDialog(IntPtr parentWindow)
                                             // Get[Algorithm]PrivateKey methods would always have returned the private key if the PrivateKey property would
                                             // But Get[Algorithm]PrivateKey methods never throw but returns null in case of error during cryptographic operations
                                             // But we want exception to be thrown when an error occurs during a cryptographic operation so that we can redisplay the certificate picker
-                                            #pragma warning disable SYSLIB0028
-                                            AsymmetricAlgorithm testKey = x509Certificate2.PrivateKey;
-                                            #pragma warning restore SYSLIB0028
+                                            throw new CryptographicException("PrivateKey is not retrievable.");
```
SamBent, line 509: Same concerns as my remark in XmlDigitalSignatureProcessor.cs.  Except this exception is caught 8 lines down, so it's not visible to user.

ryalanms:
DRTs and Microsuite tests showed no regressions vs the most recent changes in main.

src/Microsoft.DotNet.Wpf/src/WindowsBase/MS/Internal/IO/Packaging/XmlDigitalSignatureProcessor.cs
ryalanms: https://docs.microsoft.com/en-us/dotnet/api/system.security.cryptography.x509certificates.x509certificate2.privatekey?view=net-5.0. 

For 1, there is no specific language about a null key throwing, but it does say an invalid RSA, DSA, or 'unreadable' key will throw CryptographicException.  One indication that it throws if everything is null is that the callers like PackageDigitalSignatureManager.Sign catch but do not handle the null case.  

For 2/3, we can leave the 509certificate2.privatekey replacement for RC2 and check in everything else.  Or if 509certificate2.privatekey can only throw a single error code in CryptographicException when all three Get*Key methods return null, then it can be replaced now.  

For 4, I read the comment as the PrivateKey property was used as a replacement for the Get*PrivateKey methods to intentionally cause an exception, rather than returning null.

src/Microsoft.DotNet.Wpf/src/PresentationUI/MS/Internal/documents/DocumentSignatureManager.cs
ThomasGoulet73: The try/catch could probably be removed and just set requestAgain to true instead of throwing since none of the Get*Key methods are documented to throw CryptographicException.

src/Microsoft.DotNet.Wpf/src/WindowsBase/MS/Internal/IO/Packaging/XmlDigitalSignatureProcessor.cs
SamBent: That doc doesn't mention ECDsa keys at all, so I'm not sure whether PrivateKey can be an ECDsa key at all, and if so whether c.PrivateKey throws.  The doc could be literally true, or simply out-of-date since ECDsa was added.  Have you or Ashish asked the crypto team?

In any case, if we called c.PrivateKey to cause a user-visible exception, we are now throwing that exception ourselves and for compat we must throw the same exception (we do, if doc is correct) with the same message (unlikely), and that message needs localization.  Ask the crypto team what message they use, and use SR to localize it.  Plenty of examples in our code.

SamBent (approved):
Approving the version without crypto changes.

Issue #4990, opened 2021-08-03
appcrash in clr_failfast_80131623_presentationframework.dll!ms.internal.controls.webbrowserevent.documentcomplete


SamBent:
[Port of a servicing fix in .NET 4.7-4.8]

 **Problem description:**
 WPF applications intermittently crashes in various parts of that application with an Invariant.Assert/Failfast when viewing documents with the WPF WebBrowser control in DocumentComplete.
The application is clearing IE by navigating to about:Blank as recommended by the IE team.
However, rarely, the DocumentComplete for the previous document occurs after navigating to about:blank, at which point WPF crashes the process.

 **Actual behavior:** 
 ```
Description: The application requested process termination through System.Environment.FailFast(string message).
Message: Unrecoverable system error.
Stack:
at System.Environment.FailFast(System.String)
at MS.Internal.Invariant.FailFast(System.String, System.String)
at MS.Internal.Controls.WebBrowserEvent.DocumentComplete(System.Object, System.Object ByRef)
at System.RuntimeMethodHandle.InvokeMethod(System.Object, System.Object[], System.Signature, Boolean)
```

SamBent:
Fix ported to 3.1.19, 5.0.10, 6.0.

PR #4993, opened 2021-08-03
Port WebBrowser completion events fix from 4.8

SamBent:
Addresses Issue #4990

## Description

When WebBrowser navigates to about:blank, it cancels outstanding navigations.  However it may still receive completion events for those navigations, due to race conditions between the browser process and the WPF app.  When this happens, ignore the completion event (instead of failing an assert and crashing).

## Customer Impact

Intermittent crashes.

## Regression

No.

## Testing

Standard regressions testing.

## Risk

Low.  Port of .NETFx servicing fix released earlier this year.

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/PresentationFramework/MS/Internal/Controls/WebBrowserEvent.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationFramework/MS/Internal/Controls/WebBrowserEvent.cs
@@ -261,10 +261,12 @@ namespace MS.Internal.Controls
 
         public void CommandStateChange(long command, bool enable)
         {
-            if (command == NativeMethods.CSC_NAVIGATEBACK) {
+            if (command == NativeMethods.CSC_NAVIGATEBACK)
+            {
                 _parent._canGoBack = enable;
             }
-            else if (command == NativeMethods.CSC_NAVIGATEFORWARD) {
+            else if (command == NativeMethods.CSC_NAVIGATEFORWARD)
+            {
                 _parent._canGoForward = enable;
             }
         }
--- a/src/Microsoft.DotNet.Wpf/src/PresentationFramework/MS/Internal/Controls/WebBrowserEvent.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationFramework/MS/Internal/Controls/WebBrowserEvent.cs
@@ -24,8 +24,8 @@ using MS.Win32;
 using MS.Internal.AppModel;
 using MS.Internal.Interop;
 
-//In order to avoid generating warnings about unknown message numbers and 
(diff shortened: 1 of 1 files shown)


ryalanms (approved):

src/Microsoft.DotNet.Wpf/src/PresentationFramework/MS/Internal/Controls/WebBrowserEvent.cs
```diff
@@ -248,14 +248,23 @@ public void DocumentComplete(object pDisp, ref object url)
             }
         }
 
+        // if we are navigating to "about:blank", ignore completion events
+        // for any other url.  These can happen despite our attempt to cancel
+        // outstanding navigations (see WebBrowser.DoNavigate), due to race
+        // conditions between the browser and us.
+        private bool ShouldIgnoreCompletionEvent(ref object url)
```
batzen, line 255: Do we really need ref here?

src/Microsoft.DotNet.Wpf/src/PresentationFramework/MS/Internal/Controls/WebBrowserEvent.cs
```diff
@@ -248,14 +248,23 @@ public void DocumentComplete(object pDisp, ref object url)
...
+            string urlString = url as string;
+            return (_parent.NavigatingToAboutBlank &&
+                    String.Compare(urlString, WebBrowser.AboutBlankUriString, StringComparison.OrdinalIgnoreCase) != 0);
+        }
+
         public void CommandStateChange(long command, bool enable)
         {
-            if (command == NativeMethods.CSC_NAVIGATEBACK)
-            {
+            if (command == NativeMethods.CSC_NAVIGATEBACK) {
```
batzen, line 264: Shouldn't the line be wrapped before the curly brace?

src/Microsoft.DotNet.Wpf/src/PresentationFramework/MS/Internal/Controls/WebBrowserEvent.cs
SamBent: Maybe not, but copying the conventions of existing code.  There's no gain in changing it.

src/Microsoft.DotNet.Wpf/src/PresentationFramework/MS/Internal/Controls/WebBrowserEvent.cs
SamBent: Yes.  This got changed during the original port from .NET 4.8 to .NET Core 3.0, and accidentally changed back again when porting this fix, by a tool (or tool operator) who didn't notice.

Issue #4987, opened 2021-08-03
[dotnet-sdk-6.0.100-rc.1.21401.2] System.Windows.Markup.XamlParseException when reading resource file in ProyectoFinalAplicada1


Junjun-zhao:
**Application Name**: ProyectoFinalAplicada1
**OS**: Windows 10 RS5
**CPU:** X64
**.NET Build Number:** dotnet-sdk-6.0.100-rc.1.21401.2

**Verify Scenarios:**
1). Windows10 rs5 x64 + dotnet-sdk-6.0.100-rc.1.21401.2: Fail
2). Windows10 rs5 x64 + dotnet-sdk-6.0.100-rc.1.21402.7: Fail
3). Windows10 rs5 x64 +dotnet-sdk-6.0.100-preview.7.21379.14 : Pass
4).Windows10 rs5 x64 + dotnet-sdk-3.1.411: Pass

**App and App Source checking at :** https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1364942

**Repro steps:**

This is 3.1 app, but we are running it on 6.0, machine only have 6.0 installed, and DOTNET_ROLL_FORWARD_ON_NO_CANDIDATE_FX=2, so it runs on 6.0

1.Launch application “ProyectoFinalAplicada1.exe”.  

**Expected Result:**
The app is launched successfully.

**Actual Result:**
Fail to launch the app, app is crashed.

**Exception Details:**
```
System.Windows.Markup.XamlParseException
  HResult=0x80131501
  Message='Provide value on 'System.Windows.Baml2006.TypeConverterMarkupExtension' threw an exception.' Line number '6' and line position '9'.
  Source=PresentationFramework
  StackTrace:
   at System.Windows.Markup.XamlReader.RewrapException(Exception e, IXamlLineInfo lineInfo, Uri baseUri)
   at System.Windows.Markup.WpfXamlLoader.Load(XamlReader xamlReader, IXamlObjectWriterFactory writerFactory, Boolean skipJournaledProperties, Object rootObject, XamlObjectWriterSettings settings, Uri baseUri)
   at System.Windows.Markup.WpfXamlLoader.LoadBaml(XamlReader xamlReader, Boolean skipJournaledProperties, Object rootObject, XamlAccessLevel accessLevel, Uri baseUri)
   at System.Windows.Markup.XamlReader.LoadBaml(Stream stream, ParserContext parserContext, Object parent, Boolean closeStream)
[...]

PR #5004, opened 2021-08-04
System.Windows.Markup.XamlParseException when reading resource file 

ryalanms:
The failure looks like an unintended consequence of #4799. The path incorrectly contains the filename of the application assembly.  Reverting the change for now, since we are so close to snapping the release candidate for .NET 6.0.  

Fixes issue https://github.com/dotnet/wpf/issues/4987.


SamBent (approved):

ryalanms:
The scenario is an application built for .NET 3.1 that is running on a machine with only .NET 6.0 installed. 

To repro, set the environment variable `DOTNET_ROLL_FORWARD_ON_NO_CANDIDATE_FX=2' or add the equivalent setting to the application's runtime.json file:

{
  "runtimeOptions": {
    "tfm": "netcoreapp3.1",
    "framework": {
      "name": "Microsoft.WindowsDesktop.App",
      "version": "3.1.0"
    },
    "rollForwardOnNoCandidateFx" : 2 
  }
}

lindexi:
@ryalanms Sorry

ryalanms:
> @ryalanms Sorry

Thank you for all your great work, @lindexi.  We can get a modified version of this back in when we branch for 7.0.

PR #5059, opened 2021-08-11
[release/3.1] ILDasm build break

SamBent:
The release/3.1 build breaks on machines that don't have an .NET Framework SDK for version 4.6.2 or lower.

This fixes the break by asking MSBuild to look for ILDasm.exe in all versions it knows about.

### DIFF
--- a/eng/WpfArcadeSdk/tools/InjectModuleInitializer.targets
+++ b/eng/WpfArcadeSdk/tools/InjectModuleInitializer.targets
@@ -62,8 +62,8 @@
           try
           {
             var bitness = (DotNetFrameworkArchitecture)Enum.Parse(typeof(DotNetFrameworkArchitecture), DotNetBitness);
-            ILAsm = ToolLocationHelper.GetPathToDotNetFrameworkFile("ILAsm.exe", TargetDotNetFrameworkVersion.VersionLatest, bitness);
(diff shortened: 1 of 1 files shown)


ryalanms:
LGTM.  It is worth noting that this change was required because `TargetDotNetFrameworkVersion.VersionLatest` is returning the wrong value. 

https://docs.microsoft.com/en-us/dotnet/api/microsoft.build.utilities.targetdotnetframeworkversion?view=msbuild-16-netcore

https://github.com/dotnet/msbuild/issues/4354

ryalanms (approved):

SamBent:
Reading the fix for dotnet/msbuild#4354, you learn that "changing VersionLatest is a breaking change - can only be done in a major release", and that "Latest only works with MSBuild 15.1 or above".  Fortunately we require 16.1, so we can use Latest.  But we can't use VersionLatest until we require MSBuild 17.

SamBent:
This break was fixed in 5.0 by a much more elaborate change, see #2298.  It uses a fixed copy of the IL tools, copied from 4.7.2 into blob storage.  There are concerns about using the 4.8 ILDasm (which this change would do on machine with 4.8 SDK), but they were about round-tripping problems that would arise at build time.

I've built release/3.1 with this simple fix, confirming that it used the 4.8 IL tools, with no build errors.  So I think we're OK.  I'll be watching the test results for problems.

SamBent:
I'm seeing no ill effects at runtime.  As expected - this is a compile-time problem.  So I'm checking it in.

PR #5085, opened 2021-08-17
Fix Popup Focus State Not Reacting with WebView2

singhashish-wpf:
Fixes Issue 1366812
https://github.com/MicrosoftEdge/WebView2Feedback/issues/1134

## Description

If we create a popup without losing focus of the WebView2 element (e.g. by using a context menu, rather than clicking a focusable button), the popup looks and feels as if it has keyboard focus, but all keyboard inputs go to the WebView2 instead.

This does not happen if focus is lost by WebView by clicking some focusable WPF element first (e.g. clicking the button, rather than right-clicking it).

The problem is with out-of-proc, In this fix we are comparing the process IDs.
If process IDs don’t match, call the TrySetFocus() function inside the if condition for WS_EX_NOACTIVATE.

## Customer Impact
Unable to write inside the popup text box.

## Regression
No

## Testing
Build and verify the fix.

## Risk
Extra processing for fetching the pid.

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/InterOp/HwndKeyboardInputProvider.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/InterOp/HwndKeyboardInputProvider.cs
@@ -117,6 +117,13 @@ namespace System.Windows.Interop
                         // window handle for windows attached to the calling
                         // thread's queue.
                         //
+                        int thisPid = 0;
+                        int childPid = 0;
+                        UnsafeNativeMethods.GetWindowThreadProcessId(new HandleRef(null, focus), out childPid);
+                        UnsafeNativeMethods.GetWindowThreadProcessId(thisWindow, out thisPid);
+
+                        if (childPid != thisPid)
+                            UnsafeNativeMethods.TrySetFocus(thisWindow);
                         result = focus != IntPtr.Zero;
                     }
                 }


src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/InterOp/HwndKeyboardInputProvider.cs
```diff
@@ -117,6 +117,13 @@ bool IKeyboardInputProvider.AcquireFocus(bool checkOnly)
                         // window handle for windows attached to the calling
                         // thread's queue.
                         //
+                        int thisPid = 0;
+                        int childPid = 0;
+                        UnsafeNativeMethods.GetWindowThreadProcessId(new HandleRef(null, focus), out childPid);
+                        UnsafeNativeMethods.GetWindowThreadProcessId(thisWindow, out thisPid);
+
+                        if (childPid != thisPid)
+                            UnsafeNativeMethods.TrySetFocus(thisWindow);
```
SamBent, line 126: This is exactly what the comment at line 85 says we should _not_ do.  Which suggests that this will break something (even though it fixes WebView2).  This is not the right fix.

Lines 110-118 explain how VS participates in cross-thread focus behavior.  It sounds like WebView2 is doing some, but not all, of that work:  they are (presumably) calling AttachThreadInput so that GetFocus returns their window, but not reposting kb messages to our thread.   So the right fix might belong to WebView2.

vishalmsft: @SamBent The comments also says, "This is typically for the menu Popup". Other than Menu-popup, should current control always grab the focus, if AcquireFocus is called?
For multi-process application, where we cannot trust the implementation by others to do the right thing, this may become appropriate to force move the focus, if child-process isn't behaving the way we expect it to.

Issue #4631, opened 2021-06-04
Inconsistent UIA IsOffscreen and ClickablePoint properties for WPF ListViewItems that are scrolled out of view


SamBent:
* .NET Core Version: (e.g. 3.0 Preview1, or daily build number, use `dotnet --info`) 3.0+
* Windows version: (`winver`) all
* Does the bug reproduce also in WPF for .NET Framework 4.8?: Yes
* Is this bug related specifically to tooling in Visual Studio (e.g. XAML Designer, Code editing, etc...)? no
* Security issues and bugs should be reported privately, learn more via our [responsible disclosure guidelines](https://github.com/dotnet/wpf/blob/master/README.md#reporting-security-issues-and-security-bugs).

 
 
Copied from internal .NET Framework bug [1161493](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1161493), which contains additional information.

 **Problem description:**
1. Create a new WPF .NET Framework 4.8 app.
2. Add the following XAML:

        <ListView Height="60">
            <ListViewItem Content="One" />
            <ListViewItem Content="Two" />
            <ListViewItem Content="Three" />
            <ListViewItem Content="Four" />
            <ListViewItem Content="Five" />
            <ListViewItem Content="Six" />
            <ListViewItem Content="Seven" />
            <ListViewItem Content="Eight" />
            <ListViewItem Content="Nine" />
            <ListViewItem Content="Ten" />
        </ListView>

3. Run the app, verify that the ListView can only present a subset of the full item set, and scroll through the list such that all items have been viewed.

4. Point the Accessibility Insights for Windows (AIWin) tool at the list, and review the failures that the tool reports.

Result: AIWin reports a failure of "An element's IsOffScreen property must be false when its clickable point is on screen" for all items scrolled out of view. 

Expected: AIWin doesn't report failures for the default WPF experience. 

This also happens for TreeView and DataGrid.

SamBent:
Fixed in 6.0-rc1.

PR #5130, opened 2021-08-20
[release/6.0-rc1] Accessibility:  Fix ClickablePoint property when IsOffscreen is true

ryalanms:
Fixes Issue #4631

## Description

The automation ClickablePoint property shouldn't be an onscreen point, when the IsOffscreen property is true.

This is a port of @SamBent's Accessibility fix from main.  

## Customer Impact

Accessibility Insights for Windows flags this as an accessibility error.

## Testing

The test cases that claimed to check this didn't work, for 3 independent reasons.  I've repaired them in a companion PR in the test repo.  That PR is _much_ more interesting than this one.

## Risk

Low.   Changing this case to conform to the expected behavior, per UIA docs.  The new return value is the same as we return in other "no clickable point" scenarios.


SamBent (approved):

ryalanms:
Approved by Tactics via email.

PR #5182, opened 2021-08-26
Remove the unnecessary Array code in FontTypeConverter.cs

lindexi:
## Description

We can reduce the allocation of a static array. And it can run faster than traversing the array.

## Testing

Just CI.

## Risk

Low.


anjalisheel-wpf (approved):
I went through the changes, even tested them, it is working as expected, hence approving it.

dipeshmsft:
@lindexi , we have ColorTypeConverter and ImageSourceTypeConverter which have the same pattern, should we make similar changes to those classes, such that we follow the same pattern for all type converters ??

lindexi:
@dipeshmsft Thank you and I fix it in new commit.

dipeshmsft (approved):

PR #5215, opened 2021-08-29
Add the demo to debug.

lindexi:
## Description

How about adding a demo project to help us debug more easily?

## Testing

Just CI.

## Risk

Low.


vishalmsft:
@lindexi The demo is to help with setting up debugging environment faster. 
While debugging, people will be modify the demo to debug various features. 
Do you think it will cause unnecessary commits of files?

Even if we use demo, do you think it may be good idea to copy the project to output-temp directory, which shall be ignored from commits?

lindexi:
@vishalmsft  Good ideas.

dipeshmsft:
@lindexi Should we go ahead and close this PR ??

lindexi:
@dipeshmsft Yes.

Issue #4501, opened 2021-05-11
Code comments and behavior are inconsistent in Selector.cs


lindexi:
We say that "Returns false if FrameworkElement representing this item has Selector.SelectableProperty set to false", but we just check the object type

https://github.com/dotnet/wpf/blob/1c3be71312540f51d98f769b431ea3cc110844d0/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Controls/Primitives/Selector.cs#L1859-L1872

ryalanms:
Thanks, @lindexi.  Please let WPF triage assign bugs.

ThomasGoulet73:
@ryalanms: @SamBent seems to be the default assignee when creating an issue of type "Feature / API request": https://github.com/dotnet/wpf/blob/main/.github/ISSUE_TEMPLATE/feature_request.md.

dipeshmsft:
Closing the issue as the corresponding PR #5216 is merged.

Issue #5143, opened 2021-08-23
Miss the SelectableProperty property in Selector


lindexi:
The Selector says that the `ItemGetIsSelectable` will check if the `FrameworkElement` representing this item has `Selector.SelectableProperty` set to false. But actually it does nothing with SelectableProperty property. And we can not find the SelectableProperty property in Selector.

Should we add the SelectableProperty property in Selector?

https://github.com/dotnet/wpf/blob/89d172db0b7a192de720c6cfba5e28a1e7d46123/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Controls/Primitives/Selector.cs#L1859-L1872

lindexi:
Duplicate of #4501

PR #5216, opened 2021-08-29
Add the SelectableProperty in Selector.

lindexi:
Fixes Issue #4501

## Description

Adding the SelectableProperty in Selector. And we can set the item not to be selected by SelectableProperty.

See:

- https://github.com/dotnet/wpf/issues/5143
- https://github.com/dotnet/wpf/issues/4501

## Regression

No.

## Testing

Just CI and my [demo](https://github.com/lindexi/lindexi_gd/tree/ef2c5fe539e051953ad6bae969f4161a1f0b85ee/WPFDemo).

## Risk

Low.

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Controls/Primitives/Selector.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Controls/Primitives/Selector.cs
@@ -201,6 +201,33 @@ namespace System.Windows.Controls.Primitives
             return (bool) element.GetValue(IsSelectionActiveProperty);
         }
 
+        public static readonly DependencyProperty SelectableProperty =
+            DependencyProperty.RegisterAttached(
+                "SelectableProperty",
+                typeof(bool),
+                typeof(Selector),
+                new PropertyMetadata(BooleanBoxes.TrueBox));
+
+        public static bool GetSelectable(DependencyObject element)
+        {
+            if (element == null)
+            {
+                throw new ArgumentNullException(nameof(element));
+            }
+
+            return (bool) element.GetValue(SelectableProperty);
+        }
+
+        public static void SetSelectable(DependencyObject element, bool selectable)
+        {
+            if (element == null)
+            {
+                throw new ArgumentNullException(nameof(element));
+            }
+
+            element.SetValue(SelectableProperty, BooleanBoxes.Box(selectable));
+        }
+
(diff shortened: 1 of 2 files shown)


dipeshmsft:
@lindexi After digging around in some old specs, I found the following thing regarding the Selectable attached property : 

> Why is the selectable attached property bad? Well, selection happens over data items, but the attached property works with DPs, so in virtualization cases, an item could be marked as selected when there is no UI and then get un-selected when it’s UI is de-virtualized and the style as the attached property Selectable=false.

Due to this reasoning, the Selectable attached property was removed. Therefore, rather than adding the property, I think it would be better to fix the comment in this case.

Please suggest if you think otherwise.

lindexi:
@dipeshmsft Thank you. I totally agree with you.

The selection happens over data items, and we can control the attached property value by DataTemplate or binding to the DataContext. That means the Data will control the Selectable attached property.

But in virtualization cases, the attached property will be a problem.

dipeshmsft (approved):

src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Controls/Primitives/Selector.cs
```diff
@@ -1857,7 +1857,7 @@ private void ItemSetIsSelected(ItemInfo info, bool value)
         }
 
         /// <summary>
-        /// Returns false if FrameworkElement representing this item has Selector.SelectableProperty set to false.  True otherwise.
+        /// Returns false if FrameworkElement is Separator.  True otherwise.
```
dipeshmsft, line 1860:
```
/// <summary>
/// Returns false if FrameworkElement representing this item is not selectable. True otherwise. If the FrameworkElement is Separator or null, we return False.
/// </summary>
```

How does this sound ?

dipeshmsft (changes requested):

src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Controls/Primitives/Selector.cs
lindexi: Thank you @dipeshmst 

And I fix the comment

dipeshmsft (approved):

dipeshmsft:
Thank you @lindexi for bringing this to our attention and the quick responses to suggestions.

Issue #3660, opened 2020-10-16
Replace obsolete APIs used by WPF


ryalanms:
WPF is using obsolete APIs.  These need to be replaced for 6.0.  

WebRequest.WebRequest()
WebRequest.Create(Uri)
Uri.EscapeUriString(string)

```
D:\wpf7\src\Microsoft.DotNet.Wpf\src\PresentationCore\ref\PresentationCore.cs(11,9): error SYSLIB0014: 'WebRequest.WebR
equest()' is obsolete: 'Use HttpClient instead.' [D:\wpf7\src\Microsoft.DotNet.Wpf\src\PresentationCore\ref\Presentatio
nCore-ref.csproj]

D:\wpf7\src\Microsoft.DotNet.Wpf\src\PresentationCore\MS\Internal\WpfWebRequestHelper.cs(76,30): error SYSLIB0014: 'Web
Request.Create(Uri)' is obsolete: 'Use HttpClient instead.' [D:\wpf7\src\Microsoft.DotNet.Wpf\src\PresentationCore\Pres
entationCore.csproj]

D:\wpf7\src\Microsoft.DotNet.Wpf\src\ReachFramework\PrintConfig\FallbackPTProvider.cs(403,29): error SYSLIB0013: 'Uri.E
scapeUriString(string)' is obsolete: 'Uri.EscapeUriString can corrupt the Uri string in some cases. Consider using Uri.
EscapeDataString for query string components instead.' [D:\wpf7\src\Microsoft.DotNet.Wpf\src\ReachFramework\ReachFramew
ork.csproj]
```

ryalanms:
https://docs.microsoft.com/en-us/dotnet/core/compatibility/core-libraries/6.0/obsolete-apis-with-custom-diagnostics

@jeffhandley:  WPF has some uses of WebRequest and HttpWebRequest that will be very difficult to remove.  Is it possible for WPF to continue using them or will they stop working for .NET 6.0?

jeffhandley:
Those APIs will continue to function in .NET 6; you can suppress the warnings if you need to continue using them.

https://docs.microsoft.com/en-us/dotnet/fundamentals/syslib-diagnostics/syslib0014#workarounds

ryalanms:
@singhashish-wpf:  Let's defer the WebRequest and HttpWebRequest replacements to .NET 7.0.  Please leave the existing suppression for SYSLIB0014 in place.

ryalanms:
[...]

PR #5234, opened 2021-09-01
Replace Obsolete APIs

singhashish-wpf:
Fixes Issue #3660 
Main PR #4968

## Description
SYSLIB0028(https://docs.microsoft.com/en-us/dotnet/fundamentals/syslib-diagnostics/syslib0028)
X509Certificate2.PrivateKey is obsolete, Need to throw an exception when the private key can't be retrieved. 

## Regression
No

## Testing
Build, CI/CD, Locally tested DRTs/Microsuite.

### DIFF
--- a/Directory.Build.targets
+++ b/Directory.Build.targets
@@ -11,7 +11,7 @@
        https://github.com/dotnet/wpf/issues/5224 -->
   <Target Name="_RemoveFrameworkReferenceAnalyzers" AfterTargets="ResolveTargetingPackAssets">
     <ItemGroup>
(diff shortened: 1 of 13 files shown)


ryalanms:
Does this match the exception behavior of the replaced call?

singhashish-wpf:
@ryalanms 
The replaced call (cert.privateKey which is obsolete now) was able to throw multiple exceptions (CryptographicException, ArgumentNullException, ArgumentException). 
ArgumentException is being thrown in this case as the same is being thrown in an analogous situation in the same file [here](https://github.com/dotnet/wpf/blob/abc527991a2c5bc402502b779715313e5d7f8ffe/src/Microsoft.DotNet.Wpf/src/WindowsBase/MS/Internal/IO/Packaging/XmlDigitalSignatureProcessor.cs#L1088)

The exception behavior couldn't be reproduced locally after creating a certificate with null key usng makecert tool.

Issue #5226, opened 2021-08-31
Enabling SingleFile analyzer flags different pieces of code as dangerous


tlakollo:
**Problem description:**
After enabling the single file analyzer in the wpf repository there is several pieces of code that are flagged as dangerous if used in a single file configuration. The following are the problems found in the code along with the effect they would have if deployed as single file.

- [ ] [IncrementalCompileAnalyzer.cs](https://github.com/dotnet/wpf/blob/3390df7638a3fab1c23373b7cc9904677e65be2b/src/Microsoft.DotNet.Wpf/src/PresentationBuildTasks/MS/Internal/Tasks/IncrementalCompileAnalyzer.cs#L88)
Method AnalyzeInputFiles call IsFileChanged with Assembly.Location() as the input file string, internally IsFileChanged will throw if the input file is empty

- [ ] [TaskHelper.cs](https://github.com/dotnet/wpf/blob/3390df7638a3fab1c23373b7cc9904677e65be2b/src/Microsoft.DotNet.Wpf/src/PresentationBuildTasks/MS/Internal/Tasks/TaskHelper.cs#L55)
Method DisplayLogo calls GetVersionInfo with Assembly.Location as the path, later on we check if the path exists which will throw

- [ ] [WinRT.cs](https://github.com/dotnet/wpf/blob/3390df7638a3fab1c23373b7cc9904677e65be2b/src/Microsoft.DotNet.Wpf/src/PresentationFramework/MS/Internal/WindowsRuntime/Generated/WinRT.cs#L114)
Field _currenModuleDirectory does GetDirectoryName on assembly.Location() which throws a System.ArgumentException if parameter contains invalid characters, is empty, or contains only white spaces, in this case will throw since argument is empty.

- [ ] [SerializerDescriptor.cs](https://github.com/dotnet/wpf/blob/3390df7638a3fab1c23373b7cc9904677e65be2b/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Documents/Serialization/SerializerDescriptor.cs#L102)
Method CreateFromFactoryInstance creates a serializer descriptor with assemblyPath set to empty.

- [ ] [AssemblyFilter.cs](https://github.com/dotnet/wpf/blob/3390df7638a3fab1c23373b7cc9904677e65be2b/src/Microsoft.DotNet.Wpf/src/PresentationFramework/MS/Internal/AppModel/AssemblyFilter.cs#L49)
[...]

PR #5325, opened 2021-09-15
Enable Single File Analyzer

tlakollo:
Address some of the issues described in https://github.com/dotnet/wpf/issues/5226

## Description
- Enable the single file analyzer in the WPF repo to prevent a regression on the current number of single file incompatibilities
- Fix some of the issues 
- Suppress some of the patterns that don't have a fix yet

## Customer Impact
Apps might have an undesired behavior when deploying as single-file

## Testing
Build the repo to verify the analyzer doesn't flag more patterns

## Risk
Suppressed patterns can still produce undesired behavior for the user when deploying as single-file

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/PresentationBuildTasks/MS/Internal/Tasks/TaskHelper.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationBuildTasks/MS/Internal/Tasks/TaskHelper.cs
@@ -52,13 +52,10 @@ namespace MS.Internal.Tasks
         // </summary>
         internal static void DisplayLogo(TaskLoggingHelper log, string taskName)
         {
-            string acPath = Assembly.GetExecutingAssembly().Location;
-            FileVersionInfo acFileVersionInfo = FileVersionInfo.GetVersionInfo(acPath);
-
-            string avalonFileVersion = acFileVersionInfo.FileVersion;
+            Version avalonFileVersion = Assembly.GetExecutingAssembly().GetName().Version;
 
             log.LogMessage(MessageImportance.Low,Environment.NewLine);
-            log.LogMessageFromResources(MessageImportance.Low, SRID.TaskLogo, taskName, avalonFileVersion);
+            log.LogMessageFromResources(MessageImportance.Low, SRID.TaskLogo, taskName, avalonFileVersion.ToString());
             log.LogMessageFromResources(MessageImportance.Low, SRID.TaskRight);
             log.LogMessage(MessageImportance.Low, Environment.NewLine);
         }
--- a/src/Microsoft.DotNet.Wpf/src/PresentationFramework/MS/Internal/AppModel/AssemblyFilter.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationFramework/MS/Internal/AppModel/AssemblyFilter.cs
@@ -43,13 +43,12 @@ namespace MS.Internal
         //appends assembly name with file version to generate a unique entry for the assembly lookup process
         private string AssemblyNameWithFileVersion(Assembly a)
         {
-            FileVersionInfo fileVersionInfo;
             StringBuilder sb = new StringBuilder(a.FullName);
 
-            fileVersionInfo = FileVersionInfo.GetVersionInfo(a.Location);
-            if (fileVersionInfo != null && fileVersionInfo.ProductVersion != null)
+            Version assemblyVersion = a.GetName().Version;
+            if (assemblyVersion != null)
             {
-                sb.Append(FILEVERSION_STRING + fileVersionInfo.ProductVersion);
+                sb.Append(FILEVERSION_STRING + assemblyVersion.ToString());
             }
             return ((sb.ToString()).ToLower(System.Globalization.CultureInfo.InvariantCulture)).Trim();
         }
--- a/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Documents/TextRangeSerialization.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Documents/TextRangeSerialization.cs
@@ -572,9 +572,7 @@ namespace System.Windows.Documents
             if (customTextElement)
             {
                 // If the element is not from PresentationFramework, we'll need to serialize a namespace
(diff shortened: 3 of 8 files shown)


tlakollo:
@dotnet/wpf-developers If somebody could review this PR to be able to merge it would be great, its only a couple of small refactorings

src/Microsoft.DotNet.Wpf/src/PresentationBuildTasks/MS/Internal/Tasks/TaskHelper.cs
```diff
@@ -52,13 +52,10 @@ internal static class TaskHelper
         // </summary>
         internal static void DisplayLogo(TaskLoggingHelper log, string taskName)
         {
-            string acPath = Assembly.GetExecutingAssembly().Location;
-            FileVersionInfo acFileVersionInfo = FileVersionInfo.GetVersionInfo(acPath);
-
-            string avalonFileVersion = acFileVersionInfo.FileVersion;
+            Version avalonFileVersion = Assembly.GetExecutingAssembly().GetName().Version;
```
SamBent, line 55: Does Assembly.GetName().Version.ToString() always give the same result as FileVersionInfo.GetVersionInfo(path).FileVersion ?  I doubt it - I couldn't find a guarantee in the docs.

If not, this is a breaking change.  It may not matter here, if the string is only printed out for user's convenience.  But even so, avoid changing the behavior.  Can you use the old code, and only fall back to the new code if you detect the single-file case at runtime?

src/Microsoft.DotNet.Wpf/src/PresentationFramework/MS/Internal/AppModel/AssemblyFilter.cs
```diff
@@ -43,13 +43,12 @@ internal void FilterCallback(Object sender, AssemblyLoadEventArgs args)
         //appends assembly name with file version to generate a unique entry for the assembly lookup process
         private string AssemblyNameWithFileVersion(Assembly a)
         {
-            FileVersionInfo fileVersionInfo;
             StringBuilder sb = new StringBuilder(a.FullName);
 
-            fileVersionInfo = FileVersionInfo.GetVersionInfo(a.Location);
-            if (fileVersionInfo != null && fileVersionInfo.ProductVersion != null)
+            Version assemblyVersion = a.GetName().Version;
```
SamBent, line 48: Another case where the two methods of getting a version string produce different results.  This one matters.  ProductVersion can be different from FileVersion, and presumably there's a reason the old code uses ProductVersion.  The new code breaks that scenario.

Again, please use the old code, with a fallback to the new code if running as single-file.  And find out what breaks by not using ProductVersion here, so that we can warn people not to use single-file if they need that scenario.

src/Microsoft.DotNet.Wpf/src/PresentationBuildTasks/MS/Internal/Tasks/TaskHelper.cs
agocke, line 55: Isn't this the assemby version, not file version? I think we could get the file version of the executing process if that's more appropriate. It probably depends on whether we're using the apphost

src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Documents/TextRangeSerialization.cs
```diff
@@ -572,9 +572,7 @@ private static void WriteStartXamlElement(ITextRange range, ITextPointer textRea
             if (customTextElement)
             {
                 // If the element is not from PresentationFramework, we'll need to serialize a namespace
-                // Will module name always have a '.'?  If so, can remove conditional in assembly assignment below
-                int index = elementTypeStandardized.Module.Name.LastIndexOf('.');
-                string assembly = (index == -1 ? elementTypeStandardized.Module.Name : elementTypeStandardized.Module.Name.Substring(0, index));
+                string assembly = elementTypeStandardized.Assembly.GetName().Name;
```
SamBent, line 575: Another case where the two methods of getting the `assembly` string (probably) return different results.  This one matters a lot too - the serialized result gets de-serialized by some other process, and presumably the old code uses the substring method to make that de-serialization work correctly in all cases.

Again, use the old code, falling back to the new code in the single-file case.  And find out what breaks, so we can warn people.

SamBent (changes requested):

src/Microsoft.DotNet.Wpf/src/PresentationBuildTasks/MS/Internal/Tasks/TaskHelper.cs
tlakollo: Reading the FileVersion API in runtime I notice that Version is the default product version, so yeah in this case we don't want the version. Although I end up realizing that this is in the PresentationBuildTasks so single file will not affect this code path. I end up reverting the changes to leave it as it was

src/Microsoft.DotNet.Wpf/src/PresentationFramework/MS/Internal/AppModel/AssemblyFilter.cs
tlakollo: For this case I end up using the same code path that FileVersionInfo uses in linux to get the ProductVersion, by default is the assembly.version and then if there is an InformationalVersionAttribute in the assembly it gets changed to that value. I'm not fully sure is the same in windows because it uses COM and is difficult to read, but I leave this logic only for the case in which the assembly.location is empty

src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Documents/TextRangeSerialization.cs
tlakollo: In my rudimentary testing the assembly and the trimmed module matched in name
```C#
Assembly? asm = Assembly.GetAssembly(typeof(UnconditionalSuppressMessageAttribute));
foreach (Module module in asm!.Modules)
    Console.WriteLine(module.Name.Substring(0, module.Name.LastIndexOf(".")));
Console.WriteLine(asm.GetName().Name);
#System.Private.CoreLib
#System.Private.CoreLib
```
But yes, I'm not 100% sure that is always the case. I put a fallback scenario that uses the assebly name only in the case in which Module.Name is <Unknown>

PR #5372, opened 2021-09-22
Switch to 1ES servicing pools on release/6.0

jonfortescue:
Tracking  Issue: https://github.com/dotnet/core-eng/issues/14276

~Main PR~ N/A

## Description

We need to move over to 1ES servicing pools for servicing branches. This accomplishes this in release/6.0.

## Customer Impact

N/A

## Regression

None.

## Testing

N/A

## Risk

If tests fail in this PR, it's due to interactive mode not being turned on. That's being tracked here: https://github.com/dotnet/core-eng/issues/14358.


MattGal (approved):
I don't think the casing matters, but if it does somehow you've left it off line 38 of eng/pipeline.yml

jonfortescue:
@MattGal just trying to match the original files -- it doesn't matter, purely a style thing. :)

Issue #5375, opened 2021-09-23
Application not scaled with .NET Core 3.1.19


davidegiacometti:
* .NET Core Version: 3.1.19
* Windows version: 21H1
* Does the bug reproduce also in WPF for .NET Framework 4.8?: don't know

 **Problem description:**
 PowerToys users are experiencing scaling issue after update to .NET Core 3.1.19: https://github.com/microsoft/PowerToys/issues/13221
 
The issue is affecting only PowerToys Run and Image Resizer, but I can't believe this is affecting only PowerToys.

 **Actual behavior:**

 I can also repro this on an XPS13 1920x1080 @150%

<img width="960" alt="3 1 19" src="https://user-images.githubusercontent.com/25966642/134554561-25ecc6cb-4b69-4ea5-a391-acbf52a3d177.png">
 
 **Expected behavior:**

After downgrade to .NET Core 3.1.18

<img width="960" alt="3 1 18" src="https://user-images.githubusercontent.com/25966642/134554583-c5117b2d-cb00-408d-882b-0f4bc7629ff1.png">
 
 **Minimal repro:**
 Issue can be reproduced with PowerToys Run and Image Resizer.

ThomasGoulet73:
I just tried decompiling PresentationCore.dll from  and the module initializer responsible for loading DirectWriteForwarder, which is responsible for setting DPI awareness, is not there. This is a bug with the ILAsm/ILDasm module initializer injection and is probably very high priority.

ryalanms:
Thanks, @davidegiacometti and @ThomasGoulet73.  

https://dotnet.microsoft.com/download/dotnet/thank-you/sdk-3.1.119-windows-x64-binaries

I don't see the missing module initializer when building release/3.1 locally or when examining the PresentationCore.dll in 3.1.119:  dotnet-sdk-3.1.119-win-x64.zip\shared\Microsoft.WindowsDesktop.App\3.1.19.   Does the timestamp match the assembly in the SDK (8/17/2021)?

ThomasGoulet73:
ModuleInitializer is there but <Module> (The class that invokes ModuleInitializer.Initialize and is injected by ILDasm/ILAsm) is not there:

Here's the difference using dotPeek:
3.1.18:
![image](https://user-images.githubusercontent.com/51839772/134575199-b0406d83-e4d9-4414-90cf-06b33def0578.png)

3.1.19:
[...]

PR #5377, opened 2021-09-23
Revert ILDasm/ILAsm tools version to 4.6.2 to fix module initializer in PresentationCore

ryalanms:
Revert ildasm/ilasm tools to 4.6.2.  

Apply @ThomasGoulet73's fix for module initializer injection regression in release/3.1.  This fixed PresentionCore when building locally.  

Fixes https://github.com/dotnet/wpf/issues/5375.

### DIFF
--- a/eng/WpfArcadeSdk/tools/InjectModuleInitializer.targets
+++ b/eng/WpfArcadeSdk/tools/InjectModuleInitializer.targets
@@ -62,8 +62,8 @@
           try
           {
             var bitness = (DotNetFrameworkArchitecture)Enum.Parse(typeof(DotNetFrameworkArchitecture), DotNetBitness);
-            ILAsm = ToolLocationHelper.GetPathToDotNetFrameworkFile("ILAsm.exe", TargetDotNetFrameworkVersion.Latest, bitness);
-            ILDAsm = ToolLocationHelper.GetPathToDotNetFrameworkSdkFile("ILDAsm.exe", TargetDotNetFrameworkVersion.Latest, bitness);
+            ILAsm = ToolLocationHelper.GetPathToDotNetFrameworkFile("ILAsm.exe", TargetDotNetFrameworkVersion.VersionLatest, bitness);
+            ILDAsm = ToolLocationHelper.GetPathToDotNetFrameworkSdkFile("ILDAsm.exe", TargetDotNetFrameworkVersion.VersionLatest, bitness);
           }
           catch(Exception e)
           {


eng/WpfArcadeSdk/tools/InjectModuleInitializer.targets
```diff
@@ -62,8 +62,8 @@
           try
           {
             var bitness = (DotNetFrameworkArchitecture)Enum.Parse(typeof(DotNetFrameworkArchitecture), DotNetBitness);
-            ILAsm = ToolLocationHelper.GetPathToDotNetFrameworkFile("ILAsm.exe", TargetDotNetFrameworkVersion.Latest, bitness);
-            ILDAsm = ToolLocationHelper.GetPathToDotNetFrameworkSdkFile("ILDAsm.exe", TargetDotNetFrameworkVersion.Latest, bitness);
+            ILAsm = ToolLocationHelper.GetPathToDotNetFrameworkFile("ILAsm.exe", TargetDotNetFrameworkVersion.VersionLatest, bitness);
+            ILDAsm = ToolLocationHelper.GetPathToDotNetFrameworkSdkFile("ILDAsm.exe", TargetDotNetFrameworkVersion.VersionLatest, bitness);
```
SamBent, line 66: This reverts the fix for #5059, hence re-introduces that build break.  Which was incomprehensible to me and everyone else for many days.

If you must do this, is there somewhere you can announce that building release/3.1 requires having installed the .NET Framework 4.6.2 developer pack?  In a way that's discoverable by users who hit the build break.

It's also disturbing that the testing for #5059 didn't discover the problem.   There's a gap that needs filling.

ryalanms: Yes, we could add it to the requirements in the documentation and update the .vsconfig (for 3.1).

ThomasGoulet73: @SamBent See my comment for the reason it didn't break the build: https://github.com/dotnet/wpf/issues/5375#issuecomment-926116560.

In short, ILAsm fails but the Task ILAsm returns true whether or not the process exited with 0 or not. Since the ILAsm simply overrides the original dll, it doesn't break the build and the original dll is kept in the output. One quick fix would be to replace [this](https://github.com/dotnet/wpf/blob/v3.1.19/eng/WpfArcadeSdk/tools/InjectModuleInitializer/ILAsmTask.cs#L291-L292) by this:
```
Process process = Process.Start(startInfo);
process.WaitForExit();
return process.ExitCode == 0;
```

ryalanms: I put the documentation update in a separate PR: https://github.com/dotnet/wpf/pull/5402.

SamBent (approved):

PR #5390, opened 2021-09-25
Using span to replace Substring in parserscommon.cs

lindexi:
## Description

Using span to replace Substring in parserscommon.cs. It will be called multiple times to parse the geometry string.

## Customer Impact

None.

## Regression

dotnet 7.0

## Testing

Just CI.

## Risk

Low. I did not change the results of the code.


lindexi:
Close via https://github.com/dotnet/wpf/pull/6272

PR #5399, opened 2021-09-28
Avoid unnecessary boxing in GlyphRun.cs

lindexi:
## Regression

No

## Testing

Just CI.

## Risk

Low.

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Media/GlyphRun.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Media/GlyphRun.cs
@@ -1269,9 +1269,9 @@ namespace System.Windows.Media
 
             if ((_flags & GlyphRunFlags.CacheInkBounds) != 0)
             {
-                if (_inkBoundingBox != null)
+                if (_inkBoundingBox.HasValue)
                 {
-                    return (Rect)_inkBoundingBox;
+                    return _inkBoundingBox.Value;
                 }
             }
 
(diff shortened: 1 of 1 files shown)


src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Media/GlyphRun.cs
```diff
@@ -1269,9 +1269,9 @@ public Rect ComputeInkBoundingBox()
 
             if ((_flags & GlyphRunFlags.CacheInkBounds) != 0)
             {
-                if (_inkBoundingBox != null)
+                if (_inkBoundingBox.HasValue)
                 {
-                    return (Rect)_inkBoundingBox;
+                    return _inkBoundingBox.Value;
```
gfoidl, line 1274:
```suggestion
                    return _inkBoundingBox.GetValueOrDefault();
```
`Value` repeats the check for `HasValue` which is already done.

ThomasGoulet73: @gfoidl Isn't this supposed to be optimized by the JIT in .Net 6 by removing redundant checks when inlining ? Look for "GetValueOrDefault" in the [.Net 6 performance improvements blog post](https://devblogs.microsoft.com/dotnet/performance-improvements-in-net-6/)

gfoidl: Oh, you are right, the JIT will do this now for us. Thanks for the reminder! (I'm too used to that pattern with GetValueOrDefault...)

lindexi: @gfoidl It means I need not do any change?

gfoidl: @lindexi yeah, leave it as is (it's fine so).
Sorry for the confusion.

lindexi: Thank you @gfoidl

lindexi:
Thank you @vishalmsft

Issue #5448, opened 2021-10-05
Overbuild caused by PackageReference support


rainersigwald:
The `_wpftmp` build inside a normal project is built with a global property set: `$(_WpfTempProjectNuGetFilePathNoExt)`. It also now runs `ResolveProjectReferences` by default. This combines to cause overbuild of the referenced projects, because MSBuild treats projects as distinct when they have different global properties.

This can be observed in build output:

```sh-session
❯ dotnet build --no-restore -bl
Microsoft (R) Build Engine version 17.0.0-preview-21477-04+3a1e456fe for .NET
Copyright (C) Microsoft Corporation. All rights reserved.

C:\Program Files\dotnet\sdk\6.0.100-rc.2.21478.25\MSBuild.dll -bl -consoleloggerparameters:Summary -distributedlogger:Microsoft.DotNet.Tools.MSBuild.MSBuildLogger,C:\Program Files\dotnet\sdk\6.0.100-rc.2.21478.25\dotnet.dll*Microsoft.DotNet.Tools.MSBuild.MSBuildForwardingLogger,C:\Program Files\dotnet\sdk\6.0.100-rc.2.21478.25\dotnet.dll -maxcpucount -verbosity:m .\wpf_incremental.sln
  You are using a preview version of .NET. See: https://aka.ms/dotnet-core-preview
  You are using a preview version of .NET. See: https://aka.ms/dotnet-core-preview
  Library -> S:\play\wpf_incremental\Library\bin\Debug\net5.0\Library.dll
  Library -> S:\play\wpf_incremental\Library\bin\Debug\net5.0\Library.dll
  WpfApp -> S:\play\wpf_incremental\WpfApp\bin\Debug\net5.0-windows\WpfApp.dll
```

Note the doubled `Library -> S:\play\wpf_incremental\Library\bin\Debug\net5.0\Library.dll`: that project is building twice.

The global property for the `_wpftmp` project is set here:

https://github.com/dotnet/wpf/blob/fb1a859df66b4fc4eaa01be942e7006d20a41418/src/Microsoft.DotNet.Wpf/src/PresentationBuildTasks/Microsoft/Build/Tasks/Windows/GenerateTemporaryTargetAssembly.cs#L284-L286

Timeline:

1. dotnet/arcade#1581 introduced `$(_WpfTempProjectNuGetFilePathNoExt)` to .NET's Arcade SDK.
[...]

Issue #4119, opened 2021-02-02
Remove WPF's workaround for Arcade setting _WpfTempProjectNuGetFilePathNoExt


ryalanms:

tydunkel:
Just discovered that this was causing issues in our build. Adding this to the GlobalProperties resulted in out WPF projects rebuilding dependent assemblies, copying the original output from obj to bin. After copying to bin our build signs the assembly, so at that point an unsigned assembly replaced a signed assembly.

We build projects in parallel, so some projects depending on the original project are building but will be taking an unsigned version of the assembly until the newly copied assembly is signed. This resulted in build failures since some assemblies are unsigned in our nupkgs.

To fix this, we just added `_WpfTempProjectNuGetFilePathNoExt` to `GlobalPropertiesToRemove` when we build WPF projects. Easy workaround, but took awhile to figure out which property was causing the rebuild.

PR #5453, opened 2021-10-05
[release/6.0] Overbuild caused by PackageReference support (Remove WPF Arcade workaround)

ryalanms:
## Description

[release/6.0] Overbuild caused by PackageReference support (Remove WPF Arcade workaround)

WPF PackageReference support conflicted with an existing Arcade workaround.  The WPF workaround (for the Arcade workaround) changed a global property that caused project references to be rebuilt in WPF's temporary project.  

This removes the global property.
 
The issue is described here: 
https://github.com/dotnet/wpf/issues/5448
https://github.com/dotnet/wpf/issues/4119

## Customer Impact

Project references will be built a second time by the internal WPF project.  

## Regression

This is a regression from 5.0.

## Risk

Low.  

## Testing

Tested the updated WindowsDesktop SDK containing the change against AppCompat applications and WPF samples.

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -2,6 +2,12 @@
   <PropertyGroup>
     <RepositoryName>wpf</RepositoryName>
     <WindowsDesktopARM64Support>true</WindowsDesktopARM64Support>
+     <!-- WPF's new PackageReference support conflicts with the existing Arcade's WPF temp project 
+          workaround.  Disable IsWpfTempProject to prevent Arcade's WPF temp workaround from colliding 
+          with PackageReference support during WPF Framework compilation.  See 
+          https://github.com/dotnet/arcade/pull/1581.  This property can be removed when 1518 (the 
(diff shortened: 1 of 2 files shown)


rainersigwald (commented):
This change looks good.

I'm not sure I understand the order dependencies though. In the end we want three things to happen:

1. This goes into the shipping product.
2. The Arcade SDK workaround is removed.
3. All repos that use the Arcade SDK are updated to use a .NET SDK that has this fix.

Can we do 1 and 2 in parallel? Or will doing 2 break other repos unless they also get 3?

ryalanms:
> Can we do 1 and 2 in parallel? Or will doing 2 break other repos unless they also get 3?

The repos will break if they update to the .NET SDK with the WPF fix, prior to updating to the Arcade SDK with the fix.  But I believe both 1 and 2 can be done simultaneously if the version updates are done in that order (or together).

ryalanms:
The Arcade SDK update has already flowed to all repos.

Issue #4468, opened 2021-05-03
Very slow performance due to internal WPF deferred resource reference list


oatkins:
* .NET Core Version: 5.0.5
* Windows version: 10.0.19042
* Does the bug reproduce also in WPF for .NET Framework 4.8?: Yes
* Is this bug related specifically to tooling in Visual Studio (e.g. XAML Designer, Code editing, etc...)? No

**Problem description:**
Opening, closing and moving between views in a WPF application becomes extremely slow.
 
**Actual behavior:** 
Debugging tools show that considerable time is spent in [FindWeakReference()](https://source.dot.net/WindowsBase/R/97683fbd5ceeb8ca.html) iterating over members of [ResourceDictionary._deferredResourceReferences](https://source.dot.net/PresentationFramework/R/256cf1aca0bb8160.html).

![image](https://user-images.githubusercontent.com/10925800/116887869-c927cd00-abf8-11eb-842f-f8c035c8f730.png)

The list accumulates tens of thousands of entries.

**Expected behavior:**
No delays as described above. From browsing the source code, we believe that entries in `_deferredResourceReferences` are intended to be short lived.
 
**Minimal repro:**
https://github.com/oatkins/WpfDeferredResourceLookupRepro/tree/repro

(More details to come...)

batzen:
I am currently trying to solve this.
Original (add):
Count: 8000
Duration: 00:00:05.6561398
Memory: 212 MB

Original (remove):
Count: 3 (temporarily 16000)
Duration: 00:00:11.6930969
Memory: 202 MB

Modified (add):
Count: 1
Duration: 00:00:03.4468477
Memory: 165 MB

Modified (remove):
Count: 2
Duration: 00:00:00.9332898
Memory: 166 MB

Switching tabs in the repro application takes about 42 seconds in the original version and about 4 seconds in the modified version.

Please note: 
The modified WPF version was built in debug mode as i am unable to build in release mode locally. I guess it would be even faster when built in release mode.
I also used the changes from #4964.

If the WPF team would be interested in a PR i would be happy to create one.
The changes required only affect a few lines in `ResourceDictionary.cs` and a new special class which replaces `WeakReferenceList` for the storage of `DeferredResourceReference`.

batzen:
Allocations are also down from 121 MB to 95 MB during add. Didn't profile the other operations.

batzen:
[...]

PR #5610, opened 2021-10-30
Improve performance of DynamicResource usages

batzen:
Fixes Issue #4468 

## Description

The currently used implementation to keep track of DynamicResource usages is suboptimal as the list can grow quite large and scanning gets very expensive.
That's caused by two issues:
- Every call to FetchResource adds a new entry to the list
- The list implementation is slow as it does not use keys

The improved implementation only adds items with the same key once by checking for existing items before adding new ones. As the new implementation uses keys, retrieval of items is very fast. 

## Customer Impact

Slow performance.

## Testing

Tested the app from #4468

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/ResourceDictionary.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/ResourceDictionary.cs
@@ -1735,15 +1735,6 @@ namespace System.Windows
                     DeferredResourceReference deferredResourceReference;
                     if (!IsThemeDictionary)
                     {
-                        if (_ownerApps != null)
-                        {
-                            deferredResourceReference = new DeferredAppResourceReference(this, resourceKey);
-                        }
-                        else
-                        {
-                            deferredResourceReference = new DeferredResourceReference(this, resourceKey);
-                        }
-
                         // Cache the deferredResourceReference so that it can be validated
                         // in case of a dictionary change prior to its inflation
                         if (_deferredResourceReferences == null)
@@ -1751,7 +1742,24 @@ namespace System.Windows
                             _deferredResourceReferences = new DeferredResourceReferenceList();
                         }
 
-                        _deferredResourceReferences.AddOrSet(deferredResourceReference);
+                        if (_deferredResourceReferences.Get(resourceKey) is DeferredResourceReference existingDeferredResourceReference
+                            && existingDeferredResourceReference.Dictionary == this)
+                        {
+                            deferredResourceReference = existingDeferredResourceReference;
+                        }
+                        else
+                        {
+                            if (_ownerApps != null)
+                            {
+                                deferredResourceReference = new DeferredAppResourceReference(this, resourceKey);
+                            }
+                            else
+                            {
+                                deferredResourceReference = new DeferredResourceReference(this, resourceKey);
+                            }
+
+                            _deferredResourceReferences.AddOrSet(deferredResourceReference);
+                        }
                     }
                     else
                     {
--- a/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/ResourceDictionary.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/ResourceDictionary.cs
@@ -1742,7 +1742,7 @@ namespace System.Windows
                             _deferredResourceReferences = new DeferredResourceReferenceList();
                         }
 
-                        if (_deferredResourceReferences.Get(resourceKey) is DeferredResourceReference existingDeferredResourceReference
+                        if (_deferredResourceReferences.Get(resourceKey) is { } existingDeferredResourceReference
                             && existingDeferredResourceReference.Dictionary == this)
                         {
                             deferredResourceReference = existingDeferredResourceReference;
--- a/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/ResourceDictionary.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/ResourceDictionary.cs
@@ -1748,10 +1748,10 @@ namespace System.Windows
                         // in case of a dictionary change prior to its inflation
                         if (_deferredResourceReferences == null)
                         {
-                            _deferredResourceReferences = new WeakReferenceList();
+                            _deferredResourceReferences = new DeferredResourceReferenceList();
                         }
 
-                        _deferredResourceReferences.Add( deferredResourceReference, true /*SkipFind*/);
+                        _deferredResourceReferences.AddOrSet(deferredResourceReference);
                     }
                     else
                     {
@@ -1775,21 +1775,17 @@ namespace System.Windows
         {
             if (_deferredResourceReferences != null)
             {
-                foreach (Object o in _deferredResourceReferences)
-                {
+                DeferredResourceReference deferredResourceReference = _deferredResourceReferences.Get(resourceKey);
 
-                    DeferredResourceReference deferredResourceReference = o as DeferredResourceReference;
-                    if (deferredResourceReference != null && (resourceKey == null || Object.Equals(resourceKey, deferredResourceReference.Key)))
-                    {
-                        // This will inflate the deferred reference, causing it
-                        // to be removed from the list.  The list may also be
-                        // purged of dead references.
-                        deferredResourceReference.GetValue(BaseValueSourceInternal.Unknown);
-                    }
+                if (deferredResourceReference is not null)
+                {
+                    // This will inflate the deferred reference, causing it
+                    // to be removed from the list.  The list may also be
+                    // purged of dead references.
+                    deferredResourceReference.GetValue(BaseValueSourceInternal.Unknown);
                 }
             }
         }
-
 
         /// <summary>
         /// Called when the MergedDictionaries collection changes
@@ -2053,7 +2049,7 @@ namespace System.Windows
 
         #region Properties
 
-        internal WeakReferenceList DeferredResourceReferences
+        internal DeferredResourceReferenceList DeferredResourceReferences
         {
             get { return _deferredResourceReferences; }
         }
@@ -2478,10 +2474,7 @@ namespace System.Windows
             // redirect each entry toward its new owner
             if (_deferredResourceReferences != null)
             {
-                foreach (DeferredResourceReference drr in _deferredResourceReferences)
-                {
-                    drr.Dictionary = this;
-                }
+                _deferredResourceReferences.ChangeDictionary(this);
             }
         }
 
@@ -2548,7 +2541,7 @@ namespace System.Windows
         private WeakReferenceList                         _ownerFEs = null;
         private WeakReferenceList                         _ownerFCEs = null;
         private WeakReferenceList                         _ownerApps = null;
-        private WeakReferenceList                         _deferredResourceReferences = null;
+        private DeferredResourceReferenceList             _deferredResourceReferences = null;
         private ObservableCollection<ResourceDictionary>  _mergedDictionaries = null;
         private Uri                                       _source = null;
         private Uri                                       _baseUri = null;
--- a/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/SystemResources.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/SystemResources.cs
@@ -2037,7 +2037,7 @@ namespace System.Windows
     internal class DeferredResourceReferenceList : IEnumerable
     {
         private readonly object _syncRoot = new();
-        private readonly Dictionary<object,  WeakReference<DeferredResourceReference>> _entries = new();
+        private readonly Dictionary<object, WeakReference<DeferredResourceReference>> _entries = new();
         private int _potentiallyDeadEntryCount = 0;
 
         public void AddOrSet(DeferredResourceReference deferredResourceReference)
--- a/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/SystemResources.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/SystemResources.cs
@@ -2034,11 +2034,11 @@ namespace System.Windows
         #endregion Properties
     }
 
-    internal class DeferredResourceReferenceList
+    internal class DeferredResourceReferenceList : IEnumerable
     {
         private readonly object _syncRoot = new();
         private readonly Dictionary<object,  WeakReference<DeferredResourceReference>> _entries = new();
-        private int _knownDeadEntryCount = 0;
+        private int _potentiallyDeadEntryCount = 0;
 
         public void AddOrSet(DeferredResourceReference deferredResourceReference)
         {
@@ -2061,7 +2061,7 @@ namespace System.Windows
             WeakReference<DeferredResourceReference> weakReference;
             lock (_syncRoot)
             {
-                weakReference = _entries[resourceKey];
+                _entries.TryGetValue(resourceKey, out weakReference);
             }
 
             if (weakReference is null)
@@ -2075,8 +2075,10 @@ namespace System.Windows
             }
             else
             {
-                ++_knownDeadEntryCount;
+                ++_potentiallyDeadEntryCount;
             }
+
+            PurgeIfRequired();
 
             return null;
         }
@@ -2093,9 +2095,19 @@ namespace System.Windows
                     }
                     else
                     {
-                        ++_knownDeadEntryCount;
+                        ++_potentiallyDeadEntryCount;
                     }
                 }
+            }
+
+            PurgeIfRequired();
+        }
+
+        private void PurgeIfRequired()
+        {
+            if (_potentiallyDeadEntryCount > 25)
+            {
+                Purge();
             }
         }
 
@@ -2103,21 +2115,39 @@ namespace System.Windows
         {
             lock (_syncRoot)
             {
-                List<object> deadEntries = new(_knownDeadEntryCount);
-                _knownDeadEntryCount = 0;
+                List<object> deadKeys = new(Math.Min(_potentiallyDeadEntryCount, _entries.Count));
+                _potentiallyDeadEntryCount = 0;
 
                 foreach (KeyValuePair<object, WeakReference<DeferredResourceReference>> entry in _entries)
                 {
                     if (entry.Value.TryGetTarget(out _) == false)
                     {
-                        deadEntries.Add(entry.Key);
+                        deadKeys.Add(entry.Key);
                     }
                 }
 
-                foreach (object deadKey in deadEntries)
+                foreach (object deadKey in deadKeys)
                 {
                     _entries.Remove(deadKey);
                 }
+            }
+        }
+
+        public IEnumerator GetEnumerator()
+        {
+            lock (_syncRoot)
+            {
+                var list = new List<DeferredResourceReference>(_entries.Count);
+
+                foreach (KeyValuePair<object, WeakReference<DeferredResourceReference>> entry in _entries)
+                {
+                    if (entry.Value.TryGetTarget(out DeferredResourceReference deferredResourceReference))
+                    {
+                        list.Add(deferredResourceReference);
+                    }
+                }
+
+                return list.GetEnumerator();
             }
         }
     }
--- a/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/SystemResources.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/SystemResources.cs
@@ -1739,8 +1739,9 @@ namespace System.Windows
                 {
                     // Note that we are replacing the _keyorValue field
                     // with the value and deleting the _dictionary field.
-                    _keyOrValue = value;
                     RemoveFromDictionary();
+                    // Update after removal from dictionary as we need the key for proper removal
+                    _keyOrValue = value;
                 }
 
                 // Freeze if this value originated from a style or template
@@ -2031,6 +2032,94 @@ namespace System.Windows
         }
 
         #endregion Properties
+    }
+
+    internal class DeferredResourceReferenceList
+    {
+        private readonly object _syncRoot = new();
+        private readonly Dictionary<object,  WeakReference<DeferredResourceReference>> _entries = new();
+        private int _knownDeadEntryCount = 0;
+
+        public void AddOrSet(DeferredResourceReference deferredResourceReference)
+        {
+            lock (_syncRoot)
+            {
+                _entries[deferredResourceReference.Key] = new WeakReference<DeferredResourceReference>(deferredResourceReference);
+            }
+        }
+
+        public void Remove(DeferredResourceReference deferredResourceReference)
+        {
+            lock (_syncRoot)
+            {
+                _entries.Remove(deferredResourceReference.Key);
+            }
+        }
+
+        internal DeferredResourceReference Get(object resourceKey)
+        {
+            WeakReference<DeferredResourceReference> weakReference;
+            lock (_syncRoot)
+            {
+                weakReference = _entries[resourceKey];
+            }
+
+            if (weakReference is null)
+            {
+                return null;
+            }
+
+            if (weakReference.TryGetTarget(out DeferredResourceReference deferredResourceReference))
+            {
+                return deferredResourceReference;
+            }
+            else
+            {
+                ++_knownDeadEntryCount;
+            }
+
+            return null;
+        }
+
+        internal void ChangeDictionary(ResourceDictionary resourceDictionary)
+        {
+            lock (_syncRoot)
+            {
+                foreach (WeakReference<DeferredResourceReference> weakReference in _entries.Values)
+                {
+                    if (weakReference.TryGetTarget(out DeferredResourceReference deferredResourceReference))
+                    {
+                        deferredResourceReference.Dictionary = resourceDictionary;
+                    }
+                    else
+                    {
+                        ++_knownDeadEntryCount;
+                    }
+                }
+            }
+        }
+
+        private void Purge()
+        {
+            lock (_syncRoot)
+            {
+                List<object> deadEntries = new(_knownDeadEntryCount);
+                _knownDeadEntryCount = 0;
+
+                foreach (KeyValuePair<object, WeakReference<DeferredResourceReference>> entry in _entries)
+                {
+                    if (entry.Value.TryGetTarget(out _) == false)
+                    {
+                        deadEntries.Add(entry.Key);
+                    }
+                }
+
+                foreach (object deadKey in deadEntries)
+                {
+                    _entries.Remove(deadKey);
+                }
+            }
+        }
     }
 }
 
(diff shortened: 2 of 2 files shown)


src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/ResourceDictionary.cs
```diff
@@ -1735,23 +1735,31 @@ private void NotifyOwners(ResourcesChangeInfo info)
...
-
                         // Cache the deferredResourceReference so that it can be validated
                         // in case of a dictionary change prior to its inflation
                         if (_deferredResourceReferences == null)
                         {
-                            _deferredResourceReferences = new WeakReferenceList();
+                            _deferredResourceReferences = new DeferredResourceReferenceList();
+                        }
+
+                        if (_deferredResourceReferences.Get(resourceKey) is DeferredResourceReference existingDeferredResourceReference
```
vishalmsft, line 1745: This line results in Type check and null check both. As Type is well known, only null check shall be enough. 
Like
`
var existingDeferredResourceReference = _deferredResourceReferences.Get(resourceKey);
if (existingDeferredResourceReference?.Dictionary == this)
`

batzen: Fixed by using an inline check via `is { }`.

batzen:

src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/SystemResources.cs
```diff
@@ -2028,6 +2029,123 @@ internal override bool IsUnset
...
+                return null;
+            }
+
+            if (weakReference.TryGetTarget(out DeferredResourceReference deferredResourceReference))
+            {
+                return deferredResourceReference;
+            }
+            else
+            {
+                ++_potentiallyDeadEntryCount;
```
bgrainger, line 2073: Should this be `Interlocked.Increment`? The use of `_syncRoot` to protect all accesses to `_entries` implies that this class is used in a multithreaded context, and a simple increment is not threadsafe.

bgrainger: Even with a `lock`, there's a race condition with `Purge` in this method: if `Purge` runs (on another thread) while this thread is between lines 2059 and 2073, the entry will be removed from `_entries` but then this thread will increment `_potentiallyDeadEntryCount` even though the entry no longer exists.

src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/SystemResources.cs
```diff
@@ -2028,6 +2029,122 @@ internal override bool IsUnset
         #endregion Properties
     }
 
+    internal class DeferredResourceReferenceList : IEnumerable
```
RussKie, line 2032: I'm not sure what guidelines this repo follows, but it's generally preferred to have one type per file. This helps with general readability (a developer doesn't have to scroll two thousand lines), repo maintenance and discoverability (to find type ABC one needs to search for ABC file).
An added bonus, you can apply nullable annotations at the top of the file (as opposed to using #nullable enable #nullable disable intermingled with the code).

RussKie (commented):
Have you thought about how it can be tested? How did you confirm the performance is improved?

src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/SystemResources.cs
batzen: Sorry for my late response.
While i think you are right, i just followed the current state of the file.
It already contains 4 classes which seem to be related to each other.
Splitting my class out to a separate file didn't feel right here.

src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/ResourceDictionary.cs
```diff
@@ -1735,23 +1735,31 @@ private void NotifyOwners(ResourcesChangeInfo info)
...
-                            deferredResourceReference = new DeferredAppResourceReference(this, resourceKey);
-                        }
-                        else
-                        {
-                            deferredResourceReference = new DeferredResourceReference(this, resourceKey);
-                        }
-
                         // Cache the deferredResourceReference so that it can be validated
                         // in case of a dictionary change prior to its inflation
                         if (_deferredResourceReferences == null)
```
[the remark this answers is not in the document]
batzen: All the code here uses `== null`.
I think those checks should be upgraded throughout the whole repository, and we shouldn't start mixing usages everywhere.
If the WPF team says we should start mixing that's also ok, but should i only upgrade new code, or everything it touches? So whole classes/methods?

batzen:
> Have you thought about how it can be tested? How did you confirm the performance is improved?

I asked about how we are expected to write tests in multiple of my PRs, but never got an answer...

Performance improvements were checked by testing the app from #4468.

bgrainger (approved):

src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/ResourceDictionary.cs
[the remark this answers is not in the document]
dipeshmsft: No, we don't want to start mixing, we can leave this as it is, and convert them at once in another PR.

src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/SystemResources.cs
```diff
@@ -2028,6 +2029,122 @@ internal override bool IsUnset
...
+                    _entries.Remove(deadKey);
+                }
+            }
+        }
+
+        public IEnumerator GetEnumerator()
+        {
+            lock (_syncRoot)
+            {
+                var list = new List<DeferredResourceReference>(_entries.Count);
```
znakeeye, line 2134: If the list is used only for the enumerator and we have full control of its use, we would benefit from `ArrayPool`. E.g.:
```c#
var array = ArrayPool<DeferredResourceReference>.Shared.Rent(_entries.Count);
```
Obviously, we then need to `ArrayPool.Return()` the array when done enumerating. And we must keep track of the actual element `count`. This could be done in a custom enumerator which holds the (optimized) array enumerator.

batzen: Method was removed.

src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/SystemResources.cs
batzen: Made it sealded.

src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/SystemResources.cs
```diff
@@ -2032,9 +2033,103 @@ internal override bool IsUnset
 
         #endregion Properties
     }
-}
 
+    internal class DeferredResourceReferenceList
```
kronic, line 2037: add sealed

miloush: Why? There is no virtual members and it's an internal type.

kronic: @miloush
[See](https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1852)

miloush: > Sealing types can improve performance.

Not very insightful...

miloush: Also this assembly has [InternalsVisibleToAttribute](https://learn.microsoft.com/en-us/dotnet/api/system.runtime.compilerservices.internalsvisibletoattribute) so according to your link this should not be flagged.

miloush: The answer to my question is here: https://github.com/dotnet/runtime/issues/49944. I am not objecting the suggestion in this case, just want to know why we are doing it.

kronic: @miloush 
Why make a class unsealed if it has no heirs? When reading the code, you may be tempted to look for heirs.

pchaurasia14:
@batzen @miloush @kronic - We've got the results from our test pass and it looks like the PR can be taken. 
Currently, the only discussion whether or not the internal class deserves to be `sealed` or not is open (which can be addressed even later).

With that said, if there are no other concerns, I am going ahead and merging this PR. 
Let me know if you have objections.

miloush:
No objections. I don't mind `sealed` either way - as discussed, it has a potential for perf improvements, although possibly not in this usage.

pchaurasia14 (approved):

pchaurasia14:
Thanks @batzen ! 
Sorry, this took long to get merged.

lightwalker01:
@batzen Contratulations, you finally got it! 😉👍🏼🍾

znakeeye:
In which release can we expect this fix?

lindexi:
@znakeeye I think it will enter .NET 9.

harshit7962:
@batzen, we have encountered a regression due to these changes. As it turns out, we are not expecting a `null` `resourceKey` [here](https://github.com/dotnet/wpf/blob/main/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/ResourceDictionary.cs#L1782), but we are validating the same check [here](https://github.com/dotnet/wpf/blob/main/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/ResourceDictionary.cs#L707). This results in an `ArgumentNullException`. Is expecting a non-null value as `resourceKey` thought of? and if so, the validation checks no longer seems to be valid.

Please find the sample repro [here](https://github.com/harshit7962/wpfui/tree/regressionRepro). Try running the `Wpf.Ui.Gallery` application and it should throw the above mentioned exception. It is excepted to fail with .NET9, preview 7 binaries.

batzen:
@harshit7962 Will have a look as soon as i can.

batzen:
Created PR #9393 to fix the issues

harshit7962:
@batzen, thank for your fix earlier, but even with the fix we are yet to close a few regression issues that we got to know of recently. I have created a sample repro [here](https://github.com/harshit7962/Text-Grab/tree/regressionRepro). The application is expected to crash with preview 7 (or rc1) binaries of .NET 9.

Apologies for this iterative reporting, hopefully we will be closing these in a single go this time.

batzen:
@harshit7962 The first error i encounter is an exception because a `SolidColorBrush` is being used as a `Color` inside of `ButtonStyles.xaml` in line 2 where it says `<SolidColorBrush x:Key="Menu.Static.Background" Color="{DynamicResource SolidBackgroundFillColorSecondaryBrush}" />`.

This was also an error before my changes, but the exception was thrown at a different point in time.
As the line was changed in Text-Grab last year and no one seems to have noticed that the background stays transparent because the resource has the wrong type, IMHO the new behavior is much better as the error would have been very obvious.

Before my changes the exception was thrown during rendering, caught there and the program just continued:
```
Unhandled exception: System.InvalidOperationException: '#FF1C1C1C' is not a valid value for property 'Color'.
   at System.Windows.DependencyObject.GetEffectiveValue(EntryIndex entryIndex, DependencyProperty dp, RequestFlags requests)
   at System.Windows.DependencyObject.GetValueEntry(EntryIndex entryIndex, DependencyProperty dp, PropertyMetadata metadata, RequestFlags requests)
   at System.Windows.DependencyObject.GetValue(DependencyProperty dp)
   at System.Windows.Media.SolidColorBrush.get_Color()
   at System.Windows.Media.SolidColorBrush.UpdateResource(Channel channel, Boolean skipOnChannelCheck)
   at System.Windows.Media.SolidColorBrush.AddRefOnChannelCore(Channel channel)
   at System.Windows.Media.Brush.System.Windows.Media.Composition.DUCE.IResource.AddRefOnChannel(Channel channel)
   at System.Windows.Media.RenderData.System.Windows.Media.Composition.DUCE.IResource.AddRefOnChannel(Channel channel)
   at System.Windows.UIElement.RenderContent(RenderContext ctx, Boolean isOnChannel)
   at System.Windows.Media.Visual.UpdateContent(RenderContext ctx, VisualProxyFlags flags, Boolean isOnChannel)
   at System.Windows.Media.Visual.RenderRecursive(RenderContext ctx)
   at System.Windows.Media.Visual.UpdateChildren(RenderContext ctx, ResourceHandle handle)
   at System.Windows.Media.Visual.RenderRecursive(RenderContext ctx)
   at System.Windows.Media.Visual.UpdateChildren(RenderContext ctx, ResourceHandle handle)
   at System.Windows.Media.Visual.RenderRecursive(RenderContext ctx)
   at System.Windows.Media.Visual.Render(RenderContext ctx, UInt32 childIndex)
   at System.Windows.Media.CompositionTarget.Compile(Channel channel)
   at System.Windows.Media.CompositionTarget.System.Windows.Media.ICompositionTarget.Render(Boolean inResize, Channel channel)
   at System.Windows.Media.MediaContext.Render(ICompositionTarget resizedCompositionTarget)
   at System.Windows.Media.MediaContext.RenderMessageHandlerCore(Object resizedCompositionTarget)
   at System.Windows.Media.MediaContext.RenderMessageHandler(Object resizedCompositionTarget)
   at System.Windows.Media.MediaContext.Resize(ICompositionTarget resizedCompositionTarget)
   at System.Windows.Interop.HwndTarget.OnResize()
   at System.Windows.Interop.HwndTarget.HandleMessage(WindowMessage msg, IntPtr wparam, IntPtr lparam)
   at System.Windows.Interop.HwndSource.HwndTargetFilterMessage(IntPtr hwnd, Int32 msg, IntPtr wParam, IntPtr lParam, Boolean& handled)
```
 
After my changes the exception is thrown during BAML processing:
```
Unhandled exception. System.Windows.Markup.XamlParseException: 'Set property 'System.Windows.ResourceDictionary.Source' threw an exception.' Line number '17' and line position '18'.
 ---> System.InvalidOperationException: '#FF1C1C1C' is not a valid value for property 'Color'.
   at System.Windows.DependencyObject.EvaluateExpression(EntryIndex entryIndex, DependencyProperty dp, Expression expr, PropertyMetadata metadata, EffectiveValueEntry oldEntry, EffectiveValueEntry newEntry) in C:\DEV\OSS_Own\dotnet\wpf_my_prs\src\Microsoft.DotNet.Wpf\src\WindowsBase\System\Windows\DependencyObject.cs:line 1809
   at System.Windows.DependencyObject.EvaluateEffectiveValue(EntryIndex entryIndex, DependencyProperty dp, PropertyMetadata metadata, EffectiveValueEntry oldEntry, EffectiveValueEntry newEntry, OperationType operationType) in C:\DEV\OSS_Own\dotnet\wpf_my_prs\src\Microsoft.DotNet.Wpf\src\WindowsBase\System\Windows\DependencyObject.cs:line 1939
   at System.Windows.DependencyObject.UpdateEffectiveValue(EntryIndex entryIndex, DependencyProperty dp, PropertyMetadata metadata, EffectiveValueEntry oldEntry, EffectiveValueEntry& newEntry, Boolean coerceWithDeferredReference, Boolean coerceWithCurrentValue, OperationType operationType) in C:\DEV\OSS_Own\dotnet\wpf_my_prs\src\Microsoft.DotNet.Wpf\src\WindowsBase\System\Windows\DependencyObject.cs:line 1348
   at System.Windows.DependencyObject.InvalidateProperty(DependencyProperty dp, Boolean preserveCurrentValue) in C:\DEV\OSS_Own\dotnet\wpf_my_prs\src\Microsoft.DotNet.Wpf\src\WindowsBase\System\Windows\DependencyObject.cs:line 1210
   at System.Windows.DependencyObject.InvalidateProperty(DependencyProperty dp) in C:\DEV\OSS_Own\dotnet\wpf_my_prs\src\Microsoft.DotNet.Wpf\src\WindowsBase\System\Windows\DependencyObject.cs:line 1179
   at System.Windows.ResourceReferenceExpression.InvalidateTargetProperty(Object sender, EventArgs e) in C:\DEV\OSS_Own\dotnet\wpf_my_prs\src\Microsoft.DotNet.Wpf\src\PresentationFramework\System\Windows\ResourceReferenceExpression.cs:line 388
   at System.Windows.ResourceReferenceExpression.InvalidateExpressionValue(Object sender, EventArgs e) in C:\DEV\OSS_Own\dotnet\wpf_my_prs\src\Microsoft.DotNet.Wpf\src\PresentationFramework\System\Windows\ResourceReferenceExpression.cs:line 383
   at System.Windows.DependencyObject.OnInheritanceContextChanged(EventArgs args) in C:\DEV\OSS_Own\dotnet\wpf_my_prs\src\Microsoft.DotNet.Wpf\src\WindowsBase\System\Windows\DependencyObject.cs:line 2649
   at System.Windows.Freezable.AddInheritanceContext(DependencyObject context, DependencyProperty property) in C:\DEV\OSS_Own\dotnet\wpf_my_prs\src\Microsoft.DotNet.Wpf\src\WindowsBase\System\Windows\Freezable.cs:line 1180
   at System.Windows.DependencyObject.ProvideSelfAsInheritanceContext(DependencyObject doValue, DependencyProperty dp) in C:\DEV\OSS_Own\dotnet\wpf_my_prs\src\Microsoft.DotNet.Wpf\src\WindowsBase\System\Windows\DependencyObject.cs:line 877
   at System.Windows.DependencyObject.ProvideSelfAsInheritanceContext(Object value, DependencyProperty dp) in C:\DEV\OSS_Own\dotnet\wpf_my_prs\src\Microsoft.DotNet.Wpf\src\WindowsBase\System\Windows\DependencyObject.cs:line 850
   at System.Windows.ResourceDictionary.AddInheritanceContext(DependencyObject inheritanceContext, Object value) in C:\DEV\OSS_Own\dotnet\wpf_my_prs\src\Microsoft.DotNet.Wpf\src\PresentationFramework\System\Windows\ResourceDictionary.cs:line 2306
   at System.Windows.ResourceDictionary.AddInheritanceContextToValues() in C:\DEV\OSS_Own\dotnet\wpf_my_prs\src\Microsoft.DotNet.Wpf\src\PresentationFramework\System\Windows\ResourceDictionary.cs:line 2335
   at System.Windows.ResourceDictionary.set_Source(Uri value) in C:\DEV\OSS_Own\dotnet\wpf_my_prs\src\Microsoft.DotNet.Wpf\src\PresentationFramework\System\Windows\ResourceDictionary.cs:line 261
   at System.Windows.Baml2006.WpfSharedBamlSchemaContext.<>c.<Create_BamlProperty_ResourceDictionary_Source>b__344_0(Object target, Object value) in C:\DEV\OSS_Own\dotnet\wpf_my_prs\src\Microsoft.DotNet.Wpf\src\PresentationFramework\System\Windows\Markup\Baml2006\WpfGeneratedKnownProperties.cs:line 7535
   at System.Windows.Baml2006.WpfKnownMemberInvoker.SetValue(Object instance, Object value) in C:\DEV\OSS_Own\dotnet\wpf_my_prs\src\Microsoft.DotNet.Wpf\src\PresentationFramework\System\Windows\Markup\Baml2006\WpfKnownMemberInvoker.cs:line 44
   at MS.Internal.Xaml.Runtime.ClrObjectRuntime.SetValue(XamlMember member, Object obj, Object value) in C:\DEV\OSS_Own\dotnet\wpf_my_prs\src\Microsoft.DotNet.Wpf\src\System.Xaml\System\Xaml\Runtime\ClrObjectRuntime.cs:line 298
   at MS.Internal.Xaml.Runtime.ClrObjectRuntime.SetValue(Object inst, XamlMember property, Object value) in C:\DEV\OSS_Own\dotnet\wpf_my_prs\src\Microsoft.DotNet.Wpf\src\System.Xaml\System\Xaml\Runtime\ClrObjectRuntime.cs:line 284
   at System.Windows.Markup.XamlReader.RewrapException(Exception e, IXamlLineInfo lineInfo, Uri baseUri) in C:\DEV\OSS_Own\dotnet\wpf_my_prs\src\Microsoft.DotNet.Wpf\src\PresentationFramework\System\Windows\Markup\XamlReader.cs:line 509
   at System.Windows.Markup.WpfXamlLoader.Load(XamlReader xamlReader, IXamlObjectWriterFactory writerFactory, Boolean skipJournaledProperties, Object rootObject, XamlObjectWriterSettings settings, Uri baseUri) in C:\DEV\OSS_Own\dotnet\wpf_my_prs\src\Microsoft.DotNet.Wpf\src\PresentationFramework\System\Windows\Markup\WpfXamlLoader.cs:line 175
   at System.Windows.Markup.WpfXamlLoader.LoadBaml(XamlReader xamlReader, Boolean skipJournaledProperties, Object rootObject, XamlAccessLevel accessLevel, Uri baseUri) in C:\DEV\OSS_Own\dotnet\wpf_my_prs\src\Microsoft.DotNet.Wpf\src\PresentationFramework\System\Windows\Markup\WpfXamlLoader.cs:line 43
   at System.Windows.Markup.XamlReader.LoadBaml(Stream stream, ParserContext parserContext, Object parent, Boolean closeStream) in C:\DEV\OSS_Own\dotnet\wpf_my_prs\src\Microsoft.DotNet.Wpf\src\PresentationFramework\System\Windows\Markup\XamlReader.cs:line 1095
   at System.Windows.Application.LoadComponent(Object component, Uri resourceLocator) in C:\DEV\OSS_Own\dotnet\wpf_my_prs\src\Microsoft.DotNet.Wpf\src\PresentationFramework\System\Windows\Application.cs:line 468
   at Text_Grab.App.InitializeComponent() in C:\DEV\issues\wpf\5610\Text-Grab\Text-Grab\App.xaml:line 1
   at Text_Grab.App.Main()
```

harshit7962:
@batzen Agreed that the exception being thrown is the correct way, however this creates a compat problem for applications that currently use `DynamicResource` and rely on the default behaviour (not crashing) even when incorrect value is provided.
With this fix currently all those apps would not be able to update to .NET 9 without XAML changes on their end. 
I think we should find a way to ensure backward compatibility in this case.

miloush:
Can someone clarify what the migration expectations to newer versions of WPF are in general? Why are we expecting .NET Core 3 apps to run on .NET 9 without any change? Apps that cannot update to .NET 9 without recompilation can stay on .NET 8, that's why we have runtime config to specify which runtime should be used.

h3xds1nz:
I agree with @miloush here as I've been asking myself the same question.

This issue is also the prime example, essentially having a bug in your code that didn't only cause a crash thanks to the exception being swallowed. It just makes no sense in my opinion. And there's a history of such changes happening from version to version.

harshit7962:
@batzen, we are trying to evaluate the regression radius of the changes at the moment. I have created a minimal repro of the above application [here](https://github.com/harshit7962/TextGrabMinimalCompatRepro/). The following is my findings as of now. 

The crash is reported only when we include the following in our csproj.

```csproj
<ItemGroup>
    <Resource Include="ButtonStyles.xaml">
        <Generator>MSBuild:Compile</Generator>
        <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
    </Resource>
</ItemGroup>
```

Need your help in understanding this change in behavior.



Tagging @miloush and @h3xds1nz for their feedbacks and opinions.

miloush:
Is this only XAML parsing issue or runtime too? (Since it's a dynamic resource, a wrong type can be provided at runtime. If that now crashes and didn't used to, it would be even more concerning than the XAML parsing which is apparent immediately)

h3xds1nz:
1) I've tested the same stupid thing against .NET 8 and it will crash in runtime (when you set a resource ref) the same way it will crash with this PR in latest master, same spot, same call-stack; when running `EvaluateExpression` with `InvalidPropertyValue`

2) So the issue is indeed only when parsing BAML/XAML; and imho now it crashes in the correct spot and it is the correct behavior.

What I fail to understand in detail because I can't debug the previous version today is how did this change happened but I can imagine that the value was already `DeferredReference` at that point and that's why it didn't crash while evaluating.

miloush:
I agree that these are errors in the applications and not WPF, and I fully support the early, compile-time failure. However, already during one preview release there is evidence this change breaks things - and that doesn't happen very often, it is the reason why preview releases exist in the first place. If it also very late in the release cycle and I doubt any further changes to the codebase will be entertained for 9 without giving people an opportunity to test them.

I wonder whether we could have it failing compilation forward but not crash applications if they run on a newer runtime than they have been compiled with?

RussKie:
Control the behaviour with a feature switch.

miloush:
Switch is my least favorite option, we keep piling (often unnecessary) switches and never remove them. I am also not quite sure how switch would achieve a feature to be available at compile time but not at runtime (and there might be not enough time to test the switch before RTM). I believe the agreement is we do want the changes in this PR to be part of the product and if the intent is to update the PR to avoid breaking these apps without a switch, then the PR can be updated and merged back later, unless there is an urgency that I missed.

(On the other hand you could say setting a .NET runtime version in the config is a switch.)

RussKie:
Feature switches are used at runtime. 
They could _probably_ be used at design-time, but that would be a misuse of feature switches IMO.

RussKie:
You can learn more at https://learn.microsoft.com/dotnet/core/runtime-config/.

harshit7962:
> * I've tested the same stupid thing against .NET 8 and it will crash in runtime (when you set a resource ref) the same way it will crash with this PR in latest master, same spot, same call-stack; when running `EvaluateExpression` with `InvalidPropertyValue`

@h3xds1nz, regarding this, I would like to clarify the exact scenario. The application will not crash in .NET 8, if we do not include the brush `SolidBackgroundFillColorSecondaryBrush` (used [here](https://github.com/harshit7962/TextGrabMinimalCompatRepro/blob/main/ButtonStyles.xaml#L4)) directly in the project's resource dictionary (local) .

Instead if it is referenced from a resource dictionary not directly included with our application. In the repro, it is included via the `Fluent` dictionary [here](https://github.com/harshit7962/TextGrabMinimalCompatRepro/blob/main/App.xaml#L10).

Since the above mentioned dictionary is not available in .NET 8, I have tweaked the above repro to load the `wpf.ui` styles which should make the difference between .NET 8 and .NET 9 more evident.

The mentioned behavior is present in the application [here](https://github.com/harshit7962/TextGrabMinimalCompatRepro/tree/net8Repro). The application is expected to launch in .NET 8 and crash with `XamlParseException` on .NET 9.

h3xds1nz:
> @h3xds1nz, regarding this, I would like to clarify the exact scenario. The application will not crash in .NET 8, if we do not include the brush `SolidBackgroundFillColorSecondaryBrush` (used [here](https://github.com/harshit7962/TextGrabMinimalCompatRepro/blob/main/ButtonStyles.xaml#L4)) directly in the project's resource dictionary (local) .

Sorry, it was a direct reply to miloush's question, when you try to set an invalid resource reference type during runtime on `FrameworkElement` f.e. The behaviour there is the same. As I've further continued, the only difference before and after PR is what we've confirmed, that is that the BAML/XAML parsing throws an exception already instead of it being swallowed.

h3xds1nz:
@pchaurasia14 Thank you, being it an opt-out switch for .NET 9 release only is very much an acceptable solution that is consistent with the time constraints for .NET 9 in my opinion.

PR #5651, opened 2021-11-07
Remove unuse static field in ImageDrawing.cs

lindexi:
## Description

Remove unuse static field in ImageDrawing.cs

## Customer Impact

No.

## Regression

.NET 7

## Testing

Just CI

## Risk

Low.

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Media/Generated/ImageDrawing.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Media/Generated/ImageDrawing.cs
@@ -327,8 +327,6 @@ namespace System.Windows.Media
 
         internal System.Windows.Media.Composition.DUCE.MultiChannelResource _duceResource = new System.Windows.Media.Composition.DUCE.MultiChannelResource();
 
-        internal static Rect s_Rect = Rect.Empty;
-
         #endregion Internal Fields
 
 


RussKie (dismissed):

ThomasGoulet73:
This file seems to be generated and it says at the top of the file to not edit it manually: https://github.com/dotnet/wpf/blob/main/src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Media/Generated/ImageDrawing.cs#L7.

I thinkg that this field is generated from this entry in the xml: https://github.com/dotnet/wpf/blob/main/src/Microsoft.DotNet.Wpf/src/WpfGfx/codegen/mcg/xml/Resource.xml#L3426.

You can try removing the entry in the xml and then running the codegen (I haven't tried running the codegen but you can take a look at the [readme](https://github.com/dotnet/wpf/blob/main/src/Microsoft.DotNet.Wpf/src/WpfGfx/codegen/mcg/readme.txt)). It also generates some C++ code so this field might be used elsewhere.

vishalmsft:
Thank you @ThomasGoulet73 for highlighting the catch. 
Yes for generated code, editing manually is not a good idea. Will take a deeper look at XML and all generated source code.

RussKie:
Haha, I should've paid more attention
https://github.com/dotnet/wpf/blob/6542b543ff7c04a6dcabde1865cd1c336db2a1ec/src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Media/Generated/ImageDrawing.cs#L5-L10

dipeshmsft:
@lindexi, Can we go ahead and close this PR ??

lindexi:
@dipeshmsft Yes.

Issue #4869, opened 2021-07-18
FindCommandBinding is relatively slow


IAmTheCShark:
This is about `CommandManager.FindCommandBinding(object sender, RoutedEventArgs e, ICommand command, bool execute)`
![image](https://user-images.githubusercontent.com/20905599/126059847-b76ea36a-d6ad-44fc-a256-a09da2b3dfab.png)

I do bellieve that there is two things that would vastly improve performance of this method.

Right now,  "ClassCommandBindings" are stored in a Dictionary that uses `Type` as a key. This fact requires to loop through the type hirarchy every time, to find possible bindings (this would mean 5 loops for a ListBox). 

And then for each of those type related command bindings, loop through a list of bindings to find one that matches the command. 

So in general, lots of looping. That gets worse and worse the more bindings there are.

Instead, why not use a `Dictionary<ICommand, List<(Type, List<CommandBindings>)>`, 
The only looping required then is to find all bindings for the type using `InstanceOfType`. 
My main assumption here is that the vast majority of CommandBindings is not class level.

If this is too much of a resturcture, a simple `HashSet<ICommand>` that stores every command that has a class level binding, could be used as a fast path, to avoid the looping.

Second minor improvement would be to turn those Type-Checks to access the "InputBindingsInternal" into is-Checks like
`if (sender is UIElement uiElement)  localInputBindings = uiElement.InputBindingsInternal;`

PR #5693, opened 2021-11-14
FindCommandBinding improvements

deeprobin:
Might fix Issue #4869

## Description

The main focus was on the FindCommandBinding method in CommandManager.cs. Here I decided to use a ValueTuple instead of a Tuple as well as to do the check using a switch-is block, as this is most likely easier to optimize through the JIT.

Furthermore, I made other code improvements in related methods, like reducing code-nesting and using new C# features.

I don't think these make a big difference to the generated JIT asm, but still hope to have made the code a bit more readable.

/cc @IAmTheCShark

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Input/Command/CommandBinding.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Input/Command/CommandBinding.cs
@@ -7,6 +7,7 @@
 
 using System;
 using System.ComponentModel;
+using System.Diagnostics;
 using System.Windows;
 using System.Windows.Markup;
 using MS.Internal; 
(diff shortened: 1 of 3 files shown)


IAmTheCShark:
I do not believe this fixes #4869 because the algorith itself is quite inefficient to my understanding.
Looking at your changes those are mosly refactorings in favour of new language features. Or did I miss something?

deeprobin:
> 
> 
> I do not believe this fixes #4869 because the algorithm itself is quite inefficient to my understanding. Looking at your changes those are mosly refactorings in favour of new language features. Or did I miss something?

The usage of ValueTuples instead of Tuples might improve the performance a little bit. But I don't know how to benchmark WPF Code because BenchmarkDotNet does not work with this workset

lindexi:
> BenchmarkDotNet

See https://github.com/dotnet/BenchmarkDotNet/issues/1616

anjalisheel-wpf (approved):

deeprobin:
@anjalisheel-wpf What is the current status on this?

anjalisheel-wpf:
@deeprobin we ran into .NET 7 migration issues. So, taking time to merge community test pass this month.

Issue #5462, opened 2021-10-07
Microsoft.WindowsDesktop.App.Runtime.win-arm64 contains a stub "D3DCompiler_47.dll", causing issues with 3rd party libraries


kpreisser:
* .NET Core Version: .NET SDK 5.0.401
* Windows version: Windows 11 Pro Version 21H2 (Build 22000.194) ARM64
* Does the bug reproduce also in WPF for .NET Framework 4.8?: No

 
 
 **Problem description:**

When publishing a WPF app targeting .NET 5.0/6.0 for `win-arm64` (self-contained), a `D3DCompiler_47.dll` file is deployed in the application directory. However, in contrast to the file from the Windows `System32` directory which has ~7 MB and has exports like `D3DCompile`, the file deployed by the .NET SDK has only ~23 KB and doesn't seem to have exports (it comes from the [ Microsoft.WindowsDesktop.App.Runtime.win-arm64](https://www.nuget.org/packages/Microsoft.WindowsDesktop.App.Runtime.win-arm64) package).

Apparently, the presence of this stub DLL can cause issues with 3rd party libraries like CefSharp (embedded Chromium browser), because CEF also depends on `D3DCompiler_47.dll`. In some cases when starting the application, the chromium browser hangs, and the log file contains an error that it cannot find the `D3DCompile` entry point:

```
[1007/102117.812:WARNING:angle_platform_impl.cc(49)] Debug.cpp:180 (insertMessage): GL error: HIGH: Internal D3D11 error: HRESULT: 0x8007000E: Error finding D3DCompile entry point.
[1007/102117.813:ERROR:shared_context_state.cc(73)] Skia shader compilation error
```

**Note:** This doesn't occur when publishing the application for `win-x86` and `win-x64`, because there, a file with a different name `D3DCompiler_47_cor3.dll` is deployed instead of `D3DCompiler_47.dll` (https://github.com/dotnet/wpf/issues/37#issuecomment-494984585), and that file seems to be a [copy from the Windows SDK redist files](https://github.com/dotnet/wpf/blob/30dc08ae87526b3993f295a5e192a93d235847a2/src/Microsoft.DotNet.Wpf/redist/D3DCompiler/D3DCompiler.vcxproj#L46-L53) with all the necessary exports.

See also https://github.com/cefsharp/CefSharp/issues/2944#issuecomment-937595542.
 
 **Minimal repro:**
[...]

PR #5747, opened 2021-11-29
Don't include d3dcompiler_47.dll on ARM64.

singhashish-wpf:
Fixes Issue 
#5462 
Main PR 
None
## Description
Don't include d3dcompiler_47.dll on ARM64.

## Customer Impact

Without this change the d3dcompiler.dll which is shipped for arm64 is causing issues with 3rd party libraries.
 In some cases when starting the application, the chromium browser hangs, and the log file contains an error that it cannot find the D3DCompile entry point
## Regression
No

## Testing
Build and CI

## Risk
None


vishalmsft (approved):

hamarb123:
I'm having issues with this on my app. Can this be back-ported to .NET 6.0.4? I'm going to wait until next LTS for upgrade so I would be stuck not being able to support arm64 users until .NET 8 (~1.5 years).

lindexi:
@singhashish-wpf Can we cherry-pick this to dotnet 6 ? Can it work in dotnet 6?

Issue #5305, opened 2021-09-13
[Performance] Rewrite DirectWriteForwarder in C# to improve rendering performance


deeprobin:
See #4768.

The current code is written in C++/CLI and I think that can be optimized. Unfortunately, I have written very little C++/CLI myself but I suspect that the generated IL code is smaller and better performing.

See also <https://github.com/dotnet/wpf/tree/main/src/Microsoft.DotNet.Wpf/src/DirectWriteForwarder>

madewokherd:
Hi, I did this last year but I needed a small C component to get around some COM interop issues. https://github.com/madewokherd/wpf/blob/main/src/Microsoft.DotNet.Wpf/src/DirectWriteForwarder/CPP/DWriteWrapper/Factory.cs

deeprobin:
@madewokherd Oh cool 👍🏼 Is in this alrready in Pull Request?

madewokherd:
No, I did it because I needed a way to build it on Linux using open source tools for wine-mono. I didn't expect it to be useful for upstream.

It probably needs to be cleaned up. The few C functions needed could be moved to an existing dll, or maybe ComWrappers would remove the need for them. There's still TrueTypeSubsetter which I didn't touch at all, it doesn't seem to be commonly used and could be moved to pure C++ I think. And the C# code probably shouldn't be mixed in with existing C++ code.

It also needs build system updates as I just used my own build system (my requirements were to build on Linux with Framework Mono, so the existing build files were not usable for my purposes).

I'm skeptical that this will improve performance. We still have to call into dwrite.dll so the native interop is just done in a different place.

Maybe I can at least put something together so you can do performance testing, once I find some free time.

ThomasGoulet73:
Hey, this is a great idea and I've actually been working in my free time on converting DirectWriteForwarder to C# for the past couple of weeks. The performance from the move from C++/CLI shouldn't be that noticeable since, as @madewokherd stated, we're still calling into dwrite.dll. But, there are other advantages, here are some in no particular order:
- Trimming support. The .Net Trimmer does not support C++/CLI assembly analysis so it might remove code that is used in DirectWriteForwarder which can produce exceptions that are hard to diagnose (Eg. #3903).
- NativeAOT support. NativeAOT does not support C++/CLI assemblies at all, and if I remember correctly, the team responsible for NativeAOT does not plan on adding support for it. (dotnet/runtimelab#306)
[...]

PR #5765, opened 2021-12-01
Migrate DPI awareness initialization to managed

ThomasGoulet73:
Contributes #5305

## Description
Migrate DPI awareness initialization to managed.

I removed the OS check used to only do the initialization on Windows Vista or later since every version of Windows supported by .Net 6 are more recent than Windows Vista, which renders the check unnecessary.

## Customer Impact
None.

## Regression
No.

## Testing
I tested a WPF application with this PR with high DPI and the DPI awareness was initialized properly.

## Risk
Seems pretty low.


deeprobin (approved):
LGTM

dipeshmsft (approved):

PR #5803, opened 2021-12-09
Update Wpf.Cpp.props to use VS2022 for cpp compilation

vishalmsft:
Update VS and toolset version to use VS2022 (143)
Work in progress for #5733


vishalmsft:
Thanks @singhashish-wpf . These changes are included in the PR #5804. Closing this PR.

PR #5804, opened 2021-12-09
Build on VS 2022

singhashish-wpf:
Fixes Issue 
Build on VS 2022
Main PR

### DIFF
--- a/src/Microsoft.DotNet.Wpf/redist/VCRuntime/VCRuntime.vcxproj
+++ b/src/Microsoft.DotNet.Wpf/redist/VCRuntime/VCRuntime.vcxproj
@@ -33,8 +33,8 @@
   <Import Project="$(VCTargetsPath)\Microsoft.Cpp.Default.props" />
   <Import Project="$(WpfCppProps)" />
   <PropertyGroup>
-    <VCRedistCrtFolderName Condition="'$(Configuration)'=='Release'">Microsoft.VC142.CRT</VCRedistCrtFolderName>
-    <VCRedistCrtFolderName Condition="'$(Configuration)'=='Debug'">Microsoft.VC142.DebugCRT</VCRedistCrtFolderName>
+    <VCRedistCrtFolderName Condition="'$(Configuration)'=='Release'">Microsoft.VC143.CRT</VCRedistCrtFolderName>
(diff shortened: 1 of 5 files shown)


vishalmsft:
@singhashish-wpf  it will be good idea to first take the changes from eng/WpfArcadeSdk/tools/Wpf.Cpp.props . Let build publish the packages and then move to VS2022. 
The problem is with dependency. This migration won't complete without internal build complete. And internal build uses this props to set the toolset version.

src/Microsoft.DotNet.Wpf/redist/VCRuntime/VCRuntime.vcxproj
```diff
@@ -33,8 +33,8 @@
   <Import Project="$(VCTargetsPath)\Microsoft.Cpp.Default.props" />
   <Import Project="$(WpfCppProps)" />
   <PropertyGroup>
-    <VCRedistCrtFolderName Condition="'$(Configuration)'=='Release'">Microsoft.VC142.CRT</VCRedistCrtFolderName>
-    <VCRedistCrtFolderName Condition="'$(Configuration)'=='Debug'">Microsoft.VC142.DebugCRT</VCRedistCrtFolderName>
+    <VCRedistCrtFolderName Condition="'$(Configuration)'=='Release'">Microsoft.VC143.CRT</VCRedistCrtFolderName>
```
ThomasGoulet73, line 36: Is the fallback still necessary ?
https://github.com/dotnet/wpf/blob/af9ff74b557fb694f91944a609e71ed3e8b88568/src/Microsoft.DotNet.Wpf/redist/VCRuntime/VCRuntime.vcxproj#L38-L42

It seems weird to fallback to VC141 when VC143 does not exist.

singhashish-wpf: Not necessary @ThomasGoulet73 
Will be removed.

singhashish-wpf: @ThomasGoulet73 https://github.com/dotnet/wpf/pull/5810

PR #5813, opened 2021-12-14
Improve pre-commit githook

SamBent:
Fixes Issue 

Main PR 

## Description

Improvements to the hook that deters users from changing generated files:

1. The hook calls a file-comparison tool that expects its inputs to be sorted, but doesn't sort the inputs ("comm: file 1 is not in sorted order").  Fixed that by (a) sorting the dynamic list of staged files, and (b) pre-sorting the fixed list of generated files.
2. Revised the error output to give a better suggestion about what to do next.

To get the full benefit of this change, copy `eng\WpfArcadeSdk\tools\pre-commit.githook` to `.git\hooks\pre-commit`, overwriting the existing file (both paths are relative to your repo's root).  There's probably some way to cause this to happen automatically, but I couldn't find it.

## Customer Impact

No change to product code.  Developer experience is improved.

## Regression

No.

## Testing

It works for me.

## Risk

No risk.

### DIFF
--- a/eng/WpfArcadeSdk/tools/WPF_Generated_Files.txt
+++ b/eng/WpfArcadeSdk/tools/WPF_Generated_Files.txt
@@ -187,9 +187,9 @@ src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Media/Generated/Ima
 src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Media/Generated/ImageSource.cs
 src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Media/Generated/Int32Collection.cs
 src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Media/Generated/Int32CollectionConverter.cs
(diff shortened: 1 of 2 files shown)


singhashish-wpf (approved):

dipeshmsft:
The  `eng\WpfArcadeSdk\tools\pre-commit.githook` file is compied to `.git\hooks\pre-commit` at the time when `build.cmd` is run. It executes `restore-toolset.ps1` which copies the pre-commit hook file.

dipeshmsft (approved):

SamBent:
@dipeshmsft 

> The `eng\WpfArcadeSdk\tools\pre-commit.githook` file is compied to `.git\hooks\pre-commit` at the time when `build.cmd` is run. It executes `restore-toolset.ps1` which copies the pre-commit hook file.

No it doesn't.  This only happens the very first time you build the repo.  After that, restore-toolset says:
```
Detecting WPF Git hooks...
Detected existing WPF Git pre-commit hook.
```

However, deleting  `.git\hooks\pre-commit` will install the new one the next time you build:
```
Detecting WPF Git hooks...
Installing WPF Git pre-commit hook...
```

BTW, I pushed the wrong version of `pre-commit.githook`, lacking the sort (step 1a).  Fixed in latest commit.

Issue #5835, opened 2021-12-16
ContextMenu stops working if its owner is removed from the visual tree


SamBent:
[Port of a servicing fix for .NET 4.7-4.8]
 
 **Problem description:**
If user invokes a ContextMenu (e.g. right-clicking on its owner), but the owner leaves the visual tree before the ContextMenu actually displays, the ContextMenu never displays again.  This is interesting when the ContextMenu is shared by multiple owners - it won't display for other owners, even when the other owners remain in the visual tree normally.
 
 **Actual behavior:** 
Invoking the ContextMenu does nothing.
 
 **Expected behavior:**
ContextMenu displays
 
 **Minimal repro:**

huiliu-code:
Could you show a minimal example of a reproducible problem?

vishalmsft:
Hi @huiliu-code , attached is the [sample project](https://github.com/dotnet/wpf/files/7778762/TestPopupMenu2.zip). Steps are
1. Start the debugging.
2. Click on "Start Test"
3. Multiple quick right click on List Items. 
4. Once Context Menu stops showing up, it never shows up again.

The PR fix that behavior.

PR #5843, opened 2021-12-16
Port "remove ContextMenu owner" fix from 4.8

SamBent:
Addresses #5835
This is a port of a servicing fix in .NET 4.7-4.8.

## Description

There are two reasons the context menu does not display.  The first reason is that its owner gets removed from the tree after the right-button-up gesture but before the PopupControlService initiates the display logic.  The logic notices that the owner is not being displayed, and therefore doesn't display the context menu.  This is by design - you shouldn't display a context menu for an element that isn't itself visible.

The second reason is that the (shared) context menu is in a bad state, where it's marked IsOpen=true, but isn't actually being displayed.  The display logic doesn't do anything, since it thinks the menu is already open.  This is a bug, left over from the first case, which doesn't reset IsOpen to false when it aborts the display logic.   The net effect is that once a context menu doesn't open because of reason 1, it will never open again, even in cases where it should.

The fix is to reset the context menu's state when aborting the display logic.

## Customer Impact

ContextMenus stop working

## Regression

No

## Testing

Standard regression testing.

## Risk

Low. Port of a .NETFx servicing fix released earlier this year.


vishalmsft (approved):
Reviewed with the test project attached in the issue.

PR #5901, opened 2022-01-03
Change TFM version to net7.0

singhashish-wpf:
Fixes #5801 

Main PR 

## Description
Change Target Framework Version to net7.0 and move TargetFramework to Directory.Build.Props.

## Customer Impact
None

## Regression
None

## Testing
CI Build

## Risk
None

PR #5910, opened 2022-01-05
Use builtin MSBuild functions to compare versions

ThomasGoulet73:
## Description
Future-proofs target framework version increments by using builtin MSBuild functions to compare target framework versions.

Reduces the number of place to update when incrementing the target framework version.

This PR could be merged before #5901 to reduce diff or after to reduce diff of future target framework version increments.

The changes in this PR requires MSBuild 16.5 since `VersionGreaterThanOrEquals` was added in this version.

## Customer Impact
Build only.

## Regression
No.

## Testing
Local build + CI.

## Risk
None.

/cc @singhashish-wpf @RussKie


singhashish-wpf (approved):

RussKie:
Nice!

PR #5915, opened 2022-01-07
adding support for Nullable

ayushverma-ms:
Fixes # 

Main PR 

## Description
Create a C# "WPF Application" project targeting .NET 5 or .NET 6
Add a "Page Function (WPF)" item to the project.
Build Solution
Expected: No errors
Actual: CS0263 -- Partial declarations of 'PageFunction1' must not specify different base classes

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/PresentationBuildTasks/MS/Internal/MarkupCompiler/MarkupCompiler.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationBuildTasks/MS/Internal/MarkupCompiler/MarkupCompiler.cs
@@ -666,7 +666,7 @@ namespace MS.Internal
                     {
                         _ccRoot.CodeClass.IsPartial = true;
                     }
-                    addDirectiveToFile(codeStreamWriter, NULLABLE_RESTORE_DIRECTIVE);
+                    AddDirectiveToFile(codeStreamWriter, NULLABLE_RESTORE_DIRECTIVE);
                     codeProvider.GenerateCodeFromCompileUnit(ccu, codeStreamWriter, o);
 
                     codeStreamWriter.Flush();
@@ -686,7 +686,7 @@ namespace MS.Internal
         }
 
         // adds directrive to top of the file.
-        private static void addDirectiveToFile(StreamWriter streamWriter, string directive) {
+        private static void AddDirectiveToFile(StreamWriter streamWriter, string directive) {
             streamWriter.WriteLine(directive);
         }
 
--- a/src/Microsoft.DotNet.Wpf/src/PresentationBuildTasks/MS/Internal/MarkupCompiler/MarkupCompiler.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationBuildTasks/MS/Internal/MarkupCompiler/MarkupCompiler.cs
@@ -686,7 +686,8 @@ namespace MS.Internal
         }
 
         // adds directrive to top of the file.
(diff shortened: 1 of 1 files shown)


src/Microsoft.DotNet.Wpf/src/PresentationBuildTasks/MS/Internal/MarkupCompiler/MarkupCompiler.cs
```diff
@@ -685,6 +685,11 @@ private void GenerateSource()
             GenerateInternalTypeHelperImplementation();
         }
 
+        // adds directrive to top of the file.
+        private static void addDirectiveToFile(StreamWriter streamWriter, string directive) {
```
lindexi, line 689: Code Format issues.

See: https://github.com/dotnet/runtime/blob/main/docs/coding-guidelines/coding-style.md

> 1. We use Allman style braces, where each brace begins on a new line. A single line statement block can go without braces but the block must be properly indented on its own line and must not be nested in other statement blocks that use braces (See rule 18 for more details). One exception is that a using statement is permitted to be nested within another using statement by starting on the following line at the same indentation level, even if the nested using contains a controlled block.

> 13. We use PascalCasing for all method names, including local functions.

dipeshmsft (approved):

src/Microsoft.DotNet.Wpf/src/PresentationBuildTasks/MS/Internal/MarkupCompiler/MarkupCompiler.cs
```diff
@@ -685,6 +685,11 @@ private void GenerateSource()
             GenerateInternalTypeHelperImplementation();
         }
 
+        // adds directrive to top of the file.
+        private static void AddDirectiveToFile(StreamWriter streamWriter, string directive) {
```
lindexi, line 689: Code Format issues.

See: https://github.com/dotnet/runtime/blob/main/docs/coding-guidelines/coding-style.md

> 1. We use Allman style braces, where each brace begins on a new line. A single line statement block can go without braces but the block must be properly indented on its own line and must not be nested in other statement blocks that use braces (See rule 18 for more details). One exception is that a using statement is permitted to be nested within another using statement by starting on the following line at the same indentation level, even if the nested using contains a controlled block.

```suggestion
        private static void AddDirectiveToFile(StreamWriter streamWriter, string directive) 
        {
```

src/Microsoft.DotNet.Wpf/src/PresentationBuildTasks/MS/Internal/MarkupCompiler/MarkupCompiler.cs
```diff
@@ -1643,6 +1649,7 @@ private string ParentFolderPrefix
 
         private void AddLinePragma(CodeTypeMember ctm, int lineNumber)
         {
+
```
RussKie, line 1652: Please remove

src/Microsoft.DotNet.Wpf/src/PresentationBuildTasks/MS/Internal/MarkupCompiler/MarkupCompiler.cs
```diff
@@ -3609,6 +3616,9 @@ internal string SubClass
         private static string s_generatedCode_ToolName;
         private static string s_generatedCode_ToolVersion;
 
+        // For enabling nullable in the project
+        private static string NULLABLE_RESTORE_DIRECTIVE = "#nullable restore";
```
lindexi, line 3620: How about `static readonly` or `const` ?

PR #5925, opened 2022-01-09
Remove the unnecessary calculate code in ImageSource.cs

lindexi:
## Description

Should we remove the calculate when dpif equals 96.0F?

    return (double)(pixels * (96.0F / dpif));
    When dpif == 96.0F
    (double)(pixels * (96.0F / dpif)) 
      = (double)(pixels * (96.0F / 96.0F)) 
      = (double)(pixels * 1) 
      = (double)(pixels) 

## Customer Impact

None.

## Testing

Just CI

## Risk

Low.

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Media/ImageSource.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Media/ImageSource.cs
@@ -112,7 +112,7 @@ namespace System.Windows.Media
 
             if (dpif < 0.0F || FloatUtil.IsCloseToDivideByZero(96.0F, dpif))
             {
-                dpif = 96.0F;
+                return pixels;
             }
 
             return (double)(pixels * (96.0F / dpif));


src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Media/ImageSource.cs
```diff
@@ -112,7 +112,7 @@ protected static double PixelsToDIPs(double dpi, int pixels)
 
             if (dpif < 0.0F || FloatUtil.IsCloseToDivideByZero(96.0F, dpif))
             {
-                dpif = 96.0F;
+                return pixels;
```
dipeshmsft, line 115: Before returning we should typecast pixels to double.

lindexi: @dipeshmsft The int pixels will cast to double automatic

lindexi: Do you mean I should cast to double manual ?

dipeshmsft: I guess we can leave it to that.

anjalisheel-wpf (approved):

lindexi:
@anjalisheel-wpf Thank you.

PR #5930, opened 2022-01-10
Skip InitialShowDelay when replacing the current tooltip

### LINKED ISSUE #5703: .NET 6 Tooltip behavior change from .NET 5 (bug?)
* .NET Version: 6.0.100
* Windows version: 21H1 (19043.1165)
* Does the bug reproduce also in WPF for .NET Framework 4.8?: Not checked

 
 
 **Problem description:**
After upgrading a WPF app from .NET 5 (SDK 5.0.402) to .NET 6, new tooltips are no longer shown without delay when leaving a UI element by mouse for which a tooltip is currently open.

See the following XAML:
````code
<StackPanel Orientation="Horizontal"
        HorizontalAlignment="Center"
        VerticalAlignment="Center">
    <Button Content="Button 1"
        ToolTip="Button 1 tooltip"
        ToolTipService.Placement="MousePoint"
        ToolTipService.VerticalOffset="-50"
        ToolTipService.HorizontalOffset="-45"
        ToolTipService.InitialShowDelay="1000"
        ToolTipService.BetweenShowDelay="1000"
        ToolTipService.ShowDuration="4000"/>
    <Button Content="Button 2"
        ToolTip="Button 2 tooltip"
        ToolTipService.Placement="MousePoint"
        ToolTipService.VerticalOffset="-50"
        ToolTipService.HorizontalOffset="-45"
        ToolTipService.InitialShowDelay="1000"
        ToolTipService.BetweenShowDelay="1000"
        ToolTipService.ShowDuration="4000"/>
</StackPanel>
````

And the resulting tooltip behavior for both .NET 5 and .NET 6:
| .NET 5 | .NET 6 |
| ------------- | ------------- |
| WPF tooltip behavior NET 5  | WPF tooltip behavior NET 6 |

As you can see, the tooltip for "Button 2" is no longer shown immediately after entering it with the mouse as it does in the .NET 5 app. This is especially problematic in an app which uses a (WinUI-styled) NumberBox control with two adjacent spinner buttons both with their own tooltips (such as "Increase/Decrease by 0.1"):
image

I suspect this behavior change was introduced by the tooltip work for .NET 6, i.e. in PR https://github.com/dotnet/wpf/pull/4623. Is this behavior change intended or is this a bug? Personally, this looks like a bug to me as it negatively affects the user experience in my view. The tooltip for "Button 1" should no longer be shown when the mouse enters "Button 2" which has its own tooltip - and which should show immediately based on the `ToolTipService.BetweenShowDelay` setting.

@SamBent FYI as you worked on the tooltip changes for .NET 6.
 
 **Expected behavior:**
Same as in .NET 5.
 
 **Minimal repro:**
[WPF issue.zip](https://github.com/dotnet/wpf/files/7554940/WPF.issue.zip) (Just change the framework from .NET 5 to .NET 6 to see the .NET 6 behavior.)

SamBent:
Fixes #5703

## Description

Moving the mouse from a control A that is showing a tooltip, to another control B that has a tooltip, should show B's tooltip immediately (ignoring its InitialShowDelay).  This works correctly if A's tooltip is closed before checking whether B has a tooltip, which was always the case prior to .NET 6.0.  The new tooltip behavior in 6.0 can keep A's tooltip open longer, and now needs to check whether a tooltip is already open when it decides that B wants to open its tooltip.

## Customer Impact

Fixes a regression.

## Regression

Yes.

## Risk

Low.  Handles a case that couldn't happen in .NET 5.

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Controls/PopupControlService.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Controls/PopupControlService.cs
@@ -314,20 +314,29 @@ namespace System.Windows.Controls
             PendingToolTip = SentinelToolTip(o, triggerAction);
 
             // decide when to promote to current
-            int showDelay;
-            switch (triggerAction)
+            bool showNow = _quickShow;
+            if (!showNow)
             {
-                case ToolTipService.TriggerAction.Mouse:
-                case ToolTipService.TriggerAction.KeyboardFocus:
-                    showDelay = _quickShow ? 0 : ToolTipService.GetInitialShowDelay(o);
-                    break;
-                case ToolTipService.TriggerAction.KeyboardShortcut:
-                default:
-                    showDelay = 0;
-                    break;
+                switch (triggerAction)
+                {
+                    case ToolTipService.TriggerAction.Mouse:
+                        // a mouse request shows without delay if the current tooltip also came from mouse
+                        showNow = !(CurrentToolTip?.FromKeyboard ?? true);
+                        break;
+                    case ToolTipService.TriggerAction.KeyboardFocus:
+                        // a focus request shows without delay if the current tooltip also came from keyboard
+                        showNow = CurrentToolTip?.FromKeyboard ?? false;
+                        break;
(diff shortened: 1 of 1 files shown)


SamBent:
Replaced by a better fix #6058

PR #5931, opened 2022-01-10
ToolTips in popups (or any context where capture is active)

### LINKED ISSUE #5716: Comboboxitem tooltip bug
* .NET Core Version: NET 6, NET 5
* Windows version: windiws 10 1803

NET 6

 
NET 5

   ```
<Grid>
        <ComboBox Width="100" Height="20">
            <ComboBoxItem ToolTip="1">13.42</ComboBoxItem>
            <ComboBoxItem ToolTip="2">15.82</ComboBoxItem>
        </ComboBox>
    </Grid>
```

### PR DESCRIPTION
**Pull request by SamBent**
Fixes #5716

## Description

The new 6.0 tooltip behavior computes a "safe area" in which the user can move the mouse without dismissing the tooltip.  It's computed with coordinates relative to the owner's PresentationSource (hwnd), but the mouse movements were tracked relative to the input provider's PresentationSource.  These are often the same, but not when the tooltip's owner comes from a Popup (or similar context that captures mouse input away from the normal input provider).  This dismisses the tooltip immediately - the logic thinks the mouse has left the safe area.

Fixed by converting the mouse position to the correct coordinate system - relative to the tooltip owner's PresentationSource.

## Customer Impact

Fixes a regression.

## Regression

Yes.

## Risk

Low.  Applies the correct coordinate transformation before checking the mouse position.

### CODE REVIEW

File: src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Controls/PopupControlService.cs [hunk 1]
```diff
@@ -1475,6 +1475,44 @@ private void AddPoints(PointList points, in NativeMethods.RECT rect, bool rectIs
...
+            internal bool ContainsMousePoint()
+            {
+                IInputElement rootElement = _source.RootVisual as IInputElement;
+                if (rootElement != null)
+                {
+                    // get the coordinates of the current mouse point, relative to our PresentationSource
+                    System.Windows.Point pt = Mouse.PrimaryDevice.GetPosition(rootElement);
+
+                    // check whether the point lies within the hull
+                    return ContainsPoint(_source, (int)pt.X, (int)pt.Y);
```
lindexi, line 1488: Could there be a loss of precision here? How about changing the ContainsPoint method argument from int to double?

SamBent: ContainsPoint requires 16-bit integers, which is what GetPosition supplies (albeit wrapped in a Point whose fields have type double).  See the [comment](https://github.com/dotnet/wpf/pull/5931/files/e43234e81b48430b36ff24d0d29a22fe2f9b05dd#diff-e4eeb1239fe1089ce77795654002e57e7c12c2b616962d9369b70878e74b8e95R1233) for class ConvexHull.

singhashish-wpf (approved):

PR #6029, opened 2022-01-29
Replace mention of .NET Core 3 to .NET 6

kant2002:
That gives very stale look on the WPF. Designer is working too.

### DIFF
--- a/Documentation/getting-started.md
+++ b/Documentation/getting-started.md
@@ -1,13 +1,13 @@
-# Getting started with WPF for .NET Core
+# Getting started with WPF for .NET
 
-This document describes the experience of using WPF on .NET Core. The [Developer Guide](developer-guide.md) describes how to develop features and fixes for WPF.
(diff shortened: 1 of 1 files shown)


vishalmsft:
@kant2002 Is there a reason to remove the "Core" word? Removing "3" is really great and thanks for taking the efforts to update the stale documents.

kant2002:
My idea is that “.Net” about 2 year like new name. There still a bit of confusion between what is .Net and what’s difference with .Net Core. So I think better use “current” name if project is not freezed in time. And WPF has “perception” of being abandoned. So having Net Core as primary name will just strengthens this opinion. Which I do not think justified

ThomasGoulet73:
They removed the "Core" suffix in the Windows Forms repo ([getting-started.md in dotnet/winforms](https://github.com/dotnet/winforms/blob/main/docs/getting-started.md)). IIRC the "Core" suffix is only used when comparing .Net Core and .Net Framework (Exemple: ".Net Core supports XYZ while .Net Framework does not.").

vishalmsft (approved):

PR #6034, opened 2022-01-30
Update MSBuild tasks

kant2002:
This updates MSBuild to cross-platform tools. In my case that allow me to smoke test changes on Linux
where I occasionally type code.


lindexi:
Awesome

RussKie (approved):

PR #6044, opened 2022-02-01
Use generic Marshal.StructureToPtr

ThomasGoulet73:
## Description
Use generic Marshal.StructureToPtr instead of the non-generic. I removed the customs UnsafeNativeMethods.StructureToPtr which justs forwards to Marshal.StructureToPtr because it causes linker warnings because Marshal.StructureToPtr is annotated while UnsafeNativeMethods.StructureToPtr isn't.

Almost every call to Marshal.StructureToPtr were already using the generic overload without the need to change the code.

Contributes to #3811

## Customer Impact
None, same behavior as the non-generic.

## Regression
No

## Testing
I manually tested a few of the generic and non-generic and they gave the same result.

## Risk
None, same behavior as the non-generic.

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/UIAutomation/UIAutomationClient/System/Windows/Automation/Condition.cs
+++ b/src/Microsoft.DotNet.Wpf/src/UIAutomation/UIAutomationClient/System/Windows/Automation/Condition.cs
@@ -38,7 +38,8 @@ namespace System.Windows.Automation
         }
 
         // uiaCondition is one of the Uia condition structs - eg UiaCoreApi.UiaAndOrCondition
-        internal static SafeConditionMemoryHandle AllocateConditionHandle(object uiaCondition)
+        internal static SafeConditionMemoryHandle AllocateConditionHandle<T>(T uiaCondition)
+            where T : struct
         {
             // Allocate SafeHandle first to avoid failure later.
             SafeConditionMemoryHandle sh = new SafeConditionMemoryHandle();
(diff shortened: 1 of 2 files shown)


src/Microsoft.DotNet.Wpf/src/UIAutomation/UIAutomationClient/System/Windows/Automation/Condition.cs
```diff
@@ -38,7 +38,8 @@ override protected bool ReleaseHandle()
         }
 
         // uiaCondition is one of the Uia condition structs - eg UiaCoreApi.UiaAndOrCondition
-        internal static SafeConditionMemoryHandle AllocateConditionHandle(object uiaCondition)
+        internal static SafeConditionMemoryHandle AllocateConditionHandle<T>(T uiaCondition)
```
ThomasGoulet73, line 41: Using generics here allows to call the generic version of Marshal.StructureToPtr and also to avoid boxing of `uiaCondition` every call to `AllocateConditionHandle` passes a struct for `uiaCondition`.

kant2002:
Thank you!

bmarinov:
Hey all, any comment on this PR from the repo / code owners would be appreciated. The PR has been open for two months and it looks like a minor change that is not breaking anything. CC @SamBent / @fabiant3 / @vishalmsft / @singhashish-wpf 

Context: .NET Core 3.1 is EoL at the end of the year. Trimming is unstable and practically unusable in production with any subsequent .NET versions. There has been little to no progress on #3811 with duplicates being closed with no comment regarding the original issue.

singhashish-wpf (approved):

PR #6053, opened 2022-02-02
Replace resource constants with properties in WindowsFormsIntegration

ThomasGoulet73:
Contributes to dotnet/wpf#1

Depends on #6059

## Description
Replace the use of resource constants with properties in WindowsFormsIntegration. This follows the standard of other .Net projects, like [dotnet/runtime](https://github.com/dotnet/runtime).

## Customer Impact
None.

## Regression
No.

## Testing
I tested a few resources and made sure that the generated code in the property getter is the same as the original code.

## Risk
None.


eng/WpfArcadeSdk/tools/SystemResources.props
```diff
@@ -26,7 +26,7 @@
       <ClassName Condition="'$(AssemblyName)'=='ReachFramework'">System.Windows.Xps.SRID</ClassName>
       <ClassName Condition="'$(AssemblyName)'=='PresentationFramework'">System.Windows.SRID</ClassName>
       <ClassName Condition="'$(AssemblyName)'=='PresentationUI'">System.Windows.TrustUI.SRID</ClassName>
-      <ClassName Condition="'$(AssemblyName)'=='WindowsFormsIntegration'">System.Windows.SRID</ClassName>
+      <ClassName Condition="'$(AssemblyName)'=='WindowsFormsIntegration'">System.Windows.SR</ClassName>
```
ThomasGoulet73, line 29: I replaced `SRID` with `SR` because the suffix `ID` does not apply anymore. This is the same as the default in dotnet/runtime: https://github.com/dotnet/runtime/blob/main/eng/resources.targets#L8.

ThomasGoulet73:
I rebased my changes onto #6059, which this PR depends on.

Issue #5880, opened 2021-12-28
By using keyboard not able to adjust the width of the column.


Swati1700:
**GitHubTags:**#A11yUsable;#A11ySev4;#A11yTCS;#DesktopApp;#Win10;#BM-.NetCore(WPF)-Win32-July2021;#.NET Core;#Benchmark;#Keyboard;

### Environment Details:
Application Name: Azure RTOS - Trace X
Windows Version: Win10
Application Version: 6.1.9.0

### Repro Steps:

1. Open WPFDataGridSorted-master file.
2. Open WPFDataGridSorted.sln file.
3. Tab to "Start" button and hit enter.
4. Tab to First column in the table.
5. Observe that by using keyboard not able to adjust the column width.

### Actual:
By using keyboard not able to adjust the width of the "Label", "Labelled" and "Description" column.

### Expected:
By using keyboard (eg. CTRL + Right arrow/Left arrow) the width of the column should be adjustable.

### User Impact:
Keyboard only users will not able to adjust the column width and they will not able to see complete information present on the screen.

[WPFDataGridSorted-master (5).zip](https://github.com/dotnet/wpf/files/7783485/WPFDataGridSorted-master.5.zip)

https://user-images.githubusercontent.com/86710888/147543355-76c90b59-4c6a-4cd0-9cd2-14bebd774644.mp4

vishalmsft:
@Swati1700 Is this issue created as part of migrating bugs from ADO?
User Story [1383796](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1383796): WPF GridView/DataGrid column width cannot be changed using only the keyboard

Swati1700:
@vishalmsft We have logged the primary bug in the ADO and they have said that this is external to WPF. So we have logged this WPF bug in github.

miloush:
I am assuming this is expected to work only when header is focused, otherwise it would interfere with DataGrid navigation. 

Is the usage of Ctrl+arrow keys meaning column resizing part of any user guidelines? Does it mirror in RTL?

Also it might be worth supporting <kbd>Ctrl</kbd>+<kbd>Num+</kbd> for autosizing all columns like the native listview control [does](https://devblogs.microsoft.com/oldnewthing/20121030-00/?p=6223),

vishalmsft:
@miloush 

> I am assuming this is expected to work only when header is focused, otherwise it would interfere with DataGrid navigation.

[...]

PR #6054, opened 2022-02-02
WPF DataGrid/GridView  column width can be changed using the keyboard shortcut ALT+left or right arrow key

anjalisheel-wpf:
Fixes #  #5880 

Main PR  #6054 

## Description

Issue related to WPF DataGrid column width can now be changed using the keyboard shortcut "ALT + left arrow key" and "ALT + right arrow key".

## Customer Impact

Fixes a regression.

## Regression

Yes.

## Testing

Integration tests are in progress.

## Risk

Low. Add ability to adjust column width using keyboard for WPF DataGrid.

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Controls/DataGridCell.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Controls/DataGridCell.cs
@@ -964,6 +964,28 @@ namespace System.Windows.Controls
         /// </summary>
         protected override void OnKeyDown(KeyEventArgs e)
         {
+            if ((Keyboard.Modifiers & ModifierKeys.Alt) == ModifierKeys.Alt)
+            {
+                DataGridLength updatedWidth = new DataGridLength();
+
+                if (e.Key == Key.Right)
+                {
+                    updatedWidth = new DataGridLength(this.Column.ActualWidth + _columnWidthStepSize);
+                }
+                else if (e.Key == Key.Left)
+                {
+                    updatedWidth = new DataGridLength(this.Column.ActualWidth - _columnWidthStepSize);
+                }
+
+                if (Column.CanColumnResize(updatedWidth))
+                {
+                    this.Column.SetCurrentValue(DataGridColumn.WidthProperty, updatedWidth);
+                }
+
+                e.Handled = true;
+                return;
+            }
+            
             SendInputToColumn(e);
         }
 
@@ -1100,6 +1122,7 @@ namespace System.Windows.Controls
         private DataGridRow _owner;
         private ContainerTracking<DataGridCell> _tracker;
         private bool _syncingIsSelected;                    // Used to prevent unnecessary notifications
+        private const double _columnWidthStepSize = 10d;
 
         #endregion
     }
--- a/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Controls/DataGridCell.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Controls/DataGridCell.cs
@@ -970,11 +970,11 @@ namespace System.Windows.Controls
 
                 if (e.Key == Key.Right)
                 {
-                    updatedWidth = new DataGridLength(this.Column.ActualWidth + _columnWidthStepSize);
+                    updatedWidth = new DataGridLength(this.Column.ActualWidth + ColumnWidthStepSize);
                 }
                 else if (e.Key == Key.Left)
                 {
-                    updatedWidth = new DataGridLength(this.Column.ActualWidth - _columnWidthStepSize);
+                    updatedWidth = new DataGridLength(this.Column.ActualWidth - ColumnWidthStepSize);
                 }
 
                 if (Column.CanColumnResize(updatedWidth))
@@ -1122,7 +1122,7 @@ namespace System.Windows.Controls
         private DataGridRow _owner;
         private ContainerTracking<DataGridCell> _tracker;
         private bool _syncingIsSelected;                    // Used to prevent unnecessary notifications
-        private const double _columnWidthStepSize = 10d;
+        private const double ColumnWidthStepSize = 10d;
 
         #endregion
     }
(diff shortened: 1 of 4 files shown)


src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Controls/DataGridCell.cs
```diff
@@ -1100,6 +1122,7 @@ private bool NeedsVisualTree
         private DataGridRow _owner;
         private ContainerTracking<DataGridCell> _tracker;
         private bool _syncingIsSelected;                    // Used to prevent unnecessary notifications
+        private const double _columnWidthStepSize = 10d;
```
lindexi, line 1125: As the coding-style said, we should use PascalCasing to name all our constant local variables and fields.

See https://github.com/dotnet/runtime/blob/main/docs/coding-guidelines/coding-style.md

src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Controls/DataGridCell.cs
```diff
@@ -964,6 +964,28 @@ protected override void OnPreviewKeyDown(KeyEventArgs e)
         /// </summary>
         protected override void OnKeyDown(KeyEventArgs e)
         {
+            if ((Keyboard.Modifiers & ModifierKeys.Alt) == ModifierKeys.Alt)
+            {
+                DataGridLength updatedWidth = new DataGridLength();
+
+                if (e.Key == Key.Right)
+                {
+                    updatedWidth = new DataGridLength(this.Column.ActualWidth + _columnWidthStepSize);
```
lindexi, line 973: As the coding-style said, we avoid `this.` unless absolutely necessary.

See https://github.com/dotnet/runtime/blob/main/docs/coding-guidelines/coding-style.md

anjalisheel-wpf: Pushed changes as per the comments - adapted PascalCasing.

singhashish-wpf (approved):

src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Controls/ListBox.cs
```diff
@@ -350,6 +350,30 @@ protected override void OnKeyDown(KeyEventArgs e)
                 case Key.Down:
                 case Key.Right:
                     {
+                    if ((Keyboard.Modifiers & ModifierKeys.Alt) == ModifierKeys.Alt)
```
SamBent, line 353: Same here - test that only Alt is pressed.

src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Controls/ListBox.cs
```diff
@@ -350,6 +350,30 @@ protected override void OnKeyDown(KeyEventArgs e)
...
+                        {
+                            if(e.OriginalSource is GridViewColumnHeader gridViewColumnHeader)
+                            {
+                                if (key == Key.Left)
+                                {
+                                    if(gridViewColumnHeader.Column.ActualWidth > 0)
+                                    {
+                                        gridViewColumnHeader.Width = gridViewColumnHeader.Column.ActualWidth - ColumnWidthStepSize;
+                                        gridViewColumnHeader.UpdateColumnHeaderWidth(gridViewColumnHeader.Width);
```
SamBent, line 362: This is different from the mouse code, which only calls UpdateColumnHeaderWidth (without setting header.Width directly).  Why?   I'd expect this to do the same.

src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Controls/DataGridCell.cs
```diff
@@ -964,6 +964,28 @@ protected override void OnPreviewKeyDown(KeyEventArgs e)
...
+                {
+                    updatedWidth = new DataGridLength(this.Column.ActualWidth - ColumnWidthStepSize);
+                }
+
+                if (Column.CanColumnResize(updatedWidth))
+                {
+                    this.Column.SetCurrentValue(DataGridColumn.WidthProperty, updatedWidth);
+                }
+
+                e.Handled = true;
```
SamBent, line 985: You should only do these lines when e.Key is Right or Left.  As it stands, typing Alt+Up (for example) results in setting column width to 0 (default DataGridLength) and handling the event, both of which are wrong.

SamBent (changes requested):

src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Controls/ListBox.cs
anjalisheel-wpf: In Mouse we are capturing mouse position and that is giving us new width and in mouse it can increase or decrease by any width, whereas for keyboard we will be resizing by a fixed size whenever correct key combination is hit.

batzen (changes requested):

src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Controls/DataGridCell.cs
```diff
@@ -964,6 +964,28 @@ protected override void OnPreviewKeyDown(KeyEventArgs e)
...
+                    updatedWidth = new DataGridLength(this.Column.ActualWidth + ColumnWidthStepSize);
+                }
+                else if (e.Key == Key.Left)
+                {
+                    updatedWidth = new DataGridLength(this.Column.ActualWidth - ColumnWidthStepSize);
+                }
+
+                if (Column.CanColumnResize(updatedWidth))
+                {
+                    this.Column.SetCurrentValue(DataGridColumn.WidthProperty, updatedWidth);
```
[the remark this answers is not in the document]
anjalisheel-wpf: No, it does not enter inside for any other key combination.

src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Controls/DataGridCell.cs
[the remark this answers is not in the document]
anjalisheel-wpf: Yes, only when the correct combination is pressed then only it will enter inside the condition and make changes accordingly.

PR #6060, opened 2022-02-03
Replace resource constants with properties in WindowsBase

ThomasGoulet73:
Contributes to dotnet/wpf#1

Depends on #6059

## Description
Replace the use of resource constants with properties in WindowsBase. This follows the standard of other .Net projects, like [dotnet/runtime](https://github.com/dotnet/runtime).

There are a bunch of files that are shared with other projects. I temporarily changed the use of constants with using nameof of properties in WindowsBase in this commit: 86ef9fc8510ca3c787116a45509cc902a7687b5d.

## Customer Impact
None.

## Regression
No.

## Testing
I tested a few resources and made sure that the generated code in the property getter is the same as the original code.

## Risk
None.


ThomasGoulet73:
I changed shared files to use properties even in projects that do not use properties by adding properties to SR. I think this is better than my initial approach of using nameof because it makes it easier to remove workaround when changing other projects to use properties because the properties will conflict with the generated properties.

I think that this PR is now ready to review. This PR should have a pretty low risk of breaking something.

RussKie (approved):
:shipit:

anjalisheel-wpf (approved):

PR #6063, opened 2022-02-03
ToolTips in popups (or any context where capture is active) (#5931)

### LINKED ISSUE #5716: Comboboxitem tooltip bug

singhashish-wpf:
Fixes #5716

## Description

The new 6.0 tooltip behavior computes a "safe area" in which the user can move the mouse without dismissing the tooltip.  It's computed with coordinates relative to the owner's PresentationSource (hwnd), but the mouse movements were tracked relative to the input provider's PresentationSource.  These are often the same, but not when the tooltip's owner comes from a Popup (or similar context that captures mouse input away from the normal input provider).  This dismisses the tooltip immediately - the logic thinks the mouse has left the safe area.

Fixed by converting the mouse position to the correct coordinate system - relative to the tooltip owner's PresentationSource.

## Customer Impact

Fixes a regression.

## Regression

Yes.

## Risk

Low.  Applies the correct coordinate transformation before checking the mouse position.

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Controls/PopupControlService.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Controls/PopupControlService.cs
@@ -817,18 +817,6 @@ namespace System.Windows.Controls
 
         private bool MouseHasLeftSafeArea()
         {
-            // if there is no SafeArea, the mouse didn't leave it
-            if (SafeArea == null)
-                return false;
-
-            // if the current tooltip's owner is no longer being displayed, the safe area is no longer valid
-            // so the mouse has effectively left it
-            DependencyObject owner = GetOwner(CurrentToolTip);
-            PresentationSource presentationSource = (owner != null) ? PresentationSource.CriticalFromVisual(owner) : null;
-            if (presentationSource == null)
-                return true;
-
-            // if the safe area is valid, see if it still contains the mouse point
             return !(SafeArea?.ContainsMousePoint() ?? true);
         }
 
--- a/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Controls/PopupControlService.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Controls/PopupControlService.cs
@@ -90,7 +90,7 @@ namespace System.Windows.Controls
                                     if (directlyOver != null)
                                     {
                                         // Process the mouse move
-                                        OnMouseMove(directlyOver, mouseReport);
+                                        OnMouseMove(directlyOver);
                                     }
                                 }
                             }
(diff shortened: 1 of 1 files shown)


gmanny:
I'm really sorry to probably just waste your time as I'm very new to this repo, but I would appreciate you going a bit into detail on how introducing this new method, which seems to not be used anywhere else, would fix the tooltip issue? Again, don't want to waste your time, so any pointer would suffice.

I've searched the repo for occurrences of the `ContainsMousePoint` string and it's not present anywhere else except in the declaration of this method. Not in branches `release/6.0` or `main`, nor in any other file changes in any other commit.

SamBent:
This comes from a bad merge.  The original commit in PR #5931 changed `MouseHasLeftSafeArea`, including replacing a call to `SafeArea.ContainsPoint` with a call to `SafeArea.ContainsMousePoint`,  changes to the signature of MouseHasLeftSafeArea, and corresponding changes to its callers.   All these got wiped out by two later commits, attempting to merge other changes from main.

@singhashish-wpf , please restore the original commit.  (In main, and in all servicing branches.)

gmanny:
Yeah, I guess this was actually conflicting with this commit: https://github.com/dotnet/wpf/pull/5931/commits/3c729a330cc86ee74022987993d8d8710b83a7ca

dipeshmsft (approved):

amirburbea:
Tooltips are still broken in 6.0.3

redseujac75:
> Tooltips are still broken in 6.0.3 - in some ways worse than before, is it possible the merge resolution was unsuccessful? Now tooltips seem to disappear everywhere. Run the popup-repro @ https://github.com/amirburbea/PopupRepro
> 
> Change the app between .NET5 & .NET6, check out tooltips in a ListBox (left) and in a ComboBox (right). It seems broken in both, whereas before it was only broken in the ComboBox.

Confirmed!

ALL Tooltips in C#, WPF project developed in VS 2022 Community .Net 6.0.2, vanish immediately after appearing. This happens with all controls (buttons, text boxes ...) with a tooltip. This problem appeared after the most recent VS2022 update to **.Net 6.0.3**. among other things. It wasn't there with previous .Net 6 versions at least as far as buttons and textboxes were concerned.

**Notice**: the issue only occurs when I set the display scale to 125% or more, not with 100%.

Mgamerz:
Can confirm borked tooltips. It's a serious detriment to my application on my laptop, which uses tooltips a ton, and has a high dpi screen (which >100% scaling).

PR #6097, opened 2022-02-09
Adds Support for ObservableCollection<T> and Collection<T> Range APIs from dotnet/runtime

### LINKED ISSUE #1887: Support Bulk Operations in ListCollectionView via ObservableCollection<T>
To support corefx [10752](https://github.com/dotnet/corefx/issues/10752), surfaced by corefx issue [38085](https://github.com/dotnet/corefx/issues/38085).

In .NET Core 3 Preview 7, the CoreFX team introduced support for ranges via `Collection<T>` and `ObservableCollection<T>`.  WPF is not able to handle how the new implementation of `AddRange` on `Collection<T>` as the add is done as a bulk insert at the end which changes the behavior of 'ListCollectionView` which does not support bulk operations and throws an https://github.com/dotnet/wpf/blob/master/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Data/ListCollectionView.cs#L2521.  

The net effect here is that a common helper method that is used in many WPF apps (`AddRange` to an `ObservableCollection<T>` no longer binds and the replacement method's behavior has a negative and observable side effect.

Adding issue for .NET 5 consideration

ahoefling:
Fixes #1887

## Description
Removes the `CollectionChanged` exceptions from `ListCollectionVIew` when the record count is not equal to 1. This PR is required to support https://github.com/dotnet/runtime/pull/65101. 

The change to dotnet/runtime as originally attempted in 2019 and it was discovered during preview builds that it broke WPF. I have spoken with both the .NET Team and the WPF Team and both teams are ready to accept this change. See https://github.com/dotnet/runtime/pull/65101 for a complete detail of history and API changes.

## Customer Impact
If this change is accepted it will allow developers to use `AddRange()` and `RemoveRange()` APIs on their `ObservableCollection<T>`. This will reduce the call stack from event invocations by N number of items that are being added to the collection. If an application is adding 5000 items to an `ObservableCollection<T>` currently it will invoke 5000 events, with this change it will only invoke 1 event.

If this change is not accepted, then https://github.com/dotnet/runtime/pull/65101 will not be accepted as it will break just about every WPF application that uses `ListCollectionView`.

## Regression
N/A

## Testing
I had problems running the wpf project locally and wasn't able to test as I would have liked to. I wanted to submit this PR and discuss with the community if someone could give me some help testing or instructions. I think the testing is giving me issues because I need a custom build of .NET 7 and the SDK project, then I need a custom build of WPF and finally will be able to test.

## Risk
As far as I understand this PR mitigates the risk of breaking changes from https://github.com/dotnet/runtime/pull/65101. I think it is important once we get a build we can test with to use some large WPF projects as a test solution like the NuGet Package Explorer. That was the project that originally caught the issue from the first attempt in 2019

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Data/ListCollectionView.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Data/ListCollectionView.cs
@@ -1681,8 +1681,6 @@ namespace System.Windows.Data
             if (args == null)
                 throw new ArgumentNullException("args");
 
-            ValidateCollectionChangedEventArgs(args);
-
             // adding or replacing an item can change CanAddNew, by providing a
             // non-null representative
             if (!_isItemConstructorValid)
@@ -2517,42 +2515,6 @@ namespace System.Windows.Data
         //------------------------------------------------------
 
         #region Private Methods
-
-        private void ValidateCollectionChangedEventArgs(NotifyCollectionChangedEventArgs e)
-        {
-            switch (e.Action)
-            {
-                case NotifyCollectionChangedAction.Add:
-                    if (e.NewItems.Count != 1)
-                        throw new NotSupportedException(SR.Get(SRID.RangeActionsNotSupported));
-                    break;
-
-                case NotifyCollectionChangedAction.Remove:
-                    if (e.OldItems.Count != 1)
-                        throw new NotSupportedException(SR.Get(SRID.RangeActionsNotSupported));
-                    break;
-
-                case NotifyCollectionChangedAction.Replace:
-                    if (e.NewItems.Count != 1 || e.OldItems.Count != 1)
-                        throw new NotSupportedException(SR.Get(SRID.RangeActionsNotSupported));
-                    break;
-
-                case NotifyCollectionChangedAction.Move:
-                    if (e.NewItems.Count != 1)
-                        throw new NotSupportedException(SR.Get(SRID.RangeActionsNotSupported));
-                    if (e.NewStartingIndex < 0)
-                        throw new InvalidOperationException(SR.Get(SRID.CannotMoveToUnknownPosition));
-                    break;
-
-                case NotifyCollectionChangedAction.Reset:
-                    break;
-
-                default:
-                    throw new NotSupportedException(SR.Get(SRID.UnexpectedCollectionChangeAction, e.Action));
-            }
-        }
-
-
         /// <summary>
         /// Create, filter and sort the local index array.
         /// called from Refresh(), override in derived classes as needed.


SamBent (changes requested):
All this does is remove the immediate NotSupportedException when ListCollectionView gets a range event.  It doesn't add support for those events.  To do that you'd need to
- add logic to `ProcessCollectionChanged` to do the right thing for range events (currently it assumes one item)
- add logic to methods it calls that also assume one item
- make similar changes to all handlers of CollectionChanged (there are dozens)
- make changes to secondary events and virtuals, e.g. `ItemsControl.OnItemsChanged`, `ItemContainerGenerator.ItemsChanged`
- track the range event through transformations like shaping (sorting/filtering/grouping), live-shaping, cross-thread notifications, etc.

Without that work, range events will get handled by logic that isn't prepared for them, resulting in unexpected behavior - crashes, data loss or corruption, incorrect/undesired layout, etc.  Better to throw NotSupportedException up front.

(Doing all that work is a huge project - probably months of work.)

airbreather:
> To do that you'd need to
>
> (lots of stuff)
>
> Without that work, range events will get handled by logic that isn't prepared for them, resulting in unexpected behavior - crashes, data loss or corruption, incorrect/undesired layout, etc. Better to throw NotSupportedException up front.
>
> (Doing all that work is a huge project - probably months of work.)

From my perspective as an outsider, I'm only **really** interested (near-term) in seeing the dotnet/runtime PR get unblocked.  I'd really like to start using efficient versions of the "Range" APIs in other situations (`Collection<T>` is a really good abstraction for more than just WPF, of course).  I get that this change, as it stands, cannot remove the blocker, and I completely support that decision.

But is a "full, proper" implementation **really** the only way to remove the blocker?  Our .NET Framework 4.8 application has been using a subclass of `ObservableCollection<T>` that rewrites bulk `CollectionChanged` events as `Reset` for over a decade now, and I have yet to observe even the slightest imperfection in the behavior of the system (and believe me, we have pushed WPF to places that it was **never** designed to go, with black magic wizardry that dates back at least as early as .NET Framework 4.0, and I think even going back to .NET Framework 3.5 before I joined that team).

So this may be a bit of a naïve question, but would it not be sufficient to resolve this issue by doing something similar on the receiving end in `ListCollectionView` itself?  "I got a change notification for multiple items, so I will pretend that it was a `Reset`"?

The main* downside that I can see to doing it that way is that `ListCollectionView` itself is `INotifyCollectionChanged`, and so if `ListCollectionView` starts raising `Reset` events for Range operations today, then that means "full, proper" support for Range operations tomorrow could have observable side-effects on external code that's written to expect a `Reset` event.

If that's the main downside to that idea, then I imagine that the "full, proper" support could include `AppContext` switch to bring back the `Reset` behavior to allow those applications to keep working?

If there's something more to it, then what other reasons might there be not to take that lighter approach?

**I can also concede that it would cause the overall system to show suboptimal performance in many important situations, such as when an `AddRange` adds 2 items to a 1,000,000-item list, especially when the `ListCollectionView` is doing live shaping and both items would be live-filtered out anyway.*

*Still, though, as annoying as it **will** be for someone who **will** find themselves in this situation, it would only affect **new** code, where some people (at worst) have to use a wrapper collection that modifies various calls to work around `ListCollectionView`'s imperfections... hey, wait a second, that sounds familiar...*

robertmclaws:
Yes, Microsoft, please make this happen!

I have been trying to get this fixed since [**August 13, 2016**](https://github.com/dotnet/runtime/issues/18087). Throwing exceptions in an event handler is a really bad decision that should have never been allowed to happen. That single decision has been blocking people from making the right architectural decisions for over 15 years. And it has kept WPF from being the amazing platform it was supposed to be.

People know that major .NET version releases break things... that's what SemVer is for. We will fix our own applications... and we'll help fix the runtime too.

@SamBent We don't know each other, but I have been in the .NET world since the .NET 1.0 preview. I am asking you, one .NET expert to another, please stop allowing perfect to be the enemy of good, and unblock us. Once this fix is in, we have several people that will work through the codebase and make the required improvements. It will not take months, it will take weeks. We can get it into the .NET 7 release if you just let us. 🙏🙏🙏

robertmclaws:
Also, please see this comment: https://github.com/dotnet/runtime/issues/18087#issuecomment-1114849195 MAUI is already working to get it integrated in the .NET 7 timeframe.

weltkante:
This PR does not unblock the dotnet runtime changes. Without WPF handling ranges properly, as soon as the dotnet runtime changes go in, WPF will still be regressed and it'll (probably) have to be rolled back again.

airbreather:
> This PR does not unblock the dotnet runtime changes.

But if the code were to be modified to address the issues that were identified in the review feedback further up the page, then it would, wouldn't it?  Otherwise, something similar could be said about roughly **any** PR that doesn't pass an initial review.

legistek:
No offense intended to anyone, but this is a terrible idea to implement this as is. As was pointed out both here and at https://github.com/dotnet/runtime/issues/18087#issuecomment-1117545927, the issue that forced the original PR adding range support to `Collection` to be reverted will still exist and arguably be worse if this change is implemented. 

The problem was that the methods (AddRange, etc.) being added to `Collection` were taking precedence over peoples' custom `Collection` and `ObservableCollection` extension methods (a very common workaround) of the same name.. In peoples' custom methods, they used `Reset` or just iterated - either way being WPF-safe - whereas the methods being proposed for `Collection` did not. 

The same thing will happen again if you go down this route, the only difference being the exception will not throw. However, peoples' applications will still break because - at best - the notifications won't get handled properly. `DataGrid`s and the like will add only one row, where multiple were intended, for example. Again, best case scenario. 

@SamBent (who obviously knows WPF better than anyone else here) is 100% right that this absolutely will not work without a complete overhaul of how WPF handles INCC events. This is not about the perfect being the enemy of the good. This is about not breaking 20 years of legacy code. Please do not attempt this. 

Adding the full range of INCC event handling to a built-in class is a worthy goal but neither `Collection`, nor `ObservableCollection`, nor certainly WPF itself, need to be messed with in order to do this. Make a new `ObservableList` class, make sure the modern frameworks support it, and make sure it warns the hell out of you if you try to use it with WPF.

weltkante:
> > This PR does not unblock the dotnet runtime changes.
> 
> But if the code were to be modified to address the issues that were identified in the review feedback further up the page, then it would, wouldn't it? If so, then something similar could be said about roughly **any** PR that doesn't pass an initial review.

Well, yes, if you turn it into a PR that does something entirely else, namely implement the missing support, then yes that will solve the problem. My comment was just highlighting, that the PR does not attempt that, thus is not unblocking this issue.

airbreather:
> this absolutely will not work without a complete overhaul of how WPF handles INCC events. This is not about the perfect being the enemy of the good. This is about not breaking 20 years of legacy code. Please do not attempt this.

Can you please explain the flaw in my comment above, https://github.com/dotnet/wpf/pull/6097#issuecomment-1116015460?  As you say, a common workaround is to have the **sender** raise the INCC event with `Reset` (when it really wants to do something unsupported), so it seems to follow that it wouldn't be fundamentally broken for the **receiver** to act as if the event was raised with `Reset` in those cases, no?

robertmclaws:
@legistek Creating an N+1 problem is not the solution here. Fixing WPF to do the right thing in a compatible way is the solution. You CAN fix WPF to maintain compatibility in the final .NET 7 release... but to do that you need to remove this exception so the people working to fix it can be unblocked.

legistek:
Hey @airbreather I thought your idea was really interesting on its own. All the potential downsides I can think of are things you allude to. Perhaps the biggest issue that would concern me is that depending on the size of the existing collection vs. the size of what's being added or removed, either iteration or reset could be horribly inefficient. Each developer who's faced this and made their own `AddRange` extension has had to decide which one to use for their unique situation. 

But the real problem lies with [#18087](https://github.com/dotnet/runtime/issues/18087). If `AddRange` were to be re-added to `Collection` and peoples' extension methods were again to get usurped, then we're still forcing new and untested behavior - without warning - on code that was previously written and presumably optimized based on different assumptions. Will [#18087](https://github.com/dotnet/runtime/issues/18087) plus your idea break applications as badly as this PR? Probably not, but I could absolutely see it causing huge efficiency headaches, excessive database querying, etc. , if applications that were previously using one method (iteration vs. reset) suddenly and inadvertently wind up using the other. 

On the other hand, if we instead had a totally opt-in `ObservableList` class, then your idea poses very little risk to legacy code and could indeed be a good interim step prior to full INCC support in WPF if that's eventually done. Either way, what's crucial is that we be able to _choose_ whether to opt into range support rather than change a ubiquitous base class like `Collection` causing severe side-effects that can only be detected at runtime. That's why [#18087](https://github.com/dotnet/runtime/issues/18087) is fatally flawed, and no modification to WPF will change that.

ThomasGoulet73:
In my opinion, removing `ValidateCollectionChangedEventArgs` is the very last step here. There is code throughout this repo (Including ListCollectionView) that needs to be fixed before removing the validation.

Take this line of code for exemple:
https://github.com/dotnet/wpf/blob/main/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Data/ListCollectionView.cs#L1748

It expects a single added item, which will silently break if you remove the validation without changing this code.

PR #6101, opened 2022-02-10
Pin assembly versions to prevent revving those in servicing releases

RussKie:
Resolves https://github.com/dotnet/core/issues/7172#issuecomment-1034105137
This is a complimentary work to https://github.com/dotnet/winforms/pull/6667

## Proposed changes

* Lock runtime and targeting pack assemblies at 6.0.2.0 so that any applications, libraries or NuGets built against 6.0.2 or later can work. 
This assembly versions pinning is similar to https://github.com/dotnet/runtime/blob/bbc766a7c65716305919bc691d696b1e46fa6f62/eng/Versions.props#L14 and https://github.com/dotnet/aspnetcore/blob/c0575788ecadd6e5cfeb4eab635c13b5fd433d37/Directory.Build.targets#L88.
* Add a validation step when we run packaging to verify that the assemblies we package into the transport package are correctly versioned, i.e., analyzers/generators are revved for every release for sxs, and those part of the ref pack are pinned to X.Y.2.0.

## Customer Impact

Currently WPF applications built against 6.0.2 won't run on 6.0 GA or 6.0.1
With the fix applications, libraries or NuGets built against 6.0.2 or later can work. Any applications built on 6.0.2 or later that need to target 6.0.0 or 6.0.1 will need to use the workaround provided in https://github.com/dotnet/core/issues/7176. It is not ideal, but is believed to have a lesser impact than going back to 6.0.0.0 versions, and have 6.0.2 broken.

## Regression? 

- Yes 

## Risk

- Medium. With this fix we'll stop supporting the folks that compiled against 6.0.2 or later and target 6.0.0 or 6.0.1 without the workaround.
In general, since 6.0.1 and 6.0.2 are security releases, customers are encouraged to update to the latest version.

### Before

6.0.1
image
image

6.0.2
image
image

### After

A [private build](https://dev.azure.com/dnceng/internal/_artifacts/feed/general-testing-internal/NuGet/runtime.win-x64.Microsoft.DotNet.Wpf.GitHub/6.0.3-servicing.22110.10/) for this fix:

image
image

### DIFF
--- a/eng/Versions.props
+++ b/eng/Versions.props
@@ -1,7 +1,25 @@
 <?xml version="1.0" encoding="utf-8"?>
 <Project ToolsVersion="4.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
   <PropertyGroup>
-    <VersionPrefix>6.0.3</VersionPrefix>
+    <MajorVersion>6</MajorVersion>
+    <MinorVersion>0</MinorVersion>
+    <PatchVersion>3</PatchVersion>
+    <VersionPrefix>$(MajorVersion).$(MinorVersion).$(PatchVersion)</VersionPrefix>
+
+    <!--
+      Set assembly version to align with major and minor version, as for the patches and revisions should be manually
+      updated per assembly if it is serviced.
+
+      Note, any components that aren't exposed as references in the targeting pack (like analyzers/generators) those should rev
+      so that they can exist sxs. The compiler relies on different version to change assembly version for caching purposes.
+
+      Because it is possible to build on a lower version and run on a higher version, we're locking the targeting pack and the runtime
+      assembly vesions to 6.0.2.
+      There's still a risk that some customers building on 6.0.2+ and trying to run on 6.0.0 and 6.0.1 will still have issue, but
+      those issues can be mitigated with a workaround described in https://github.com/dotnet/core/issues/7176.
+      -->
+    <AssemblyVersion>$(MajorVersion).$(MinorVersion).2.0</AssemblyVersion>
+
     <PreReleaseVersionLabel>servicing</PreReleaseVersionLabel>
     <PreReleaseVersionIteration>
     </PreReleaseVersionIteration>
--- a/packaging/Microsoft.DotNet.Wpf.GitHub/Check-AssemblyVersions.ps1
+++ b/packaging/Microsoft.DotNet.Wpf.GitHub/Check-AssemblyVersions.ps1
@@ -1,0 +1,66 @@
+[CmdletBinding(PositionalBinding=$false)]
+Param(
+  [Parameter(Mandatory=$True, Position=1)]
+  [string] $NuspecFile,
+  [Parameter(Mandatory=$True, Position=2)]
+  [string] $ExpectedAssemblyVersion,
+  [Parameter(Mandatory=$True, Position=3)]
+  [string] $IsServicingRelease,
+  [Parameter(ValueFromRemainingArguments=$true)][String[]] $properties
+)
+
+$servicingRelease = $null;
+[bool]::TryParse($IsServicingRelease, [ref]$servicingRelease) | Out-Null;
+
+[xml] $xmlDoc = Get-Content -Path $NuspecFile -Force;
+
+# 
+# Verify that components that are exposed as references in the targeting packs don't have their versions revved.
+# See https://github.com/dotnet/core/issues/7172#issuecomment-1034105137 for more details.
+[xml] $xmlDoc = Get-Content -Path $NuspecFile -Force;
+
+# Iterate over files that MUST NOT have their versions revved with every release
+$nonRevAssemblies = $xmlDoc.package.files.file | `
+    Where-Object { 
+            ($_.target.StartsWith('lib\') -or $_.target.StartsWith('ref\')) `
+                -and $_.target.EndsWith('.dll', [System.StringComparison]::OrdinalIgnoreCase) `
+                -and !$_.target.EndsWith('resources.dll', [System.StringComparison]::OrdinalIgnoreCase)
+        } | `
+    Select-Object -Unique src | `
+    Select-Object -ExpandProperty src;
+
+$nonRevAssemblies | `
+    sort-object | `
+    foreach-object {
+        $assembly = $_;
+        [string] $version = ([Reflection.AssemblyName]::GetAssemblyName($assembly).Version).ToString()
+
+        Write-Host "$assembly`: $version"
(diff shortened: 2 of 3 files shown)


packaging/Microsoft.DotNet.Wpf.GitHub/Directory.Build.targets
```diff
@@ -0,0 +1,44 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project>
+  <Import Project="$([MSBuild]::GetPathOfFileAbove('Directory.Build.targets', '$(MSBuildThisFileDirectory)../'))" />
+
+  <PropertyGroup>
+    <_PowerShellExe Condition="'$(_PowerShellExe)' == ''">C:\Windows\System32\WindowsPowerShell\v1.0\powershell.exe</_PowerShellExe>
```
ThomasGoulet73, line 6: Maybe use `%WINDIR%` instead of `C:\Windows` ?
```suggestion
    <_PowerShellExe Condition="'$(_PowerShellExe)' == ''">%WINDIR%\System32\WindowsPowerShell\v1.0\powershell.exe</_PowerShellExe>
```

packaging/Microsoft.DotNet.Wpf.GitHub/Check-AssemblyVersions.ps1
```diff
@@ -0,0 +1,66 @@
...
+
+$revAssemblies | `
+    sort-object | `
+    foreach-object {
+        $assembly = $_;
+        [string] $version = ([Reflection.AssemblyName]::GetAssemblyName($assembly).Version).ToString()
+
+        Write-Host "$assembly`: $version"
+        if ($servicingRelease -and [string]::Equals($version, $ExpectedAssemblyVersion)) {
+            throw "$assembly is not versioned correctly. '$version' is not expected."
```
ThomasGoulet73, line 63: I think this should be the same error message as [here](https://github.com/dotnet/wpf/pull/6101/files#diff-5260553429151724d0c7833a4c405ce7865877280619ed64356fa7b95cc543b9R40).
```suggestion
            throw "$assembly is not versioned correctly. Expected: '$ExpectedAssemblyVersion', found: '$version'."
```

RussKie: `$ExpectedAssemblyVersion` == `$version` here. This checks that these two are **not** the same. Thus this message is wrong: "_D:\a\_work\1\s\artifacts\packaging\Release\x64\Microsoft.DotNet.Wpf.GitHub\lib\net6.0\PresentationCore.dll is not versioned correctly. Expected: '6.0.2.0', found: '6.0.2.0'._"

RussKie:
Here's a test package: https://dev.azure.com/dnceng/internal/_artifacts/feed/general-testing-internal/NuGet/runtime.win-x64.Microsoft.DotNet.Wpf.GitHub/6.0.3-servicing.22110.10/

singhashish-wpf (approved):

packaging/Microsoft.DotNet.Wpf.GitHub/Check-AssemblyVersions.ps1
ThomasGoulet73: Oh you're right, sorry about that.

dotMorten:
> Any applications built on 6.0.2 or later that need to target 6.0.0 or 6.0.1 will need to use the workaround provided in https://github.com/dotnet/core/issues/7176. It is not ideal, but is believed to have a lesser impact than going back to 6.0.0.0 versions, and have 6.0.2 broken.

Is it really? In my mind 6.0.2 is already broken. Why can't you roll the version back to 6.0.0?

eng/Versions.props
```diff
@@ -1,7 +1,25 @@
...
+
+      Note, any components that aren't exposed as references in the targeting pack (like analyzers/generators) those should rev
+      so that they can exist sxs. The compiler relies on different version to change assembly version for caching purposes.
+
+      Because it is possible to build on a lower version and run on a higher version, we're locking the targeting pack and the runtime
+      assembly vesions to 6.0.2.
+      There's still a risk that some customers building on 6.0.2+ and trying to run on 6.0.0 and 6.0.1 will still have issue, but
+      those issues can be mitigated with a workaround described in https://github.com/dotnet/core/issues/7176.
+      -->
+    <AssemblyVersion>$(MajorVersion).$(MinorVersion).2.0</AssemblyVersion>
```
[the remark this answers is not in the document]
batzen: Version here must be 6.0.0.0 or simply 6.0. At least from my point of view and past experience with version pinning.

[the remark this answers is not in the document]
RussKie: > Same comment as on the WinForms PR, using the `.2` patch version should be conditioned on whether we're building for 6.0, and should not apply to 7.0 and forwards.

This is a 6.0 branch. For the 7.0 branch it is locked at Major.Minor.0.0, see https://github.com/dotnet/wpf/pull/6099.
Is there a situation where you expect 6.0 spill into another branch?

RussKie:
> Is it really? In my mind 6.0.2 is already broken. Why can't you roll the version back to 6.0.0? We now don't have to ensure people have .NET 6 installed, but also have the very latest version of it installed. People have to change their build systems or code just to use a class library built with the latest VS.

If we roll back to 6.0.0.0 - we will have 6.0.2 release broken, and any artifacts produced against 6.0.2 will require a recompilation (app, dlls, NuGets).
If we lock at 6.0.2.0 - we prevent running 6.0.2+ artifacts on 6.0.0 and 6.0.1 releases. However both 6.0.1 and 6.0.2 are security releases, and customers (especially enterprise) are encouraged to update to the latest version, and over time we don't expect may 6.0.0 or 6.0.1 installations.
Unfortunately this is a lose-lose situation, and we're choosing the long term viability.

dsplaisted:
@dotMorten To add to what @RussKie said:

- If we revert to the version number 6.0.0, then all DLLs built using the 6.0.2 reference assemblies will be broken.  I don't think there would be any other workaround besides rebuilding, so any NuGet packages published like this would be irreversibly broken and would need to be updated.
- If we leave the version number at 6.0.2, there is a workaround that lets you build an app or library that can run on 6.0.0 or 6.0.1, documented here: https://github.com/dotnet/core/issues/7176
- Whichever change we make, it won't be until next month's release, so by then a lot more of the runtimes should have been updated to at least 6.0.2 anyway.

So that's some of the reasoning behind why we think the least bad thing to do is to probably leave the version numbers at 6.0.2.

RussKie:
Approved over email

PR #6122, opened 2022-02-14
Remove or replace policheck violations from the code

dipeshmsft:
Removes words that are policheck violations : "crazy", etc.

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/PresentationCore/MS/Win32/UnsafeNativeMethodsPenimc.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationCore/MS/Win32/UnsafeNativeMethodsPenimc.cs
@@ -154,7 +154,7 @@ namespace MS.Win32.Penimc
             // Instantiating PimcManager using "new PimcManager()" results
             // in calling CoCreateInstanceForApp from an immersive process
             // (like designer). Such a call would fail because PimcManager is not
-            // in white list for that call. Hence we call CoCreateInstance directly.
+            // in allow list for that call. Hence we call CoCreateInstance directly.
             // Note: Normally WPF is not supported for immersive processes
             // but designer is an exception.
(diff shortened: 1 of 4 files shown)


vishalmsft (approved):

src/Microsoft.DotNet.Wpf/src/PresentationFramework/MS/Internal/Data/PropertyPathWorker.cs
```diff
@@ -1468,13 +1468,13 @@ private bool MatchIndexerParameters(PropertyInfo pi, IndexerParameterInfo[] aryI
                 {
                     // We can't recognize such properties in general, but we can
                     // recognize the most common cases - properties declared by .Net
-                    // types on a whitelist.
+                    // types on a allowlist.
```
RussKie, line 1471:
```suggestion
                    // types on an allowlist.
```

src/Microsoft.DotNet.Wpf/src/PresentationCore/MS/Win32/UnsafeNativeMethodsPenimc.cs
```diff
@@ -154,7 +154,7 @@ private static IPimcManager3 CreatePimcManager()
             // Instantiating PimcManager using "new PimcManager()" results
             // in calling CoCreateInstanceForApp from an immersive process
             // (like designer). Such a call would fail because PimcManager is not
-            // in white list for that call. Hence we call CoCreateInstance directly.
+            // in allow list for that call. Hence we call CoCreateInstance directly.
```
RussKie, line 157:
```suggestion
            // in the allowlist for that call. Hence we call CoCreateInstance directly.
```

RussKie (approved):
👍 with suggestions

RussKie (approved):

PR #6155, opened 2022-02-19
Command Base class Added (implements ICommand)

### LINKED ISSUE #6151: No Base Implementation For ICommand Interface  for WPF apps
Implementation needed for Command Base Class That works for WPF Command Binding as it works for Xamarin Forms

Ibrahimogod:
Fixes #6151 

## Description
WeakEventManager Class Added to manage Events,
ICommand Interface Implemented with Command Class to be Base Class That takes an Action to Execute 

## Customer Impact

make it easy for developer to use MVVM with Command Binding in WPF

## Testing
Tested Successfully on a wpf app

## Risk
Just make sure you don't have a class with "Command" where you Use System.Windows.Input

### DIFF
--- a/eng/WpfArcadeSdk/tools/ApiCompat.targets
+++ b/eng/WpfArcadeSdk/tools/ApiCompat.targets
@@ -21,7 +21,7 @@
 
   <PropertyGroup>
     <!-- Turn off default API Compat targets in favor of WPF ones. -->
-    <RunApiCompat>false</RunApiCompat>
+    <RunApiCompat>true</RunApiCompat>
   </PropertyGroup>
 
   <ItemGroup>
--- a/src/Microsoft.DotNet.Wpf/src/PresentationCore/PresentationCore.csproj
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationCore/PresentationCore.csproj
@@ -356,6 +356,8 @@
     <Compile Include="System\Windows\IContentHost.cs" />
     <Compile Include="System\Windows\IDataObject.cs" />
     <Compile Include="System\Windows\IInputElement.cs" />
+    <Compile Include="System\Windows\Input\Command\Command.cs" />
+    <Compile Include="System\Windows\Input\Command\LowEventManager.cs" />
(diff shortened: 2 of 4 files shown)


eng/WpfArcadeSdk/tools/ApiCompat.targets
```diff
@@ -21,7 +21,7 @@
 
   <PropertyGroup>
     <!-- Turn off default API Compat targets in favor of WPF ones. -->
-    <RunApiCompat>false</RunApiCompat>
+    <RunApiCompat>true</RunApiCompat>
```
lindexi, line 24: Could you turn off the `RunApiCompat` ?

lindexi:
Thank you @Ibrahimogod , but I think this change is worth discussing. Today, most third-party libraries have their own custom command types. I think it's hard to add the `Command` type in WPF repo to make everyone happy.

ThomasGoulet73:
I agree with @lindexi and there are also other consideration about adding this to WPF:
1. ICommand is not specific to WPF, it exists in the base class library of .Net. If a default implementation of ICommand were to exist, it would probably be in [dotnet/runtime](https://github.com/dotnet/runtime).
2. One utility of ICommand is that it's not UI dependent. Which makes it useful when you separate your view from your logic but when your logic needs something from the UI framework, an ICommand implementation from WPF in this case, the separation becomes less useful.
3. There are many third-party implementation of ICommand that are used a lot. My recommedation would be to use one of those. I would recommend [Prism](https://github.com/PrismLibrary/Prism)'s DelegateCommand.

Thank you for the PR though. I'll let the WPF team decide whether they want this or not.

singhashish-wpf:
@Ibrahimogod Thank you for the PR, your interest is highly appreciated in making wpf better and easy for developers. However, for the same reasons stated by @lindexi and @ThomasGoulet73 it doesn't seems appropriate to take this change. If you have any further reasoning around this, wpf team along with the wpf community will be happy to discuss it. Otherwise, we request you to close the PR and the related issue.

Issue #3546, opened 2020-08-05
XPS documents in .NET Core can't be opened


wstaelens:
Besides the [impersonation limitation I reported earlier](https://github.com/dotnet/core/issues/5021) we have also found that XPS documents coming from a virtual printer can't be opened in .NET Core. 

We first have to extract the XPS files, combine all part files to a single file and then create the XPS file again. Then we can open an XPS file in .NET Core. 

Is this by design or a bug? 

Because we don't have to do this in .NET Full Framework.

ryalanms:
Thank you for the report, @olowk.  Which versions of .NET Core and Windows are you using?

olowk:
I am using .net core 3.1 and Windows 10 updated to latest version.

wstaelens:
@GrabYourPitchforks @terrajobst  did you ever test this?
We are still having these issues after all these months. 

.NET SDK 5.0.202 
.NET runtime 5.0.5

ThomasGoulet73:
@wstaelens Could you provide a XPS file that you cannot open on .Net Core. I tried to reproduce your issue but it worked on my machine.

Thank you.

wstaelens:
Files were a bit too big, zip file 200mb can be found here on onedrive: https://1drv.ms/u/s!ArFv7v_tfAlngrQFZNhqp6HU-vkDMQ?e=WWhie5 

You will find:
```
00 pdf printed
01 captured incoming job
02 xps dump
03 xps modified for .net core
```

`00` is the pdf that I print to V4 driver.

`01` is the job that I captured as .prn.

`02` is an xps file using .content extension that we work with but you can just rename it to .xps. This XPS file can be opened and read in .NET Framework. Not in .NET Core / .NET 5

`03` is an xps file using .content extension that we work with but you can just rename it to .xps. This XPS file has been modified so that .NET Core / .NET 5 can open it!

When you go to `02` and `03`  and rename the .content(=.xps) to .zip you will see that in the received XPS job **[Content_Types].xml**, **DiscardControls.xml** and **FixedDocumentSequence.fdseq** are **directories** containing many separate files. (same for `/Documents/1/FixedDocument.fdoc` etc...)

.NET Framework is able to open an .XPS with all these separate files. <br />In .NET Core/.NET 5 we cannot open this. We first have to convert the many separate files into single files.

This is the problem that we are facing (besides XPS being terribly slow).

ThomasGoulet73:
Hey @wstaelens,

[...]

PR #6171, opened 2022-02-23
[Managed DWrite] Migrate Factory to managed

ThomasGoulet73:
Contributes to #5305.

## Description
This is the first part of #5305. I migrated only Factory to managed for now. This should allow to migrate DirectWriteForwarder to C# by migrating everything related to Factory because almost everything related to DWrite starts from Factory.

For now, the bindings to DWrite are generated by dotnet/ClangSharp and modified by hand (I plan to use dotnet/ClangSharp without modifications when I'm able to tune it like I want).

This is only a small chunk of the migration of DirectWriteForwarder to C# but hopefully this will allow to fully migrate it.

## Customer Impact

## Regression
No.

## Testing
Local build + CI + I tested a few apps to make sure that the text rendering is working.

## Risk
Low. For the most part, it is a copy of the C++ code manually rewritten to C# with as little changes as possible.


src/Microsoft.DotNet.Wpf/src/DirectWriteForwarder/CPP/DWriteWrapper/TextAnalyzer.cpp
```diff
@@ -20,7 +20,7 @@ namespace MS { namespace Internal { namespace Text { namespace TextInterface
         __in_ecount(length) const WCHAR* text,
         UINT32                           length,
         CultureInfo^                     culture,
-        Factory^                         factory,
+        IDWriteFactory*                  pDWriteFactory,
```
ThomasGoulet73, line 23: I changed `Factory` to `IDWriteFactory` to be able to remove Factory from DirectWriteForwarder. `Factory` was only used to call `DWriteFactoryAddRef` and then use the returned `IDWriteFactory`. I moved the responsibility of calling `DWriteFactoryAddRef` to the caller the method.

src/Microsoft.DotNet.Wpf/src/DirectWriteForwarder/CPP/DWriteWrapper/Factory.h
```diff
@@ -25,159 +25,18 @@ namespace MS { namespace Internal { namespace Text { namespace TextInterface
     /// <summary>
     /// The root factory interface for all DWrite objects.
     /// </summary>
-    private ref class Factory sealed : public CriticalHandle
+    private ref class InternalFactory abstract sealed
```
ThomasGoulet73, line 28: I renamed `Factory` to `InternalFactory` to avoid conflicts in PresentationCore since I added a new `Factory` class with the same namespace.

src/Microsoft.DotNet.Wpf/src/Common/src/Interop/Windows/Kernel32/Interop.GetModuleHandle.cs
```diff
@@ -0,0 +1,14 @@
...
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System;
+using System.Runtime.InteropServices;
+
+internal static partial class Interop
+{
+    internal static partial class Kernel32
+    {
+        [DllImport(Libraries.Kernel32, CharSet = CharSet.Unicode, SetLastError = true, ExactSpelling = true)]
```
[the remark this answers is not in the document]
kant2002: I would not recommend this in this specific repo given answer from Aaron https://github.com/dotnet/runtime/issues/60595#issuecomment-1043888326

I have my experiement with that package inside [WinForms](https://github.com/kant2002/winforms/tree/kant/dllimportgenerator) but would say that friction required, and risk given that this is glacially moving repo, is not jusified yet.

src/Microsoft.DotNet.Wpf/src/Common/src/Interop/Windows/Interop.HRESULT_FROM_WIN32.cs
```diff
@@ -0,0 +1,19 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System.Runtime.CompilerServices;
+
+internal static partial class Interop
+{
+    // Implementation of HRESULT_FROM_WIN32 macro
+    [MethodImpl(MethodImplOptions.AggressiveInlining)]
```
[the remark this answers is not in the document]
ThomasGoulet73: I took this code from dotnet/runtime [here](https://github.com/dotnet/runtime/blob/main/src/libraries/Common/src/Interop/Windows/Interop.HRESULT_FROM_WIN32.cs) which has `AggressiveInlining`. The only changes were returning 0 when errorCode <= 0 (The code in dotnet/runtime assumes that HRESULT_FROM_WIN32 is only called when errorCode is not an error, which is different than the macro).

hez2010:
So can someone review and merge this PR so we can push forward getting rid of C++/CLI in favor of making WPF AOT-compatible?

Symbai:
> So can someone review and merge this PR

Lol, you must be new here 😂. Welcome to the WPF repo where it takes months before a PR get's reviewed. So sit down, relax, take a coke or a coffee while waiting and enjoy your experience.

src/Microsoft.DotNet.Wpf/src/PresentationCore/MS/internal/Interop/DWrite/IDWriteFactory.cs
```diff
@@ -0,0 +1,58 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System;
+using System.Runtime.CompilerServices;
+
+namespace MS.Internal.Interop.DWrite
+{
+    internal unsafe struct IDWriteFactory : IUnknown
```
kant2002, line 9: I would not try to replicate COM inheritance here. If you look how RCW/CCW with ComWrapppers implemented [in runtime](https://github.com/dotnet/runtime/pull/63969/files#diff-24e48862e3b82f52e7fa04f22700b1c976a012bfeb08a246406f4e5ec579699bR53), or even [more lean](https://github.com/dotnet/winforms/pull/6685/files#diff-9e75e0c456be18af94a8953ab771dd422b5ca1836345043a1f9212fce8f02f93R13) in WinForms.

For example `IUnknown` interface not needed, since you can use regular casts instead of QueryInterface and `Marshal.AddRef(IntPtr)`/`Marshal.Release(IntPtr)`

src/Microsoft.DotNet.Wpf/src/PresentationCore/MS/internal/Interop/DWrite/IDWriteFactory.cs
```diff
@@ -0,0 +1,58 @@
...
+    {
+        public void** lpVtbl;
+
+        public int QueryInterface(Guid* riid, void** ppvObject)
+        {
+            return ((delegate* unmanaged[Stdcall]<IDWriteFactory*, Guid*, void**, int>)(lpVtbl[0]))((IDWriteFactory*)Unsafe.AsPointer(ref this), riid, ppvObject);
```
kant2002, line 15: `[Stdcall]` is default call convention, and no need to specify it here.

No need to use explicit types like `IDWriteFactory*` where you can just have `IntPtr` or `void*`

kant2002 (commented):
Left comments about RCW created.

kant2002:
For the @dipeshmsft or other team members.
 https://github.com/dotnet/windowsdesktop/blob/c7e42392fa9470cc0e19ff85811f3c0ed53e0972/pkg/windowsdesktop/sfx/Microsoft.WindowsDesktop.App.Ref.sfxproj#L69-L70

This is almost indication that DirectWriteForwarder can be sliced and diced as we like. If I understand correctly, no ref assemblies, no contract guarantees from MS.

src/Microsoft.DotNet.Wpf/src/PresentationCore/MS/internal/Interop/DWrite/IDWriteFactory.cs
ThomasGoulet73: I would prefer to leave it like this. I try to stay closer to the code generated by ClangSharp which maps the header files to C#. This offers some kind of type safety when using generics (Like [here](https://github.com/dotnet/wpf/pull/6171/files#diff-3945071e1113941167d91a8269971018f18b75ca6457fda26b6ac9baabafa1d3R24)). I also try to avoid using `Marshal.AddRef(IntPtr)/Marshal.Release(IntPtr)` because we can just call the methods implemented in the struct or in `IUnknown`.  The drawback is that the code is slightly larger but I think the type safety and cleaner code (In my opinion) is worth it.

src/Microsoft.DotNet.Wpf/src/PresentationCore/MS/internal/Interop/DWrite/IDWriteFactory.cs
ThomasGoulet73: Fixed in 557eea0394c625106e6d7f87fb35426987b47277. I tweaked ClangSharp to not add Stdcall.

miloush:
I would also like to point out that there is [DWriteCore](https://docs.microsoft.com/en-us/windows/win32/directwrite/dwritecore-overview) which should have compatible API (except the factory is createdy by `DWriteCoreCreateFactory`). It would be good if there could be a flag or something to use that one instead.

miloush:
@kant2002 @ThomasGoulet73 I don't have much experience with RCW development, so please bear with me - I would like to learn why this approach is better than just relying on the RCW generated by the runtime.

i.e. what are the benefits of

```c#
internal unsafe struct IDWriteFactory : IUnknown
    {
        public void** lpVtbl;

        public int QueryInterface(Guid* riid, void** ppvObject)
        {
            return ((delegate* unmanaged<IDWriteFactory*, Guid*, void**, int>)(lpVtbl[0]))((IDWriteFactory*)Unsafe.AsPointer(ref this), riid, ppvObject);
        }

        public uint AddRef()
        {
            return ((delegate* unmanaged<IDWriteFactory*, uint>)(lpVtbl[1]))((IDWriteFactory*)Unsafe.AsPointer(ref this));
        }

        public uint Release()
        {
            return ((delegate* unmanaged<IDWriteFactory*, uint>)(lpVtbl[2]))((IDWriteFactory*)Unsafe.AsPointer(ref this));
        }

        ...

        public int CreateTextAnalyzer(IDWriteTextAnalyzer** textAnalyzer)
        {
            return ((delegate* unmanaged<IDWriteFactory*, IDWriteTextAnalyzer**, int>)(lpVtbl[21]))((IDWriteFactory*)Unsafe.AsPointer(ref this), textAnalyzer);
        }
```

over

```C#
[ComImport, Guid(UuidOf.IDWriteFactory), InterfaceType(ComInterfaceType.InterfaceIsIUnknown)]
internal interface IDWriteFactory
{
    ...
    IDWriteTextAnalyzer CreateTextAnalyzer();
}
```

?

kant2002:
@miloush

> I would also like to point out that there is [DWriteCore](https://docs.microsoft.com/en-us/windows/win32/directwrite/dwritecore-overview) which should have compatible API

That's very nice and if that's closely aligned maybe that's best way to use it. Still we have to wait for the feedback from the team, which almost does not receive funding on this area.

> i.e. what are the benefits of

Benefits of hand-written RCW/CCW is that this make code linker friendly, and trimming friendly. Without that ILTrim does not know what COM interfaces it should brings. Also NativeAOT does not support built-in COM, so if this PR lands, and subsequent works too, then **maybe** can have WPF in NativeAOT. Right now because of that built-in COM trimming during R2R publish disabled for WPF apps, making them unnecessary larger.

ThomasGoulet73:
@dotnet/wpf-developers: Could this PR be considered for the next test pass?

I think it's a valuable PR since it's a first step for a migration to managed DirectWriteForwarder to allow linker and NativeAOT friendliness. I think it would be good to merge it early in .Net 7 previews to allow testing. I understand that this PR is rather large but it is the smallest changes I could think of for a starting point for the migration (I'm open to suggestions if someone can think of way to split this PR in smaller changes).

Symbai:
Looks like they don't. Several PR's got queued for test now but this not. @ThomasGoulet73 I hope you still got patience.

kant2002:
@AaronRobinsonMSFT obviously no concerns about future discussion here, I use this PR + sibling as baseline for my work WPF in NativeAOT https://github.com/dotnet/wpf/issues/5305#issuecomment-1148121102 (branch in this comment and screenshot below) so the PR which you find earlier(https://github.com/dotnet/wpf/pull/6587) is not something which I am using right now for my experiments.

src/Microsoft.DotNet.Wpf/src/PresentationCore/MS/internal/Text/TextInterface/Factory.cs
```diff
@@ -0,0 +1,342 @@
...
+            _wpfFontCollectionLoader = new FontCollectionLoader(
+                                                               fontSourceCollectionFactory,
+                                                               _wpfFontFileLoader
+                                                               );
+
+            _fontSourceFactory = fontSourceFactory;
+
+            IntPtr pIDWriteFontFileLoaderMirror = Marshal.GetComInterfaceForObject(
+                                                    _wpfFontFileLoader,
+                                                    typeof(IDWriteFontFileLoaderMirror));
```
AaronRobinsonMSFT, line 64: I've some doubts about a few of the low-level COM work here, not because it is wrong but because I think relying on built-in COM interop first is safer for the port and then it can be revisited. The counter is NativeAOT, but the above line is not AOT friendly. I think it might make more sense for the first pass to be reliant on the built-in COM interop and then go back through and make it AOT friendly. It will take longer but higher chances of success.

kant2002: I support any progress even if it’s not immediate and via some checkpoints. But that’s not my PR but @ThomasGoulet73 so let he speak his mind.

wstaelens:
> > So can someone review and merge this PR
> 
> Lol, you must be new here 😂. Welcome to the WPF repo where it takes months before a PR get's reviewed. So sit down, relax, take a coke or a coffee while waiting and enjoy your experience.

months? Years 😄 😭

wstaelens:
@singhashish-wpf  and what about https://github.com/dotnet/wpf/issues/3546 and https://github.com/dotnet/runtime/issues/51929 ? Will it be postponed again to .net 8 or will it be .net 36? 

or is it not valuable enough? 

https://github.com/dotnet/runtime/issues/78629 

@Symbai you have too wait again some months or years, it is too impactful 😆

driver1998:
Speaking of valuable and impactful things, https://github.com/dotnet/runtime/issues/18087 should also have some love... Also held back because "it will break WPF".

ThomasGoulet73:
@singhashish-wpf: I'm glad this is on the team's radar for .Net 8.

@wstaelens and @driver1998: I'm sorry that the issues you want fixed don't have updates but I have to agree with what @Symbai said here https://github.com/dotnet/wpf/discussions/6542#discussioncomment-4321221 that commenting on random issues/PRs with a request for status on unrelated issues/PRs is the wrong way to go about asking for status from the different .Net teams. It just clutters the comments with unrelated things.

miloush:
Also note https://github.com/dotnet/runtime/issues/66674. I imagine that once that is in place, we would want to take advantage of it for COM wrappers in WPF. Wouldn't it make more sense to focus the efforts on that generator to meet the goals we have here and then switch rather than migrate twice?

ThomasGoulet73:
@miloush I don't think this PR should depend on the work on COM source generator. This implementation is closer to the C++ code and mostly trim-friendly.

AaronRobinsonMSFT:
> @miloush I don't think this PR should depend on the work on the COM source generator.

Agree.

> This implementation is closer to the C++ code and mostly trim-friendly.

I would like to discourage this perspective for now and focus entirely on the C++/CLI to C# transition. Porting this over is going to get less trim-friendly before it gets better and the primary goal should be on moving away from C++/CLI not making anything trim-friendly. Having it as a secondary goal is fine, but removing C++/CLI should be the first goal in the process.

miloush:
I was suggesting that the COM source generator could be made to generate trim-friendly code - so that others can benefit from it too - and then it could be used instead of manually crafted signatures.

I completely support an extra step midway of C++/CLI to C# transition before additional changes.

Symbai:
> This PR is a valuable one and an impactful one too. There has been a long time since this is up, we plan to take this in .NET8 once the internal review and testing is complete.

Is this still on the radar for .NET 8 or did we missed the time frame once again? This PR is indeed very valuable.

src/Microsoft.DotNet.Wpf/src/Common/src/Interop/Windows/Kernel32/Interop.FreeLibrary.cs
```diff
@@ -0,0 +1,14 @@
...
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System;
+using System.Runtime.InteropServices;
+
+internal static partial class Interop
+{
+    internal static partial class Kernel32
+    {
+        [DllImport(Libraries.Kernel32, SetLastError = true, ExactSpelling = true)]
```
huoyaoyuan, line 11: Now `GeneratedDllImport` is renamed to `LibraryImport` and generally available to third party code. We should consider to use it like BCL.

src/Microsoft.DotNet.Wpf/src/PresentationCore/MS/internal/Text/TextInterface/DWriteLoader.cs
```diff
@@ -0,0 +1,100 @@
...
+
+            if (hKernel32 != IntPtr.Zero)
+            {
+                if (Kernel32.GetProcAddress(hKernel32, "AddDllDirectory") != IntPtr.Zero)
+                {
+                    // All supported platforms newer than Vista SP2 shipped with dwrite.dll.
+                    // On Vista SP2, the .NET servicing process will ensure that a MSU containing 
+                    // dwrite.dll will be delivered as a prerequisite - effectively guaranteeing that 
+                    // this following call to LoadLibraryEx(dwrite.dll) will succeed, and that it will 
+                    // not be susceptible to typical DLL planting vulnerability vectors.
```
huoyaoyuan, line 75: We no longer support Vista or lower OS for a long time. The fallback path can be removed.

src/Microsoft.DotNet.Wpf/src/PresentationCore/MS/internal/Text/TextInterface/DWriteLoader.cs
```diff
@@ -0,0 +1,100 @@
...
+                    // Fall back to using plain ol' LoadLibrary
+                    // There is risk that this call might fail, or that it might be
+                    // susceptible to DLL hijacking. 
+                    hDWriteLibrary = Kernel32.LoadLibraryW("dwrite.dll");
+                }
+            }
+
+            if (hDWriteLibrary != IntPtr.Zero)
+            {
+                DWriteCreateFactory = (delegate* unmanaged<int, void*, void*, int>)Kernel32.GetProcAddress(hDWriteLibrary, "DWriteCreateFactory");
```
huoyaoyuan, line 90: The `NativeLibrary` class provides decent support of `DllSearchPath`. P/Invoke to `LoadLibrary` and `FreeLibrary` can be removed then.

src/Microsoft.DotNet.Wpf/src/PresentationCore/ModuleInitializer.cs
```diff
@@ -19,6 +20,13 @@ public static void Initialize()
     {
         IsProcessDpiAware();
 
+        DWriteLoader.LoadDWrite();
+
+        AppDomain.CurrentDomain.ProcessExit += static (object sender, EventArgs e) =>
+        {
+            DWriteLoader.UnloadDWrite();
+        };
```
huoyaoyuan, line 28: Is this still required since AppDomain unloadability is no longer applicable? `AssemblyLoadContext.Unloading` could also be handled.

huoyaoyuan (commented):
Nice work! The comments are available as follow-up tasks to improve code quality to match the line with BCL.
This should get merged ASAP to unblock follow-up work.

src/Microsoft.DotNet.Wpf/src/PresentationCore/MS/internal/Text/TextInterface/DWriteLoader.cs
huoyaoyuan: No. Setting compatibility flag to Vista SP0 (The lowest supported on Win10) and the program can still see the functions.

src/Microsoft.DotNet.Wpf/src/Common/src/Interop/Windows/Kernel32/Interop.FreeLibrary.cs
ThomasGoulet73: Since this repo does not use `LibraryImport` yet, I don't want this large PR to be the first to use it. I think it would be best to do this in a smaller PR or in a PR the only migrates `DllImport` to `LibraryImport`.

src/Microsoft.DotNet.Wpf/src/PresentationCore/MS/internal/Text/TextInterface/DWriteLoader.cs
ThomasGoulet73: This code was recently removed from the C++/CLI sources by @kant2002 in #6538, so I'll make sure to also remove it in my PR.

src/Microsoft.DotNet.Wpf/src/PresentationCore/ModuleInitializer.cs
ThomasGoulet73: I wrote this code by checking what `atexit` was mapped to in C#/IL when compiling C++/CLI to managed (It also added the same handler to `AppDomain.CurrentDomain.DomainUnload` but since there is only one AppDomain I didn't bother to add it). I doesn't seem useful to me but I decided to add it since it seemed harmless to me and I wanted to stay as close as possible to the C++/CLI code.

src/Microsoft.DotNet.Wpf/src/PresentationCore/MS/internal/Text/TextInterface/DWriteLoader.cs
ThomasGoulet73: I was always under the impression that `NativeLibrary` should be used when it should work cross platform but since WPF is Windows only, `NativeLibrary` would only add unnecessary overhead instead of direct P/Invoke. (I don't know if this is the official recommendation, this is only from memory so it might be wrong)

ThomasGoulet73:
Thanks for the feedback @huoyaoyuan! I don't think your suggestions are blocker so I won't do anything for now (Except syncing my PR to main to remove dead Vista or lower code). I'm trying to keep my changes as small as possible and as close as a direct port as possible to facilitate the review. I'll make sure to go back to your suggestions when/if this PR is merged or if the WPF team wants these changes to merge this PR.

@singhashish-wpf or @dipeshmsft: Is it still in the team's plan to take this PR in .Net 8 ?

src/Microsoft.DotNet.Wpf/src/PresentationCore/MS/internal/Text/TextInterface/DWriteLoader.cs
AaronRobinsonMSFT: > but since WPF is Windows only, NativeLibrary would only add unnecessary overhead instead of direct P/Invoke. (I don't know if this is the official recommendation, this is only from memory so it might be wrong)

I would advocate for using `NativeLibrary` and official .NET APIs as opposed to calling OS specific functions. The overhead here is minimal at best and would be a one time cost relative to how often the function pointer itself. Calling the `LoadLibrary()` and `GetProcAddress` directly force a lot of thinking about flags and system state. Using `NativeLibrary` directly avoids any potential confusion and handles all security concerns.

ThomasGoulet73: Done in 49550d13a0069b9529630e5f37671825b02b1d73

src/Microsoft.DotNet.Wpf/src/PresentationCore/MS/internal/Text/TextInterface/DWriteLoader.cs
ThomasGoulet73: Done in 19bd765911aac689f883d5495834ca0de97f3705

singhashish-wpf:
UPDATE:
This was included in the recent test pass.
We are currently analyzing the new test failures due to this change. As of now we have 4 new DRT failures and 15 new Microsuite failure reports with this change.


| Feature     | Test Case                |
|-------------|--------------------------|
| 3D          | /Name="DrtBasic3D()"     |
|             | /Name="DrtElement3D()"   |
| Annotations | /Name="DrtAnnotations()" |
| Imaging     | /Name="DrtImaging()"     |




| Feature      | Test Case                                                                                           |
|--------------|-----------------------------------------------------------------------------------------------------|
| 3D           | /Name="AmbientLight" /SubArea="BVT"                                                                 |
|  Annotations | /Name="MenuSuite.menu1.fixed.inkstickynote" /SubArea="MenuSuite"                                    |
|              | /Name="MenuSuite.menu1.fixed.stickynote" /SubArea="MenuSuite"                                       |
|              | /Name="MultiNoteSuite.multinote3_5.fixed.inkstickynote" /SubArea="MultiNoteSuite"                   |
|              | /Name="MultiNoteSuite.multinote3_5.fixed.stickynote" /SubArea="MultiNoteSuite"                      |
|              | /Name="MultiSegmentSuite.multisegment12_4.fixed./fds=false.nonvisible" /SubArea="MultiSegmentSuite" |
|              | /Name="MultiSegmentSuite.multisegment12_4.fixed./fds=false.visible" /SubArea="MultiSegmentSuite"    |
|              | /Name="MultiSegmentSuite.multisegment12_4.fixed.nonvisible" /SubArea="MultiSegmentSuite"            |
|              | /Name="MultiSegmentSuite.multisegment12_4.fixed.visible" /SubArea="MultiSegmentSuite"               |
|              | /Name="PaginatorSuite.adp_getpage2" /SubArea="PaginatorSuite"                                       |
|              | /Name="StylingSuite.style8.fixed.inkstickynote" /SubArea="StylingSuite"                             |
|              | /Name="StylingSuite.style8.fixed.stickynote" /SubArea="StylingSuite"                                |
| DataServices | /Name="Element3DDatabind(BindingUnderA3DViewport)" /SubArea="Controls"                              |
| Editing      | /Name="IMEInsertDeleteTest_Japanese" /SubArea="IME"                                                 |
| Text         | /Name="Display_CT_FontEmbedding5-3" /SubArea="Display_CT_FontEmbeddingTests"                        |

kant2002:
Thank you guys!!

SmRiley:
> Thank you guys!!
Thank you very much for your contribution to promoting this feature. I am also very interested in this feature, but I don’t quite understand the above reply. Is the progress of this PR already at the final stage?

kant2002:
I can always negatively talk, how WPF team is moving slow, we still do not have all DRT ported, etc, etc, etc. It maybe because we have bad technical team, or other way around because we have bad management, or somebody from MS do not give enough money on the project. I simply don't know.

What I know from reply of Ashish is that this issue is getting attention, and we need only fix couple tests to start talking what else is needed. I really appreciate that even after long time of idling this issue may start moving. So I decide to say thanks to WPF team to paying attention to this PR.

driver1998:
This is the first baby step of the C++/CLI -> C# migration throughout WPF, still, good to hear things finally start moving. Thanks to the team!

pchaurasia14:
>  Is the progress of this PR already at the final stage?

@SmRiley - Well, it's hard to say right now. Once we investigate, isolate and fix the failing tests (either in the PR or the tests themselves), we will be in a better shape to consider the next steps with this PR. 

We want to ensure that we fix the known issues first and take this PR early on so it gives us enough room to troubleshoot any future compat problems. 

@kant2002 - I think some of the tests are already available on the [test](https://github.com/dotnet/wpf-test) repo (if you'd like to take a crack at them)

ThomasGoulet73:
I'm glad this PR is getting considered by the team.

@pchaurasia14 I tried running the tests locally but it didn't seem to work, I'll try again in a couple of days. Let me know if the team is able to isolate the bug. Thanks.

singhashish-wpf:
cc:// @harshit7962

harshit7962:
In [Annotations Test](https://github.com/dotnet/wpf-test/tree/main/src/Test/Annotations/DRT/Annotations), we are calling the `InitializeComponent` method on `FixedDocument1Class` [here](https://github.com/dotnet/wpf-test/blob/main/src/Test/Annotations/DRT/Annotations/pageviewer.xaml.cs#L91).
This makes subsequent calls to `factory.cs` and it's `CreteFontFile` method, and here we pass fontFileLoader as a nullptr, which in turn throws an ArgumentNullException.
On comparing it with the same method call in `Factory.cpp` [here](https://github.com/dotnet/wpf/blob/main/src/Microsoft.DotNet.Wpf/src/DirectWriteForwarder/CPP/DWriteWrapper/Factory.cpp#L133), we are passing `_wpfFontFileLoader` as the fontFileLoader. Will passing the same as the parameter here solve the issue? or did you omit it foreseeing some other failure?

cc:// @ThomasGoulet73

ThomasGoulet73:
Thank you @harshit7962. That does seem like a mistake, it was probably an oversight when trying to get it to build and then I forgot to change it.

I pushed a commit that fixes the null parameter, let me know if it fixes the tests.

I was able to run DrtAnnotations but I can't get it to fail, the tests passes before and after my last commit. DrtAnnotations also seem to pass in the CI (If it's the same thing?).

harshit7962:
DRTAnnotations and related tests are passing after your fix. We will be checking the other failures as of now.

src/Microsoft.DotNet.Wpf/src/PresentationCore/MS/internal/Text/TextInterface/Factory.cs
AaronRobinsonMSFT: Okay. Let's see how this goes. Converting C++ directly to C# isn't all that good of an idea in general because it tends to generate poor and unsafe C#. It should be a priority to change this prior to moving onto the next section though.

AaronRobinsonMSFT (approved):

pchaurasia14 (approved):

src/Microsoft.DotNet.Wpf/src/PresentationCore/MS/internal/Text/TextInterface/Factory.cs
```diff
@@ -0,0 +1,342 @@
...
+using MS.Internal.Interop;
+using MS.Internal.Interop.DWrite;
+using MS.Internal.Text.TextInterface.Interfaces;
+
+namespace MS.Internal.Text.TextInterface
+{
+    /// <summary>
+    /// The root factory interface for all DWrite objects.
+    /// </summary>
+    internal unsafe class Factory
```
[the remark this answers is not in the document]
ThomasGoulet73: I think this is a mistake. I think I was trying to reuse the class `NativeIUnknownWrapper` but Factory.cpp does more things in ReleaseHandle than what `NativeIUnknownWrapper.ReleaseHandle` does. I'll try to fix it soon and make it closer to C++ and we can clean it up later if needed.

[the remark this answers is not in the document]
ThomasGoulet73: I pushed a new commit where I redid in C# what `Factory.ReleaseHandle` did in C++. I took a different approach where `ReleaseHandle` is in a private class that inherits `NativeIUnknownWrapper` where it also unregisters and releases `FontFileLoader` and `FontCollectionLoader` in `ReleaseHandle`.

I verified that everything the `FontFileLoader` and `FontCollectionLoader` where unregistered and released when `Factory` is garbage collected in a sample application that forces it to be garbage collected.

singhashish-wpf:
All internal tests are green now for this PR, also the due process is complete. We are good to merge this one. Tagging @dotnet/dotnet-wpf-triage @dotnet/wpf-developers for any final comments or suggestions.

singhashish-wpf (approved):

singhashish-wpf:
Thank you @ThomasGoulet73 for all your hard work and patience! Apologies on behalf of the team for the delays. Thank you to all the contributors who commented, reviewed, debugged, and fixed test failures.

wstaelens:
Yay! Updates to WPF! 🥳

ThomasGoulet73:
Big thanks to the team (Even if there was a delay 😄) and to everyone who reviewed this PR!

psmulovics:
BIG one down - well done all who contributed, reviewed, merged, etc this one!

PR #6182, opened 2022-02-25
Update GenAPI additional parameter usage

ViktorHofer:
https://github.com/dotnet/arcade/commit/e44123d2796c8333cf90f90b51dfd9d04d224bea switched GenAPI to run in-proc as an msbuild task. Because of that, the usage is slightly different when additional parameters are passed in to GenAPI.


singhashish-wpf:
@ViktorHofer Could you please take care of the conflict?

ViktorHofer:
You already did so :)

singhashish-wpf (approved):

Issue #4768, opened 2021-06-29
UIElement.Measure is very slow.


deeprobin:
* .NET Core Version: 5.0.6
* Windows version: 21H1 (Build 19043.928)
* Does the bug reproduce also in WPF for .NET Framework 4.8?: Yes
* Is this bug related specifically to tooling in Visual Studio (e.g. XAML Designer, Code editing, etc...)?: No

 
 
 **Problem description:**
UIElement.Measure this very slow
 
 **Actual behavior:** 
![image](https://user-images.githubusercontent.com/10373452/123769736-e2cc5480-d8c9-11eb-9368-5540156292f4.png) 

 **Expected behavior:**
Fast measuring for WPF rendering.
If there is nothing more that can be optimized in the measure method, it might make sense to cache the measures according to certain parameters, since this is very slow with large visual trees. 
Compared to Windows Forms, Windows Forms is superior in terms of performance here.

Would it not make sense to write the measure method in native C++ somehow to optimize the performance accordingly or is the overhead here too large?
 
 **Minimal repro:**
```xaml
<Grid>
    <Grid.ColumnDefinitions>
        <ColumnDefinition Width="*"/>
        <!-- ... Many column definitions -->
    </Grid.ColumnDefinitions>
    <Grid.RowDefinitions>
        <RowDefinition Heigth="*"/>
        <!-- ... Many row definitions -->
    </Grid.RowDefinitions>
    <Button Grid.Row="0" Grid.Column="0">Test</Button>
    <Button Grid.Row="1" Grid.Column="0">Test</Button>
    <Button Grid.Row="0" Grid.Column="2">Test</Button>
    <Button Grid.Row="2" Grid.Column="3">Test</Button>
    <Button Grid.Row="12" Grid.Column="0">Test</Button>
    <Button Grid.Row="7" Grid.Column="8">Test</Button>
    <!-- ... big visual tree -->
</Grid>```

deeprobin:
![image](https://user-images.githubusercontent.com/10373452/123772721-596a5180-d8cc-11eb-9613-489d9427db4b.png)
This are the hot spots of UIElement.Measure. Would it perhaps be useful to take a closer look at the TextAnalyzer performance-wise?

There is the fastest font renderer in the Rust language. Would make sense to create bindings for it and let [fontdue](https://github.com/mooman219/fontdue) do the work. 
Or what do you think?

Symbai:
[...]

PR #5571, opened 2021-10-24
Adding GreaterThanZero to DoubleUtil.cs

lindexi:
## Description

Benchmark code:

```csharp
    public class Program
    {
        static void Main(string[] args)
        {
            BenchmarkRunner.Run<Program>();
        }

        [Benchmark]
        [ArgumentsSource(nameof(GetArguments))]
        public bool TestGreaterThanZero(double value)
        {
            return DoubleUtil.GreaterThanZero(value);
        }

        [Benchmark]
        [ArgumentsSource(nameof(GetArguments))]
        public bool TestGreaterThanZero2(double value)
        {
            return DoubleUtil.GreaterThanZero2(value);
        }

        [Benchmark]
        [ArgumentsSource(nameof(GetArguments))]
        public bool TestGreaterThan(double value)
        {
            return DoubleUtil.GreaterThan(value, 0);
        }

        public IEnumerable<double> GetArguments()
        {
            const int count = 10;
            for (int i = 0; i < count; i++)
            {
                yield return i;
            }

            for (double i = 0; i < count; i++)
            {
                yield return i / count;
            }

            for (int i = 0; i < count; i++)
            {
                yield return -i;
            }

            for (double i = 0; i < count; i++)
            {
                yield return -i / count;
            }
        }
    }

    public static class DoubleUtil
    {
        /// <summary>
        /// GreaterThanZero - Returns whether or not the value is greater than zero
        /// </summary>
        /// <param name="value"></param>
        /// <returns></returns>
        public static bool GreaterThanZero2(double value)
        {
            // IsZero = Math.Abs(value) < 10.0 * DBL_EPSILON
            // value > 0 && !IsZero(value) == value > 0 && !(Math.Abs(value) < 10.0 * DBL_EPSILON)
            // = !(value < 10.0 * DBL_EPSILON)
            return value >= 10.0 * DBL_EPSILON;
        }

        public static bool GreaterThanZero(double value)
        {
            if (value > 0)
            {
                // AreClose(double value1, double value2)
                // = (|value1-value2| / (|value1| + |value2| + 10.0)) < DBL_EPSILON
                //
                // AreClose(value, 0) = (|value-0| / (|value| + 0 + 10.0)) < DBL_EPSILON
                // = value / (value + 10.0) < DBL_EPSILON
                double eps = (value + 10.0) * DBL_EPSILON;
                double delta = value;
                return !((-eps < delta) && (eps > delta));
            }
            else
            {
                return false;
            }
        }

[...]

PR #6219, opened 2022-03-07
Optimize Measure and Arrange

deeprobin:
See #4768 (this does not fix the issue - only supports the performance problem)

## Description

In this PR I try to improve the Measure and Arrange method performance-wise. I have also moved certain code blocks to other methods for readability.

## Testing

I unfortunately could not test it due to build issues.

### ContextLayoutManager.From (Risk 1/10)
I have reduced the ContextLayoutManager.From calls.
Normally this should fit but I can't confirm this 100%.

### DoubleUtil.AreClose changed behavior (Risk 6/10)
I have changed the behavior of DoubleUtil.AreClose. This definitely needs to be tested.

Since this method is called frequently I tried to reduce the corresponding JIT codegen.
I managed to do this, but there is a certain risk, even if according to the summary the method should not return an exact result.

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/UIElement.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/UIElement.cs
@@ -552,7 +552,7 @@ namespace System.Windows
         /// giving it opportunity to compute its DesiredSize.<para/>
         /// This method will return immediately if child is not Dirty, previously measured
         /// and availableSize is the same as cached. <para/>
-        /// This method also resets the IsMeasureinvalid bit on the child.<para/>
+        /// This method also resets the IsMeasureInvalid bit on the child.<para/>
         /// In case when "unbounded measure to content" is needed, parent can use availableSize
         /// as double.PositiveInfinity. Any returned size is OK in this case.
         /// </remarks>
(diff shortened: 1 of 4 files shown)


src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/UIElement.cs
```diff
@@ -562,159 +562,169 @@ internal static void PropagateResumeLayout(Visual parent, Visual v)
...
+                    // Cache desired size
                     _desiredSize = desiredSize;
 
-                    //notify parent if our desired size changed (watefall effect)
-                    if (!MeasureDuringArrange
-                       && !DoubleUtil.AreClose(prevSize, desiredSize))
+                    // Notify parent if our desired size changed (waterfall effect), otherwise return
+                    if (MeasureDuringArrange || DoubleUtil.AreClose(prevSize, desiredSize)) return;
+
+                    GetUIParentOrICH(out var parent, out var contentHost); //only one will be returned
```
lindexi, line 690:
```suggestion
                    GetUIParentOrContentHost(out var parent, out var contentHost); //only one will be returned
```

src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/FrameworkElement.cs
```diff
@@ -4306,223 +4306,221 @@ protected sealed override Size MeasureCore(Size availableSize)
...
+            {
+                mm.maxHeight = RoundLayoutValue(mm.maxHeight, dpi.DpiScaleY);
+                mm.maxWidth = RoundLayoutValue(mm.maxWidth, dpi.DpiScaleX);
+                mm.minHeight = RoundLayoutValue(mm.minHeight, dpi.DpiScaleY);
+                mm.minWidth = RoundLayoutValue(mm.minWidth, dpi.DpiScaleX);
+            }
 
-                LayoutTransformData ltd = LayoutTransformDataField.GetValue(this);
+            LayoutTransformData ltd = LayoutTransformDataField.GetValue(this);
+            {
```
lindexi, line 4360: Can we remove the brackets?

src/Microsoft.DotNet.Wpf/src/Shared/MS/Internal/DoubleUtil.cs
```diff
@@ -57,12 +59,7 @@ internal static class DoubleUtil
         /// <param name="value2"> The second double to compare. </param>
         public static bool AreClose(double value1, double value2)
         {
-            //in case they are Infinities (then epsilon check does not work)
-            if(value1 == value2) return true;
-            // This computes (|value1-value2| / (|value1| + |value2| + 10.0)) < DBL_EPSILON
-            double eps = (Math.Abs(value1) + Math.Abs(value2) + 10.0) * DBL_EPSILON;
-            double delta = value1 - value2;
-            return(-eps < delta) && (eps > delta);
+            return Math.Abs(value1 - value2) < 10.0 * DBL_EPSILON;
```
lindexi, line 62: The bad news is I tested to replace Math.Abs, but the result will be changed. 

In my other PR, I have tried to do it that is  similar your code, but I failed. See https://github.com/dotnet/wpf/pull/5571

src/Microsoft.DotNet.Wpf/src/Shared/MS/Internal/DoubleUtil.cs
```diff
@@ -83,7 +80,7 @@ public static bool AreClose(double value1, double value2)
         /// <param name="value2"> The second double to compare. </param>
         public static bool LessThan(double value1, double value2)
         {
-            return (value1 < value2) && !AreClose(value1, value2);
+            return value1 < value2 && !AreClose(value1, value2);
```
lindexi, line 83: I think adding `()` makes the meaning clearer.

lindexi (commented):
Awesome, you make the code clearer.

lindexi:
> src\Microsoft.DotNet.Wpf\src\Shared\MS\Internal\DoubleUtil.cs(20,22): error CS0234: (NETCORE_ENGINEERING_TELEMETRY=Build) The type or namespace name 'Intrinsics' does not exist in the namespace 'System.Runtime' (are you missing an assembly reference?)

src/Microsoft.DotNet.Wpf/src/Shared/MS/Internal/DoubleUtil.cs
deeprobin: Will the result be changed essentially or are these just minor details? As described in the comment, the method is not expected to deliver exact results.

lindexi: @deeprobin Just minor details.

PR #6245, opened 2022-03-11
Fix setting DPI awareness

AustinWise:
## Description

Currently DPI awareness will be enabled *only if* a `DisableDpiAwarenessAttribute` attribute is applied to the entry assembly. Which does not make a lot of sense. This also does not match the behavior of .NET 6.

This PR changes the logic to always enable DPI awareness *unless* a `DisableDpiAwarenessAttribute` attribute is applied to the entry assembly.

## Customer Impact

WPF applications are not DPI aware. This can be seen by:

```
dotnet new wpf
dotnet run
```

And then checking the DPI Awareness column in Task Manager. On .NET 6 the DPI awareness is "System". In .NET 7 preview 1, it is "Unaware".

## Regression

This regression was introduced by #5765, which migrated the DPI awareness enablement code from C++ to C#.

## Testing

```
dotnet new wpf
```

Then replace the contents of `MainWindow.xml.cs` with:

```csharp
using System.Windows;
using System.Runtime.InteropServices;

namespace test
{
    /// <summary>
    /// Interaction logic for MainWindow.xaml
    /// </summary>
    {
        [DllImport("User32", EntryPoint = "IsProcessDPIAware", CharSet = CharSet.Auto, SetLastError = true)]
        internal static extern bool IsProcessDPIAware();

        public MainWindow()
        {
            InitializeComponent();
            this.Title = IsProcessDPIAware().ToString();
        }
    }
}
```

Results of running this program on different versions of WPF:

|.NET Version|title|
|--|--|
|6.0.3|True|
|7 Preview 1|False|
|This PR|True|

## Risk

Low.

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/PresentationCore/ModuleInitializer.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationCore/ModuleInitializer.cs
@@ -25,11 +25,18 @@ internal static class ModuleInitializer
 
     private static void IsProcessDpiAware()
     {
+        bool disableDpiAware = false;
+
         // By default, Application is DPIAware.
         Assembly assemblyApp = Assembly.GetEntryAssembly();
 
         // Check if the Application has explicitly set DisableDpiAwareness attribute.
         if (assemblyApp != null && Attribute.IsDefined(assemblyApp, typeof(System.Windows.Media.DisableDpiAwarenessAttribute)))
+        {
+            disableDpiAware = true;
+        }
+
+        if (!disableDpiAware)
         {
             // DpiAware composition is enabled for this application.
             SetProcessDPIAware_Internal();


src/Microsoft.DotNet.Wpf/src/PresentationCore/ModuleInitializer.cs
```diff
@@ -26,17 +26,19 @@ public static void Initialize()
...
     {
         // By default, Application is DPIAware.
-        Assembly assemblyApp = Assembly.GetEntryAssembly();
+        bool dpiAware = true;
 
         // Check if the Application has explicitly set DisableDpiAwareness attribute.
+        Assembly assemblyApp = Assembly.GetEntryAssembly();
         if (assemblyApp != null && Attribute.IsDefined(assemblyApp, typeof(System.Windows.Media.DisableDpiAwarenessAttribute)))
         {
-            // DpiAware composition is enabled for this application.
```
ThomasGoulet73, line 34: Could you keep the comments at the same place as the previous C++ code? You can use this code for reference:
https://github.com/dotnet/wpf/blob/560f243910f407df33fd5cdf06e2a857684f4bc7/src/Microsoft.DotNet.Wpf/src/DirectWriteForwarder/main.cpp#L184-L205

ThomasGoulet73 (commented):
Thank you! This is definitely a mistake on my part and I believe that this is the right fix.

src/Microsoft.DotNet.Wpf/src/PresentationCore/ModuleInitializer.cs
AustinWise: I changed it it to match the original C++ code more closely. I renamed `bDisableDpiAware` to `disableDpiAware`, so that it matches typical C# conventions.

dipeshmsft (approved):

PR #6252, opened 2022-03-13
Optimize the code of OSVersionHelper

lindexi:
## Description

We can not run the WPF application in the system which version lower than win7. And it will return fast without call the RtlVerifyVersionInfo method.

## Customer Impact

Improve performance?

## Regression

None.

## Testing

Just CI.

## Risk

Low.


dipeshmsft:
@lindexi Do we have a benchmark for this?

lindexi:
@dipeshmsft Sorry, I do not know how to test it. Can you tell me?

PR #6254, opened 2022-03-14
Eliminate memory copy when reading font data

### LINKED ISSUE #6236: Poor performance for `<Resource>` fonts
* .NET Version: 6.0
* Windows version: Windows 10 21H2 (19044.1526)
 
 **Problem description:**

When using a font that is embedded as a `<Resource>` (see [fonts as resource items](https://docs.microsoft.com/en-us/dotnet/desktop/wpf/advanced/packaging-fonts-with-applications?view=netframeworkdesktop-4.8#adding-fonts-as-resource-items)), the performance of rendering text can be extremely poor.

This was first noted in a production .NET 4.8 WPF Application; a detailed writeup of the problem and a solution that was found is at https://faithlife.codes/blog/2019/06/improving-wpf-text-display-performance/.
 
 **Actual behavior:**

dotTrace shows that very large amounts of time are spent in `MS.Internal.Text.TextInterface.FontFileStream.ReadFileFragment`. Additionally, large amounts of memory are allocated and the program spends a lot of time in GC.

image
 
 **Expected behavior:**

Rendering a `<Resource>` font is extremely efficient because the resource is already mmap'ed into memory and returning its data could be as simple as a pointer addition.
 
 **Minimal repro:**

An example repro is at https://github.com/bgrainger/EmbeddedFontPerformance. Clone the repo, build the code, and scroll the text box. You'll notice that it can take several seconds to draw the next screen of text (depending on how large the window is).

The following text complements that repo and describes how it reproduces the problem:

This reproduces most easily with the [Noto Sans CJK font](https://github.com/googlefonts/noto-cjk) (perhaps because of the large number of glyphs that have to be read?).

Embed the font:

```xml
  <ItemGroup>
    <Resource Include="NotoSansCJKtc-Regular.otf" />
  </ItemGroup>
```

Then create a text box that references the font and contains a large amount of Chinese text:

```xml
    <Grid>
        <ScrollViewer>
            <TextBlock FontFamily="./#Noto Sans CJK TC" TextWrapping="Wrap">
            <!-- place Chinese text here -->
            </TextBlock>
        </ScrollViewer>
    </Grid>
```

Run the app and scroll the ScrollViewer.

bgrainger:
Fixes #6236

## Description

Using a `<Resource>` font in WPF currently causes a lot of allocation, GC pinning, and memory copying. This is entirely avoidable by returning a pointer to the already-loaded font data to DWrite.

## Customer Impact

Without this fix, customers are forced to hook DWrite APIs and supply their own font-loading implementation; e.g., https://faithlife.codes/blog/2019/06/improving-wpf-text-display-performance/.

## Regression

Not a regression.

## Testing

Manual testing using a test app: https://github.com/bgrainger/EmbeddedFontPerformance

## Risk

Risk is low. This fix has been prototyped in a commercial WPF application that's been shipping for 5 years (see blog post above). Since the font source is an `UnmanagedMemoryStream`, we can rely on its data being present in memory at least as long as we hold a reference to the stream. (And since, practically speaking, the font data is backed by the assembly that contains it being mmap'ed into the process' address space, it's going to be safe to dereference it for the lifetime of the application.)

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/DirectWriteForwarder/CPP/DWriteWrapper/FontFileStream.cpp
+++ b/src/Microsoft.DotNet.Wpf/src/DirectWriteForwarder/CPP/DWriteWrapper/FontFileStream.cpp
@@ -17,17 +17,15 @@ namespace MS { namespace Internal { namespace Text { namespace TextInterface
         // guarantee that this problem will be fixed so we will use the GetUnmanagedStream(). Note: This path will only 
         // be taken for embedded fonts among which XPS is a main scenario. For local fonts we use DWrite's APIs.
         _fontSourceStream = fontSource->GetUnmanagedStream();
+        _fontSourcePointer = _fontSourceStream->PositionPointer - _fontSourceStream->Position;
         try
         {
             _lastWriteTime = fontSource->GetLastWriteTimeUtc().ToFileTimeUtc();
-        }    
+        }
         catch(System::ArgumentOutOfRangeException^) //The resulting file time would represent a date and time before 12:00 midnight January 1, 1601 C.E. UTC. 
         {
             _lastWriteTime = -1;
-        }        
-
-        // Create lock to control access to font source stream.
-        _fontSourceStreamLock = gcnew Object();
+        }
     }
 
     FontFileStream::~FontFileStream()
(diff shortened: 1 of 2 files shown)


bgrainger:
Performance data obtained by running the test app, maximising on a 1600x1200 screen, then scrolling the textbox all the way to the bottom.

## Before

image

## After

image

## Comparison

image

singhashish-wpf:
This PR is being considered for next community test pass.

PR #6268, opened 2022-03-17
Remove unnecessary string and string[] allocations from MS.Internal.ContentType

stephentoub:
## Description

MS.Internal.ContentType is incurring a multitude of unnecessary allocations.  It's parsing input by Substring'ing along it and Trim'ing along it to find the next piece, but only a subset of those strings are actually saved. It's using string.Split just to get two strings, resulting in a string[].  Etc.

## Customer Impact

Apps using this indirectly incur unnecessary overhead.

## Regression

No

## Testing

CI

## Risk

Minimal.


src/Microsoft.DotNet.Wpf/src/WindowsBase/MS/Internal/ContentType.cs
```diff
@@ -526,10 +525,12 @@ private static int GetLengthOfParameterValue(string s, int startIndex)
 
                 while (!found)
                 {
-                    length = s.IndexOf('"', ++length);
+                    int startingLength = ++length;
+                    length = s.Slice(startingLength).IndexOf('"');
```
danmoseley, line 529:
```suggestion
                    int startingLength = length + 1;
                    length = s.Slice(startingLength).IndexOf('"') + 1;
```

danmoseley (approved):

src/Microsoft.DotNet.Wpf/src/WindowsBase/MS/Internal/ContentType.cs
stephentoub: That won't work correctly.

PR #6270, opened 2022-03-17
Don't allocate fallback name in XamlNamespace.GetXamlType unless it's needed

stephentoub:
## Description

XamlNamespace.GetXamlType is always concatenating a string together, but that string is only used on a fallback path.  So only create it on the fallback path.

## Customer Impact

Unnecessary string allocation.

## Regression

No

## Testing

CI

## Risk

Minimal


lindexi (approved):
Thank you.

dipeshmsft (approved):

PR #6272, opened 2022-03-17
Use span slice instead of substring in AbbreviatedGeometryParser.ReadNumber

stephentoub:
## Description

AbbreviatedGeometryParser.ReadNumber is using Substring to pass to Convert.ToDouble the relevant portion of the string to parse. We can instead just pass in a span slice to double.Parse (which Convert.ToDouble delegates to after a null check that's irrelevant here).

## Customer Impact

Unnecessary allocations.

## Regression

No

## Testing

CI

## Risk

Minimal


src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Media/ParsersCommon.cs
```diff
@@ -435,11 +435,13 @@ void SkipDigits(bool signAllowed)
             }
             else
             {
-                string subString = _pathString.Substring(start, _curIndex - start);
-
                 try
                 {
-                    return System.Convert.ToDouble(subString, _formatProvider);
+#if PRESENTATION_CORE
```
ThomasGoulet73, line 440: Could you use `NET` instead of PRESENTATION_CORE since it's .Net Core specific not PresentationCore project specific (In case in the future another .Net Core project references this file).
```suggestion
#if NET
```

ThomasGoulet73 (commented):
This PR is a duplicate of #5390 but in my opinion the changes in this PR are slightly better since it has less ifdefed code.

We should either close #5390 and keep this PR or close this PR and move the changes from this PR to #5390.

/cc @lindexi

src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Media/ParsersCommon.cs
lindexi: Thank you @ThomasGoulet73 

Could you review my code @stephentoub  ?

stephentoub: > Could you review my code @stephentoub ?

Thanks, @lindexi.  As @ThomasGoulet73 noted, we don't need to duplicate all that code.  Really all that's needed is to swap the AsSpan for Substring.  I chose to duplicate the one line in the name of readability, but we shouldn't need to duplicate the whole block as your PR has.  I very much appreciate your efforts (and I hope others of your PRs can be merged), but I think we should close #5390 in favor of this one.

stephentoub: > Could you use NET instead of PRESENTATION_CORE since it's .Net Core specific not PresentationCore project specific (In case in the future another .Net Core project references this file).

I agree with the reasoning why, but I chose PRESENTATION_CORE because it's already used with ifdefs multiple times in this file.  As such, I'll stick with the current ifdef.  If this file were to be included in other projects, I expect other ifdefs would also want to be revisited (even though this one is for a slightly different purpose).

ThomasGoulet73: That's true but this means that PresentationBuildTasks, which multi-targets .Net Framework and .Net Core, will use Substring in both target frameworks when it could could use the span version for .Net Core.

dipeshmsft: @stephentoub I agree with @ThomasGoulet73's argument, that we should use NET here instead of PRESENTATION_CORE flag as this code is not specific to PresentationCore.dll.

dipeshmsft (approved):

PR #6135, opened 2022-02-16
Fix MilCodeGen

ThomasGoulet73:
## Description
This PR aims to fix MilCodeGen which seems to be very outdated and it doesn't seem like it ever worked in this repo (It probably worked in the internal repo for Microsoft).

Since it does not seem like it ever worked, I did these changes just by deducing how it's supposed to work so this PR might be totally wrong.

I noticed a couple of changes in the generated code other than white-space and license header where I don't know whether the source of truth is the new generated code or the existing generated code (It's mostly comments that differ between both).

This PR unblocks PRs like this one: #5651.

## Customer Impact
None.

## Regression
No.

## Testing
I tested this fix locally using this command `build -projects "src\Microsoft.DotNet.Wpf\src\WpfGfx\codegen\mcg\mcg.proj"` and by looking at the changes and doing a normal build afterwards.

## Risk
I think the risk is pretty low since it's only used for development in this repo and the worse that can happen is to be unable to run MilCodeGen (It didn't work before this PR so it can only be an improvement). All the generated code is checked out in Git so it's easier to validate that the generated code is good.

PR #6274, opened 2022-03-17
Run some built-in .NET SDK perf analyzers over several WPF assemblies

stephentoub:
## Description

Over the last several releases of .NET, the SDK has added a multitude of analyzers to help find and flag opportunities for perf improvements.  I ran several of those analyzers over a few of the assemblies and applied the auto-fixers; I also code reviewed each fix and tweaked a handfull.

## Customer Impact

Many of these changes contribute small but meaningful reductions in overhead.

## Regression

No

## Testing

CI

## Risk

Minimal


src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Media/Generated/DoubleCollection.cs
```diff
@@ -723,7 +723,7 @@ internal string ConvertToString(string format, IFormatProvider provider)
 
                 if (i != _collection.Count-1)
                 {
-                    str.Append(" ");
```
ThomasGoulet73, line 726: You should revert the changes made in generated files (See [this list of generated files](https://github.com/dotnet/wpf/blob/main/eng/WpfArcadeSdk/tools/WPF_Generated_Files.txt)). These files are generated by MilCodeGen which is not in a state to be modified or runned in this repo. #6135 aims to allow to modify and run MilCodeGen which would allow to apply this change [here](https://github.com/dotnet/wpf/blob/main/src/Microsoft.DotNet.Wpf/src/WpfGfx/codegen/mcg/generators/ManagedStruct.cs#L375) and add proper `<auto-generated>` file header.

stephentoub: The fact that these are generated doesn't seem to be a good reason not to make such changes.  I've updated the templates accordingly.  When they again become runnable, my changes shouldn't result in diffs.

src/Microsoft.DotNet.Wpf/src/PresentationCore/MS/internal/IO/Packaging/ByteRangeDownloader.cs
```diff
@@ -417,8 +417,7 @@ private ByteRangeDownloader(Uri requestedUri, SafeWaitHandle eventHandle)
             }
 
             // Ensure uri is correct scheme (http or https) Do case-sensitive comparison since Uri.Scheme contract is to return in lower case only.
-            if (String.Compare(requestedUri.Scheme, Uri.UriSchemeHttp, StringComparison.Ordinal) != 0
-                    && String.Compare(requestedUri.Scheme, Uri.UriSchemeHttps, StringComparison.Ordinal) != 0)
+            if (!string.Equals(requestedUri.Scheme, Uri.UriSchemeHttp, StringComparison.Ordinal) && !string.Equals(requestedUri.Scheme, Uri.UriSchemeHttps, StringComparison.Ordinal))
```
bgrainger, line 420:
```suggestion
            if (requestedUri.Scheme != Uri.UriSchemeHttp && requestedUri.Scheme != Uri.UriSchemeHttps)
```

Ordinal string comparison can be simplified. (Pattern match `if (requestedUri.Scheme is not "http" and not "https")` could be even clearer, but doesn't support using the named constants.)

dipeshmsft (approved):

PR #6288, opened 2022-03-21
Avoid unnecessary allocations/code in ReaderWriterLockWrapper

stephentoub:
## Description

By default the code was using structs as the IDisposables returned to the caller, resulting in those structs being boxed on every access.  The code was also for some reason maintaining two code paths based on a switch, but the only impact of that would be on whether new instances would be allocated on each call, which is not something we need to guard with a switch (that's seemingly unrelated).  The code is also allocating a new delegate for every enter and exit operation.  This commit simplifies all that and avoids all that unnecessary allocation and duplication.

## Customer Impact

Unnecessary allocation and overhead

## Regression

No

## Testing

CI

## Risk

Minimal. Only issue would be if callers were for some reason depending on a different object reference being returned from the Read/WriteLock properties on every access.

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/Shared/MS/Internal/ReaderWriterLockWrapper.cs
+++ b/src/Microsoft.DotNet.Wpf/src/Shared/MS/Internal/ReaderWriterLockWrapper.cs
@@ -6,19 +6,14 @@
 //
 //
 // Description:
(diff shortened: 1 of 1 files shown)


stephentoub:
Closing, as this has been open for two and a half years.

Issue #6289, opened 2022-03-10
After updating to .Net 6.0.3 through VS2022 Tooltips in WPF immediately vanish 


redseujac75:
### Description
All Tooltips in C#, WPF project developed in VS 2022 Community .Net 6.0.2, vanish immediately after appearing. This happens with all controls (buttons, text boxes ...) with a tooltip. This problem appeared after updating to .Net 6.0.3. It wasn't there with previous .Net 6 versions.

### Configuration

* Which version of .NET is the code running on? .NET 6.0.3
* What OS and version, and for Linux, what distro? Windows 10 Pro 21H2 build 19044.1586
* What is the architecture (x64, x86, ARM, ARM64)? x64 and x86
* Do you know whether it is specific to that configuration? NO
 
### Regression?

* Did this work in a previous build or release of .NET, or from .NET Framework? If you can try a previous release or build to find out, that can help us narrow down the problem. If you don't know, that's OK.
The problem didn't exist in versions of .NET 6 preceding 6.0.3

mikeKuester:
Our Azure DevOps build pipeline still uses .NET 6.0.2 (image windows2022). It would be very helpful if you @ Microsoft could assure that the responsive team updates .NET **after** this bug is fixed. In the moment the production build is not broken.

ThomasGoulet73:
According to @SamBent's comment [here](https://github.com/dotnet/wpf/pull/6300#issuecomment-1076850512), this is being worked on by the WPF team.

SamBent:
Closing this as a dupe of #6220.

PR #6300, opened 2022-03-23
Fix Popup regression on high DPI

ThomasGoulet73:
Fixes dotnet/wpf#6289

## Description
Fixes a regression where popups would vanish when using high DPI settings. The current code in .Net 6.0.3 uses the position of the mouse to check if the mouse was inside an array of points. The problem was that the position of the mouse was not DPI aware while the array of points was DPI aware. I fixed this by transforming the position of the mouse to have a point which is DPI aware.

## Customer Impact
Fixes popup vanishing immediately.

## Regression
Regression introduced in .Net 6.0.3 with #6063 which was a backport of #5931.

## Testing
I tested locally using the sample provided here: https://github.com/dotnet/wpf/pull/6063#issuecomment-1064069083. I used two monitors with different DPI settings and after the changes in this PR, the issue was fixed.

## Risk
Low.

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Controls/PopupControlService.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Controls/PopupControlService.cs
@@ -1562,7 +1562,8 @@ namespace System.Windows.Controls
                 if (rootElement != null)
                 {
                     // get the coordinates of the current mouse point, relative to our PresentationSource
-                    System.Windows.Point pt = Mouse.PrimaryDevice.GetPosition(rootElement);
+                    Matrix transform = PresentationSource.FromVisual(_source.RootVisual).CompositionTarget.TransformToDevice;
+                    System.Windows.Point pt = transform.Transform(Mouse.PrimaryDevice.GetPosition(rootElement));
 
                     // check whether the point lies within the hull
                     return ContainsPoint(_source, (int)pt.X, (int)pt.Y);


ThomasGoulet73:
@dotnet/wpf-developers Could this PR be prioritized ? This fixes a regression introduced in .Net 6.0.3 which broke most popups on screens with high DPI. Hopefully this could get backported to the release/6.0 branch and included in .Net 6.0.4 (If this PR is considered to be the right fix and accepted by the WPF team, obviously). Thank you!

SamBent:
@ThomasGoulet73 - I'm preparing a better fix in .NET Framework, which we'll port to .NET Core when it's ready (probably this week).  I started out doing something similar to yours, but discovered it had numerical instability and performance issues, both arising from the many layers of transforms involved.  Fortunately, there's a way to avoid all that.

We should have announced that we were working on this, to save you some trouble.  Sorry about that.

BTW, your diagnosis is backwards - the points in the convex hull use a physical pixel coordinate system (i.e. not DPI-aware), whereas the mouse point uses logical pixels (i.e. DPI-aware).  Or maybe we just differ in what "DPI-aware" means.  Not a big deal - the important thing is that they use different coordinate systems when scaling is > 100%, and that's what causes the bug.

ThomasGoulet73:
> @ThomasGoulet73 - I'm preparing a better fix in .NET Framework, which we'll port to .NET Core when it's ready (probably this week). I started out doing something similar to yours, but discovered it had numerical instability and performance issues, both arising from the many layers of transforms involved. Fortunately, there's a way to avoid all that.

That's great, I'm looking forward to see what the right fix is. I didn't notice "numerical instability and performance issues" but this might be because I only did manual testing for this PR.

> We should have announced that we were working on this, to save you some trouble. Sorry about that.

No worries. But yes, in the future, a quick "We're working on this" would be nice to avoid having the WPF team and people from the community work on the same thing at the same time.

> BTW, your diagnosis is backwards - the points in the convex hull use a physical pixel coordinate system (i.e. not DPI-aware), whereas the mouse point uses logical pixels (i.e. DPI-aware). Or maybe we just differ in what "DPI-aware" means. Not a big deal - the important thing is that they use different coordinate systems when scaling is > 100%, and that's what causes the bug.

That makes sense. We have the same meaning of "DPI-aware", I just thought that the convex hull was dpi scaled and the mouse was not dpi scaled. Thank you for the explanation!

Closing this PR since it's being worked on by the WPF team.

PR #6334, opened 2022-03-31
Added correct WPFArtifactsPath for x64

dipeshmsft:
Fixes # 

Main PR 

## Description

Earlier the wpf-debug.targets did not work with applications built on x64. In this fix, when the PlatformTarget is set to x64, correct WPF artifacts path is being pointed for loading the modules.

## Regression
No

## Testing
Tested the loaded modules in both scenarios ( x86, x64 )

## Risk
None

### DIFF
--- a/eng/wpf-debug.targets
+++ b/eng/wpf-debug.targets
@@ -13,7 +13,10 @@
   <PropertyGroup>
     <WpfConfig Condition="'$(WpfConfig)'==''">Debug</WpfConfig>
     <RuntimeIdentifier>win-$(PlatformTarget)</RuntimeIdentifier>
-    <WPFArtifactsPath>$([System.IO.Path]::GetFullPath('$(MSBuildThisFileDirectory)\..\'))\artifacts\packaging\$(WpfConfig)\Microsoft.DotNet.Wpf.GitHub.$(WpfConfig)</WPFArtifactsPath>
+    
+    <WpfPackagesCommonPath>$([System.IO.Path]::GetFullPath('$(MSBuildThisFileDirectory)\..\'))\artifacts\packaging</WpfPackagesCommonPath>
+    <WPFArtifactsPath Condition="'$(PlatformTarget)'=='x86'">$(WpfPackagesCommonPath)\$(WpfConfig)\Microsoft.DotNet.Wpf.GitHub.$(WpfConfig)</WPFArtifactsPath>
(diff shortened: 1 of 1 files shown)


eng/wpf-debug.targets
```diff
@@ -13,7 +13,10 @@
   <PropertyGroup>
     <WpfConfig Condition="'$(WpfConfig)'==''">Debug</WpfConfig>
     <RuntimeIdentifier>win-$(PlatformTarget)</RuntimeIdentifier>
-    <WPFArtifactsPath>$([System.IO.Path]::GetFullPath('$(MSBuildThisFileDirectory)\..\'))\artifacts\packaging\$(WpfConfig)\Microsoft.DotNet.Wpf.GitHub.$(WpfConfig)</WPFArtifactsPath>
+    
+    <WpfPackagesCommonPath>$([System.IO.Path]::GetFullPath('$(MSBuildThisFileDirectory)\..\'))\artifacts\packaging</WpfPackagesCommonPath>
+    <WPFArtifactsPath Condition="'$(PlatformTarget)'=='x86'">$(WpfPackagesCommonPath)\$(WpfConfig)\Microsoft.DotNet.Wpf.GitHub.$(WpfConfig)</WPFArtifactsPath>
+    <WPFArtifactsPath Condition="'$(PlatformTarget)'=='x64'">$(WpfPackagesCommonPath)\$(WpfConfig)\x64\Microsoft.DotNet.Wpf.GitHub.$(WpfConfig)</WPFArtifactsPath>
```
lindexi, line 19: How about the `Arm` ?

How about the code below?

```xml
<WPFPlatformTargetArtifacts Condition="'$(PlatformTarget)'=='x64'">x64\</WPFPlatformTargetArtifacts>

<WPFArtifactsPath>$([System.IO.Path]::GetFullPath('$(MSBuildThisFileDirectory)\..\'))\artifacts\packaging\$(WpfConfig)\$(WPFPlatformTargetArtifacts)Microsoft.DotNet.Wpf.GitHub.$(WpfConfig)</WPFArtifactsPath>
```

dipeshmsft: Right, I missed about ARM64. Thanks @lindexi for pointing that out. I will update the PR.

dipeshmsft: @lindexi, regarding ARM64, it still uses the same directory as AnyCPU, hence we don't need to add a separate clause for that.

singhashish-wpf (approved):

PR #6332, opened 2022-03-31
ToolTip closes immediately in high DPI

SamBent:
Fixes #6220

## Description

In high DPI environment, most tooltips opened by the mouse close immediately.

This is because the SafeArea uses unscaled coordinates, while the current mouse point uses scaled coordinates.  In high DPI, these are different, leading to false negatives from the containment test "is the mouse within the SafeArea".

Fixed by using unscaled coordinates for the mouse point.  (And incidentally improving the perf by avoiding a lot of unnecessary transformations into, and out of, scaled coordinates.)

## Customer Impact

Make ToolTips great again (in high DPI).

## Regression

Regression in PR #5931.

## Testing

Tested by hand, but thoroughly, in .NET 4.8.1.  The automated tests don't cover high DPI, unless you explicitly run them on a machine whose primary monitor is high DPI.

This is a straightforward port of the .NET Fx fix, but should be tested in .NET Core anyway.

PR #6370, opened 2022-04-06
use unscaled mouse coordinates for convex hull test (#6332)

### LINKED ISSUE #6220: Tooltip flicker in Ribbon Elements that are inside menu/dropdown
* .NET Core Version: 6.0.101
* Windows version: 21H1 OS Build : 19043.1526

 
 
 **Problem description:**
 In .NET6 if you have a tooltip on a combobox/menu item it flickers and can't be read by user ( see picture https://imgur.com/a/RxFvAPP )
 **Actual behavior:** 
 Tooltip shows for just a few milliseconds on the screen and closes itself.
 **Expected behavior:**
 Tooltip does not close itself unless user moves the mouse
 **Minimal repro:**
https://github.com/SebiCiuca/RibbonToolTipFlicker

Steps to reproduce:
Open app .NET 5 - all tooltips for RibbonButton/RibbonMenuItem/RibbonGallery can be read by user ( they do not close unless move move)
Open app .NET 6 - tooltip from RibbonMenuItem/RibbonGallery can't be read by user they close in a matter of milliseconds. RibbonButton tooltip still works

singhashish-wpf:
Fixes #6220

## Description

In high DPI environment, most tooltips opened by the mouse close immediately.

This is because the SafeArea uses unscaled coordinates, while the current mouse point uses scaled coordinates.  In high DPI, these are different, leading to false negatives from the containment test "is the mouse within the SafeArea".

Fixed by using unscaled coordinates for the mouse point.  (And incidentally improving the perf by avoiding a lot of unnecessary transformations into, and out of, scaled coordinates.)

## Customer Impact

Make ToolTips great again (in high DPI).

## Regression

Regression in PR #5931.

## Testing

Tested by hand, but thoroughly, in .NET 4.8.1.  The automated tests don't cover high DPI, unless you explicitly run them on a machine whose primary monitor is high DPI.

This is a straightforward port of the .NET Fx fix but should be tested in .NET Core anyway.

## Risk
Low

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/wpf/pull/6370)

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Controls/PopupControlService.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Controls/PopupControlService.cs
@@ -1558,39 +1558,56 @@ namespace System.Windows.Controls
             // Test whether the current mouse point lies within the convex hull
             internal bool ContainsMousePoint()
             {
-                IInputElement rootElement = _source.RootVisual as IInputElement;
-                if (rootElement != null)
+                // get the coordinates of the current mouse point, relative to the Active source
+                PresentationSource mouseSource = Mouse.PrimaryDevice.CriticalActiveSource;
+                System.Windows.Point pt = Mouse.PrimaryDevice.NonRelativePosition;
+
+                // translate the point to our source's coordinates, if necessary
+                // (e.g. if the tooltip's owner comes from a window with capture,
+                // such as the popup of a ComboBox)
+                if (mouseSource != _source)
                 {
-                    // get the coordinates of the current mouse point, relative to our PresentationSource
-                    System.Windows.Point pt = Mouse.PrimaryDevice.GetPosition(rootElement);
-
-                    // check whether the point lies within the hull
-                    return ContainsPoint(_source, (int)pt.X, (int)pt.Y);
-
-                    // NOTE: GetPosition doesn't actually return the position of the current mouse point,
-                    // but rather the last recorded position.  (See MouseDevice.GetScreenPositionFromSystem,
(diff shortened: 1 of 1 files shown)


src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Controls/PopupControlService.cs
```diff
@@ -1558,39 +1558,56 @@ private void AddPoints(PointList points, in NativeMethods.RECT rect, bool rectIs
...
-                    // when the ComboBox is close to the edge of the main window so that a tooltip from its
-                    // popup content extends beyond the main window.
-                    //      This can only be fixed by changing MouseDevice.GetScreenPositionFromSystem to
-                    // use a "better way" to find the current mouse position, which allegedly needs work from the OS.
-                    // But we can live with this behavior, because
-                    //  * this is a corner case - tooltips from popup content that extend beyond the main window
-                    //  * the effect is transient - the tooltip will close when the user dismisses the popup
-                    //  * there's no accessibility issue - WCAG 2.1 only requires that the tooltip stays open under
-                    //      proscribed conditions, not that it has to close when the conditions cease to apply
+                    System.Windows.Point ptScreen = PointUtil.ClientToScreen(pt, mouseSource);
```
dreddy-work, line 1570: you may want to use direct mapping if it helps perf: ref:https://github.com/dotnet/winforms/blob/8b099dc847ccc153782b7c1591f677ff779ae9a8/src/System.Windows.Forms.Primitives/src/Interop/User32/Interop.MapWindowPoints.cs

dipeshmsft (approved):

PR #6374, opened 2022-04-06
PickFolders option for OpenFileDialog (minimal)

### LINKED ISSUE #438: WPF alternative for WinForms FolderBrowserDialog

miloush:
For discussion. Fixes #438. Further discussion in #4039. Previous attempt: #6351

## Description

Allows developers to ask the user to select a folder, a feature known as `FolderBrowserDialog` in WinForms.

This PR is the minimal code changes needed to introduce the feature. It closely reflects the underlying API design, i.e. the ability to specify `FOS_PICKFOLDERS` on the `IOpenFileDialog`. This is a backwards compatible solution - no public API has been moved or removed and no existing behavior has changed.

A custom option `OPTION_PICKFOLDERS` is introduced that sets `FOS_PICKFOLDERS` and clears `OFN_FILEMUSTEXIST` when the dialog is prepared.

### New visible members

In `OpenFileDialog`:
* `public bool PickFolders { get; set; }`

### Legacy behavior

Legacy code path (pre-Vista) ignores the flag and shows standard file open dialog.

## Customer Impact

Without this fix, users have to either use WinForms' `FolderBrowserDialog`, resort to 3rd party libraries or re-implement the whole `IFileDialog` API. #438 is currently the top voted issue in the repository.

## Regression

No.

## Testing

Compiled custom _PresentationFramework.dll_ and tested in 6.0.2 x86 app that both folder browsing and file browsing works as expected.

## Risk

Low - existing behavior not affected, minimal code changes. Setting `Filter` on a dialog with `PickFolders` will throw `COMException` on `ShowDialog()`.

The bit used by `OPTION_PICKFOLDERS` is currently unused in both legacy and `IOpenFileDialog` API, but it might get allocated in the future. This would however not break this feature as the bits are masked out (same happened in the past with `OPTION_ADDEXTENSION`).

User code that processes "used" file dialogs could now be handed an instance that unexpectedly contains folders in the `FileName(s)` properties. This, however, would have to be explicitly caused using new code. Unexpected file operations on folders (i.e. trying to open them) yield IO exceptions that the existing code should be already handling.

/cc @ThomasGoulet73 @batzen

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/PresentationFramework/Microsoft/Win32/FileDialog.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationFramework/Microsoft/Win32/FileDialog.cs
@@ -1792,6 +1792,19 @@ namespace Microsoft.Win32
             // Force no mini mode for the SaveFileDialog.
             // Only accept physically backed locations.
             FOS options = ((FOS)Options & c_VistaFileDialogMask) | FOS.DEFAULTNOMINIMODE | FOS.FORCEFILESYSTEM;
+
+            // FILEMUSTEXIST is set by default in OpenFileDialog.
+            // While the combination of PICKFOLDERS and FILEMUSTEXIST is valid,
+            // it currently does not let users select anything in the dialog.
+            // We therefore disable FILEMUSTEXIST for folder selection.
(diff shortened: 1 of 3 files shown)


Symbai:
@miloush Reason why this is still being a draft?

PR #6443, opened 2022-04-15
Use Dictionary instead of Hashtable in WeakEventTable

zgabi:
## Description

It's currently using a Hashtable with struct keys (except the _managerTable, but better to change all of them), which means every operation on the Hashtable boxes the struct. Switching to a Dictionary removes those costs.
The Hashtable indexer returns null when key not exists, so in this case TryGetValue should be used on the Dictionary object.

## Customer Impact

Less unnecessary boxing means less pauses due to GC.

## Regression

No

## Testing

Just CI

## Risk

Similar as in #4731

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/WindowsBase/MS/Internal/WeakEventTable.cs
+++ b/src/Microsoft.DotNet.Wpf/src/WindowsBase/MS/Internal/WeakEventTable.cs
@@ -108,7 +108,11 @@ namespace MS.Internal
         /// </summary>
         internal WeakEventManager this[Type managerType]
         {
-            get { return (WeakEventManager)_managerTable[managerType]; }
+            get
+            {
+                _managerTable.TryGetValue(managerType, out WeakEventManager result);
+                return result;
+            }
             set { _managerTable[managerType] = value; }
         }
 
@@ -120,7 +124,8 @@ namespace MS.Internal
             get
             {
                 EventNameKey key = new EventNameKey(eventSourceType, eventName);
-                return (WeakEventManager)_eventNameTable[key];
+                _eventNameTable.TryGetValue(key, out WeakEventManager result);
+                return result;
             }
 
             set
@@ -138,7 +143,7 @@ namespace MS.Internal
             get
             {
                 EventKey key = new EventKey(manager, source);
-                object result = _dataTable[key];
+                _dataTable.TryGetValue(key, out object result);
                 return result;
             }
 
@@ -297,7 +302,7 @@ namespace MS.Internal
                 {
                     // copy the keys into a separate array, so that later on
                     // we can change the table while iterating over the keys
-                    ICollection ic = _dataTable.Keys;
+                    ICollection<EventKey> ic = _dataTable.Keys;
                     EventKey[] keys = new EventKey[ic.Count];
                     ic.CopyTo(keys, 0);
 
@@ -331,14 +336,11 @@ namespace MS.Internal
                     Debug.Assert(_toRemove.Count == 0, "to-remove list should be empty");
                     _inPurge = true;
 
-                    // enumerate the dictionary using IDE explicitly rather than
-                    // foreach, to avoid allocating temporary DictionaryEntry objects
-                    IDictionaryEnumerator ide = _dataTable.GetEnumerator() as IDictionaryEnumerator;
-                    while (ide.MoveNext())
+                    foreach (KeyValuePair<EventKey, object> kvp in _dataTable)
                     {
-                        EventKey key = (EventKey)ide.Key;
+                        EventKey key = kvp.Key;
                         object source = key.Source;
-                        foundDirt |= key.Manager.PurgeInternal(source, ide.Value, purgeAll);
+                        foundDirt |= key.Manager.PurgeInternal(source, kvp.Value, purgeAll);
 
                         // if source has been GC'd, remove its data
                         if (!purgeAll && source == null)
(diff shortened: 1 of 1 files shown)


src/Microsoft.DotNet.Wpf/src/WindowsBase/MS/Internal/WeakEventTable.cs
```diff
@@ -108,7 +108,11 @@ internal IDisposable WriteLock
         /// </summary>
         internal WeakEventManager this[Type managerType]
         {
-            get { return (WeakEventManager)_managerTable[managerType]; }
+            get
+            {
+                _managerTable.TryGetValue(managerType, out WeakEventManager result);
+                return result;
+            }
             set { _managerTable[managerType] = value; }
```
paulozemek, line 116: I think that a remove should be done if value is null.

zgabi: you mean the result variable?
I think it is never null, when key exitst in the Hashset.
TryGetValue is needed, because the key maybe does not exists in the Dictionary. (Hashset returned null in this case, Dictionary throws exception when I use _managerTable[managerType])

paulozemek: I am talking about the set. What if the user passes "null" as the value?
We shouldn't keep a key/value pair in the dictionary for a null value.

zgabi: Ok, so in the setter... I did not want to make any functional change, but for me that is OK :)
This setter is called from `WeakEventManager.SetCurrentManager`, which is protected, so accessible for users (not internal). I think that method should check the argument.

paulozemek: OK... sounds good to me.

src/Microsoft.DotNet.Wpf/src/WindowsBase/MS/Internal/WeakEventTable.cs
```diff
@@ -297,7 +302,7 @@ private bool Purge(bool purgeAll)
                 {
                     // copy the keys into a separate array, so that later on
                     // we can change the table while iterating over the keys
-                    ICollection ic = _dataTable.Keys;
+                    ICollection<EventKey> ic = _dataTable.Keys;
```
stephentoub, line 305: Dictionary supports removal during enumeration.  This copying of the keys shouldn't be necessary.

paulozemek: No... if you try to remove during enumeration you get: Collection was modified; enumeration operation may not execute.
I think you are confusing ConcurrentDictionary (that can be modified during enumeration, even by other threads) with the plain Dictionary, which cannot be modified during enumeration.

ThomasGoulet73: You can remove from a dictionary during enumeration ([sharplab](https://sharplab.io/#v2:C4LgTgrgdgPgAgJgAwFgBQiCM65IARyYAsA3OugCICWAxsFQPZQCGYAngDxVTAA0BmJAD48AE1r0mrNngC8eKAFMA7nmp1GLdlx79CwgBQBKdAG90eS3lN5M/AESZ7eAL68LVmwgcJnbj5Y2AMwOQX7uaFbWeEQOROEB0QCsDkkJaC5kaOgAZgxgisw0ABZ4BgBurHgA1uV43GISmtImaOaRVlQ5ZbUAdADSijKy8pitUVHiGlLsvQBKigC2DOWKBn2DbEZZLuRoeQVFpRVV1UP1UI3TWmwDQwDOre1RXT3nI3gI4xOWU5I38yWKzWZy2Oz2hAAnAYAMrAMDcADmvQAUgxuAZ7A5+H9mrNNo9tkA)) but in this case you can't. It removes items from last to first using index, which I don't think is possible without copying the keys to an array. I don't know if there is a valid reason for removing from last to first, other than using index to remove items.

paulozemek: You are right... I see this is a difference between the .NET Framework and the .NET Core.

zgabi:
Any news on this PR? It is quite trivial and another similar PR was also merged. (#4731)

singhashish-wpf:
This PR is being considered for next community test pass.

stephentoub:
The mechanical changes look fine.

One possible concern, though (which I've not investigated in depth), is that Hashtable and Dictionary have different thread-safety guarantees.  Dictionary is not.  Hashtable supports any number of concurrent reads with up to one concurrent write.  Normally when these types are used as an implementation detail, that difference isn't relevant, as the containing types themselves don't make any thread-safety guarantees... but in this case I see the type has ReadLock/WriteLock properties, which suggests instances are used concurrently and the thread-safety guarantees could come into play.

To merge this, someone should audit the usage to ensure those guarantees aren't actually being relied on.

singhashish-wpf:
@zgabi This PR is failing some of the internal tests which are not yet public in the wpf-test repo, Hence no further details can be provided on the failures until the tests are open sourced to the wpf-test repo. As of now this PR is a no merge, until the tests are public and the failures are resolved.
Since Test migration is already planned in upcoming sprints, we'll inform you once the failing tests are in public domain.

PR #6500, opened 2022-04-29
Avoid Hashtable-related allocations in AccessorTable

stephentoub:
## Description

Use a strongly-typed Dictionary instead.

- The key is a struct.  Every time an item is added to the table or a lookup into the table is performed, the key is getting boxed. Using Dictionary avoids all this boxing.
- The struct key lacks a strongly-typed Equals.  Every comparison results in a boxing.  Gave the key a strongly-typed Equals.
- Every cleanup operation is allocating an enumerator.  Dictionary has a struct-based enumerator.
- Every cleanup operation is allocating an array to store the items to be removed (even if there aren't any to be removed).  Dictionary supports removal during enumeration, so this can be removed entirely.

## Customer Impact

Unnecessary allocations

## Regression

No

## Testing

CI

## Risk

Minimal

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/wpf/pull/6500)

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/PresentationFramework/MS/Internal/Data/AccessorTable.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationFramework/MS/Internal/Data/AccessorTable.cs
@@ -16,7 +16,9 @@
 
 using System;
 using System.Collections;
+using System.Collections.Generic;
 using System.ComponentModel;                // IBindingList
+using System.Diagnostics;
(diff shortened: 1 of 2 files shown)


dipeshmsft (approved):

dipeshmsft:
We have taken this PR for the current Community Test Pass. Thanks for the contribution.

PR #6518, opened 2022-05-02
Avoid allocating `Stack<BranchNode>` just to peek at it

stephentoub:
## Description

When peeking to see if the BranchNodeStack is empty, it may force it into existence.

## Customer Impact

Unnecessary allocation

## Regression

No

## Testing

CI

## Risk

Minimal

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/wpf/pull/6518)

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/EventRoute.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/EventRoute.cs
@@ -321,7 +321,7 @@ namespace System.Windows
             branchNode.Node = node;
             branchNode.Source = source;
             
-            BranchNodeStack.Push(branchNode);
+            (_branchNodeStack ??= new Stack<BranchNode>(1)).Push(branchNode);
         }
 
         /// <summary>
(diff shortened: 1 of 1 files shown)


src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/EventRoute.cs
```diff
@@ -321,7 +321,7 @@ public void PushBranchNode(object node, object source)
             branchNode.Node = node;
             branchNode.Source = source;
             
-            BranchNodeStack.Push(branchNode);
+            (_branchNodeStack ??= new Stack<BranchNode>()).Push(branchNode);
```
IAmTheCShark, line 324: The original Stack was created with a size of 1. I think you overlooked that?

stephentoub: Thanks.

dipeshmsft (approved):

PR #6523, opened 2022-05-04
[DevDiv VSO: 1425279] WPF should ignore transitory/ignorable DWM error

### LINKED ISSUE #5688: WPF does not ignore ignorable DWM errors and instead crashes the process
The gist of the problem is WPF does not ignore STATUS_MESSAGE_LOST in their call to DWM's flush function. This message is ignorable and in fact should be ignored as DWM will recycle itself if it crashes and become available again. WPF not ignoring this forces a SyncFlush failure in VS which forces a UI thread exception to be thrown which crashes the product potentially losing unsaved user data. 

The DWM folks in an internal email (quoted below) suggested this error (and possibly all others) should be treated as transitory/ignored, as WPF is already doing for SOME other error codes in this same code path.

This crash is responsible for approximately 50% of the SyncFlush Watson bucket which is the #1 crash in Dev17.

From an internal email thread with the DWM folks (emphasis on 'all errors should be in this bucket', mine):

> Yes, STATUS_MESSAGE_LOST is the NTSTATUS that is returned by LPC, whereas the other value is the HRESULT version of the same thing. The only difference is the extra facility bit that’s OR’ed in.
> 
> Here, you’ll notice that WPF doesn’t want to do anything different – it just bubbles the error out. Also notice that we’re filtering some known errors already, and doing nothing special when these are encountered. My point is, **all errors should be in this bucket**, and at that point there is no longer a reason to even check the return code of the method. It’s basically a best-effort method: if it works, you want to wait for it to complete, and if it doesn’t work then there’s nothing to wait for.
>

singhashish-wpf:
Fixes #5688
[DevDiv VSO: 1425279] WPF should ignore transitory/ignorable DWM error

Main PR 
None
## Description
[DevDiv VSO: 1425279] WPF should ignore transitory/ignorable DWM error code (STATUS_MESSAGE_LOST) from DwmFlush

An app can opt-in to the behavior of ignoring all DwmFlush errors by setting a regkey in
HKCU\Software\Microsoft\Avalon.Graphics\IgnoreDwmFlushErrors
or
HKLM\Software\Microsoft\Avalon.Graphics\IgnoreDwmFlushErrors
whose name is the full path to the .exe that wants to opt-in, and whose DWORD value is 1.

## Regression
No

## Testing
Tested on .NET framework and Core both with internal test passes.

## Risk
None

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/WpfGfx/core/uce/rendertargetmanager.cpp
+++ b/src/Microsoft.DotNet.Wpf/src/WpfGfx/core/uce/rendertargetmanager.cpp
@@ -1186,6 +1186,10 @@ CRenderTargetManager::WaitForDwm(
             || hr == WGXERR_UCE_CHANNELSYNCABANDONED)
         {
             hr = S_OK;
+        }        
+        else if (ShouldIgnoreDwmFlushErrors())
+        {
+            hr = S_OK;
         }
         else 
         {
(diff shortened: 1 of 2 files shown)


dipeshmsft (approved):

miloush:
Is there a reason why this is opt-in?

omariom:
@singhashish-wpf 
So the fix is either machine-wide or user-wide.
Why not make it application specific as well?

PR #6528, opened 2022-05-04
Reuse bool box objects in UncommonField<bool>.SetValue

stephentoub:
## Description

In some scenarios, these UncommonField<bool>s end up being set very frequently, and each time the value being passed in gets boxed.  We can avoid that by caching and reusing a box for each of the only two possible values (though it looks like the default value is almost always false and so it mainly only boxes true).

## Customer Impact

Unnecessary allocation

## Regression

No

## Testing

CI

## Risk

Minimal.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/wpf/pull/6528)

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/WindowsBase/System/Windows/UncommonField.cs
+++ b/src/Microsoft.DotNet.Wpf/src/WindowsBase/System/Windows/UncommonField.cs
@@ -4,6 +4,7 @@
 
 using System.Collections.Generic;
 using System.Runtime.CompilerServices;
+using MS.Internal.KnownBoxes;
 using MS.Internal.WindowsBase;  // for FriendAccessAllowed
 
 namespace System.Windows
@@ -56,8 +57,8 @@ namespace System.Windows
 
                     if (typeof(T) == typeof(bool))
                     {
-                        // There are only two possible values. Use shared boxed instances rather than creating new objects for each SetValue call.
-                        valueObject = Unsafe.As<T, bool>(ref value) ? (s_true ??= true) : (s_false ??= false);
+                        // Use shared boxed instances rather than creating new objects for each SetValue call.
+                        valueObject = BooleanBoxes.Box(Unsafe.As<T, bool>(ref value));
                     }
                     else
                     {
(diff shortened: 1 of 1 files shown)


src/Microsoft.DotNet.Wpf/src/WindowsBase/System/Windows/UncommonField.cs
```diff
@@ -136,6 +146,11 @@ internal int GlobalIndex
         private int _globalIndex;
         private bool _hasBeenSet;
 
+        /// <summary>A lazily boxed false value.</summary>
+        private static object s_false;
```
ThomasGoulet73, line 150: You could use [MS.Internal.KnownBoxes.BooleanBoxes](https://github.com/dotnet/wpf/blob/main/src/Microsoft.DotNet.Wpf/src/WindowsBase/MS/Internal/KnownBoxes.cs#L11) which is already used in a bunch of places in this repo.

stephentoub: Wasn't aware that existed.  Even better.

src/Microsoft.DotNet.Wpf/src/WindowsBase/System/Windows/UncommonField.cs
```diff
@@ -54,12 +53,24 @@ public void SetValue(DependencyObject instance, T value)
                 // Set the value if it's not the default, otherwise remove the value.
                 if (!EqualityComparer<T>.Default.Equals(value, _defaultValue))
                 {
-                    instance.SetEffectiveValue(entryIndex, null /* dp */, _globalIndex, null /* metadata */, value, BaseValueSourceInternal.Local);
+                    object valueObject;
+
+                    if (typeof(T) == typeof(bool))
```
lindexi, line 58: May I know the performance difference between `typeof(T) == typeof(bool)` and `T is bool`? Thank you.

stephentoub: `typeof(T) == typeof(bool)` and `value is bool` should both be optimized to a const value by the JIT in a release build and tier 1 compilation.  In tier 0, it's possible `value is bool` could box.

lindexi: Thank you very much.

dipeshmsft (approved):

Issue #5697, opened 2021-11-16
RootNamespace should be set explicitly in WPF inner builds


AArnott:
* .NET Core Version: 6.0.100
* Windows version: Windows 11 (21H2, 22000.318)
* Does the bug reproduce also in WPF for .NET Framework 4.8?: Yes/No
* Is this bug related specifically to tooling in Visual Studio (e.g. XAML Designer, Code editing, etc...)? No

 **Problem description:**

A typical WPF project file does not explicitly set the `RootNamespace` MSBuild property but rather inherits it from the `$(MSBuildProjectName)`.

WPF's inner build creates a temporary project file with a random suffix added to the project name. This means that the *default RootNamespace changes* to not match the actual `RootNamespace` from the outer build.
Two problems with that:

1. The `RootNamespace` in the inner build does not match the value used by the outer build. This may not be a customer facing problem if the inner build does not ever include targets that care what that property is set to. `EmbeddedResource` items are embedded using logical names based on `RootNamespace`, so as one example if that step is done within an inner build, then the inner build has a corrupted result. 
2.  The `RootNamespace` changes to a new random value with each and every build. This breaks incremental build and forces WPF's inner build to actually compile the result with each and every build (with no source changes) when Nerdbank.GitVersioning is installed in the project, since this property serves as an input to a generated source file.

See #5458 for other work targeting improving incremental build performance.

 **Actual behavior:**

The `RootNamespace` property is not set in the generated tmpproj.csproj file.
 
 **Expected behavior:**

If `RootNamespace` is explicitly set in the project file, its definition should be copied into the tmpproj
If `RootNamespace` is *not* explicitly set in the project file, the generated tmpproj *should* set it explicitly to `$(_TargetAssemblyProjectName)`.
 
For example we were able to workaround the problem by adding this to our `Directory.Build.targets` file:

```xml
  <!-- Incremental build perf improvements -->
  <PropertyGroup>
    <!-- Fix RootNamespace in WPF inner builds so they are constant across builds. -->
    <RootNamespace Condition=" $(RootNamespace.EndsWith('_wpftmp')) ">$(_TargetAssemblyProjectName)</RootNamespace>
  </PropertyGroup>
```

Kuldeep-MS:
@AArnott - I have added a fix [here](https://github.com/dotnet/wpf/pull/7557). Can you please give it a try to see if the issue is fixed or not?

AArnott:
[...]

Issue #4421, opened 2021-04-16
WPF XAML file can't see types generated by Razor Source Generator


Eilon:
* .NET Core Version: 6.0-preview3
* Windows version: N/A
* Does the bug reproduce also in WPF for .NET Framework 4.8?: N/A
* Is this bug related specifically to tooling in Visual Studio? No (interestingly, this _does_ work in VS IntelliSense, but not when compiling!)

 **Problem description:**

In .NET 6 we are enabling authoring of WPF apps in a "Blazor Desktop" scenario where Blazor web UI (HTML+CSS) is hosted in a web view in a WPF app. Such projects contain a mix of XAML (WPF), Blazor/Razor (HTML+CSS), and other code artifacts (C#, etc.). This is done with a new `<BlazorWebView ... />` WPF control being built by the ASP.NET team.

The XAML compiler, however, is unable to see the types generated by the Razor Source Generator. (Note: Using a source generator is new for Blazor in .NET 6; previously it was done in other ways).

For the WPF BlazorWebView control we’d like to support a XAML syntax like this to specify a Razor component to load:

```xaml 
<Window ...
        xmlns:local="clr-namespace:BlazorWpfApp">
   <DockPanel>
        ...
        <blazor:BlazorWebView HostPage="wwwroot\index.html" Services="{StaticResource services}">
            <blazor:BlazorWebView.RootComponents>
                <blazor:RootComponent Selector="#app" ComponentType="{x:Type local:Main}" />
            </blazor:BlazorWebView.RootComponents>
        </blazor:BlazorWebView>
    </DockPanel>
</Window>
```
 
Where `local:Main` is a class that is generated by the Razor compiler as part of the project’s compilation. There is a file `Main.razor` that generates a type Main in the same namespace. The Razor-generated class is visible from C# code, but the XAML compiler is unable to find.

From @captainsafia in an email thread:

> We'll need to do some work on the WPF size to react to the changes made in the Razor SDK.

 **Actual behavior:**

In VS Intellisense you can see the Razor-generated types (good):
![image](https://user-images.githubusercontent.com/202643/115072267-34109e80-9eac-11eb-80da-7d71b4c41f0d.png)

But when you try to compile the app, the XAML compiler can't see the same type anymore (bad):

```
...\MainWindow.xaml(17,55): error MC3050: Cannot find the type 'local:Main'. Note that type names are case sensitive. Line 17 Position 55.
```
 
 **Expected behavior:**

[...]

PR #6535, opened 2022-05-05
Fix build when using Razor source generator

ThomasGoulet73:
Fixes dotnet/wpf#4421
Fixes dotnet/wpf#5697

## Description
Fixes build when using Razor source generator by adding RootNamespace to the list of property added to the generated project.

Without this PR, the Razor source generator would generate files in the wrong namespace (`WpfBlazorRepro_5s0cnboz_wpftmp ` instead of `WpfBlazorRepro`).

Fixes incremental build where the RootNamespace is changing on each build and therefore is recognized as a fresh build (Not incremental).

## Customer Impact
Without this PR, customers cannot build a WPF application with the Razor source generator (Maybe others?).

## Regression
No. To my knowledge, it never worked.

## Testing
Tested by building the repro project in the issue (https://github.com/Eilon/WPFXamlRazorGeneratorBug). This PR fixed the build.

Tested the repro here: https://github.com/dotnet/wpf/issues/5697#issuecomment-1458411584. This PR makes the repro build incrementally.

## Risk
Low.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/wpf/pull/6535)

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/PresentationBuildTasks/Microsoft.WinFX.targets
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationBuildTasks/Microsoft.WinFX.targets
@@ -454,6 +454,7 @@
                 Analyzers="@(Analyzer)"
                 TemporaryTargetAssemblyProjectName="$(_TemporaryTargetAssemblyProjectName)"
                 MSBuildProjectExtensionsPath="$(MSBuildProjectExtensionsPath)"
+                RootNamespace="$(RootNamespace)"
                  >
 
           <Output TaskParameter="TemporaryAssemblyForLocalTypeReference" PropertyName="_AssemblyForLocalTypeReference" />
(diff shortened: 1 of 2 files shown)


Nirmal4G:
We really need to rethink the whole `GenerateTemporaryTargetAssembly` using a different solution like source generators. Instead of adding every single property over time. First was the project name, then file, then output paths and now namespace!

ThomasGoulet73:
I rebased my changes on main and validated that this PR fixes dotnet/wpf#5697. I will link my PR to this issue.

ThomasGoulet73:
@Nirmal4G I agree with you that this build process is not optimal. This PR is more like a patch instead of fixing the problem at its root which is that the build process is weird. Unfortunately, this is the best I can do without changing a bunch of things since the build process requires BAML compilation which requires a built assembly so that it can output resources to add to said assembly. For now this requires 2 build but I have changes locally that instead of triggering a whole build it just adds the resources to the assembly in the first build using Cecil but it is not PR-ready.

rchauhan18:
@ThomasGoulet73, thanks for your contribution

PR #6538, opened 2022-05-05
Simplify dwrite.dll loading process

kant2002:
All supported OS always has `dwrite.dll`.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/wpf/pull/6538)

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/Shared/cpp/dwriteloader.cpp
+++ b/src/Microsoft.DotNet.Wpf/src/Shared/cpp/dwriteloader.cpp
@@ -11,33 +11,7 @@ namespace WPFUtils
 #endif
 HMODULE LoadDWriteLibraryAndGetProcAddress(void **pfncptrDWriteCreateFactory)
 {
-    HMODULE hDWriteLibrary = nullptr;
-    
-    // KB2533623 introduced the LOAD_LIBRARY_SEARCH_SYSTEM32 flag. It also introduced
-    // the AddDllDirectory function. We test for presence of AddDllDirectory as an 
-    // indirect evidence for the support of LOAD_LIBRARY_SEARCH_SYSTEM32 flag. 
-    HMODULE hKernel32 = GetModuleHandle(L"kernel32.dll");
-    if (hKernel32 != nullptr)
-    {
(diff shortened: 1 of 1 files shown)


src/Microsoft.DotNet.Wpf/src/Shared/cpp/dwriteloader.cpp
```diff
@@ -7,37 +7,9 @@
...
-        {
-            // LOAD_LIBRARY_SEARCH_SYSTEM32 is not supported on this OS. 
-            // Fall back to using plain ol' LoadLibrary
-            // There is risk that this call might fail, or that it might be
-            // susceptible to DLL hijacking. 
-            hDWriteLibrary = LoadLibrary(L"dwrite.dll");
-        }
-    }
-    
+    HMODULE hDWriteLibrary = LoadLibraryEx(L"dwrite.dll", nullptr, LOAD_LIBRARY_SEARCH_SYSTEM32);
```
lindexi, line 12: Windows 7, Windows Server 2008 R2, Windows Vista and Windows Server 2008:  This value requires [KB2533623](https://support.microsoft.com/kb/2533623) to be installed. See https://docs.microsoft.com/en-us/windows/win32/api/libloaderapi/nf-libloaderapi-loadlibraryexa

kant2002: Argh! I miss that. But only Windows 7 SP1 is possible applicable since other OS [does not supported](https://github.com/dotnet/core/blob/main/release-notes/7.0/supported-os.md) by .NET 7
And even that seems to be [KB2533623](https://support.microsoft.com/kb/2533623) required for .NET see https://github.com/dotnet/docs/issues/20459

But I hardly can understand is ESU has this patch or not.

kant2002: That should be slated for .NET 8 then. That way Win 7 would be out of support at release date (Nov 2023) and not important anymore.

dipeshmsft:
@kant2002 thanks for your contribution.

PR #6547, opened 2022-05-06
Ignore NotImplementedException from ITaskbarList

### LINKED ISSUE #2377: Setting the TaskbarItemInfo leads to System.NotImplementedException when no user is logged in
* .NET Core Version: All that support WPF
* Windows version: Windows 7 and later

 **Problem description:**
Take a WPF application which attempts to set the `TaskbarItemInfo` property of the main application window on Windows 7 or later.  In some situations this can lead to an unhandled `System.NotImplementedException` which crashes the application.

The apparent cause is attempting to run such an application when no user is logged in (e.g. remote desktop / Citrix / Windows Server).

An identical issue was present (and later [fixed](https://github.com/DevExpress/DevExpress.Mvvm.Free/blob/d8f957007cd409dd7e678876640b666fcf0c6467/DevExpress.Mvvm.UI/Services/TaskbarButtonService/Native/TaskbarInfoApplicator.cs#L11)) in [DevExpress MVVM toolkit](https://www.devexpress.com/Support/Center/Question/Details/T528105/notimplementedexception-in-taskbarbuttonservice-when-no-user-is-logged-in).  

 **Actual behavior:** 
The following line throws a `System.NotImplementedException`:
https://github.com/dotnet/wpf/blob/c90f457bfeb8d7737bb32f9e088aa2d9f8aff1d2/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Window.cs#L6291
```
System.NotImplementedException: The method or operation is not implemented.
   at MS.Internal.AppModel.ITaskbarList.HrInit()
   at System.Windows.Window.ApplyTaskbarItemInfo()
   at System.Windows.Window.WindowFilterMessage(IntPtr hwnd, Int32 msg, IntPtr wParam, IntPtr lParam, Boolean& handled)
   at System.Windows.Interop.HwndSource.PublicHooksFilterMessage(IntPtr hwnd, Int32 msg, IntPtr wParam, IntPtr lParam, Boolean& handled)
```

 **Expected behavior:**
If the code in the try block containing the `taskbarList.HrInit();` call fails, the exception should be caught and handled accordingly.  If setting the `TaskbarItemInfo` property is not possible, even on a supported operating system, the code should simply return without setting the `TaskbarItemInfo` property.

 **Minimal repro:**
Add the following lines to an empty WPF desktop application:
```
    <Window.TaskbarItemInfo>
        <TaskbarItemInfo Description="Test" />
    <Window.TaskbarItemInfo>
```
Attempt to run the application on a machine where no user is logged in.

miloush:
Fixes #2377

## Description

In some cases, not closer specified in the documentation, [ITaskbarList::HrInit](https://docs.microsoft.com/en-us/windows/win32/api/shobjidl_core/nf-shobjidl_core-itaskbarlist-hrinit) can fail. When that happens, WPF applications setting `Window.TaskbarItemInfo` crash without possibility of gracefully handling the error.

This PR fixes the issue by checking for the result of HrInit, and not proceeding with taskbar customization if it fails, as per the documentation. "Not proceeding" simply means returning from the `ApplyTaskbarItemInfo` method the same way it is done in the method in other cases (e.g. API not supported by OS).

## Customer Impact

Application crashes when it sets `TaskbarItemInfo` and the API fails. Reports in #2377 indicate this happens in cases when no user is logged in, a custom shell without taskbar is used, or under terminal services.

## Regression

No.

## Testing

Built and verified that `TaskbarItemInfo` is still working on 7.0-preview3. Closed Explorer and verified the app crashes before the fix and does not crash after the fix, and that the taskbar info still gets applied when Explorer is restarted. Verified the error is appears in the trace output.

## Risk

I believe the risk is minimal. There is no changes to public API. There is a behavior change in an internal interface (no longer throws exceptions on error), but the fixed method is the only callsite currently using it. The handle to `ITaskbarList` is released in the `finally` block as per the documentation.

A valid worry is that throwing an exception is actually the desired outcome when taskbar is not available and application tries to use it. However, since the code already simply ignores other cases when this API is not available and developers cannot easily catch or avoid the exception, ignoring the error seems to be a consistent and justifiable behavior.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/wpf/pull/6547)

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Window.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Window.cs
@@ -637,7 +637,7 @@ namespace System.Windows
                     // Explorer being non-responsive should be a transient issue.  Post back to apply the full TaskbarItemInfo.
                     _taskbarRetryTimer.Start();
                 }
-                else if (hr == (HRESULT)Win32Error.ERROR_INVALID_WINDOW_HANDLE)
+                else if (hr == (HRESULT)Win32Error.ERROR_INVALID_WINDOW_HANDLE || hr == HRESULT.E_NOTIMPL)
                 {
                     // We'll get this when Explorer's not running.  This means there's no Shell to integrate with.
                     if (TraceShell.IsEnabled)
@@ -6292,6 +6292,7 @@ namespace System.Windows
                 return;
             }
 
+            HRESULT hr = HRESULT.S_OK;
             if (_taskbarList == null)
             {
                 // If we don't have a handle and there isn't a TaskbarItemInfo, then we don't have anything to apply or remove.
@@ -6304,12 +6305,19 @@ namespace System.Windows
                 try
                 {
                     taskbarList = (ITaskbarList)Activator.CreateInstance(Type.GetTypeFromCLSID(new Guid(CLSID.TaskbarList)));
-                    taskbarList.HrInit();
+
+                    hr = taskbarList.HrInit();
+                    if (hr != HRESULT.S_OK)
+                    {
+                        // Taskbar not available (no user logged in, running under terminal service, custom shell, etc.)
+                        HandleTaskbarListError(hr);
+                        return;
+                    }
 
                     // This QI will only work on Win7.
                     _taskbarList = (ITaskbarList3)taskbarList;
                     taskbarList = null;
-}
+                }
                 finally
                 {
                     Utilities.SafeRelease(ref taskbarList);
@@ -6334,7 +6342,6 @@ namespace System.Windows
             }
 
             // Apply (or clear) all aspects of the TaskbarItemInfo to this Window.
-            HRESULT hr = HRESULT.S_OK;
             hr = RegisterTaskbarThumbButtons();
 
             if (hr.Succeeded)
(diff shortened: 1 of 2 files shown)


miloush:
While I tried to avoid blank `catch` statement, I wasn't comfortable the error did not get reported anywhere, Added trace output.

src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Window.cs
```diff
@@ -6304,12 +6305,19 @@ private void ApplyTaskbarItemInfo()
...
                     taskbarList = (ITaskbarList)Activator.CreateInstance(Type.GetTypeFromCLSID(new Guid(CLSID.TaskbarList)));
-                    taskbarList.HrInit();
+
+                    hr = taskbarList.HrInit();
+                    if (hr != HRESULT.S_OK)
+                    {
+                        // Taskbar not available (no user logged in, running under terminal service, custom shell, etc.)
+                        HandleTaskbarListError(hr);
+                        return;
+                    }
```
RussKie, line 6315: What about releasing the COM?
image

miloush: Thank you @RussKie, as noted in the description, the interface is released in the finally block below.

RussKie: Ah, I totally missed it. Thank you.

RussKie (approved):

dipeshmsft (approved):

dipeshmsft:
@miloush, this PR has been included in the current test pass. Will share the results once the test pass is complete.

PR #6550, opened 2022-05-06
GridSplitter should not exclusively rely on its logical parent

### LINKED ISSUE #6490: GridSplitter does not work when IsItemsHost is true
* .NET Core Version: 5.0

`GridSplitter` does not work, when it is added programmatically to `ItemsControl.Items` that uses a `Grid` with `IsItemsHost="true"`
 
Reason:
[`GridSplitter.InitializeData` ](https://source.dot.net/#PresentationFramework/System/Windows/Controls/GridSplitter.cs,493)checks if `Parent` is a `Grid`. In this case the `LogicalParent` is the `ItemsControl`.
Shouldn't it check the `VisualParent`?
[repro.zip](https://github.com/dotnet/wpf/files/8571712/repro.zip)

IAmTheCShark:
`GridSplitter` should rely on its `VisualParent` not it's `LogicalParent`

Fixes #6490

## Description
`GridSplitter` needs to interact with it's parent, this however should not exclusively be the LogicalParent.

## Customer Impact

GridSplitters do not work in `IsItemsHost` scenarios.

## Testing

I copied the modified code since I havent managed to build WPF yet.

## Risk

I dont think it is possible to use a `ControlTemplate` for a `Grid`, so `VisualParent` should generally be the more suitable solution

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/wpf/pull/6550)

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Controls/GridSplitter.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Controls/GridSplitter.cs
@@ -490,7 +490,7 @@ namespace System.Windows.Controls
         // Initialize the data needed for resizing
         private void InitializeData(bool ShowsPreview)
         {
-            Grid grid = VisualTreeHelper.GetParent(this) as Grid;
+            Grid grid = FindParentGrid();
 
             // If not in a grid or can't resize, do nothing
             if (grid != null)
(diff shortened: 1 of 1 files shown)


miloush:
As discussed in the issue, I would be more comfortable if the current check for a logical parent was in place and visual parent was only checked if the logical parent failed, because currently the `GridSplitter` works even if the Grid is not the immediate visual parent.

IAmTheCShark:
> miloush

I tried to reproduce a situation where LogicalParent = Grid and VisualParent != Grid and failed.
Since a Grid cannot have a ControlTemplate, I do not know how you'd be able to acchieve this.

miloush:
The parent element is in charge of the logical tree.

```C#
public class MyGrid : Grid
{
    public MyGrid()
    {
        GridSplitter splitter = new GridSplitter();
        AddLogicalChild(splitter);
        HostVisual visual = new HostVisual();
        visual.Children.Add(splitter);
        AddVisualChild(visual);
    }
}
```

For `splitter`, the logical parent `is Grid` but visual parent is `HostVisual`.

IAmTheCShark:
@miloush updated the PR accordingly. Logical parent is checked first now.
Also I learned something new today, thank you for this example.

IAmTheCShark:
Im closing this PR because I am not 100% certain that this is a good idea anymore.

Issue #6525, opened 2022-05-04
WPF Characters render differently in XPS document and on screen ( VisualContainer ) in WPF Application


dipeshmsft:
* .NET Core Version: (e.g. 3.0 Preview1, or daily build number, use `dotnet --info`) :  **Any ( 5.0 , 6.0 )**
* Windows version: (`winver`)   **Any**
* Does the bug reproduce also in WPF for .NET Framework 4.8?:   **Yes**
* Is this bug related specifically to tooling in Visual Studio (e.g. XAML Designer, Code editing, etc...)? If yes, please file the issue via the instructions [here](https://docs.microsoft.com/visualstudio/ide/how-to-report-a-problem-with-visual-studio?view=vs-2019)   **No**
* Security issues and bugs should be reported privately, learn more via our [responsible disclosure guidelines](https://github.com/dotnet/wpf/blob/main/README.md#reporting-security-issues-and-security-bugs).

 

Output on Screen : 
![on-screen](https://user-images.githubusercontent.com/85861525/166672314-d70301ab-d03a-4eaa-b61c-0c6f3829f6bd.png)

Output in XPS File :
![on-xps](https://user-images.githubusercontent.com/85861525/166672309-e2b395c6-b72f-44e5-9052-58332714b88a.png)

 
 **Problem description:**
When we render WPF characters as an XPS document, the output is incorrectly rendered as compared to that on the screen, thus affecting the visual fidelity of the applications.

 **Actual behavior:** 
Incorrectly sized outputs when XPS document is created.
 
 **Expected behavior:**
 The output on both screen and XPS document is similar.

PR #6554, opened 2022-05-09
[Port] Fixes rounding error while glyphrun serialization

dipeshmsft:
Fixes #  #6525 
Port of PR #6295 

Main PR 

## Description
When we serialize a GlyphRun to an XPS file, the advance widths for each glyph are rounded to an integer separately, thus producing a significant difference between the rendered output in the file and on the screen. This PR fixes the rounding, by taking into account the cumulative rounding error from the previous glyphs while rounding the current glyph.

## Customer Impact
Affects the visual fidelity of WPF Applications.

## Regression

No

## Testing

Ad-Hoc testing. Tried with different sample apps.

## Risk
Less


leecow:
Not a security fix so does not meet the 5.0 bar.

Issue #5944, opened 2022-01-13
Popup element are reversed left and right in Windows 11


ChengYen-Tang:
* .NET Core Version: .net 5~6(e.g. 3.0 Preview1, or daily build number, use `dotnet --info`)
* Windows version: Windows 11
* Does the bug reproduce also in WPF for .NET Framework 4.8?: Yes/No
* Is this bug related specifically to tooling in Visual Studio : No
* Security issues and bugs should be reported privately, learn more via our [responsible disclosure guidelines](https://github.com/dotnet/wpf/blob/main/README.md#reporting-security-issues-and-security-bugs).

 
 
 **Problem description:**
I found that using the Popup element and setting the Placement to left will actually appear on the right and the right will appear on the left
This problem happens on windows 11, windows 10 is normal
Note: Visual studio menu also has this problem 

 **Actual behavior:** 
Windows 11:
![image](https://user-images.githubusercontent.com/32535494/149258036-5b5536c7-b13e-4582-8010-143b0580e65a.png)
![image](https://user-images.githubusercontent.com/32535494/149258167-40457884-e544-46c6-9aeb-a5902301cefe.png) 

 **Expected behavior:**
 Windows 10:
![image](https://user-images.githubusercontent.com/32535494/149258583-aeaf06e3-ff61-4481-9633-b614c659b04f.png)
![image](https://user-images.githubusercontent.com/32535494/149258822-bc7b53b1-422a-4307-9a8d-f835530d986a.png)

 **Minimal repro:**

lindexi:
It's Looks like a Windows problem, not the WPF issues.

ChengYen-Tang:
Seems yes, but I have a few questions
1. This problem has existed since the release of Win11, but it has not been improved
2. My colleagues use Qt to develop desktop applications, this problem does not appear in Qt
3. I have been using Win11 for a while, and most programs do not have this problem
4. I don't know how to give feedback to the windows department

miloush:
I believe this works as designed. There is a setting in Windows that controls on which side the menu appears.

This used to be part of _Control Panel_ > _Table PC settings_ > _Other_ tab > _Handedness_. 

![image](https://user-images.githubusercontent.com/10546952/149672141-4867b8bc-3572-4771-958f-623a44a4ae68.png)

However, I no longer have this tab on Win11. Try _Settings_ > _Bluetooth & Devices_ > _Pen & Windows Ink_ > _Choose which hand you write with_.

[...]

Issue #4641, opened 2021-06-08
Popup's placement propery is wrong


yanshouwang:
* .NET Core Version: 3.0, 3.1, 5.0
* Windows version: (`winver`) 20H2 (19042.985)
* Does the bug reproduce also in WPF for .NET Framework 4.8?: Yes
 
 **Problem description: Popups in WPF show's in wrong place, when set to left the popup shows at right, set to right shows at left, set to up shows at top-right (which should shows at top-left) and set to bottom shows at bottom-right (which should shows at bottom-left).
 
 **Actual behavior:** 
![image](https://user-images.githubusercontent.com/24455689/121114231-f1cd6300-c845-11eb-9550-7a5be7efe186.png)

 **Expected behavior:**
![image](https://user-images.githubusercontent.com/24455689/121114258-fe51bb80-c845-11eb-8f0d-eba0d0fae5c6.png)

 **Minimal repro:**

daisyTian0517:
@yanshouwang
I tested your code on my desktop, and the Popups are on right place. This should be because your device is in tablet mode. **Soultion:**
Please go to your `Settings` -> `System` -> `Tablet`  , then choose `Don't swtict to tablet mode` for the `When I use this device as a tablet`.

yanshouwang:
@daisyTian0517 No, my device is a PC, and the settings is alreay set to PC mode, I think this is related to the windows version or .NET sdk version. This issue doesn't occured on my workmates' computer which the windows and .net sdk version is lower then mine.

daisyTian0517:
It may be not caused by SDK version or winVersion, my winVersion is 20H2, my wpf project target platform is  .NET Framework 4.8.  Do your workmates use the same machines as you?

yanshouwang:
@daisyTian0517 yes, do you installed .net 5.0 SDK? I aslo install .NET 5.0 SDK

yanshouwang:
@daisyTian0517 and this problem even affect my visual studio, but I tried some winform applications, which doesn't have this problem, so I think this only happened in WPF apps.
![image](https://user-images.githubusercontent.com/24455689/121634102-9950cc80-cab6-11eb-9ea7-d3548cb4e8ac.png)
![image](https://user-images.githubusercontent.com/24455689/121634184-b6859b00-cab6-11eb-84a0-e9554d012215.png)

daisyTian0517:
[...]

PR #6557, opened 2022-05-09
Public the Menu Parameters set API

lindexi:
Fixes https://github.com/dotnet/wpf/issues/5944 https://github.com/dotnet/wpf/issues/4641

## Description

We can ignore the System settings by the code:

            SystemParameters.StaticPropertyChanged += (sender, eventArgs) =>
            {
                SystemParameters.MenuDropAlignment = false;
            };

cc @miloush

## Testing

CI and My demo

## Risk

Low.


miloush:
Wouldn't be possible for the setters to actually set the system parameters? How do users cancel the override? Could we rather not fix the consumers to use `MenuDropAlignmentKey` as a dynamic resource and override it in app resources?

PR #6612, opened 2022-05-22
Fix definition of ITfContextOwnerCompositionServices

kant2002:
See https://docs.microsoft.com/en-us/windows/win32/api/msctf/nn-msctf-itfcontextownercompositionservices
ITfContextOwnerCompositionServices inherited from IUnknown, so most likely this is broke any text composition because `TerminateComposition` has slot with index 3, but instead of that it was on index 7.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/wpf/pull/6612)


kant2002:
I think you are right. I file an issue with docs team https://github.com/MicrosoftDocs/sdk-api/pull/1172

kant2002:
I double check and in Windows SDK headers also indicates that `ITfContextOwnerCompositionServices` derived from `ITfContextComposition`. I do not have specific issue. I just migrate code from built-in COM to ComWrappers and notice this discrepancy between docs and code, and most likely incorrectly place blame on WPF code.

PR #6619, opened 2022-05-25
Replace resource constants with properties in UIAutomationProvider

ThomasGoulet73:
Contributes to dotnet/wpf#1

## Description
Replace the use of resource constants with properties in UIAutomationProvider. This follows the standard of other .Net projects, like [dotnet/runtime](https://github.com/dotnet/runtime).

## Customer Impact
None.

## Regression
No.

## Testing
Local build + CI.

## Risk
None.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/wpf/pull/6619)

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/UIAutomation/UIAutomationProvider/System/Windows/Automation/Provider/AutomationInteropProvider.cs
+++ b/src/Microsoft.DotNet.Wpf/src/UIAutomation/UIAutomationProvider/System/Windows/Automation/Provider/AutomationInteropProvider.cs
@@ -61,7 +61,7 @@ namespace System.Windows.Automation.Provider
         /// <returns>base raw element for specified window</returns>
         public static IRawElementProviderSimple HostProviderFromHandle ( IntPtr hwnd )
         {
-            ValidateArgument(hwnd != IntPtr.Zero, nameof(SRID.HwndMustBeNonNULL));
+            ValidateArgument(hwnd != IntPtr.Zero, nameof(SR.HwndMustBeNonNULL));
             return UiaCoreProviderApi.UiaHostProviderFromHwnd(hwnd);
         }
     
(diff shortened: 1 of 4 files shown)


src/Microsoft.DotNet.Wpf/src/UIAutomation/UIAutomationProvider/System/Windows/Automation/Provider/AutomationInteropProvider.cs
```diff
@@ -215,7 +215,7 @@ private static void ValidateArgumentNonNull(object obj, string argName)
         // Throw an argument Exception with a generic error
         private static void ThrowInvalidArgument(string argName)
         {
-            throw new ArgumentException(SR.Format(SRID.GenericInvalidArgument, argName));
+            throw new ArgumentException(SR.Format(SR.GenericInvalidArgument, argName));
```
ThomasGoulet73, line 218: This is bug in the current main branch. SR.Format expects the first parameter to be the resource, **not** the resource name: https://github.com/dotnet/wpf/blob/main/src/Microsoft.DotNet.Wpf/src/Common/src/System/SR.cs#L81-L89

RussKie (approved):

src/Microsoft.DotNet.Wpf/src/UIAutomation/UIAutomationProvider/System/Windows/Automation/Provider/AutomationInteropProvider.cs
dipeshmsft: @ThomasGoulet73 , what do you mean by 'this is a bug' ? Is the usage here in AutomationInteropProvider.cs wrong ?? Or are you referrring that SR.Format needs to be fixed ??

ThomasGoulet73: @dipeshmsft Sorry I forgot to answer your question. This specific usage of SR.Format in this file is wrong. Before this PR, `SRID.GenericInvalidArgument` was the key of the resource not the resource itself and SR.Format is a simple wrapper around string.Format. So this code was doing something like this:
`throw new ArgumentException(string.Format("GenericInvalidArgument", argName));`
which is wrong.

dipeshmsft: However, with your PR this is fixed. So, I am marking this as resolved. And thanks for explaining this.

dipeshmsft:
@ThomasGoulet73 , this PR has been included in the current test pass. Will share the results once the test pass is complete.

PR #6656, opened 2022-06-07
Remove sorcery for getting consistent names in the traces

kant2002:
That helps for trimming #3811

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/wpf/pull/6656)

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/PresentationCore/MS/internal/MediaTrace.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationCore/MS/internal/MediaTrace.cs
@@ -15,7 +15,8 @@ namespace MS.Internal
 
     internal class MediaTrace : IDisposable
     {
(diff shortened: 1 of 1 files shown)


kant2002:
For any reviewer, that traces can be enabled only from code, since TraceSwitch are not read from `app.config` at all in .NET.

ThomasGoulet73:
It's also worth noting that this class is only used when PresentationCore is built in Debug:
https://github.com/dotnet/wpf/blob/381768982cf9371ab7dc8841e5cbc8cf9b79d16f/src/Microsoft.DotNet.Wpf/src/PresentationCore/MS/internal/MediaTrace.cs#L14

The class is also internal so Assembly.GetCallingAssembly() will always return PresentationCore.dll.

dipeshmsft (approved):

dipeshmsft:
@kant2002, this PR has been included in the current test pass. Will share the results once the test pass is complete.

PR #6691, opened 2022-06-16
Fix Invalid_IInputElement resource

ThomasGoulet73:
## Description
Fixes a small bug in the Invalid_IInputElement resource. UIElement3D is not included in the message even if it is supported.
Fixes a small bug in some usage of the Invalid_IInputElement resource. Some usage were not passing an argument in the Get method so the format would return the message unformated.

Also added a couple of comment cleanup that did not include UIElement3D in the supported list.

## Customer Impact
Give proper exception message.

## Regression
No. This is probably an old "bug" when UIElement3D was introduced.

## Testing
Local build + CI.

## Risk
Low. Exception message could change but I don't think it is breaking

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/wpf/pull/6691)


src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/PresentationSource.cs
```diff
@@ -129,10 +129,10 @@ public static void AddSourceChangedHandler(IInputElement element, SourceChangedE
                 throw new ArgumentNullException("element");
             }
 
-            // Either UIElement or ContentElement
+            // Either UIElement, ContentElement or UIElement3D.
             if (!InputElement.IsValid(element))
             {
-                throw new ArgumentException(SR.Get(SRID.Invalid_IInputElement), "element");
+                throw new ArgumentException(SR.Get(SRID.Invalid_IInputElement, element.GetType()), nameof(element));
```
ThomasGoulet73, line 135: Here is an example of a call to SR.Get using Invalid_IInputElement and not passing a parameter for the format.

Here's the content of Invalid_IInputElement: https://github.com/dotnet/wpf/blob/41bdb1071eb62fb04bfb57d88d94e79dedc6fdbc/src/Microsoft.DotNet.Wpf/src/PresentationCore/Resources/Strings.resx#L1289

dipeshmsft:

PR #6693, opened 2022-06-17
Remove SafeReflectionInvoker

ThomasGoulet73:
## Description
Removes SafeReflectionInvoker which was useful in .Net Framework because those methods were annotated with `SecuritySafeCriticalAttribute` for partial trust but since .Net Core does not support partial trust, this level of indirection is useless and only hurts performance because of the indirection and the methods are annotated to never be inlined by the JIT.

I removed `IsInSystemXaml` and `IsSystemXamlNonPublic` which are internal and unused but there are comments that it is by design that they still exist. Those comments dates back to the initial commit of System.Xaml in this repo so it probably comes from the .Net Framework version of WPF. I removed them in a separate commit so it's easy to revert if the WPF does not want to remove those methods.

## Customer Impact
Should improve performance a bit (Though I wasn't able to properly measure it).

## Regression
No.

## Testing
Local build + CI + I tested a simple app.

## Risk
Low. It is the same code but without indirection.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/wpf/pull/6693)


dipeshmsft:
We have completed the test pass for this PR and the tests passed. However, we would like to keep `IsInSystemXaml` and `IsSystemXamlNonPublic` until we find out more details on that. 

@ThomasGoulet73 can you please revert the change for removing `IsInSystemXaml` and `IsSystemXamlNonPublic` .

Once the changes are done, we will go ahead and merge the PR.

ThomasGoulet73:
@dipeshmsft Thank you, I'll try to do this later today.

ThomasGoulet73:
@dipeshmsft Done.

pchaurasia14:
Thank you @ThomasGoulet73 !

PR #6700, opened 2022-06-17
Reduce allocations when tracing routed events

bgrainger:
## Description

I was tracing an application with dotMemory and noticed a huge amount of temporary `object[]` being allocated by `EventRoute.InvokeHandlersImpl`.

image

This particular application had attached a listener to `PresentationTraceSources.RoutedEventSource` to trace some diagnostic information. While it's expected that doing so will add some overhead, it shouldn't be this great.

## Customer Impact

Merging this PR will greatly reduce allocations for customers whose applications add a listener to `PresentationTraceSources.RoutedEventSource.Listeners`. 

## Regression

No.

## Testing

Tested with dotMemory after shipping this fix in https://github.com/Faithlife/wpf.

## Risk

Minimal.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/wpf/pull/6700)

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/EventRoute.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/EventRoute.cs
@@ -193,7 +193,8 @@ namespace System.Windows
                     
                     // Invoke listeners
 
-                    if( TraceRoutedEvent.IsEnabled )
+                    var traceRoutedEventIsEnabled = TraceRoutedEvent.IsEnabled;
+                    if ( traceRoutedEventIsEnabled )
                     {
                         _traceArguments ??= new object[3];
                         _traceArguments[0] = _routeItemList[i].Target;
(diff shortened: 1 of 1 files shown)


src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/EventRoute.cs
```diff
@@ -195,25 +195,28 @@ private void InvokeHandlersImpl(object source, RoutedEventArgs args, bool reRais
...
+                            _traceArguments);
                     }
                     
                     _routeItemList[i].InvokeHandler(args);
 
                     if( TraceRoutedEvent.IsEnabled )
                     {
+                        _traceArguments ??= new object[3];
+                        _traceArguments[0] = _routeItemList[i].Target;
+                        _traceArguments[1] = args;
```
bgrainger, line 214: I'm assuming that `TraceRoutedEvent.IsEnabled` can be set to true concurrently, e.g., by attaching a listener on another thread. If that's not the case, and this can't change while the method is running, these redundant lines could be deleted.

Or we could read the `IsEnabled` property once into a local.

src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/EventRoute.cs
```diff
@@ -552,6 +555,11 @@ internal void Clear()
         // Stores Source Items for separated trees
         private FrugalStructList<SourceItem> _sourceItemList;
 
+        // Stores arguments that are passed to TraceRoutedEvent.Trace (to reduce allocations)
+        private object[] _traceArguments;
+        private static readonly object _true = true;
```
lindexi, line 560: Can you use the `BooleanBoxes` here ?

bgrainger: Great suggestion; didn't realise that class was available in WPF.

rchauhan18:
Thank you @bgrainger for the contribution.

PR #6714, opened 2022-06-22
DataGridColumnBorder caching of Freezable brushes is not actually caching anything at all

### LINKED ISSUE #6713: DataGridHeaderBorder.CacheFreezable method does not actually cache freezable brushes
Code: https://github.com/dotnet/wpf/blob/89d172db0b7a192de720c6cfba5e28a1e7d46123/src/Microsoft.DotNet.Wpf/src/Themes/Shared/Microsoft/Windows/Themes/DataGridHeaderBorder.cs#L365

The logic is backwards in that it only caches if the indexed value is _not_ null, versus caching if the indexed value _is_ null. This will result in freezable values never being cached and any calls to GetCachedFreezable will always return a null value causing brushes to be created every time a rendering pass is performed for the specific theme.

The code should instead be:

```csharp
        private static void CacheFreezable(Freezable freezable, int index)
        {
            Debug.Assert(freezable.IsFrozen, "Cached Freezables should be frozen.");

            lock (_cacheAccess)
            {
                if (_freezableCache[index] == null)
                {
                    _freezableCache[index] = freezable;
                }
            }
        }
```

ameltzerMSFT:
Fixes #6713 

## Description

CacheFreezable() logic is incorrect in that it checks if the indexed value is NOT null before caching versus if it IS null. This means that values will never actually be cached and any calls to GetCachedFreezable() will always return null causing the Brush to be regenerated rather than reused on every rendering pass.

## Customer Impact

Unnecessary computation on every rendering pass. Possible memory and brush leak.

## Regression

Does not appear to be a regression.

## Testing

Validated fix locally in private implementation of DataGridColumnBorder

## Risk

This should be minimal. DataGridColumnBorder code was clearly implemented assuming that brushes will be cached but they never actually are. There are various assertions in place already to ensure proper behavior here.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/wpf/pull/6714)

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/Themes/Shared/Microsoft/Windows/Themes/DataGridHeaderBorder.cs
+++ b/src/Microsoft.DotNet.Wpf/src/Themes/Shared/Microsoft/Windows/Themes/DataGridHeaderBorder.cs
@@ -158,7 +158,7 @@ namespace Microsoft.Windows.Themes
         /// </summary>
         public static readonly DependencyProperty SeparatorVisibilityProperty =
             DependencyProperty.Register("SeparatorVisibility", typeof(Visibility), typeof(DataGridHeaderBorder), new FrameworkPropertyMetadata(Visibility.Visible));
-        
+
         #endregion
 
         #region Layout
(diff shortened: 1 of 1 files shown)


src/Microsoft.DotNet.Wpf/src/Themes/Shared/Microsoft/Windows/Themes/DataGridHeaderBorder.cs
```diff
@@ -368,7 +368,7 @@ private static void CacheFreezable(Freezable freezable, int index)
 
             lock (_cacheAccess)
             {
-                if (_freezableCache[index] != null)
+                if (_freezableCache[index] == null)
```
dipeshmsft, line 371: On going through the code I agree that the logic earlier was not right. But with the current changes, any call to CacheFreezable won't be able to over-write the cache value once stored. 

Probably need to look a bit more to check if that happens / is a valid scenario.

dipeshmsft: I went through the code and this change is fine imo. As the CacheFreezable is only called when we get null from calling GetCachedFreezable. So, once cached we won't be overwriting the cache.

dipeshmsft (approved):

dipeshmsft:
@ameltzerMSFT , thanks for fixing the incorrect caching logic

PR #6749, opened 2022-07-03
Fix bind null name, error bind will throw NullReferenceException 

### LINKED ISSUE #4969: Two ways binding NullReferenceException
# Configuration

|Name|Version|
|:-:|:-:|
|dotnet tool| 5.0.302 |
|target framework| net452, net461, net472 |

# Problem

When I use background binding, provided `Path`, `Source` and set mode `TwoWay`, always throw NullReferenceException , but in net5.0 no exception, but how can i do when in NET Framework?

Whether i use `SetBinding` or `BindingOperations.SetBinding`.

# StackTrack

```
   at MS.Internal.AvTrace.AntiFormat(String s)
```

# Code

```csharp
var q = new Binding();
q.Path = new PropertyPath("RotateTransform");
q.Source = obj.RotateTransformDesigner;
q.Mode = BindingMode.TwoWay;
Bd.SetBinding(Control.RenderTransformProperty, q);
BindingOperations.SetBinding(Bd, Control.RenderTransformProperty, q);//ERROR
```

Cricle:
Fixes #4969 

Main PR

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/WindowsBase/MS/Internal/AvTrace.cs
+++ b/src/Microsoft.DotNet.Wpf/src/WindowsBase/MS/Internal/AvTrace.cs
@@ -402,6 +402,9 @@ namespace MS.Internal
         // replace { and } by {{ and }} - call if literal string will be passed to Format
         static public string AntiFormat(string s)
         {
+            if (string.IsNullOrEmpty(s))
+                return s;
+
             int formatIndex = s.IndexOfAny(FormatChars);
             if (formatIndex < 0)
                 return s;


miloush:
It would be good it you could fill the template a bit including whether/how you tested it.

As for the PR, I would also suggest the `AntiFormat` is removed on arguments in `AppendFormat`.

bgrainger:
It doesn't seem right to me to just ignore `null`s here; better would be to figure out when/why `null` is being passed to this method, and fix that. (It seems like an invariant is being violated, not that this method should be updated to accept `null`.)

miloush:
@bgrainger The code escapes some characters and if the string is `null`, the correct escaped version is `null`. It's intended use is for messages in string formatting, which are not allowed to be null, but there is no reason why these concerns shouldn't be separated.

The reason why `null` is being passed is because the method is incorrectly called in the `AppendFormat` method I mentioned for _values_, which are allowed to be `null`, not _messages_. I agree fixing that should be part of the PR.

Cricle:
Would it be better?
```csharp
if (string.IsNullOrEmpty(s))
    return s;
//Change it to
Debug.Assert(s!=null,"some message");
```
And fix call error.

Issue #6777, opened 2022-07-12
System.OverflowException in PresentationFramework.dll in System.Windows.Shell.WindowChromeWorker


lindexi:
* .NET Core Version: All
* Windows version: All
* Does the bug reproduce also in WPF for .NET Framework 4.8?: Yes
 
 **Problem description:**

We have been getting a crash when using the mouse on certain windows in our WPF application. This Stackoverflow issue appears with exactly the same error messages as our own so I believe it to be relevant. https://stackoverflow.com/questions/33287542/what-would-cause-wm-nchittest-lparam-to-overflow-a-32-bit-integer

This issue has also been reported to the Visual Studio community site and a simple test application has been provided. https://developercommunity.visualstudio.com/content/problem/167357/overflow-exception-in-windowchrome.html

The crashes occur on a Windows 10 Pro Version 1709 OS Build 16299.309, win7 sp1, Windows10 21H2 19044.1645

I run the application with .NET 6.0.5 and @contre run the application with  .NET Framework 4.6.2

 
 **Actual behavior:** 

```
System.OverflowException
  HResult=0x80131516
  Message=Arithmetic operation resulted in an overflow.
  Source=PresentationFramework
  StackTrace:
   at System.Windows.Shell.WindowChromeWorker._HandleNCHitTest(WM uMsg, IntPtr wParam, IntPtr lParam, Boolean& handled)
   at System.Windows.Shell.WindowChromeWorker._WndProc(IntPtr hwnd, Int32 msg, IntPtr wParam, IntPtr lParam, Boolean& handled)
   at System.Windows.Interop.HwndSource.PublicHooksFilterMessage(IntPtr hwnd, Int32 msg, IntPtr wParam, IntPtr lParam, Boolean& handled)
   at MS.Win32.HwndWrapper.WndProc(IntPtr hwnd, Int32 msg, IntPtr wParam, IntPtr lParam, Boolean& handled)
   at MS.Win32.HwndSubclass.DispatcherCallbackOperation(Object o)
   at System.Windows.Threading.ExceptionWrapper.InternalRealCall(Delegate callback, Object args, Int32 numArgs)
   at System.Windows.Threading.ExceptionWrapper.TryCatchWhen(Object source, Delegate callback, Object args, Int32 numArgs, Delegate catchHandler)
   at System.Windows.Threading.Dispatcher.LegacyInvokeImpl(DispatcherPriority priority, TimeSpan timeout, Delegate method, Object args, Int32 numArgs)
[...]

PR #6779, opened 2022-07-12
Fix System.OverflowException in WindowChromeWorker._HandleNCHitTest

lindexi:
Fixes https://github.com/dotnet/wpf/issues/6777

## Description

The lParam may be the int 64 value which will throw OverflowException when cast to int 32. We only use the int32 part inside the int64, so it's safe for us to use ToInt64 first and then cast to int32.

## Customer Impact

See https://github.com/dotnet/wpf/issues/6777

## Regression

None.

## Testing

Just the CI.

## Risk

Low. It is safe for us to use IntPtr.ToInt64 and cast the long to int.


miloush:
The comments are not very useful on their own. If you want to include any, it would be better to say what's going on, like "// avoid overflow for negative coordinates (#6777)".

Otherwise looks good to me.

lindexi:
Thank you @miloush and I fixed it.

anjalisheel-wpf (approved):

dipeshmsft:
@lindexi thank you for the contribution. @miloush thanks for your review comments.

lindexi:
Thank you @dipeshmsft

Issue #6452, opened 2022-04-18
WeakEventManager lock issue


zgabi:
.NET 6.0.202
Windows Server 2022 21H2 20348.643
Does the bug reproduce also in WPF for .NET Framework 4.8?: yes

 **Problem description:**
Collection was modified exception in `WeakEventTable `cleanup.
It seems that the problem occures only when `EnableWeakEventMemoryImprovements` is true. 

_dataTable is changed in the PurgeInternal code in this loop:
https://github.com/dotnet/wpf/blob/89d172db0b7a192de720c6cfba5e28a1e7d46123/src/Microsoft.DotNet.Wpf/src/WindowsBase/MS/Internal/WeakEventTable.cs#L336-L348

Somewhere here:
https://github.com/dotnet/wpf/blob/bdfe29afcaa879c55aa3ed6541698cdc42be4dd8/src/Microsoft.DotNet.Wpf/src/WindowsBase/System/Windows/WeakEventManager.cs#L406-L445

 **Actual behavior:** 
I receive the following exception sometimes which terminates the whole process (of course I could catch it in the `Dispatcher.UnhandledException` event handler):

System.InvalidOperationException: Collection was modified; enumeration operation may not execute.
   at System.Collections.Hashtable.HashtableEnumerator.MoveNext()
   at MS.Internal.WeakEventTable.Purge(Boolean purgeAll)
   at MS.Internal.CleanupHelper.DoCleanup(Boolean forceCleanup)
   at MS.Internal.CleanupHelper.OnCleanupTick(Object sender, EventArgs e)
   at System.Windows.Threading.DispatcherTimer.FireTick()
   at System.Windows.Threading.ExceptionWrapper.InternalRealCall(Delegate callback, Object args, Int32 numArgs)
   at System.Windows.Threading.ExceptionWrapper.TryCatchWhen(Object source, Delegate callback, Object args, Int32 numArgs, Delegate catchHandler)
 
 **Expected behavior:**
No unhandled exception
 
 **Minimal repro:**
Open the attached project, execute it and press the button several times.
[WeakEventManagerTest.zip](https://github.com/dotnet/wpf/files/8506598/WeakEventManagerTest.zip)

lindexi:
It seems to be a bug

Camios:
Also reproducible on .NET 8 and .NET 10

kmaddock:
The problem is in WeakEventManager.Purge:
```
[...]

PR #6790, opened 2022-07-13
Adding the lock to keep the WeakEventTable ThreadSafe

lindexi:
Fixes https://github.com/dotnet/wpf/issues/6452

## Description

The `_dataTable` is the field in WeakEventTable which will be read and written in a thread-unsafe code. See https://github.com/dotnet/wpf/issues/6452

## Customer Impact

Fix the WeakEventTable thread safe issues.

## Regression

None.

## Testing

Just CI.

## Risk

The lock may cause the performance issues.

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/WindowsBase/MS/Internal/WeakEventTable.cs
+++ b/src/Microsoft.DotNet.Wpf/src/WindowsBase/MS/Internal/WeakEventTable.cs
@@ -137,15 +137,21 @@ namespace MS.Internal
         {
             get
             {
-                EventKey key = new EventKey(manager, source);
-                object result = _dataTable[key];
-                return result;
+                using (ReadLock)
+                {
+                    EventKey key = new EventKey(manager, source);
+                    object result = _dataTable[key];
+                    return result;
+                }
             }
 
             set
             {
-                EventKey key = new EventKey(manager, source, true);
-                _dataTable[key] = value;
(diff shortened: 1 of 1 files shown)


ThomasGoulet73:
I think those places do not lock by design, they delegate the responsibility of locking to the caller I suppose. That means that the places that now lock in these changes might already be locked.

Here's an example:
https://github.com/dotnet/wpf/blob/df1e6313039c58d5d7e10bb4ab5379c45c302646/src/Microsoft.DotNet.Wpf/src/WindowsBase/System/Windows/WeakEventManager.cs#L274-L276

In this example, it locks `Table.WriteLock` and then uses the indexer for `Table` which in your changes now call locks `Table.WriteLock`.

lindexi:
@ThomasGoulet73 How about replace ConcurrentDictionary to it?

src/Microsoft.DotNet.Wpf/src/WindowsBase/MS/Internal/WeakEventTable.cs
```diff
@@ -177,13 +183,16 @@ internal bool IsCleanupEnabled
         internal void Remove(WeakEventManager manager, object source)
         {
             EventKey key = new EventKey(manager, source);
```
IAmTheCShark, line 185: Nitpick: This is the only place where you create the key outside of the lock.

Issue #6792, opened 2022-07-12
Visual Studio 17.2.6 Breaks CommunityToolkit.Mvvm


jamiehankins:
**Version Used**:  17.2.6

**Steps to Reproduce**:

1. Create a WPF exe project targeting `net6.0-windows`
2. Reference CommunityToolkit.Mvvm like so:
`<PackageReference Include="CommunityToolkit.Mvvm" Version="7.1.2" />`
4. Create a simple class inheriting from `ObservableObject`, like so:
```
using CommunityToolkit.Mvvm.ComponentModel;

namespace ObservablePropertyTest
{
    public partial class TestVM : ObservableObject
    {
        [ObservableProperty]
        private string _testString = string.Empty;
    }
}
```

**Expected Behavior**:
In 17.2.5, it built and ran fine.

**Actual Behavior**:
```
Rebuild started...
1>------ Rebuild All started: Project: ObservablePropertyTest, Configuration: Debug Any CPU ------
Restored C:\git\ObservablePropertyTest\ObservablePropertyTest.csproj (in 2 ms).
1>C:\git\ObservablePropertyTest\CommunityToolkit.Mvvm.SourceGenerators\CommunityToolkit.Mvvm.SourceGenerators.ObservablePropertyGenerator\__KnownINotifyPropertyChangedOrChangingArgs.cs(12,27,12,70): error CS0101: The namespace 'CommunityToolkit.Mvvm.ComponentModel.__Internals' already contains a definition for '__KnownINotifyPropertyChangedOrChangingArgs'
1>C:\git\ObservablePropertyTest\CommunityToolkit.Mvvm.SourceGenerators\CommunityToolkit.Mvvm.SourceGenerators.ObservablePropertyGenerator\__KnownINotifyPropertyChangedOrChangingArgs.cs(7,6,7,51): error CS0579: Duplicate 'global::System.CodeDom.Compiler.GeneratedCode' attribute
1>C:\git\ObservablePropertyTest\CommunityToolkit.Mvvm.SourceGenerators\CommunityToolkit.Mvvm.SourceGenerators.ObservablePropertyGenerator\__KnownINotifyPropertyChangedOrChangingArgs.cs(8,6,8,52): error CS0579: Duplicate 'global::System.Diagnostics.DebuggerNonUserCode' attribute
[...]

PR #6680, opened 2022-06-14
Fix build when using source generators (#6534)

singhashish-wpf:
Fixes dotnet/wpf#6522
Fixes https://github.com/dotnet/roslyn/issues/60257

## Description
Fixes the build when using source generators.

## Customer Impact
Without this PR, customers cannot build a WPF application with source generators.

## Regression
No. To my knowledge, it never worked.

## Testing
Tested by building a blank WPF application that uses the Regex source generator. This PR fixed the build.

## Risk
Low.

/cc @RussKie (It looks like your hunch about the `Analyzers` property was right)

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/wpf/pull/6680)

PR #6534, opened 2022-05-04
Fix build when using source generators

ThomasGoulet73:
Fixes dotnet/wpf#6522
Fixes https://github.com/dotnet/roslyn/issues/60257

## Description
Fixes the build when using source generators.

## Customer Impact
Without this PR, customers cannot build a WPF application with source generators.

## Regression
No. To my knowledge, it never worked.

## Testing
Tested by building a blank WPF application that uses the Regex source generator. This PR fixed the build.

## Risk
Low.

/cc @RussKie (It looks like your hunch about the `Analyzers` property was right)

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/wpf/pull/6534)

PR #6793, opened 2022-07-14
Fix build when using source generators coming from NuGet packages.

ThomasGoulet73:
Fixes dotnet/wpf#6792

## Description
Fixes build of project using source generators coming from NuGet packages. I fixed this by only including analyzers that are coming from a FrameworkReference in the temporary target assembly project. Those analyzers are the ones that are not automatically added in the temporary target assembly project because `ResolveTargetingPackAssets` is never called (We could investigate calling this target in the future but, to me, it seemed riskier than the changes in this PR).

## Customer Impact
Fixes build.

## Regression
This is a regression introduced in .Net 6.0.7 by #6534 which was backported by #6680. Those were changes from me, sorry about the regression.

## Testing
Tested locally by building a project which uses Regex source generator, WinForms source generator and [CommunityToolkit.Mvvm 7.1.2](https://www.nuget.org/packages/CommunityToolkit.Mvvm/7.1.2) source generator (This is an example of a source generator that currently fails on .Net 6.0.7 SDK).

## Risk
Low.


ThomasGoulet73:
@dotnet/wpf-developers: This should probably be considered to be backported to .Net 6.0.

candritzky:
This is just to confirm that this patch will also fix the problems we're having with WPF class libraries in tandem with Refit.

dipeshmsft (approved):

Issue #6829, opened 2022-07-20
WPF tocuh in Window with StylusPlugIn may throw InvalidOperationException 


lindexi:
* .NET Core Version: 6.0.121.56705
* Windows version: win10 21H2
* Does the bug reproduce also in WPF for .NET Framework 4.8?: No

 
 **Problem description:**

After https://github.com/dotnet/wpf/pull/2891 the `Stylus Input` will call the `GetAndCacheTransformToDeviceMatrix` to get the Matrix to calculate.

https://github.com/dotnet/wpf/blob/ab8e3f026bc4319aae72e0b3b0383acbc500aeef/src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Input/Stylus/Common/StylusLogic.cs#L417-L435

After we changed the moniter DPI and then we touch the application immediately, the `GetAndCacheTransformToDeviceMatrix` will call `hwndSource.CompositionTarget.TransformToDevice` and then throw the System.InvalidOperationException: The calling thread cannot access this object because a different thread owns it.

https://github.com/dotnet/wpf/blob/ab8e3f026bc4319aae72e0b3b0383acbc500aeef/src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Input/Stylus/Common/StylusLogic.cs#L427

Because we can not find the new DpiScale in `_transformToDeviceMatrices`
 
 **Actual behavior:** 

```
Application: Application.exe
CoreCLR Version: 6.0.121.56705
.NET Version: 6.0.1
Description: The process was terminated due to an unhandled exception.
Exception Info: System.InvalidOperationException: The calling thread cannot access this object because a different thread owns it.
   at System.Windows.Threading.Dispatcher.ThrowVerifyAccess()
   at System.Windows.Threading.Dispatcher.VerifyAccess()
   at System.Windows.Threading.DispatcherObject.VerifyAccess()
   at System.Windows.Media.CompositionTarget.VerifyAPIReadOnly()
   at System.Windows.Interop.HwndTarget.get_TransformToDevice()
   at System.Windows.Input.StylusLogic.GetAndCacheTransformToDeviceMatrix(PresentationSource source)
   at System.Windows.Input.StylusWisp.WispLogic.GetTabletToViewTransform(PresentationSource source, TabletDevice tabletDevice)
   at System.Windows.Input.PenContexts.InvokeStylusPluginCollection(RawStylusInputReport inputReport)
[...]

PR #6840, opened 2022-07-22
Fix get TransformToDevice in Stylus Input thread will throw the InvalidOperationException

lindexi:
Fixes https://github.com/dotnet/wpf/issues/6829

## Description

See https://github.com/dotnet/wpf/issues/6829

The `Stylus Input` thread will enter the GetAndCacheTransformToDeviceMatrix method to get the Matrix. But after we change the DPI, the `Stylus Input` thread can not find the Matrix in `_transformToDeviceMatrices`. It means that the `Stylus Input` thread should get the Matrix from `HwndTarget.TransformToDevice` property which only allow get from UI Thread.

I fix it by dispatch to UI Thread. To improve performance, we will calculate it `CurrentDpiScale` property.

## Customer Impact

Fix tocuh in Window with StylusPlugIn may throw InvalidOperationException

## Regression

Yes, see https://github.com/dotnet/wpf/pull/2891

## Testing

Just CI

## Risk

Low.

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Input/Stylus/Common/StylusLogic.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Input/Stylus/Common/StylusLogic.cs
@@ -419,16 +419,42 @@ namespace System.Windows.Input
             var hwndSource = source as HwndSource;
             Matrix toDevice = Matrix.Identity;
 
-            if (hwndSource?.CompositionTarget != null)
+            HwndTarget compositionTarget = hwndSource?.CompositionTarget;
+            if (compositionTarget != null)
             {
-                // If we have not yet seen this DPI, store the matrix for it.
-                if (!_transformToDeviceMatrices.ContainsKey(hwndSource.CompositionTarget.CurrentDpiScale))
+                // We can calculate the Matrix fast without any cache and avoid any thread issues.
+                // If others inherit HwndTarget, the following rules may not apply. Because they can override the TransformToDevice property.
+                // Fix https://github.com/dotnet/wpf/issues/6829
+                Type compositionTargetType = compositionTarget.GetType();
+                if (compositionTargetType == typeof(HwndTarget))
                 {
-                    _transformToDeviceMatrices[hwndSource.CompositionTarget.CurrentDpiScale] = hwndSource.CompositionTarget.TransformToDevice;
-                    Debug.Assert(_transformToDeviceMatrices[hwndSource.CompositionTarget.CurrentDpiScale].HasInverse);
+                    var hwndTarget = (HwndTarget) compositionTarget;
+                    DpiScale2 currentDpiScale = hwndTarget.CurrentDpiScale;
+                    toDevice.Scale(currentDpiScale.DpiScaleX, currentDpiScale.DpiScaleY);
                 }
+                else
+                {
+                    // If we have not yet seen this DPI, store the matrix for it.
+                    if (!_transformToDeviceMatrices.ContainsKey(compositionTarget.CurrentDpiScale))
+                    {
+                        // The Stylus Input thread will enter this case.
+                        if (compositionTarget.Dispatcher.CheckAccess())
+                        {
+                            _transformToDeviceMatrices[compositionTarget.CurrentDpiScale] = compositionTarget.TransformToDevice;
+                        }
+                        else
+                        {
+                            // We have to run it in UI Thread, see https://github.com/dotnet/wpf/issues/6829
+                            compositionTarget.Dispatcher.Invoke(() =>
+                                _transformToDeviceMatrices[compositionTarget.CurrentDpiScale] =
+                                    compositionTarget.TransformToDevice);
+                        }
 
-                toDevice = _transformToDeviceMatrices[hwndSource.CompositionTarget.CurrentDpiScale];
+                        Debug.Assert(_transformToDeviceMatrices[compositionTarget.CurrentDpiScale].HasInverse);
+                    }
+
+                    toDevice = _transformToDeviceMatrices[compositionTarget.CurrentDpiScale];
+                }
             }
 
             return toDevice;


src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Input/Stylus/Common/StylusLogic.cs
```diff
@@ -419,16 +419,41 @@ protected Matrix GetAndCacheTransformToDeviceMatrix(PresentationSource source)
             var hwndSource = source as HwndSource;
             Matrix toDevice = Matrix.Identity;
 
-            if (hwndSource?.CompositionTarget != null)
+            HwndTarget compositionTarget = hwndSource?.CompositionTarget;
+            if (compositionTarget != null)
```
miloush, line 423:
```C#
if (hwndSource?.CompositionTarget is HwndTarget compositionTarget)
```

src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Input/Stylus/Common/StylusLogic.cs
```diff
@@ -419,16 +419,41 @@ protected Matrix GetAndCacheTransformToDeviceMatrix(PresentationSource source)
...
+                // We can calculate the Matrix fast without any cache and avoid any thread issues.
+                // If others inherit HwndTarget, the following rules may not apply. Because they can override the TransformToDevice property.
+                // Fix https://github.com/dotnet/wpf/issues/6829
+                Type compositionTargetType = compositionTarget.GetType();
+                if (compositionTargetType == typeof(HwndTarget))
                 {
-                    _transformToDeviceMatrices[hwndSource.CompositionTarget.CurrentDpiScale] = hwndSource.CompositionTarget.TransformToDevice;
-                    Debug.Assert(_transformToDeviceMatrices[hwndSource.CompositionTarget.CurrentDpiScale].HasInverse);
+                    DpiScale2 currentDpiScale = compositionTarget.CurrentDpiScale;
+                    toDevice.Scale(currentDpiScale.DpiScaleX, currentDpiScale.DpiScaleY);
```
miloush, line 432: is the result is not cached?

src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Input/Stylus/Common/StylusLogic.cs
```diff
@@ -419,16 +419,41 @@ protected Matrix GetAndCacheTransformToDeviceMatrix(PresentationSource source)
...
                 }
+                else
+                {
+                    // If we have not yet seen this DPI, store the matrix for it.
+                    if (!_transformToDeviceMatrices.ContainsKey(compositionTarget.CurrentDpiScale))
```
miloush, line 437: `TryGetValue` to avoid second lookup in case it exists

src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Input/Stylus/Common/StylusLogic.cs
lindexi: @miloush Do you think cache is fast than calculation?

src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Input/Stylus/Common/StylusLogic.cs
lindexi: @miloush Thank you.

src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Input/Stylus/Common/StylusLogic.cs
```diff
@@ -419,16 +419,41 @@ protected Matrix GetAndCacheTransformToDeviceMatrix(PresentationSource source)
...
+                        if (compositionTarget.Dispatcher.CheckAccess())
+                        {
+                            _transformToDeviceMatrices[compositionTarget.CurrentDpiScale] = compositionTarget.TransformToDevice;
+                        }
+                        else
+                        {
+                            // We have to run it in UI Thread, see https://github.com/dotnet/wpf/issues/6829
+                            compositionTarget.Dispatcher.Invoke(() =>
+                                _transformToDeviceMatrices[compositionTarget.CurrentDpiScale] =
+                                    compositionTarget.TransformToDevice);
```
miloush, line 449: wouldn't of the overloads with arguments be better to avoid creating a closure?

lindexi: @miloush Thank you. This code branch rarely enters. Maybe keeping it simple is a good thing. Do you agree?

miloush:
I am wondering why is the class a `DispatcherObject` and whether the special case for `HwndTarget` is worth the performance gain. If `HwndTarget` is updated in the future to support other kinds of transforms, someone will have to magically remember that this code needs to be updated.

lindexi:
> If `HwndTarget` is updated in the future to support other kinds of transforms, someone will have to magically remember that this code needs to be updated.

@miloush  In fact, I think it is more correct to use CurrentDpiScale. Thank you. 

Why we use `TransformToDevice` ? Cc @rladuca

rchauhan18:
Hey @lindexi ,
I'm having trouble reproducing this issue using the steps you provided. I might be overlooking something, so could you assist me by sharing a sample repository to clarify?
<details>
<summary>Here is the code I am using:</summary>
<br>

```
{
    public MainWindow()
    {
        InitializeComponent();
        
        this.StylusPlugIns.Add(new CustomStylusPlugin());
    }
   
}

public class CustomStylusPlugin : StylusPlugIn
{
    {
        base.OnStylusDown(rawStylusInput);

        // Get the points
        StylusPointCollection stylusPoints = rawStylusInput.GetStylusPoints();

        // Output points
        foreach (StylusPoint point in stylusPoints)
        {
            System.Diagnostics.Debug.WriteLine($"X: {point.X}, Y: {point.Y}");
        }
    }
}
```
</details>

lindexi:
@rchauhan18 Do you changed the DPI? Please make sure the `_transformToDeviceMatrices` do not contain the `CurrentDpiScale`.

rchauhan18:
@lindexi Yes, I am able to add a break point in ```GetAndCacheTranformToDeviceMatrix``` method. However, the issue I'm encountering is that when I change the DPI value from the settings, the application's DPI value doesn't update dynamically. To see the change, I have to restart the application. Therefore, the count always remains at one.

lindexi:
@rchauhan18 How about multi-screen with different DPI configuration? Moving the app between the two screens.

LickBag:
> @rchauhan18 Do you change the DPI? Please make sure the `_transformToDeviceMatrices` do not contain the `CurrentDpiScale`.
> 
> And could you debug and add the break point in `GetAndCacheTransformToDeviceMatrix` method? Can the `Stylus Input` thread enter the `GetAndCacheTransformToDeviceMatrix` method? If the `Stylus Input` thread can enter the `GetAndCacheTransformToDeviceMatrix` method, it means that the `Stylus Input` thread will visit the `hwndSource.CompositionTarget.TransformToDevice` which will throw InvalidOperationException.

Stylus Input Thread
I met this problem, the `Stylus Input` thread can enter the `GetAndCacheTransformToDeviceMatrix` method, and throw the InvalidOperationException.
Please Fix It.

lindexi:
@LickBag Could you test my custom WPF version?

1. Install or download the assets resouce from `dotnetCampus.WPF.Resource` package in your .NET 6 project.

```xml
    <PackageVersion Include="dotnetCampus.WPF.Resource" Version="6.0.4-alpha07-test06" />
```

2. Publish your project with self contained

3. Copy and replace the dll from `C:\Users\<Your UserName>\.nuget\packages\dotnetcampus.wpf.resource\6.0.4-alpha07-test06\Assets\` folder

PR #6873, opened 2022-08-09
Sort Datagrid Column by Keyboard F3

### LINKED ISSUE #6737: A11y_.NET Core_WPF_ExpenseItIntro_ExpenseReportfor_Keyboard: The table column headers are not accessible by keyboard.
“[Check out Accessibility Insights!](https://accessibilityinsights.io/) - Identify accessibility bugs before check-in and make bug fixing faster and easier.”

**GitHubTags:**#A11yMAS;#A11yTCS;#.NET Core[WPF]-Win32-June2022;#.NET Core;#A11ySev2;#WCAG2.1.1;#DesktopApp;#Win11;#Keyboard;#Benchmark;

### Environment Details:
.NET Core
WPF
Operating System: Windows 11 Enterprise 21H2

### Steps to Reproduce:

1. Launch VS 2019 Int Preview
2. Navigate to Getting started and right click on the Expense it intro and click on build
3. Then right click on Expense it Intro then Click on Debug-->Start New Instance. 
4. Navigate to Names list and select any name and click on View.
5. 'Expense Report for' screen will open
6. Tab till table and observe the issue

### Actual:
The table column headers "ExpenseType" and "Amount" are not accessible by keyboard. While using mouse and clicking on this column headers sorting functionality is performed but we are not able to use sorting functionality by using keyboard.

### Expected:
The table column headers "ExpenseType" and "Amount" should be accessible by keyboard. So that the sorting functionality should be accessible by using keyboard.

### User Impact:
Keyboard users will not be able to sort the data by using keyboard.
[WPF keyboard.zip](https://github.com/dotnet/wpf/files/9008397/WPF.keyboard.zip)

singhashish-wpf:
Fixes #6737 

## Description
Datagrid column headers are not accessible to keyboard users for performing sorting.
Provided a way (F3) on datagrid cell to sort the column for keyboard users.

## Customer Impact
Keyboard users won't be able to sort datagrid columns.

## Regression
No

## Testing
Ad hoc testing via sample app.

## Risk
Low

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/wpf/pull/6873)


src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Controls/DataGridCell.cs
```diff
@@ -985,7 +985,15 @@ protected override void OnKeyDown(KeyEventArgs e)
...
             }
-            
+
+            else if(e.Key == Key.F3)
+            {
+                if (Column.CanUserSort)
+                {
+                    Column.DataGridOwner.PerformSort(Column);
+                }
+            }
```
dipeshmsft, line 997: I think we should mark the event as handled here.

src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Controls/DataGridCell.cs
```diff
@@ -985,7 +985,17 @@ protected override void OnKeyDown(KeyEventArgs e)
...
-            
+
+            else if(e.Key == Key.F3)
+            {
+                if (Column.CanUserSort)
+                {
+                    Column.DataGridOwner.PerformSort(Column);
+                }
+                e.Handled = true;
+                return;
```
Symbai, line 996: You should either merge the IF-case like that `else if(e.Key == Key.F3 && Column.CanUserSort)` or move these two lines within the last IF. Because if a column cannot be sorted and  a user presses F3, you still set `e.Handled` to `true` and return so the input is never send to column.

PR #7033, opened 2022-09-13
Generated the AssemblyRecord code in AssemblyHelper to improve startup performance

lindexi:
## Description

The AssemblyHelper will create the `AssemblyRecord[]` by UncommonAssembly. And the code will be run when the application startup.

And I think it's worth doing this optimization to improve startup performance. And I don't think these Uncommon Assembly will be added or changed.

Benchmark result:

``` ini

BenchmarkDotNet=v0.13.2, OS=Windows 10 (10.0.19044.1645/21H2/November2021Update)
Intel Core i7-6700 CPU 3.40GHz (Skylake), 1 CPU, 8 logical and 4 physical cores
.NET SDK=6.0.400
  [Host]     : .NET 6.0.8 (6.0.822.36306), X64 RyuJIT AVX2
  DefaultJob : .NET 6.0.8 (6.0.822.36306), X64 RyuJIT AVX2


```
|         Method |      Mean |    Error |   StdDev | Ratio |
|--------------- |----------:|---------:|---------:|------:|
|   OriginalCode | 281.07 ns | 4.154 ns | 3.469 ns |  1.00 |
| OptimizedCode1 |  36.25 ns | 0.703 ns | 1.030 ns |  0.13 |
| OptimizedCode2 |  18.35 ns | 0.449 ns | 0.584 ns |  0.07 |

## Customer Impact

Slightly slow startup performance issues.

## Regression

None.

## Testing

Just CI and the benchmark code.

## Risk

Very low. The code is changed equivalently. No behavior was changed.


dipeshmsft:
@lindexi, this PR looks good, but our only concern with this PR is that hard-coding the list will make the maintenance harder. Thoughts?

lindexi:
Thank you @dipeshmsft 

I agree with you. But I don't think it's possible to add new assemblies. In other words, this is the constant.

dipeshmsft:
Thanks @lindexi for your contributions in performance.

miloush:
> And I don't think these Uncommon Assembly will be added or changed.

Why do you think so? They have been changed in the past, cf. https://referencesource.microsoft.com/#WindowsBase/Base/MS/Internal/AssemblyHelper.cs,d11803a9eea58251

lindexi:
@miloush What is `cf` means?

lindexi:
@miloush Thank you. You are right. Can I ask you for advice?

miloush:
Well the fact that they are both in the same file helps a bit with the maintenance. I would have put a comment on the enum that the static AssemblyHelper constructor needs to be updated when changes are done to the enum.

Also why didn't you just do

```c#
_records = new AssemblyRecord[5];
_records[0].Name = "System.Drawing.Common,";
_records[1].Name = "System.Private.Xml,";
...
```

However since this is merged now, I guess you can leave it be.

PR #7181, opened 2022-10-11
Use ArgumentNullException.ThrowIfNull in System.Xaml

ThomasGoulet73:
## Description
Use ArgumentNullException.ThrowIfNull in System.Xaml which is shorter and cleaner (IMO). The IL is also smaller and the JIT should be able to better inline the code since it moves the throw into a non-inlineable method.

In the future we could remove throws of ArgumentNullException inside a null-coalescing operator ( `_value = value ?? throw new ArgumentNullException(nameof(value))`) by using something similar to WinForms' `OrThrowIfNull`:
https://github.com/dotnet/winforms/blob/13d4d7813ed0d894c239e2f696106c166d130c43/src/System.Windows.Forms.Primitives/src/System/Windows/Forms/Internals/ArgumentValidation.cs#L11-L15

## Customer Impact
Might be faster.

## Regression
No.

## Testing
Local build + CI.

## Risk
Low.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/wpf/pull/7181)

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/System.Xaml/System/Windows/Markup/ArrayExtension.cs
+++ b/src/Microsoft.DotNet.Wpf/src/System.Xaml/System/Windows/Markup/ArrayExtension.cs
@@ -46,10 +46,7 @@ namespace System.Windows.Markup
         /// <param name="elements">The array to write</param>
         public ArrayExtension(Array elements)
         {
-            if (elements == null)
-            {
-                throw new ArgumentNullException(nameof(elements));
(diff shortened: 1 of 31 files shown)


dipeshmsft:
@ThomasGoulet73, thank you for this contribution.

ThomasGoulet73:
Thanks @dipeshmsft

Issue #7125, opened 2022-09-27
My WPF project has a namespace error


344089386:
Error Tip:
No type or namespace name "Windows" in namespace "test.wpf.system" (missing assembly reference?)

This error occurs when my project contains the System namespace.
I predict this is due to the implicit namespaces of.net6.
I tried `<DisableImplicitNamespaceImports>true</DisableImplicitNamespaceImports>` and `<Using Remove="XXX.System" />`  but it didn't work.

![image](https://user-images.githubusercontent.com/22017199/191935560-2c0b601d-507b-481b-921a-fd2807f9cca7.png)

Repeat steps：
build project

Error Code：
https://github.com/344089386/WpfTest

miloush:
I don't think this has anything to do with implicit usings. You are introducing `Test.Wpf.System` namespace in `UserInfo.cs` and your window is in the `Test.Wpf` namespace. All `System` namespaces then resolve to `Test.Wpf.System` which has precedence.

So don't do that. Technically you could argue that the code generator should use global namespace alias, though I think you would have to present a compelling justification for such change.

344089386:
This is a large engineering System, and there are other hierarchies that use similar System namespaces. This problem occurs when WPF references other libraries.

1, if you change the namespace of other class libraries, then the change will be a little big.

2. I think there should be a default setting that you can configure and disable.

miloush:
The problem is if you put e.g. a Window into that hierarchy. The libraries cannot inherit anything from the System namespace and they cannot be used by anything that is in the same hierarchy, unless specific and not very common syntax is used. That seems like an unfortunate choice of namespaces.

That said, the resource generator for example does use global namespace alias, so it is not an unreasonable expectation. If we do that, we might break some applications that rely on this behavior, but I don't expect that to be of a significant risk.

344089386:
Can I only change the name of the namespace now? Is there a middle ground?

miloush:
Not that I am aware of, you would have to change the root namespace, in your example put your App and MainWindow into TestUI.Wpf namespace.

344089386:
All right. Thank you for your trouble. I'm gonna go change the namespace

chris-kruining:
> I don't think this has anything to do with implicit usings. You are introducing `Test.Wpf.System` namespace in `UserInfo.cs` and your window is in the `Test.Wpf` namespace. All `System` namespaces then resolve to `Test.Wpf.System` which has precedence.
> 
[...]

Issue #4014, opened 2021-01-15
Enable Top-Level Namespaces for Xaml Classes in Visual Basic


A-Badry:
**Targeting:** Visual Basic + XAML

**Abstract**
Being able to create a xaml class under a top-level namespace other than the project’s Root Namespace in Visual Basic.
For example: The desired is:  x:Class="SomeNamespace.MyClass" creates 'SomeNamespace' as a top-level (global) namespace.
What happens for now:   x:Class="SomeNamespace.MyClass" creates 'SomeNamespace' as a nested namespace under the project's RootNamespace.

**Why do we need this?**
Consider the following scenario, you have reusable code files that are added to your Visual Studio projects as links, these files contains xaml files that define some controls, some of these controls are used within the others. That means you need to declare clr-namespace of some of these controls within xaml code of the others. Here, if the clr-namespace contains the root-namespace of current project (and it should, as long as you are using VB), then the reusable file cannot be shared across projects anymore. The solution for this is to be able to define xaml classes under another top-level namespace so it can be reusable across different projects when referencing its file as link.

**A solution, hard on developers:**
Is to erase the ‘Root namespace’ field found in the project properties page. In this case any namespace declared within x:Class declaration will be top-level (Global). Anyhow in this case a namespace declaration should be added manually to all code files, which is a minus in productivity.

**Another impractical solution:**
In your xaml, define x:class like this: x:Class=“SomeNamespac.ClassName”.
In related code-behind file, include the class into Global.SomeNamespace namespace.
Go to path obj\Debug and change namespace in files ClassName.g.i.vb and ClassName.g.vb to Global.SomeNamespace.
Done! now the xaml class is in a separate top-level namespace, but that will last only until the next time the IDE generates its auto generated files (when you clean project or rebuild it).

**The perfect solution**:
Is to make the Visual Basic's XAML interpreter acts as follows:
* x:Class="SomeNamespace.MyClass" means generating SomeNamespace as a top-level namespace not nesting it under the root namespace of the project, this is the major idea in this feature suggestion.
* x:Class="MyClass" means generating 'MyClass' under the root namespace of the project (this is the case already), and make that equivalent to: x:Class="RootNamespace.MyClass" (this is not the case now).
* x:Class="RootNamespace.SomeNamespace.MyClass" puts 'SomeNamespace' as a nested namespace under the project's root namespace.
 
Many Thanks!

miloush:
[...]

PR #7196, opened 2022-10-13
Fully qualify types in code generated by markup compiler

### LINKED ISSUE #7173: namespace collision in clr-namespace
* .NET Core Version: 7.0-rc1
* Windows version: 11
* Does the bug reproduce also in WPF for .NET Framework 4.8?: Don't know, not tested
 
 **Problem description:**
  `MainWindow.xaml(28, 27): [CS0234] The type or namespace name 'Wpf' does not exist in the namespace 'Comlink.CefSharp' (are you missing an assembly reference?)`

This is the error I get, the error itself is "correct". that namespace does not contain Wpf. However, .NET shouldn't even be looking in that namespace in the first place, since I have defined an assembly in my xaml file. 

to give some more details:
`/src/View/MainWindow.xaml`
```c#
<Window 
    x:Class="Comlink.Example.View.MainWindow"
    xmlns:cefSharp="clr-namespace:CefSharp.Wpf;assembly=CefSharp.Wpf"
    mc:Ignorable="d"
    Title="Comlink examples" 
    Height="450" 
    Width="800"
>
    <cefSharp:ChromiumWebBrowser x:Name="CefSharpBrowser" Address="https://google.com" />
</Window>
```

`/src/View/MainWindow.xaml`
```c#
using System.Windows;
using Comlink.Core;
using Comlink.CefSharp;

namespace Comlink.Example.View
{
    {
        public MainWindow()
        {
            InitializeComponent();
        }

        private void Initialize()
        {
            Kaas kaas = new Kaas();
            Proxy<Kaas> proxied = Core.Comlink.Proxy(kaas);
            
            CefSharpBrowser.Expose(kaas);
            dynamic remote2 = Core.Comlink.Wrap<ClientSchema>(CefSharpBrowser.ToEndpoint());
            remote2.SomeMethod("a string value");
        }
    }

    public class Kaas
    {

    }

    public class ClientSchema
    {

    }
}
```

`csproj file`
```xml
  <PropertyGroup>
    <TargetFramework>net7.0-windows</TargetFramework>
    <PublishReadyToRun>true</PublishReadyToRun>
    <UseWPF>true</UseWPF>
    <NoWarn>$(NoWarn);1591</NoWarn>
    <StartupObject>Comlink.Example.View.App</StartupObject>
    <Nullable>enable</Nullable>

    <RuntimeIdentifier Condition="'$(RuntimeIdentifier)' == ''">$(NETCoreSdkRuntimeIdentifier)</RuntimeIdentifier>
    <SelfContained Condition="'$(SelfContained)' == ''">false</SelfContained>
    <Platforms>x64</Platforms>
  </PropertyGroup>

  <ItemGroup>
    <PackageReference Include="CefSharp.Wpf.NETCore" Version="106.0.260" />
  </ItemGroup>

  <ItemGroup>
    <ProjectReference Include="..\Comlink.Core\Comlink.Core.csproj" />
    <ProjectReference Include="..\Comlink.CefSharp\Comlink.CefSharp.csproj" />
  </ItemGroup>

  <ItemGroup>
    <ApplicationDefinition Include="src\View\App.xaml">
      <Generator>MSBuild:Compile</Generator>
      <XamlRuntime>Wpf</XamlRuntime>
      <SubType>Designer</SubType>
    </ApplicationDefinition>
  </ItemGroup>

  <ItemGroup>
    <Page Update="src\View\MainWindow.xaml">
      <Generator>MSBuild:Compile</Generator>
      <XamlRuntime>Wpf</XamlRuntime>
      <SubType>Designer</SubType>
    </Page>
  </ItemGroup>

</Project>
```

It seems to me that the generated g code in the obj folder is what is causing the naming collision between my referenced project and the package. I don't really understand why though because I did add the `assembly` attribute in the xaml file.

If you want to take a deeper look at the source files of this project you can find it in https://github.com/chris-kruining/comlink-sharp/tree/main/Comlink.Example

miloush:
Fixes #7173 (verified) #7125 (verified) and likely #4014 (asked for repro)

## Description

While XAML compiler treats types as fully qualified, the generated code behind does not use fully qualified types. As a result, language compilers can misinterpreted where a type should come from.

This PR fixes the issue by fully qualifying all types used in generated code that is placed in the user namespace, using global alias. Notably, code in `XamlGeneratedNamespace` is not affected (apart from attributes), since it is a root namespace on its own and no `using`s are added.

## Customer Impact

Not taking this fix makes it impossible to use WPF in namespace hierarchies where a subnamespace name is also a root namespace name.

## Regression

No.

## Testing

Built and compiled the repro projects provided. Ensured that applications using IComponentConnector, IStyleConnector and splash screen compile and run for both C# and VB.NET.

I am assuming the markup compiler is covered by tests (functional rather than textual). If that is not the case, let me know and we need to create some.

## Risk

Medium.

First, it is a breaking change in the sense that there was a hypothetical possibility that someone could have injected their own replacements of `System.*` types into the generated code, taking advantage of the current behavior. This would no longer work, but I believe that risk is minimal as the types would not be recognized as compatible by the type binding system. 

Generated code will be more verbose, consuming more space. The change affects the build process of everything and should any errors be introduced, developers will most likely be unable to workaround them.

On the plus side, the fix only enforces intended semantics of the XAML compiler, it is not meant to change any existing behavior.

### Review notes

For correctness, it is mostly about whether some instances where a global alias should have been used were forgotten. These would not be found by testing currently compiling applications.

For VB.NET, additional issues are caused by the default namespace that needs to be added to types before a global alias can be used. While I cannot think of a C# situation where this PR would break things that already worked, it is definitely a possibility for VB.NET due to the default namespace. It would be good to get as much of the VB.NET community as possible to check real-world projects compile as expected.

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/PresentationBuildTasks/MS/Internal/MarkupCompiler/MarkupCompiler.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationBuildTasks/MS/Internal/MarkupCompiler/MarkupCompiler.cs
@@ -901,7 +901,7 @@ namespace MS.Internal
                 _ccRoot.StyleConnectorFn = new CodeMemberMethod();
                 _ccRoot.StyleConnectorFn.Name = CONNECT;
                 _ccRoot.StyleConnectorFn.Attributes = MemberAttributes.Public | MemberAttributes.Final;
-                _ccRoot.StyleConnectorFn.PrivateImplementationType = new CodeTypeReference(KnownTypes.Types[(int)KnownElements.IStyleConnector]);
+                _ccRoot.StyleConnectorFn.PrivateImplementationType = new CodeTypeReference(KnownTypes.Types[(int)KnownElements.IStyleConnector], CodeTypeReferenceOptions.GlobalReference);
 
                 // void IStyleConnector.Connect(int connectionId, object target) {
                 //
@@ -985,7 +985,7 @@ namespace MS.Internal
 
                     // EventSetter eventSetter;
                     //
-                    CodeVariableDeclarationStatement cvdsES = new CodeVariableDeclarationStatement(KnownTypes.Types[(int)KnownElements.EventSetter], EVENTSETTER);
+                    CodeVariableDeclarationStatement cvdsES = new CodeVariableDeclarationStatement(new CodeTypeReference(KnownTypes.Types[(int)KnownElements.EventSetter], CodeTypeReferenceOptions.GlobalReference), EVENTSETTER);
                     _ccRoot.StyleConnectorFn.Statements.Insert(0, cvdsES);
                 }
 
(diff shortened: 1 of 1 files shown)


CyrusNajmabadi:
Any tests that show what is generated here?

PR #7229, opened 2022-10-26
Add NavigationService.RemoveForwardEntry

### LINKED ISSUE #7215: [API Proposal] NavigationService.RemoveForwardEntry
Currently when navigating backwards, content is added to forward stack. While we can remove entries from back stack, we cannot do the same for forward stack.

```C#
public class NavigationService
{
    // Removes the next journal entry from forward history, if there is one
    public JournalEntry RemoveForwardEntry()
}
public class Frame
{
    // Removes the next journal entry from forward history, if there is one
    public JournalEntry RemoveForwardEntry()
}
```

Symbai:
Fixes #7215

Main PR

## Description

There is already an API to remove the first entry from back stack. This PR allows developers to remove the first entry from forward stack. This is currently impossible.

## Customer Impact

Customers can now remove items from forward stack.

## Regression

No regression

## Risk

No risk

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Navigation/Journal.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Navigation/Journal.cs
@@ -271,6 +271,27 @@ namespace System.Windows.Navigation
             UpdateView();
             return removedEntry;
         }
+        
+        /// <summary>
+        /// Remove the top JournalEntry from forward entry
+        /// </summary>
+        // Not a true "remove" 
+        internal JournalEntry RemoveForwardEntry()
+        {
+            Debug.Assert(ValidateIndexes());
+            int index = _currentEntryIndex; // start from current but do not change it
+            do
+            {
+                if (++index >= TotalCount)
+                {
+                    return null;
+                }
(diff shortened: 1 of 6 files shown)


src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Navigation/Journal.cs
```diff
@@ -433,7 +454,7 @@ internal void NavigateTo(JournalEntry target)
             int index = _journalEntryList.IndexOf(target);
 
             // When navigating back to a page which contains a previously navigated frame a 
-            // saved journal entry is replayed to restore the frame�s location, in many cases 
+            // saved journal entry is replayed to restore the frames location, in many cases
```
lindexi, line 457:
```suggestion
            // saved journal entry is replayed to restore the frame's location, in many cases 
```

lindexi:
Look good to me

miloush:
Can you comment a bit on whether/how this was tested and why this is no risk?

Symbai:
> Can you comment a bit on whether/how this was tested and why this is no risk?

No risk because it doesn't affect any existing code, its a new API. Tested locally.

Issue #6346, opened 2022-04-03
OpenFileDialog.ShowReadOnly does not work


miloush:
* .NET Core Version: 7.0.0-preview.2.22152.2
* Windows version: 22581.1.amd64fre.ni_release.220318-1623
* Does the bug reproduce also in WPF for .NET Framework 4.8?: Yes
 
 **Problem description:**

The `OpenFileDialog` ignores the `ShowReadOnly` property on the Vista code path.
 
**Actual behavior:**

No check box in the dialog.
 
 **Expected behavior:**

A check box in the dialog as per documentation.

A dropdown option on the Open button would also be acceptable:
<img src="https://user-images.githubusercontent.com/10546952/161445476-4e0af62e-5353-43d3-ab72-eda6425bda84.png">

 **Minimal repro:**

```C#
OpenFileDialog ofd = new OpenFileDialog();
ofd.ShowReadOnly = true;
ofd.ShowDialog();
```

lindexi:
Sorry, what is the `Vista code path` ?

miloush:
There are two code paths for the file dialogs, one uses GetOpenFileName API while the other one uses IFileDialog API:

https://github.com/dotnet/wpf/blob/3ccd6f5545fbebf775262af446f2aff819a2226c/src/Microsoft.DotNet.Wpf/src/PresentationFramework/Microsoft/Win32/FileDialog.cs#L823-L827

The legacy one is used on Windows XP only. The Vista one is used everywhere else:

https://github.com/dotnet/wpf/blob/3ccd6f5545fbebf775262af446f2aff819a2226c/src/Microsoft.DotNet.Wpf/src/PresentationFramework/Microsoft/Win32/FileDialog.cs#L1830-L1833

.NET Core never supported Windows XP, unlike .NET Framework. However, you can still get the legacy behavior with .NET Core using compatibility settings:

<img width="247" alt="image" src="https://user-images.githubusercontent.com/10546952/161503596-922e0922-649d-4f19-b66a-d6d971fd94c7.png">

PR #7244, opened 2022-11-01
OpenFolderDialog

### LINKED ISSUE #438: WPF alternative for WinForms FolderBrowserDialog

miloush:
Fixes #438. Further discussion in #4039. Previous attempt: #6351

~This PR contains runtime breaking changes. For minimal implementation of #438 without breaking changes see #6374.~ EDIT: Introducing a base class is not breaking. 

## Description

Allows developers to ask the user to select a folder, a feature known as `FolderBrowserDialog` in WinForms. This PR 
1) removes all legacy (Windows XP) dialogs
2) splits the existing inheritance and moves common members up the hierarchy, so that members only available or working for files are not present for the folder browsing.

Public members before:
Before

Public members after:
After

Note that this abstraction does not exist in the native code, there is only a flag to pick folders on FileOpenDialog (akin to #6374).

What used to be `FileDialog.Options` is now strictly for the native API usage, so there is no longer need to mix and mask them.

## Example usage

```C#
OpenFolderDialog dialog = new OpenFolderDialog();
dialog.Multiselect = true;
dialog.ShowDialog();
string[] folders = dialog.FileNames;
```

## Compatibility requirements and limitations

The goal for this PR was to preserve compile-time compatibility, but not run-time compatibility. Old _PresentationFramework.dll_ assembly cannot be replaced by a new one, the application using it has to be recompiled (since members moved around types).

Anything that compiled before should still compile without changes. Most notably, while developers cannot derive from `FileDialog` due to internal abstract members, deriving from `CommonDialog` is possible. If this scenario was not required to be supported, I would further remove the following members as part of removing support for legacy dialogs:

* `protected CommonDialog.HookProc`
* `private CommonDialog.MoveToScreenCenter`

Not that inserting a class before `FileDialog` is the only place where class can be inserted if compatibility with .NET Framework API is to be preserved. If that was not a requirement, it would be possible to insert a class after `FileDialog`, which would be a bit more natural. Finally, if .NET Framework API surface was not a requirement AND we would be happy polluting the base class of possible inheritors of `CommonDialog`, we could treat `CommonDialog` newly as for `IFileDialog` API only and not insert any new classes in the hierarchy.

.NET Framework **baseline** contains changes to `FileDialog` that would be breaking if developers could inherit from the dialog, but as noted above, that is not allowed.

## Naming considerations

* **CommonItemDialog** is a name coming from [the documentation](https://learn.microsoft.com/en-us/windows/win32/shell/common-file-dialog). The native dialog works with any `IShellItem`s, not only file-based ones. I envision in the future if the code was extended to support arbitrary shell items, `CommonItemDialog` could provide the selection as shell items ~and the derived classes would utilize those for `FileNames` resp. `FolderNames` instead of doing their own conversion~. EDIT: `FileNames` were now lifted up to `CommonItemDialog`, shared for both files and folders.

* **OpenFolderDialog** is for parity with `OpenFileDialog`. Note that the save dialog no longer supports folder selection, so there is no need for an extra empty `FolderDialog` class in-between.

* **FileName(s)** makes the design much simpler (than having separate FolderNames) and reflects the native API. Note also that when dereferencing is disabled, results may contain shortcut files pointing to folders.

## Customer Impact

Without this fix, users have to either use WinForms' `FolderBrowserDialog`, resort to 3rd party libraries or re-implement the whole `IFileDialog` API. #438 is currently the top voted issue in the repository.

## Regression

No.

Known issues in the already existing code that are not addressed by this PR:
* read-only check box not available #6346
* `ValidateNames` is used to do several unrelated things - alleged invalid filename checking (that does not seem to be happening or necessary); processing file names, adding extensions, showing prompts etc.; and as a flag to native API to try create a dummy file to ensure user has rights to access the selected path.
* `AddExtension` works differently from the native API extension enforcements, so it is not realized as an API flag in the PR
* `CreatePrompt` and `OverwritePrompt` are also implemented differently than the native `FOS_CREATEPROMPT` resp. `FOS_OVERWRITEPROMPT`, notably create prompt only works for open dialogs, not save dialogs where it is defined in the WPF API, so they are not realized as API flags in the PR

## Testing

Compiled and manually tested on 7.0.0-rc.2.22472.3. Ensured overwrite and create prompts did not regress, cancelling the dialog did not regress restoring file names and filter index.

## Risk

Medium risk - No public API removed, a new base class introduced. Current applications using the types will keep working without recompilation. Code using reflection even on public types might break if it is not flattening them.

Applications with Windows XP compatibility flag when recompiled will still work, however, the dialogs will use the `IFileDialog` API rather than the legacy dialogs.

/cc @Symbai @ThomasGoulet73 @fabiant3 @batzen @lindexi

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/wpf/pull/7244)


Symbai:
Previously many WPF users had used Windows Api Codec Pack's folder dialog including myself. This dialog allowed users to optionally select multiple folders. Your PR only allows selecting a single folder. Selecting multiple can be quite helpful in any applications where you would have to add multiple folders. With your PR an end user would have to open the dialog over and over again just to select multiple sub folders.

So I suggest adding the following members to the folder dialog:
```
MultiSelect
FolderNames
SafeFolderNames
```

Also `CheckIfPathExists` is quite important. Without that a user could enter anything in the folder dialog bottom path field and click on 'OK'. Thus, developers would always have to check the result.

miloush:
Great point on multiselect, will fix that! It's the save dialog that used to support both multiselect and pick folders, but does neither anymore.

> Without that a user could enter anything in the folder dialog bottom path field and click on 'OK'. 

Have you tried that with Windows Api Codec Pack? Because the `FOS_PICKFOLDERS` implies `FOS_FILEMUSTEXIST` which implies `FOS_PATHMUSTEXIST`, my guess is since Vista SP1.

Symbai:
> Because the `FOS_PICKFOLDERS` implies `FOS_FILEMUSTEXIST` which implies `FOS_PATHMUSTEXIST`

You are correct, forget about my comment on `CheckIfPathExists`.

src/Microsoft.DotNet.Wpf/src/PresentationFramework/Microsoft/Win32/OpenFolderDialog.cs
```diff
@@ -0,0 +1,326 @@
...
+    using System.Text;
+    using System.Windows;
+
+    using MS.Internal.AppModel;
+    using MS.Internal.Interop;
+    using MS.Internal.PresentationFramework;
+    using MS.Win32;
+
+    /// <summary>
+    ///  Represents a common dialog box that allows the user to open one or more file(s).
```
lindexi, line 29:
```suggestion
    ///  Represents a common dialog box that allows the user to open one or more folder(s). 
```

src/Microsoft.DotNet.Wpf/src/PresentationFramework/Microsoft/Win32/OpenFolderDialog.cs
```diff
@@ -0,0 +1,326 @@
...
+        //---------------------------------------------------
+        //
+        // Private Properties
+        //
+        //---------------------------------------------------
+        #region Private Properties
+
+        /// <summary>
+        ///  Gets a string containing the full path of the file selected in 
+        ///  the file dialog box.
```
lindexi, line 285: Should replace the `file` to `folder`?

```suggestion
        ///  the folder dialog box.
```

src/Microsoft.DotNet.Wpf/src/PresentationFramework/Microsoft/Win32/OpenFolderDialog.cs
```diff
@@ -0,0 +1,326 @@
...
+            get
+            {
+                if (_fileNames == null)     // No filename stored internally...
+                {
+                    return String.Empty;    // So we return String.Empty
+                }
+                else
+                {
+                    // Return the first filename in the array if it is non-empty.
+                    if (_fileNames[0].Length > 0)
```
lindexi, line 298: Do you means `_fileNames.Length > 0`?

```suggestion
                    if (_fileNames.Length > 0)
```

Symbai: List is null when it does not contain entries as comments in line ~140 are telling. It's likely impossible for it to be an empty list at this point. Null check is done above. So, I guess the length check is for the filename and not for the array. But I dont really know in which case it can have an empty string as @miloush said that path & exist check is in place. To my understanding an empty string cannot be a valid path nor file/folder can exist. In which case the check might be redundant.

miloush: I haven't changed this code, it is the same in FileDialog. I agree this looks wrong, but also in order to trigger the bug you would have to convince the dialog to OK without any files/folders selected, which should not be possible with the current API surface. That said, allowing Multiselect for folders quite complicates things and I might need to review this code.

miloush:
As for multiselect, the complexity with lifting up file names to `CommonItemDialog` is in the file dialog needing to revert filter index if the OK event is cancelled. There is an architectural struggle between clean separation and knowing the API will not be extended anymore. I might also keep `FileName(s)` for both file and folder dialogs rather than separate `FolderName(s)`.

src/Microsoft.DotNet.Wpf/src/PresentationFramework/Microsoft/Win32/OpenFolderDialog.cs
miloush: rewritten in the latest commit, thanks @lindexi!

miloush:
I believe this PR now conforms to the API review.

miloush:
Reverted the FileNames unification to separate FolderNames.

Note that there is FileOk event. Are we happy with that one or should it also be disunified to FolderOk?

src/Microsoft.DotNet.Wpf/src/PresentationFramework/Microsoft/Win32/CommonItemDialog.cs
```diff
@@ -0,0 +1,693 @@
...
+        //
+        //---------------------------------------------------
+        #region Internal Properties
+
+        //   If multiple files are selected, we only return the first filename.
+        /// <summary>
+        ///  Gets a string containing the full path of the file selected in 
+        ///  the file dialog box.
+        /// </summary>
+        private protected string CriticalFileName
```
dipeshmsft, line 380: We should change these from File -> Item for both properties and members.
Like MutableFileNames, CriticalFileName, _fileNames, etc.

src/Microsoft.DotNet.Wpf/src/PresentationFramework/Microsoft/Win32/OpenFileDialog.cs
```diff
@@ -230,14 +190,11 @@ public bool ShowReadOnly
...
-                // OFN_HIDEREADONLY is the inverse of our property,
-                // so negate the results of GetOption...
-                return !GetOption(NativeMethods.OFN_HIDEREADONLY);
+                return _showReadOnly;
             }
             set
             {
-                // ... and SetOption.
-                SetOption(NativeMethods.OFN_HIDEREADONLY, !value);
+                _showReadOnly = false;
```
dipeshmsft, line 197: Can you add a comment here as well explaining the reason for doing this.

src/Microsoft.DotNet.Wpf/src/PresentationFramework/Microsoft/Win32/SaveFileDialog.cs
```diff
@@ -224,10 +200,9 @@ internal override bool PromptUserIfAppropriate(string fileName)
                 return false;
             }
          
-            bool fExist = File.Exists(Path.GetFullPath(fileName));
+            bool fExist = File.Exists(fileName);
```
dipeshmsft, line 203: What if the user writes a relative path in the dialog text box ? Will this still work ?

src/Microsoft.DotNet.Wpf/src/PresentationFramework/Microsoft/Win32/CommonItemDialog.cs
```diff
@@ -0,0 +1,693 @@
...
+        //
+        //---------------------------------------------------
+        #region Protected Methods
+
+        /// <summary>
+        /// Handles the IFileDialogEvents.OnFileOk callback.
+        /// </summary>
+        protected virtual void OnItemOk(CancelEventArgs e) { }
+
+        //  Because this class, FileDialog, is the parent class for both OpenFileDialog
```
dipeshmsft, line 205: Comment needs to be updated here.

src/Microsoft.DotNet.Wpf/src/PresentationFramework/Microsoft/Win32/FileDialog.cs
```diff
@@ -1099,41 +481,48 @@ internal virtual bool PromptUserIfAppropriate(string fileName)
...
-        /// </summary>
-        internal string[] FileNamesInternal
+        #region Internal and Protected Methods
+
+        private protected override void PrepareDialog(IFileDialog dialog)
         {
-            get
+            base.PrepareDialog(dialog);
+
+            dialog.SetFileName(CriticalFileName);
```
dipeshmsft, line 492: SetFileName is also called in CommonItemDialog. I don't think we need to call it here again.

src/Microsoft.DotNet.Wpf/src/PresentationFramework/Microsoft/Win32/FileDialog.cs
```diff
@@ -309,7 +163,7 @@ public string[] SafeFileNames
 
         //   If multiple files are selected, we only return the first filename.
         /// <summary>
-        ///  Gets or sets a string containing the full path of the file selected in 
+        ///  Gets or sets a string containing the full path of the file or folder selected in
```
dipeshmsft, line 166: Comment should be reverted here.

src/Microsoft.DotNet.Wpf/src/PresentationFramework/Microsoft/Win32/CommonItemDialog.cs
miloush: But these are strings, not `IShellItem`s

src/Microsoft.DotNet.Wpf/src/PresentationFramework/Microsoft/Win32/CommonDialog.cs
```diff
@@ -275,13 +276,14 @@ protected virtual void CheckPermissionsToShowDialog()
         //---------------------------------------------------
         #region Internal Methods
 
+        // This method is not used by IFileDialog API. Keeping for API compatibility with .NET Framework.
         /// <summary>
         ///  Centers the given window on the screen. This method is used by HookProc
         ///  to center the dialog on the screen before it is shown.  We can't mark it
         ///  private because we need to call it from our derived classes like
```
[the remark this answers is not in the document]
dipeshmsft: Sorry for the confusion. I was talking about the comment 
> _We can't mark it private because ...._

src/Microsoft.DotNet.Wpf/src/PresentationFramework/Microsoft/Win32/SaveFileDialog.cs
```diff
@@ -135,42 +132,21 @@ public override void Reset()
...
-            get
-            {
-                return GetOption(NativeMethods.OFN_OVERWRITEPROMPT);
-            }
-            set
-            {
-
-                SetOption(NativeMethods.OFN_OVERWRITEPROMPT, value);
-            }
-        }
```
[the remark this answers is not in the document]
miloush: Can you clarify the question? Win32 uses FOS_OVERWRITEPROMPT and FOS_CREATEPROMPT, but we don't use it, we have our own mechanism for these two prompts, see `PromptUserIfAppropriate`. I would love to get rid of it and use the system prompts, but amongst others it changes the message box text and I thought it is an unnecessary risk. If you agree this is fine, I am happy to remove the WPF prompts.

src/Microsoft.DotNet.Wpf/src/PresentationFramework/Microsoft/Win32/SaveFileDialog.cs
miloush: To answer your question yes, see docs for `File.Exists`:

> The path parameter is permitted to specify relative or absolute path information. Relative path information is interpreted as relative to the current working directory.

When you look at the code for `File.Exists`, you will see the first thing it does is it gets the full path: https://referencesource.microsoft.com/#mscorlib/system/io/file.cs,6fef4ff3e625580b

src/Microsoft.DotNet.Wpf/src/PresentationFramework/Microsoft/Win32/SaveFileDialog.cs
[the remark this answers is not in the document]
dipeshmsft: What I meant was, there can be other checks that can be employed by Win32 apart from what we are checking in `PromptUserIfAppropriate`.

src/Microsoft.DotNet.Wpf/src/PresentationFramework/Microsoft/Win32/CommonItemDialog.cs
```diff
@@ -0,0 +1,693 @@
...
+            // - Force no mini mode for the SaveFileDialog.
+            SetOption(FOS.DEFAULTNOMINIMODE, true);
+
+            //
+            // Initialize additional properties
+            // 
+            _fileNames = null;
+            _title.Value = null;
+            _initialDirectory.Value = null;
+
```
dipeshmsft, line 469: We should set `ClientGuid = null` here as well.

src/Microsoft.DotNet.Wpf/src/PresentationFramework/Microsoft/Win32/CommonDialog.cs
```diff
@@ -247,6 +247,7 @@ protected virtual IntPtr HookProc(IntPtr hwnd, int msg, IntPtr wParam, IntPtr lP
             return IntPtr.Zero;
         }
 
+        /// <summary>
```
dipeshmsft, line 250: An extra summary tag

src/Microsoft.DotNet.Wpf/src/PresentationFramework/Microsoft/Win32/CommonItemDialog.cs
dipeshmsft: IMHO, this will be consistent with the approach we are taking. Like 'CriticalItemName' makes more sense in CommonItemDialog and is generic enough for both files and folders.

src/Microsoft.DotNet.Wpf/src/PresentationFramework/Microsoft/Win32/SaveFileDialog.cs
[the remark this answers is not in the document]
dipeshmsft: After going through the docs, now I think that this change is valid and it will not change the behavior of the dialog.

src/Microsoft.DotNet.Wpf/src/PresentationFramework/Microsoft/Win32/CommonDialog.cs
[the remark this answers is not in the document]
miloush: Fixed in https://github.com/dotnet/wpf/pull/7916/commits/543c8098a03629f522d8d86a9666b673a0e0a045

src/Microsoft.DotNet.Wpf/src/PresentationFramework/Microsoft/Win32/CommonItemDialog.cs
```diff
@@ -0,0 +1,693 @@
...
+        //
+        //---------------------------------------------------
+        #region Public Properties
+
+        //   The actual flag is FOS_NODEREFERENCELINKS (set = do not dereference, unset = deref) - 
+        //   while we have true = dereference and false=do not dereference.  Because we expose
+        //   the opposite of the Windows flag as a property to be clearer, we need to negate 
+        //   the value in both the getter and the setter here.
+        /// <summary>
+        ///  Gets or sets a value indicating whether the dialog box returns the location
```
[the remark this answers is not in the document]
miloush: Technically it is correct since it applies to files only. I added a note saying _Not all dialogs allow users to select shortcuts._ Let me know if you were looking for a specific wording. Fixed in https://github.com/dotnet/wpf/pull/7916/commits/543c8098a03629f522d8d86a9666b673a0e0a045

src/Microsoft.DotNet.Wpf/src/PresentationFramework/Microsoft/Win32/CommonItemDialog.cs
miloush: Updated in https://github.com/dotnet/wpf/pull/7916/commits/ef7eb70b508066dffc48c04f7683b5d437804d43

src/Microsoft.DotNet.Wpf/src/PresentationFramework/Microsoft/Win32/CommonItemDialog.cs
miloush: Fixed in https://github.com/dotnet/wpf/pull/7916/commits/543c8098a03629f522d8d86a9666b673a0e0a045

src/Microsoft.DotNet.Wpf/src/PresentationFramework/Microsoft/Win32/FileDialog.cs
miloush: Fixed in https://github.com/dotnet/wpf/pull/7916/commits/543c8098a03629f522d8d86a9666b673a0e0a045

src/Microsoft.DotNet.Wpf/src/PresentationFramework/Microsoft/Win32/CommonItemDialog.cs
miloush: Good catch! Fixed in https://github.com/dotnet/wpf/pull/7916/commits/38ab72d13467b15166947fb783a5cb7ae4bd0079

src/Microsoft.DotNet.Wpf/src/PresentationFramework/Microsoft/Win32/CommonDialog.cs
miloush: Fixed in https://github.com/dotnet/wpf/pull/7916/commits/543c8098a03629f522d8d86a9666b673a0e0a045

src/Microsoft.DotNet.Wpf/src/PresentationFramework/Microsoft/Win32/OpenFileDialog.cs
miloush: Reverted in https://github.com/dotnet/wpf/pull/7916/commits/38ab72d13467b15166947fb783a5cb7ae4bd0079 (but still with no effect)

src/Microsoft.DotNet.Wpf/src/PresentationFramework/Microsoft/Win32/FileDialog.cs
miloush: Done in https://github.com/dotnet/wpf/pull/7916/commits/ffe135e61299ae37393d7741dbfae9253b0e11ea

miloush:
Since the review related to the new properties as well, all fixes have been pushed to https://github.com/dotnet/wpf/pull/7916.

miloush:
@dipeshmsft backported

src/Microsoft.DotNet.Wpf/src/PresentationFramework/Microsoft/Win32/CommonItemDialog.cs
```diff
@@ -0,0 +1,692 @@
...
+        }
+
+        /// <summary>
+        ///  Restores the current directory to its original value if the user
+        ///  changed the directory while searching for files.
+        ///
+        ///  This property is only valid for SaveFileDialog;  it has no effect
+        ///  when set on an OpenFileDialog.
+        /// </summary>
+        public bool RestoreDirectory
```
dipeshmsft, line 136: As per the new approved API we have to move this back to `FileDialog`. Can you move this back as well ?

PR #7295, opened 2022-11-17
Adding static constructor to StyleHelper class

### LINKED ISSUE #7205: NullReferenceException is thrown in the System.Windows.StyleHelper.GetExpression method
* .NET Core Version: **.NET 7.0 SDK (v7.0.100-rc.2)**
* Windows version: **Version 21h1 (OS Build 19043.2130)**
* Does the bug reproduce also in WPF for .NET Framework 4.8?: **No**
* Is this bug related specifically to tooling in Visual Studio (e.g. XAML Designer, Code editing, etc...)?: **No**
 
 **Problem description:** An attempt to call the BindingOperations.GetBindingExpression method in .NET 7.0 leads to Null Reference Exception.
 
 **Actual behavior:** 

NullRerferenceException is thrown right at the moment when the following code is executed: 

```cs
public App() {
    var viewSource = new CollectionViewSource();
    var be = BindingOperations.GetBindingExpression(viewSource, CollectionViewSource.SourceProperty);
}
```
The callstack indicates that the error is thrown in the System.Windows.StyleHelper.GetExpression method: 

**Callstack:**

```callstack
PresentationFramework.dll!System.Windows.StyleHelper.GetExpression(System.Windows.DependencyObject d, System.Windows.DependencyProperty dp)
PresentationFramework.dll!System.Windows.Data.BindingOperations.GetBindingExpressionBase(System.Windows.DependencyObject target, System.Windows.DependencyProperty dp)
PresentationFramework.dll!System.Windows.Data.BindingOperations.GetBindingExpression(System.Windows.DependencyObject target, System.Windows.DependencyProperty dp)
WpfApp1.dll!WpfApp1.App.App() Line 18
WpfApp1.dll!WpfApp1.App.Main()
```

 
 **Expected behavior:**

No errors should be thrown.
 
 **Minimal repro:**

Download and run the project. The error should be thrown right at the startup.

The project and screencast in the attachment illustrate the problematic behavior. 

[dxSample.zip](https://github.com/dotnet/wpf/files/9785782/dxSample.zip)
[screencast.mp4](https://user-images.githubusercontent.com/31204429/195839191-507189c1-bfc6-4c99-bd38-2fa719ca9bfe.mp4)

### LINKED ISSUE #7291: In the .net7 CollectionViewSource has a different behaviour than in the .net6
Basically, in .net7 when you run the attached project and press the button, it will display a list of items. If you run the unit test it will throw an exception on this line 

` EntityCollectionView = new CollectionViewSource { Source = Entities }.View;`

Because the View is null.

If you change the TargetFramework to net6 the test passes because View contains an instance. 

Is this a bug?

Regards
Daniel Kaczmarek

[WpfApp2.zip](https://github.com/dotnet/core/files/10012439/WpfApp2.zip)

Kuldeep-MS:
Fixes #7205
Fixes #7291

## Description

Adding static constructor to StyleHelper class to do the required initialization i.e. registering the "alternative Expression storage". Also, adding check to "RegisterAlternateExpressionStorage" function to avoid re-registration being called.

## Customer Impact

An attempt to call the BindingOperations.GetBindingExpression method in .NET 7.0 leads to Null Reference Exception.

## Regression

Yes

## Testing

In progress

## Risk

NA

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/wpf/pull/7295)

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/StyleHelper.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/StyleHelper.cs
@@ -41,6 +41,12 @@ namespace System.Windows
 
     internal static class StyleHelper
     {
+        static StyleHelper()
+        {
+            // Register for the "alternative Expression storage" feature, since
+            // we store Expressions in per-instance StyleData.
+            RegisterAlternateExpressionStorage();
+        }
         //  ===========================================================================
         //  These methods are invoked when a Style/Template cache needs to be updated
         //  ===========================================================================
@@ -5466,9 +5472,13 @@ namespace System.Windows
         //
         internal static void RegisterAlternateExpressionStorage()
         {
-            DependencyObject.RegisterForAlternativeExpressionStorage(
+            if(_getExpression == null)
+            {
+                DependencyObject.RegisterForAlternativeExpressionStorage(
                                 new AlternativeExpressionStorageCallback(GetExpressionCore),
                                 out _getExpression);
+            }
+            
         }
 
         private static Expression GetExpressionCore(


batzen (approved):

src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/StyleHelper.cs
```diff
@@ -5466,9 +5472,13 @@ internal static DependencyObject GetChild(List<DependencyObject> styledChildren,
         //
         internal static void RegisterAlternateExpressionStorage()
         {
-            DependencyObject.RegisterForAlternativeExpressionStorage(
+            if(_getExpression == null)
```
lindexi, line 5475: After we register in static constructor, the `_getExpression` never be null when the other code call this method. And I think the `RegisterAlternateExpressionStorage` can be private and only call from static constructor.

pchaurasia14: @lindexi - Actually this method gets called from `Style` class's static ctor as well. That's when we'll have a case where `_getExpression` has already setup via `StyleHelper`'s static ctor. Hence, this check prevents dual registration of the delegate.

miloush (approved):
Checked both repros are fixed.

dipeshmsft (approved):

PR #7297, opened 2022-11-17
Surround Keyboard Datagrid F3 sort with switch

### LINKED ISSUE #7288: Wpf DataGrid in .NET7 takes away F3 and automatically sorts. - Breaking change.
Our application uses Wpf DataGrid and was working fine till .NET 6.0
When upgraded to .NET 7, press F3 the application expectation is to take to Main window's Search box, but the grid takes over and does the sort of a column where F3 was pressed.

This breaks the functionality.

Is there a way to disable this?

singhashish-wpf:
Fixes #7288 

Main PR https://github.com/dotnet/wpf/pull/6873

## Description
F3 sort on Datagrid was causing compatibility issues. Surrounding the F3 datagrid sort feature around an opt Out switch.

## Customer Impact
Customers can run into problem with F3 handling in datagrid scenarios. Unintended sorting rather than a specifically expected behaviour.

## Regression
Yes

## Testing
Local testing by sample datagrid app.

## Risk
Low

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/wpf/pull/7297)

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Controls/DataGridCell.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Controls/DataGridCell.cs
@@ -23,7 +23,7 @@ namespace System.Windows.Controls
     /// </summary>
     public class DataGridCell : ContentControl, IProvideDataGridColumn
     {
-        private static bool IsDataGridKeyboardSortFeatureEnabled = true;
+        private static bool WPFIsDataGridKeyboardSortFeatureDisabled;
 
         #region Constructors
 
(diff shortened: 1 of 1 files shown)


zlatanov:
> opt opt switch

Is this opt in or opt out feature? Looking at the changes it seems to be opt in?

singhashish-wpf:
> > opt opt switch
> 
> Is this opt in or opt out feature? Looking at the changes it seems to be opt in?

Opt in.

singhashish-wpf:
> > opt opt switch
> 
> Is this opt in or opt out feature? Looking at the changes it seems to be opt in?

Opt Out. Default is set to true. If you need to opt out then you need to disable in the config.

singhashish-wpf:
> So AppContext.TryGetSwitch sets it's out param to true by default? The documentation says it's false if the switch is not found.
> 
> Aren't switches like this supposed to have a prefix hinting at WPF?

It was wrongly assumed that the variable won't be updated in case the switch doesn't exist, however as you said it is set to false. So modified the code to update the switch name and the condition.

zlatanov:
DataGridKeyboardSortFeature isn't really descriptive of what this switch is being used for in my opinion. It doesn't indicate that we expect a bool, or if it does if true means enable or disable.

zlatanov:
Also, shouldn't you make `IsDataGridKeyboardSortDisabled` readonly?

zlatanov (approved):

Kuldeep-MS:
The changes of this PR will be taken care by https://github.com/dotnet/wpf/pull/7349

PR #7308, opened 2022-11-23
Revert "Replacing expensive type checks with "is" checks (#4964)"

Kuldeep-MS:
Fixes https://github.com/dotnet/wpf/issues/7205
Fixes https://github.com/dotnet/wpf/issues/7291

[Main PR ] (https://github.com/dotnet/wpf/pull/7295)

## Description

Revert "Replacing expensive type checks with "is" checks (#4964)"

## Customer Impact

An attempt to call the BindingOperations.GetBindingExpression method in .NET 7.0 leads to Null Reference Exception.

## Regression

Yes

## Testing

In progress

## Risk

NA

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/wpf/pull/7308)

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/PresentationCore/MS/internal/SynchronizedInputHelper.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationCore/MS/internal/SynchronizedInputHelper.cs
@@ -297,18 +297,21 @@ namespace MS.Internal
 
         internal static void RaiseAutomationEvents()
         {
-            if (InputManager.ListeningElement is UIElement e)
+            if (InputElement.IsUIElement(InputManager.ListeningElement))
             {
+                UIElement e = (UIElement)InputManager.ListeningElement;
                 //Raise InputDiscarded automation event
                 SynchronizedInputHelper.RaiseAutomationEvent(e.GetAutomationPeer());
             }
-            else if (InputManager.ListeningElement is ContentElement ce)
+            else if (InputElement.IsContentElement(InputManager.ListeningElement))
             {
+                ContentElement ce = (ContentElement)InputManager.ListeningElement;
                 //Raise InputDiscarded automation event
                 SynchronizedInputHelper.RaiseAutomationEvent(ce.GetAutomationPeer());
             }
-            else if (InputManager.ListeningElement is UIElement3D e3D)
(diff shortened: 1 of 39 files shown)


IAmTheCShark:
@batzen

batzen:
@Kuldeep-MS may i ask why the changes have to be reverted? There is already a PR open that fixes the initialization issue.

singhashish-wpf:
> @Kuldeep-MS may i ask why the changes have to be reverted? There is already a PR open that fixes the initialization issue.

Hello @batzen 
We'll be taking the fix in the main branch. Reverting from the .NET7 branch just to give some extra time for the fix to be tested and the servicing time/difficulties which might arise due to any further regression.

batzen:
I don't think that's the best way to solve the situation.

The fix is small and simple. The issues reported are fixed by the other PR.

If reverting changes that cause issues is the preferred approach, at least the changes made for accepting F3 to sort DataGrid columns should also be reverted.

I don't think that's a sustainable approach.

singhashish-wpf:
Agreed that the approach of reverting the fixes which cause regression isn't sustainable.
The fix for this does mitigate the known regression but the confidence in the fix will come after thorough testing which will require time. (We have less time due to .NET7 servicing deadlines). Timing and Servicing Bug Bar were primary reasons for creating this PR for .NET7 branch as a placeholder just in case the testing reports further failures.

Please note that this PR hasn't been assigned any labels which are required for it to be considered for servicing. This is dependent on time/test reports that we'll take the revert PR or the actual fix in the .NET 7 branch. 

Since this was a significant improvement to the performance, we are inclined to take the fix rather than reverting. This PR is just to buy more time should the testing report any further issues.

IAmTheCShark:
@singhashish-wpf i guess this is somewhat new territory, so it might help clarify in a more general matter how you deal with issues like this when they arrise. Initially "the community" only saw "you" reverting a contribution without any further ellaboration.

singhashish-wpf:
Sure @IAmTheCShark ! 
We will provide more elaboration on such PRs going forwards.

singhashish-wpf:
Closing this as the main fix is in the servicing and main branch.

PR #7350, opened 2022-12-08
fix(column-resize): adding opt-out switch for column resize and column sorting features by keyboard inputs

### LINKED ISSUE #7322: InputBindings (KeyBinding) behaviour change in net7.0-windows
* .NET Core Version: 7.0
* Windows version: 10
 **Problem description:**
 We use Alt+Right/Left keys alot to navigate between pages in our application, the commands we use are run by InputBindings (KeyBinding).  After updating to .net7 it does not work anymore.
 **Actual behavior:** 
Alt + Right does not call the command
 **Expected behavior:**
  Alt + Right should call the command.
 **Minimal repro:**
In the demo attached in Browse.xaml code behind there are two commands with breakpoint, you will notice that when you run the app and press Alt+Right or Left the breakpoints are never hit while the focus is inside the ListBox. If you stop the application and change the Target Framework from net7.0-windows to net6.0-windows those breakpoints are always called.

[WpfApp3.zip](https://github.com/dotnet/wpf/files/10105445/WpfApp3.zip)

Kuldeep-MS:
Fixes #7322 

[Main PR ](https://github.com/dotnet/wpf/pull/7349)

## Description

adding opt-out switch for column resize and column sorting features by keyboard inputs

## Customer Impact

Without this changes the features will be enabled by default and any user defined input bindings for Alt + Left/Right arrow keys or F3 function key will not work. Also, user will not have any option to turn of the feature.

## Regression

Yes

## Testing

In-progress

## Risk

NA

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/wpf/pull/7350)

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Controls/DataGridCell.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Controls/DataGridCell.cs
@@ -23,6 +23,9 @@ namespace System.Windows.Controls
     /// </summary>
     public class DataGridCell : ContentControl, IProvideDataGridColumn
     {
+        private static readonly bool IsDataGridKeyboardSortDisabled;
+        private static readonly bool OptOutOfGridColumnResizeUsingKeyboard;
+
         #region Constructors
 
         /// <summary>
@@ -45,6 +48,9 @@ namespace System.Windows.Controls
 
             EventManager.RegisterClassHandler(typeof(DataGridCell), LostFocusEvent, new RoutedEventHandler(OnAnyLostFocus), true);
             EventManager.RegisterClassHandler(typeof(DataGridCell), GotFocusEvent, new RoutedEventHandler(OnAnyGotFocus), true);
+
+            AppContext.TryGetSwitch("System.Windows.Controls.DisableDataGridKeyboardSort", out IsDataGridKeyboardSortDisabled);
+            AppContext.TryGetSwitch("System.Windows.Controls.OptOutOfGridColumnResizeUsingKeyboard", out OptOutOfGridColumnResizeUsingKeyboard);
         }
 
         /// <summary>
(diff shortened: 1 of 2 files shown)


singhashish-wpf:
@zlatanov Tagging for review, as the other PR changes are being merged with this one.

src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Controls/DataGridCell.cs
```diff
@@ -966,43 +972,44 @@ protected override void OnKeyDown(KeyEventArgs e)
...
+                    ModifierKeys modifierKeys = Keyboard.Modifiers & ModifierMask;
+                    if (((e.SystemKey == Key.Right) || (e.SystemKey == Key.Left)) && (modifierKeys == ModifierKeys.Alt))
                     {
-                        updatedWidth = new DataGridLength(Column.ActualWidth + ColumnWidthStepSize);
-                    }
-                    else if (e.SystemKey == Key.Left)
-                    {
-                        updatedWidth = new DataGridLength(Column.ActualWidth - ColumnWidthStepSize);
-                    }
+                        DataGridLength updatedWidth = new DataGridLength();
```
zlatanov, line 980:
```suggestion
                        DataGridLength updatedWidth;
```

src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Controls/DataGridCell.cs
```diff
@@ -966,43 +972,44 @@ protected override void OnKeyDown(KeyEventArgs e)
...
-                    if (Column != null)
-                    {
-                        if (Column.CanColumnResize(updatedWidth))
+                        if (e.SystemKey == Key.Right)
                         {
-                            Column.SetCurrentValueInternal(DataGridColumn.WidthProperty, updatedWidth);
+                            updatedWidth = new DataGridLength(Column.ActualWidth + ColumnWidthStepSize);
                         }
-                        e.Handled = true;
+                        else if (e.SystemKey == Key.Left)
```
zlatanov, line 986:
```suggestion
                        else
```

No need for the `else if` because the check has already been done in the outer if.

src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Controls/DataGridCell.cs
```diff
@@ -966,43 +972,44 @@ protected override void OnKeyDown(KeyEventArgs e)
...
+                        {
+                            updatedWidth = new DataGridLength(Column.ActualWidth - ColumnWidthStepSize);
+                        }
+
+                        if (Column != null)
```
zlatanov, line 991: I don't understand why Column can be null here. If you see line 1005:
```c#
if ((e.Key == Key.F3) && Column.CanUserSort)
```
where it is not checked. Either the column can never be null here, or line 1005 we need to add a null check as well?

src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Controls/DataGridCell.cs
```diff
@@ -45,6 +48,9 @@ static DataGridCell()
 
             EventManager.RegisterClassHandler(typeof(DataGridCell), LostFocusEvent, new RoutedEventHandler(OnAnyLostFocus), true);
             EventManager.RegisterClassHandler(typeof(DataGridCell), GotFocusEvent, new RoutedEventHandler(OnAnyGotFocus), true);
+
```
zlatanov, line 51: nit: two empty lines, if not intentional, you should probably leave only one.

zlatanov (changes requested):

src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Controls/DataGridCell.cs
Kuldeep-MS: Added null check at line 1005

src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Controls/DataGridCell.cs
Kuldeep-MS: one empty line is added intentionally

src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Controls/DataGridCell.cs
Kuldeep-MS: Made the changes as suggested

src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Controls/DataGridCell.cs
Kuldeep-MS: Made the changes as suggested

zlatanov (approved):

singhashish-wpf (approved):

pchaurasia14:
Servicing approved via email

PR #7404, opened 2022-12-30
Fix widespread null-reference exception in WPF MediaContext.RenderMessageCore

glenn-slayden:
As stated in the existing code comment, "can be null if we are not resizing..."  Yes it can, and frequently is. This is the cause of numerous WPF crashes reported across the web:

- https://www.infragistics.com/community/forums/f/ultimate-ui-for-wpf/107396/application-crash-when-user-try-to-enter-text-in-xammulticolumncomboeditor
- https://stackoverflow.com/questions/70799521/random-crash-in-wpf-application-invalid-index-at-ms-win32-unsafenativemethods
- https://stackoverflow.com/questions/2008226/wpf-application-crash
- https://github.com/goatcorp/FFXIVQuickLauncher/issues/635
- https://social.msdn.microsoft.com/Forums/vstudio/en-US/25e07b53-ce57-44a0-983f-23e8d4219674/wpf-application-crashes?forum=wpf
- https://stackoverflow.com/questions/8378051/unexplained-error-in-system-windows-media-composition-duce-channel

...and hundreds of others.

Fixes # 

Main PR 

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/wpf/pull/7404)


glenn-slayden:
This fix will need further investigation.

PR #7406, opened 2023-01-01
Implement IList on PageContentCollection and DocumentReferenceCollection

### LINKED ISSUE #966: Feature ASK: Implement IList on PageContentCollection and DocumentReferenceCollection
Copied from VSO: 776611

This comes from user feedback:
https://developercommunity.visualstudio.com/content/problem/418015/fixeddocument-pagecontent-bug.html
image

REPRO:
In a WPF project in Visual Studio, construct a file like in the following screenshot:

RESULT:
The content of FixedDocument is always shown as an error.

EXPECT:
No error squiggles or error messages

aubiyko:
Fixes #966

## Description

Microsoft Visual Studio (and Blend) XAML designer seems to follow [Collections and collection types for XAML](https://learn.microsoft.com/en-us/dotnet/desktop/xaml-services/collections-and-types), which means that collections are recognized as `IList` or `IDictionary`.  `PageContentCollection` and `DocumentReferenceCollection` do not implement either of them. This leads to `Invalid Markup` error in designer and weird errors in code editor/Error List. These types have `Add` method and special handling to make them work. The code will compile and run just fine.

So instead of fixing visual designer this MR adds `ILIst` to them.

## Customer Impact

No visual preview in designer and error(s) that are not actual errors in code editor.

## Regression

No. As I can tell, this bug originates in the initial release of WPF for Microsoft .NET Framework.

## Testing

Manual.

## Risk

NA.

### DIFF
--- a/src/Microsoft.DotNet.Wpf/cycle-breakers/PresentationFramework/PresentationFramework.cs
+++ b/src/Microsoft.DotNet.Wpf/cycle-breakers/PresentationFramework/PresentationFramework.cs
@@ -10224,12 +10224,25 @@ namespace System.Windows.Documents
         void System.Windows.Markup.IAddChild.AddChild(object value) { }
         void System.Windows.Markup.IAddChild.AddText(string text) { }
     }
-    public sealed partial class PageContentCollection : System.Collections.Generic.IEnumerable<System.Windows.Documents.PageContent>, System.Collections.IEnumerable
+    public sealed partial class PageContentCollection : System.Collections.IList, System.Collections.Generic.IEnumerable<System.Windows.Documents.PageContent>, System.Collections.IEnumerable
     {
         internal PageContentCollection() { }
         public int Count { get { throw null; } }
         public System.Windows.Documents.PageContent this[int pageIndex] { get { throw null; } }
+        bool System.Collections.IList.IsFixedSize { get { throw null; } }
+        bool System.Collections.IList.IsReadOnly { get { throw null; } }
+        object System.Collections.IList.this[int index] { get { throw null; } set { } }
+        bool System.Collections.ICollection.IsSynchronized { get { throw null; } }
+        object System.Collections.ICollection.SyncRoot { get { throw null; } }
         public int Add(System.Windows.Documents.PageContent newPageContent) { throw null; }
+        int System.Collections.IList.Add(object value) { throw null; }
+        void System.Collections.IList.Clear() { }
+        bool System.Collections.IList.Contains(object value) { throw null; }
+        int System.Collections.IList.IndexOf(object value) { throw null; }
+        void System.Collections.IList.Insert(int index, object value) { }
+        void System.Collections.IList.Remove(object value) { }
+        void System.Collections.IList.RemoveAt(int index) { }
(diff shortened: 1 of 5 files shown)


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/dotnet/wpf?pullRequest=7406) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/dotnet/wpf?pullRequest=7406) before we can accept your contribution.<br/><br/>:x: aubiyko [sign now](https://cla.dotnetfoundation.org/dotnet/wpf?pullRequest=7406)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/dotnet/wpf?pullRequest=7406) it.</sub>

spadapet:
Glad to see a PR with a fix for this issue after a few years. There is more user feedback about the issue now:
* https://developercommunity.visualstudio.com/t/Adding-PageContent-to-FixedDocument-caus/10251925

miloush:
Well the collections are not really ILists, they are more like IReadOnlyLists with an Add method (and the Add method should be enough for this to work). This should be fixed either on the processing side, or by some agreement, such as ContentWrapperAttribute.

aubiyko:
> Well the collections are not really ILists, they are more like IReadOnlyLists with an Add method (and the Add method should be enough for this to work). This should be fixed either on the processing side, or by some contract, such as ContentWrapperAttribute.

While I do agree with you @miloush, there are a number of issues. Firstly, [Collections and collection types for XAML](https://learn.microsoft.com/dotnet/desktop/xaml-services/collections-and-types) clearly states that .NET XAML recognizes only `IList`, `IDictionary` or `Array` as collections. That means `IReadOnlyList` and such are a no go. I tried to emulate such behavior with `IList` to make visual preview work.
Secondly, traces of this particular bug lead to .NET Framework and Visual Studio 2008. (Also note new WinUI 3 has [no designer at all](https://github.com/microsoft/microsoft-ui-xaml/issues/5917)) I doubt that designer will be fixed any time soon.

miloush:
> clearly states

That is not my reading of the documentation, it says "based on". See also [LookupCollectionKind](https://learn.microsoft.com/en-us/dotnet/api/system.xaml.xamltype.lookupcollectionkind):
> the default implementation attempts to find the Add method of the underlying type by reflection.

This corresponds to the code which just tries to get any Add method on the underlaying type that would work. Moreover, these two specific collections are explicitly recognized as collections:

https://github.com/dotnet/wpf/blob/11ff3862bdfb0a4a85961aa43faae2496efb5d78/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Markup/Baml2006/WpfXamlType.cs#L498-L503

My point is the XAML services already correctly recognize these types as collections. And it works at runtime, so the designer doesn't seem to be doing very good job at showing what happens at runtime. Either way I don't think implementing IList here only to workaround a bug elsewhere is a reasonable solution. We should push for this to be fixed where it is broken.

PR #4177, opened 2021-02-12
Replaced some Dictionaries and Hashtables with Hashsets

ThomasGoulet73:
I replaced some Dictionaries and Hashtables that only use the key as a way to filter out duplicates. I replaced them with HashSet to improve performance.

I used the following code to benchmark it :
https://gist.github.com/ThomasGoulet73/99559a6cd49a04dc70a413e9668eee22

Here are the results (First row is 10 items, second is 1000 and third is 100000) :

Dictionary vs HashSet:
|     Method |           enumerable |           Mean |         Error |        StdDev |    Ratio | RatioSD |    Gen 0 |    Gen 1 |    Gen 2 |   Allocated |
|----------- |--------------------- |---------------:|--------------:|--------------:|---------:|--------:|---------:|---------:|---------:|------------:|
| Dictionary | Syste(...)ring] [48] |       742.2 ns |      14.71 ns |      13.76 ns |     1.00 |    0.00 |   0.2270 |   0.0010 |        - |     1.39 KB |
| Dictionary | Syste(...)ring] [48] |    60,340.7 ns |     231.42 ns |     216.47 ns |    81.32 |    1.42 |  18.8599 |   4.6997 |        - |   116.13 KB |
| Dictionary | Syste(...)ring] [48] | 6,990,405.3 ns | 114,642.24 ns | 101,627.36 ns | 9,405.37 |  234.34 | 796.8750 | 750.0000 | 742.1875 | 10303.12 KB |
|    HashSet | Syste(...)ring] [48] |       558.8 ns |       1.55 ns |       1.45 ns |     0.75 |    0.01 |   0.1907 |        - |        - |     1.17 KB |
|    HashSet | Syste(...)ring] [48] |    47,343.1 ns |     173.14 ns |     161.96 ns |    63.81 |    1.22 |  14.2822 |   2.8076 |        - |    87.76 KB |
|    HashSet | Syste(...)ring] [48] | 5,982,708.5 ns |  80,465.49 ns |  75,267.47 ns | 8,062.82 |  161.97 | 703.1250 | 648.4375 | 648.4375 |  7945.69 KB |

Hashtable vs HashSet:
|        Method |           enumerable |            Mean |         Error |        StdDev |     Ratio | RatioSD |    Gen 0 |    Gen 1 |    Gen 2 | Allocated |
|-------------- |--------------------- |----------------:|--------------:|--------------:|----------:|--------:|---------:|---------:|---------:|----------:|
[...]

PR #7486, opened 2023-02-01
Replace Dictionary with HashSet

ThomasGoulet73:
Follow up to dotnet/wpf#4177

## Description
I found more places where we used a Dictionary without using the value just to filter out duplicates.

See dotnet/wpf#4177 for benchmarks.

## Customer Impact
Better performance.

## Regression
No.

## Testing
Local build + CI.

## Risk
Low.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/wpf/pull/7486)

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/PresentationCore/MS/internal/FontCache/FontSourceCollection.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationCore/MS/internal/FontCache/FontSourceCollection.cs
@@ -159,10 +159,8 @@ namespace MS.Internal.FontCache
                                     }
                                 }
 
-                                foreach (string file in Directory.GetFiles(_uri.LocalPath))
-                                {
-                                    fontPaths.Add(file);
-                                }
+                                fontPaths.UnionWith(Directory.EnumerateFiles(_uri.LocalPath));
+
                                 files = fontPaths;
                             }
 }
--- a/src/Microsoft.DotNet.Wpf/src/PresentationCore/MS/internal/FontCache/FontSourceCollection.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationCore/MS/internal/FontCache/FontSourceCollection.cs
@@ -137,7 +137,7 @@ namespace MS.Internal.FontCache
                             {
                                 // fontPaths accumulates font file paths obtained from the registry and the file system
                                 // This collection is a set, i.e. only keys matter, not values.
-                                Dictionary<string, object> fontPaths = new Dictionary<string, object>(512, StringComparer.OrdinalIgnoreCase);
+                                HashSet<string> fontPaths = new HashSet<string>(512, StringComparer.OrdinalIgnoreCase);
 
                                 using (RegistryKey fontsKey = Registry.LocalMachine.OpenSubKey(InstalledWindowsFontsRegistryKey))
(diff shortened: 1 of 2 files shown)


miloush (approved):
LGTM (still a good sanity check to make sure the affected codepath runs as expected)

src/Microsoft.DotNet.Wpf/src/PresentationCore/MS/internal/FontCache/FontSourceCollection.cs
```diff
@@ -154,16 +154,16 @@ private void SetFontSources()
...
+                                            fontPaths.Add(fileName);
                                         }
                                     }
                                 }
 
                                 foreach (string file in Directory.GetFiles(_uri.LocalPath))
                                 {
-                                    fontPaths[file] = null;
+                                    fontPaths.Add(file);
                                 }
```
halgab, line 165: Why not do directly `fontPaths.UnionWith(Directory.GetFiles(_uri.LocalPath))`?

dipeshmsft: @ThomasGoulet73 any comments ?

dipeshmsft:
LGTM, labeling it for test pass.

src/Microsoft.DotNet.Wpf/src/PresentationCore/MS/internal/FontCache/FontSourceCollection.cs
IAmTheCShark: @halgab and maybe even better use `EnumerateFiles` ?

ThomasGoulet73: I forgot about this. I did a quick benchmark with both of your suggestions and it looks like using Directory.EnumerateFiles shows a nice improvement and that UnionWith is the same as a foreach when using Directory.EnumerateFiles:
|                  Method |     Mean |   Error |  StdDev | Ratio | RatioSD |   Gen0 | Allocated | Alloc Ratio |
|------------------------ |---------:|--------:|--------:|------:|--------:|-------:|----------:|------------:|
|         ForEachGetFiles | 174.7 μs | 3.42 μs | 6.75 μs |  1.00 |    0.00 | 3.1738 |  10.09 KB |        1.00 |
|   ForEachEnumerateFiles | 161.9 μs | 2.97 μs | 2.77 μs |  0.90 |    0.05 | 2.6855 |   8.65 KB |        0.86 |
|       UnionWithGetFiles | 166.6 μs | 3.29 μs | 5.02 μs |  0.95 |    0.05 | 3.1738 |  10.13 KB |        1.00 |
| UnionWithEnumerateFiles | 161.3 μs | 2.96 μs | 3.29 μs |  0.90 |    0.05 | 2.6855 |   8.65 KB |        0.86 |

I decided to go with using UnionWith and Directory.EnumerateFiles since it's cleaner and faster (Directory.GetFiles just wraps the result of Directory.EnumerateFiles in an array and since we just add the result to an hashset, the array is just wasted memory and makes it iterate the result twice).

PR #7503, opened 2023-02-06
Remove the old bug report template

lindexi:
## Description

We will find two bug report template which will confuse developers. And I remove the old template.

Reference https://github.com/dotnet/wpf/pull/7496

## Customer Impact

None.

## Regression

None.

## Testing

None.

## Risk

None.


lindexi:
Duplicate of #7497

Issue #5576, opened 2021-10-20
`IncludePackageReferencesDuringMarkupCompilation` causes build errors with WPF projects


las-nsc:
The reason for enabling this setting is to include code from a source generator ([BoilerPlateZero](https://github.com/IGood/boilerplatezero)), but that wasn't needed to cause the build to break.

**Version Used**: 
.NET 5.0-windows with C#9 with VS2019 16.11.0 Enterprise

**Steps to Reproduce**:

1. Take a relatively complex solution using WPF which builds with no errors
2. Add in `Directory.Build.props` or `*.csproj`
a. `<AssemblyName>MyProductName.$(MSBuildProjectName)</AssemblyName>`
b. `<IncludePackageReferencesDuringMarkupCompilation>true</IncludePackageReferencesDuringMarkupCompilation>`
3. Build project

**Expected Behavior**:
Should build with no errors.

**Actual Behavior**:
Fails to builds with errors such as:
|Severity|Code|Description|
|---|---|---|
|Error	|CS0006	|Metadata file '`[ProjectDir]`\bin\Debug\ref\\`[AssemblyName]`.dll' could not be found.|
|Error	|CS0006	|Metadata file '`[ProjectDir]`\bin\Debug\\`[AssemblyName]`.dll' could not be found.|
|Error	|MC2000	|Unknown build error, 'Could not find file '`[ProjectDir]`\obj\Debug\\`[AssemblyName]`.dll'.|
|Error	|MC3050	|Cannot find the type '`[CustomClass]`'. Note that type names are case sensitive.|
|Error	|MC3074	|The tag '`[CustomClass]`' does not exist in XML namespace 'clr-namespace:`[Namespace]`'.|

NOTE: In all cases, the `[ProjectDir]`/`[AssemblyName]`is the same as the source of the error (I assume due to the iterative way WPF is built) and the `[CustomClass]`es which aren't found in the `[Namespace]`s are also within the same project.

This occurs in two separate solutions using `$(MSBuildProjectName)` in `AssemblyName`, and the only way I could get it to build was to override `Directory.Build.props` with an explicit `AssemblyName` tag in the `*.csproj` (not using `$(MSBuildProjectName)`). 

I tried to make a sample solution without any private code, but for some reason it was lacking the complexity which causes it to fail, sorry.

vishalmsft:
@ryalanms please help with providing the reference to those PR, which fixed for .NET 6.
[...]

PR #7557, opened 2023-02-21
Fixing the AssemblyName in temporary project file for GenerateTemporaryTargetAssembly task

### LINKED ISSUE #5711: Regression for custom AssemblyName in WPF apps
### Issue Description

The PropertyGroup item `AssemblyName` cannot use macros for WPF app. This is regression, all was fine for previous MSBuild version (16)

### Steps to Reproduce
Add to \*.csproj something like :

```xml
        <MainName>$(MSBuildProjectName.Replace('-NetCore', ''))</MainName>
        <AssemblyName>$(MainName)</AssemblyName>
```

Example project attached:

[WpfApp1.zip](https://github.com/dotnet/msbuild/files/7504674/WpfApp1.zip)

### Expected Behavior

Should compile/build

### Actual Behavior

Error occurs :

```
  MainWindow.xaml(10, 10): [MC3074] The tag 'DelayTextBox' does not exist in XML namespace 'clr-namespace:WpfApp1'. Line 10 Position 10.
```

When mentioned above part of \*.csproj is deleted all compiles/build again...

### Analysis

Important note : to reproduce, ensure that `bin` and `obj` directories are deleted.

### Versions & Configurations

17.0.0.52104

Kuldeep-MS:
Fixes 
#5711
#5576

## Description

When running the `GenerateTemporaryTargetAssmebly` task, a temporary project file gets created from the user defined project and in the process of creation, two instances of `AssemblyName` gets inserted in the file. 

First instance will have evaluated value for assembly name (i.e. all the macros will be resolved) whereas second instance will still have the unresolved macros in the `AssemblyName`.

If `MSBuildProjectName` is being used computation of `AssemblyName`, it will cause MSBuild to pick current project name for naming the assembly and because temporary project file is being used we will have temporary project file name as assembly name and eventually no local references will be found in MarkupCompilation2.

Removing the second instance of `AssemblyName` from aforementioned project file resolves the issue as `MSBuild` will have only evaluated `AssemblyName`. 

## Customer Impact

Projects are not building if `$(MSBuildProjectName)` is being used in `AssemblyName`

## Regression

Yes

## Testing

In Progress

## Risk

In evaluation

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/wpf/pull/7557)

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/PresentationBuildTasks/Microsoft/Build/Tasks/Windows/GenerateTemporaryTargetAssembly.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationBuildTasks/Microsoft/Build/Tasks/Windows/GenerateTemporaryTargetAssembly.cs
@@ -279,6 +279,9 @@ namespace Microsoft.Build.Tasks.Windows
                     ( "_TargetAssemblyProjectName", Path.GetFileNameWithoutExtension(CurrentProject) ),
                 };
 
+                //Removing duplicate AssemblyName
+                RemovePropertiesByName(xmlProjectDoc, nameof(AssemblyName));
+
                 AddNewProperties(xmlProjectDoc, properties);
 
                 // Save the xmlDocument content into the temporary project file.
@@ -568,10 +571,11 @@ namespace Microsoft.Build.Tasks.Windows
         #region Private Methods
 
         //
-        // Remove specific items from project file. Every item should be under an ItemGroup.
+        // Remove specific entity from project file.
         //
-        private void RemoveItemsByName(XmlDocument xmlProjectDoc, string sItemName)
+        private void RemoveEntityByName(XmlDocument xmlProjectDoc, string sItemName, string groupName)
         {
+            
             if (xmlProjectDoc == null || String.IsNullOrEmpty(sItemName))
             {
                 // When the parameters are not valid, simply return it, instead of throwing exceptions.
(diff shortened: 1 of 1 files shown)


src/Microsoft.DotNet.Wpf/src/PresentationBuildTasks/Microsoft/Build/Tasks/Windows/GenerateTemporaryTargetAssembly.cs
```diff
@@ -671,6 +674,110 @@ private void RemoveItemsByName(XmlDocument xmlProjectDoc, string sItemName)
 
         }
 
+        //
+        // Remove specific property from project file. Every property should be under an PropertyGroup.
+        //
+        private void RemovePropertiesByName(XmlDocument xmlProjectDoc, string sPropertyName)
```
pchaurasia14, line 680: Since the only difference between `RemoveItemsByName` and `RemovePropertiesByName` is the rootnode name comparison, we can refactor it into a common method and pass it as a parameter.

Kuldeep-MS: The reasoning behind making a separate function is to avoid modification of behavior and signature of existing function to avoid any unforeseen regressions.

PR #7560, opened 2023-02-22
Fixing the AssemblyName in temporary project file for GenerateTemporaryTargetAssembly task

### LINKED ISSUE #5711: Regression for custom AssemblyName in WPF apps

Kuldeep-MS:
Fixes 
#5711
#5576

Main PR https://github.com/dotnet/wpf/pull/7557

## Description

When running the `GenerateTemporaryTargetAssmebly` task, a temporary project file gets created from the user defined project and in the process of creation, two instances of `AssemblyName` gets inserted in the file. 

First instance will have evaluated value for assembly name (i.e. all the macros will be resolved) whereas second instance will still have the unresolved macros in the `AssemblyName`.

If `MSBuildProjectName` is being used computation of `AssemblyName`, it will cause MSBuild to pick current project name for naming the assembly and because temporary project file is being used we will have temporary project file name as assembly name and eventually no local references will be found in MarkupCompilation2.

Removing the second instance of `AssemblyName` from aforementioned project file resolves the issue as `MSBuild` will have only evaluated `AssemblyName`. 

## Customer Impact

Projects are not building if `$(MSBuildProjectName)` is being used in `AssemblyName`

## Regression

Yes

## Testing

In Progress

## Risk

In evaluation

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/wpf/pull/7560)

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/PresentationBuildTasks/Microsoft/Build/Tasks/Windows/GenerateTemporaryTargetAssembly.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationBuildTasks/Microsoft/Build/Tasks/Windows/GenerateTemporaryTargetAssembly.cs
@@ -279,6 +279,9 @@ namespace Microsoft.Build.Tasks.Windows
                     ( "_TargetAssemblyProjectName", Path.GetFileNameWithoutExtension(CurrentProject) ),
                 };
 
+                //Removing duplicate AssemblyName
+                RemovePropertiesByName(xmlProjectDoc, nameof(AssemblyName));
+
                 AddNewProperties(xmlProjectDoc, properties);
 
                 // Save the xmlDocument content into the temporary project file.
@@ -568,10 +571,11 @@ namespace Microsoft.Build.Tasks.Windows
         #region Private Methods
 
         //
-        // Remove specific items from project file. Every item should be under an ItemGroup.
+        // Remove specific entity from project file.
         //
-        private void RemoveItemsByName(XmlDocument xmlProjectDoc, string sItemName)
+        private void RemoveEntityByName(XmlDocument xmlProjectDoc, string sItemName, string groupName)
         {
+            
             if (xmlProjectDoc == null || String.IsNullOrEmpty(sItemName))
             {
                 // When the parameters are not valid, simply return it, instead of throwing exceptions.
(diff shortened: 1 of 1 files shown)


miloush:
The PR claims to fix #5697 but it is not clear that the `RootNamespace` will behave as expected in the issue.

miloush:
The code changes itself look alright, but why do we end up with two AssemblyName elements in the project file in the first place? Wouldn't it be better not to generate the second one rather than remove it?

Kuldeep-MS:
> The PR claims to fix #5697 but it is not clear that the `RootNamespace` will behave as expected in the issue.

@miloush - Thank you for bringing up this to my notice. I had followed with the author of #5697 and found that this don't fixes his issue. I forgot to remove the tag from the PR, I will do it now.

ThomasGoulet73:
@miloush We need to generate an `AssemblyName` property in the generated .csproj because it needs to be evaluated using the original .csproj instead of the generated .csproj in case it uses properties like `MSBuildProjectName` which would differ between the 2.

ThomasGoulet73:
dotnet/wpf#5697 should be fixed by my PR dotnet/wpf#6535. I will validate that my PR fixes the issue and link my PR to the issue if it does.

pchaurasia14:
@miloush - The  `GenerateTemporaryTargetAssembly` task currently does the following - 
- Load the user-defined csproj 
- Prepends new properties in it
- Save the csproj as the temporary proj

The problem associated arises when user defined csproj has a `AssemblyName` that references `$(MsBuildProjName)`. 
The evaluation of this property in context of the temporary project would produce the name of the current temporary project instead of the user-defined `AssemblyName`.

The current PR intends to remove the `AssemblyName` property (with the `$(MsBuildProjName)` literal) coming from csproj and instead use the already computed property of `AssemblyName` (this computed property has the correct value of `AssemblyName`).

Kuldeep-MS:
Closing this PR as of now. Will reopen if required

PR #7500, opened 2023-02-04
Replace String.CompareOrdinal to string.Equals in KnownTypes

lindexi:
## Description

We can use `string.Equals(string1, string2, StringComparison.Ordinal)` to check for equality, see https://learn.microsoft.com/en-us/dotnet/csharp/how-to/compare-strings and https://learn.microsoft.com/en-us/dotnet/standard/base-types/best-practices-strings

> Use the [String.Compare](https://learn.microsoft.com/en-us/dotnet/api/system.string.compare) and [String.CompareTo](https://learn.microsoft.com/en-us/dotnet/api/system.string.compareto) methods to sort strings, not to check for equality.

Reference https://github.com/dotnet/wpf/issues/6588

Cc @Kuldeep-MS 

## Customer Impact

Small performance improvements

## Regression

None.

## Testing

Just CI.

## Risk

Low.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/wpf/pull/7500)

PR #7680, opened 2023-03-28
Replace String.CompareOrdinal to string.Equals in KnownTypes Part 2

lindexi:
## Description

We can use `string.Equals(string1, string2, StringComparison.Ordinal)` to check for equality, see https://learn.microsoft.com/en-us/dotnet/csharp/how-to/compare-strings and https://learn.microsoft.com/en-us/dotnet/standard/base-types/best-practices-strings

Reference https://github.com/dotnet/wpf/pull/7500

## Customer Impact

## Regression

None.

## Testing

Just CI.

## Risk

Low.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/wpf/pull/7680)

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Markup/KnownTypes.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Markup/KnownTypes.cs
@@ -3826,15 +3826,15 @@ namespace System.Windows.Markup
             switch (id)
             {
                 case KnownElements.ColumnDefinition:
-                    if (String.CompareOrdinal(propName, "MinWidth") == 0)
+                    if (string.Equals(propName, "MinWidth", StringComparison.Ordinal))
                         converterId = KnownElements.LengthConverter;
-                    else if (String.CompareOrdinal(propName, "MaxWidth") == 0)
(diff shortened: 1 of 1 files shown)


dipeshmsft (approved):

anjali-wpf (approved):

dipeshmsft:
Hey @lindexi, I think we can create an issue for this if there will be more similar PRs.

lindexi:
@dipeshmsft Can I find any similar issues?

dipeshmsft:
Let me take a look, if I find a similar issue, I will ping you back here.

dipeshmsft:
@lindexi, I wasn't able to look at another similar issue. However, we have run the test pass on this fix. No new fails. 
Thanks for your continued support. 

Also, you have mentioned in the Customer Impact section that we may have small performance improvements, do you have any benchmarks for the same ?

lindexi:
> Also, you have mentioned in the Customer Impact section that we may have small performance improvements, do you have any benchmarks for the same ?

@dipeshmsft Sorry, I didn't measure the performance in wpf. I'm just measuring it in my demo project.

PR #7684, opened 2023-03-29
Enabling Hardware Acceleration on RDP

### LINKED ISSUE #3215: Enable h/w acceleration in RDP
https://github.com/dotnet/wpf/blob/e7cad628c82f8761cac2070c80601ae8e75139a6/src/Microsoft.DotNet.Wpf/src/WpfGfx/core/common/display.cpp#L2128-L2134

Remove that check for `m_fNonLocalDevicePresent` and DirectX h/w acceleration in RDP will start working. 

In general, the overall role of `m_fNonLocalDevicePresent` may have to be re-evaluated and updated. <<Doubt this will be a big change.>>

This limitation (is s/w rendering in RDP) is no longer needed in WDDM driver model, though it made sense historically when XPDM drivers were in the mix in the ecosystem for .NET Framework. 

**Stretch goal:** Prove this works well and is safe in .NET 5 and then port the change to .NET 4.8. All the remote working software devs who use Visual Studio over RDP would love the improvement. 

/cc @fabiant3 , @SamBent - I’d be happy to contribute if you are open to take this change. Not sure whether the timing is right wrt .NET 5 schedule etc.

Kuldeep-MS:
Fixes #3215

## Description

Enabling Hardware Acceleration on RDP for exes which can opt-in.

## Customer Impact

Without this fix, customers won't be able to get the benefits of Hardware Acceleration on RDP.

## Regression

No

## Testing

In progress

## Risk

Risk is Low as the applications which want to have hardware acceleration on RDP need to Opt-in by setting AppContext Switch (`Switch.System.Windows.Media.EnableHardwareAccelerationForRdp`) to true.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/wpf/pull/7684)

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Media/MediaSystem.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Media/MediaSystem.cs
@@ -79,6 +79,10 @@ namespace System.Windows.Media
                 }
                 s_refCount++;
             }
+
+            // Setting renderOption for Hardware acceleration in RDP as per appcontext switch.
+            UnsafeNativeMethods.RenderOptions_EnableHardwareAccelerationInRdp(CoreAppContextSwitches.EnableHardwareAccelerationInRdp);
+
             // Consider making MediaSystem.ConnectTransport return the state of transport connectedness so
             // that we can initialize the media system to a disconnected state.
 
--- a/src/Microsoft.DotNet.Wpf/src/WpfGfx/core/common/d3dutils.cpp
+++ b/src/Microsoft.DotNet.Wpf/src/WpfGfx/core/common/d3dutils.cpp
@@ -232,10 +232,11 @@ GetMinimalTextureDesc(
         }
 
         // If format is ok, restore HR from width/height check
-        if (hr == S_OK)
+        if(SUCCEEDED(hr))
         {
             hr = hrWH;
         }
+        
     }
 
 #if DBG
@@ -315,8 +316,3 @@ GetSuperiorSurfaceFormat(
 
     return D3DFMT_UNKNOWN;
 }
-
-
-
-
-
--- a/src/Microsoft.DotNet.Wpf/src/WpfGfx/core/common/renderoptions.cpp
+++ b/src/Microsoft.DotNet.Wpf/src/WpfGfx/core/common/renderoptions.cpp
@@ -23,10 +23,17 @@ RenderOptions_IsSoftwareRenderingForcedForProcess()
     return RenderOptions::IsSoftwareRenderingForcedForProcess();
 }
 
+void WINAPI
+RenderOptions_EnableHardwareAccelerationInRdp(BOOL fEnable)
+{
+    RenderOptions::EnableHardwareAccelerationInRdp(fEnable);
+}
+
 // m_cs must be entered before accessing m_fForceSoftware because multiple
 // managed threads plus the render thread could try to access it
 static CCriticalSection m_cs;
 static bool m_fForceSoftware;
+static bool m_fHwAccelerationInRdpEnabled;
 
 //+---------------------------------------------------------------------------------
 //
@@ -41,6 +48,7 @@ HRESULT
 RenderOptions::Init()
 {
     m_fForceSoftware = false;
+    m_fHwAccelerationInRdpEnabled = false;
     RRETURN(m_cs.Init());
 }
 
@@ -94,4 +102,30 @@ RenderOptions::IsSoftwareRenderingForcedForProcess()
     return m_fForceSoftware;
 }
 
+//+---------------------------------------------------------------------------------
+//
+//  RenderOptions::EnableGraphicHWAccelerationInRdp
+//
+//  Synopsis:   Sets whether or not Hardware Acceleration should be enabled for RDP.
+//
+//----------------------------------------------------------------------------------
+void
+RenderOptions::EnableHardwareAccelerationInRdp(BOOL fEnable)
+{
+    CGuard<CCriticalSection> guard(m_cs);
+    m_fHwAccelerationInRdpEnabled = !!fEnable;
+}
+
+//+---------------------------------------------------------------------------------
+//
+//  RenderOptions::IsHardwareAccelerationInRdpEnabled
+//
+//  Synopsis:   return whether or not Hardware Acceleration for RDP is enabled.
+//
+//----------------------------------------------------------------------------------
+BOOL
+RenderOptions::IsHardwareAccelerationInRdpEnabled()
+{
+    return m_fHwAccelerationInRdpEnabled;
+}
 
--- a/src/Microsoft.DotNet.Wpf/src/WpfGfx/core/common/renderoptions.h
+++ b/src/Microsoft.DotNet.Wpf/src/WpfGfx/core/common/renderoptions.h
@@ -18,6 +18,10 @@ namespace RenderOptions
     void ForceSoftwareRenderingForProcess(BOOL fForce);
 
     BOOL IsSoftwareRenderingForcedForProcess();
+
+    void EnableHardwareAccelerationInRdp(BOOL fEnable);
+
+    BOOL IsHardwareAccelerationInRdpEnabled();
 };
 
 
--- a/src/Microsoft.DotNet.Wpf/src/WpfGfx/core/hw/hwbitmapcolorsource.cpp
+++ b/src/Microsoft.DotNet.Wpf/src/WpfGfx/core/hw/hwbitmapcolorsource.cpp
@@ -1421,16 +1421,17 @@ CHwBitmapColorSource::ComputeRealizationParameters(
             &uLevels
             );
 
-        Assert(pDevice->GetMinimalTextureDesc(
-            &d3dsdRequired,
-            TRUE,
-            (GMTD_CHECK_ALL |
-             (TextureAddressingAllowsConditionalNonPower2Usage(
-                 oRealizationParams.dlU.d3dta,
-                 oRealizationParams.dlV.d3dta) ?
-              GMTD_NONPOW2CONDITIONAL_OK : 0)
-            )
-            ) == S_OK);
+        Assert(SUCCEEDED(pDevice->GetMinimalTextureDesc(
+        &d3dsdRequired,
+        TRUE,
+        (GMTD_CHECK_ALL |
+            (TextureAddressingAllowsConditionalNonPower2Usage(
+                oRealizationParams.dlU.d3dta,
+                oRealizationParams.dlV.d3dta) ?
+            GMTD_NONPOW2CONDITIONAL_OK : 0)
+        )
+        )));
+        
     }
     #endif
 
--- a/src/Microsoft.DotNet.Wpf/src/PresentationCore/MS/internal/CoreAppContextSwitches.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationCore/MS/internal/CoreAppContextSwitches.cs
@@ -380,6 +380,21 @@ namespace MS.Internal
 
         #endregion
 
+        #region EnableHardwareAccelerationInRdp
+
+        internal const string EnableHardwareAccelerationInRdpSwitchName = "Switch.System.Windows.Media.EnableHardwareAccelerationInRdp";
+        private static int _enableHardwareAccelerationInRdp;
+        public static bool EnableHardwareAccelerationInRdp
+        {
+            [MethodImpl(MethodImplOptions.AggressiveInlining)]
+            get
+            {
+                return LocalAppContext.GetCachedSwitchValue(EnableHardwareAccelerationInRdpSwitchName, ref _enableHardwareAccelerationInRdp);
+            }
+        }
+
+        #endregion
+
     }
 #pragma warning restore 436
 }
--- a/src/Microsoft.DotNet.Wpf/src/PresentationCore/System/AppContextDefaultValues.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationCore/System/AppContextDefaultValues.cs
@@ -60,6 +60,7 @@ namespace System
             LocalAppContext.DefineSwitchDefault(CoreAppContextSwitches.ShouldNotRenderInNonInteractiveWindowStationSwitchName, false);
             LocalAppContext.DefineSwitchDefault(CoreAppContextSwitches.DoNotUsePresentationDpiCapabilityTier3OrGreaterSwitchName, false);
             LocalAppContext.DefineSwitchDefault(CoreAppContextSwitches.AllowExternalProcessToBlockAccessToTemporaryFilesSwitchName, false);
+            LocalAppContext.DefineSwitchDefault(CoreAppContextSwitches.EnableHardwareAccelerationInRdpSwitchName, false);
         }
     }
 #pragma warning restore 436
--- a/src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Media/UnsafeNativeMethodsMilCoreApi.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Media/UnsafeNativeMethodsMilCoreApi.cs
@@ -208,7 +208,10 @@ namespace MS.Win32.PresentationCore
                 bool fForce);
 
             [DllImport(DllImport.MilCore, EntryPoint = "RenderOptions_IsSoftwareRenderingForcedForProcess")]
-            internal unsafe static extern bool RenderOptions_IsSoftwareRenderingForcedForProcess();            
+            internal unsafe static extern bool RenderOptions_IsSoftwareRenderingForcedForProcess();
+
+            [DllImport(DllImport.MilCore, EntryPoint = "RenderOptions_EnableHardwareAccelerationInRdp")]
+            internal unsafe static extern bool RenderOptions_EnableHardwareAccelerationInRdp(bool value);                 
 
             [DllImport(DllImport.MilCore, EntryPoint = "MilResource_CreateCWICWrapperBitmap")]
             internal unsafe static extern int /* HRESULT */ CreateCWICWrapperBitmap(
--- a/src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Media/UnsafeNativeMethodsMilCoreApi.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Media/UnsafeNativeMethodsMilCoreApi.cs
@@ -211,7 +211,7 @@ namespace MS.Win32.PresentationCore
             internal unsafe static extern bool RenderOptions_IsSoftwareRenderingForcedForProcess();
 
             [DllImport(DllImport.MilCore, EntryPoint = "RenderOptions_EnableHardwareAccelerationInRdp")]
-            internal unsafe static extern bool RenderOptions_EnableHardwareAccelerationInRdp(bool value);                 
+            internal unsafe static extern void RenderOptions_EnableHardwareAccelerationInRdp(bool value);                 
 
             [DllImport(DllImport.MilCore, EntryPoint = "MilResource_CreateCWICWrapperBitmap")]
             internal unsafe static extern int /* HRESULT */ CreateCWICWrapperBitmap(
--- a/src/Microsoft.DotNet.Wpf/src/WpfGfx/core/common/display.cpp
+++ b/src/Microsoft.DotNet.Wpf/src/WpfGfx/core/common/display.cpp
@@ -2125,7 +2125,7 @@ CDisplaySet::GetGraphicsAccelerationCaps(
         *pulDisplayUniqueness = m_ulDisplayUniquenessLoader;
     }
 
-    if (m_rgpDisplays.GetCount() == 0 || m_fNonLocalDevicePresent)
+    if (m_rgpDisplays.GetCount() == 0 || (!RenderOptions::IsHardwareAccelerationInRdpEnabled() && m_fNonLocalDevicePresent))
     {
         //
         // No display - no acceleration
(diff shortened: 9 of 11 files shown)


batzen:
I don't think using the registry and requiring the full exe path is a good solution.
How am i supposed to use this properly in apps that don't have an installer or during development?

Kuldeep-MS:
@batzen - Yes, you're correct. While the registry solution may not be suitable for all users, we have discovered that several switches in the unmanaged code rely on this approach. [Link](https://github.com/dotnet/wpf/blob/639e7807f186f4dd6c996c967d99413f29a1d560/src/Microsoft.DotNet.Wpf/src/WpfGfx/core/common/display.cpp#L88-L95)

 

Although using AppContextSwitch might have been a better option, we need to investigate its viability in the unmanaged part of WPF.

JLucGarnier:
Hello, 
what WPF version will embed the fix to this issue?
do you have any rough estimated time to get a public access to this version?  
Thank you.

JLuc

pchaurasia14:
@JLucGarnier - We are targeting this for .NET 8. We're only left with the last steps here - To integrate an opt-in switch mechanism. 
The current challenge is the integration of AppContextSwitches which flows back to unmanaged part of the codebase. 
We are working to see how can we get this implemented asap.

src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Media/MediaSystem.cs
```diff
@@ -79,6 +79,17 @@ public static bool Startup(MediaContext mc)
                 }
                 s_refCount++;
             }
+
+             /// <summary>
+            /// Set renderOption if Hardware acceleration is enabled for RDP
+            // </summary>
+
+            bool isHardwareAccelerationEnabledForRDP = false;
+            AppContext.TryGetSwitch("System.Windows.Media.isHardwareAccelerationEnabledForRDP", out isHardwareAccelerationEnabledForRDP);
```
lindexi, line 88: How about use Pascal case as the configuration string?

```suggestion
            AppContext.TryGetSwitch("System.Windows.Media.IsHardwareAccelerationEnabledForRDP", out isHardwareAccelerationEnabledForRDP);
```

I found that other code is using Pascal case as the configuration string.

https://github.com/dotnet/wpf/blob/2804219c720fd90e5cbba9df535eb2ed5c8a6dd1/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Controls/DataGridCell.cs#L52

Kuldeep-MS: @lindexi - Thank you for your review. I have made the required change.

src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Media/MediaSystem.cs
```diff
@@ -79,6 +79,17 @@ public static bool Startup(MediaContext mc)
                 }
                 s_refCount++;
             }
+
+             /// <summary>
+            /// Set renderOption if Hardware acceleration is enabled for RDP
+            // </summary>
```
miloush, line 85: You don't put XML comments inside a member block. Just use // without summary

src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Media/MediaSystem.cs
```diff
@@ -79,6 +79,17 @@ public static bool Startup(MediaContext mc)
...
+
+            bool isHardwareAccelerationEnabledForRDP = false;
+            AppContext.TryGetSwitch("System.Windows.Media.IsHardwareAccelerationEnabledForRDP", out isHardwareAccelerationEnabledForRDP);
+            if(isHardwareAccelerationEnabledForRDP)
```
miloush, line 89: space after if

src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Media/MediaSystem.cs
```diff
@@ -79,6 +79,17 @@ public static bool Startup(MediaContext mc)
                 }
                 s_refCount++;
             }
+
+             /// <summary>
+            /// Set renderOption if Hardware acceleration is enabled for RDP
+            // </summary>
+
+            bool isHardwareAccelerationEnabledForRDP = false;
+            AppContext.TryGetSwitch("System.Windows.Media.IsHardwareAccelerationEnabledForRDP", out isHardwareAccelerationEnabledForRDP);
```
miloush, line 88: Technically the .NET naming guidelines states acronyms of 3 and more letters should be cased, Rdp as Tcp

src/Microsoft.DotNet.Wpf/src/WpfGfx/core/common/renderoptions.cpp
```diff
@@ -41,6 +48,7 @@ HRESULT
 RenderOptions::Init()
 {
     m_fForceSoftware = false;
+    m_fHwAccelerationForRDPEnabled = false;
```
miloush, line 51: shouldn't we make this opt out and default to true?

src/Microsoft.DotNet.Wpf/src/WpfGfx/core/common/renderoptions.h
```diff
@@ -18,6 +18,10 @@ namespace RenderOptions
     void ForceSoftwareRenderingForProcess(BOOL fForce);
 
     BOOL IsSoftwareRenderingForcedForProcess();
+
+    void SetGraphicHwAccelerationForRDP(BOOL value);
+
+    BOOL IsGraphicHWAccelerationForRDPEnabled();
```
miloush, line 24: why is the Set method `Hw` but the Is method `HW`?

src/Microsoft.DotNet.Wpf/src/WpfGfx/core/hw/hwbitmapcolorsource.cpp
```diff
@@ -1421,7 +1421,20 @@ CHwBitmapColorSource::ComputeRealizationParameters(
             &uLevels
             );
 
-        Assert(pDevice->GetMinimalTextureDesc(
+        if(RenderOptions::IsGraphicHWAccelerationForRDPEnabled())
```
miloush, line 1424: How are the two branches different? (i.e which success HR result do you expect that should fail without RDP enabled?)

src/Microsoft.DotNet.Wpf/src/WpfGfx/core/common/d3dutils.cpp
```diff
@@ -232,10 +232,19 @@ GetMinimalTextureDesc(
...
         // If format is ok, restore HR from width/height check
-        if (hr == S_OK)
+        if (RenderOptions::IsGraphicHWAccelerationForRDPEnabled())
         {
-            hr = hrWH;
+            if(SUCCEEDED(hr))
+            {
+                hr = hrWH;
+            }
+        } else {
```
miloush, line 241: the format in the file seems to use new lines around else

miloush (commented):
Clearly this needs to go through DRT in RDP.

I guess the biggest question is why is this opt-in rather than opt-out?

miloush:
The idea that no display = no hw acceleration is also questionable to me.

miloush:
The other flag that triggers `m_fNonLocalDevicePresent` is `DISPLAY_DEVICE_MIRRORING_DRIVER`. If Windows 7 is still supported OS for .NET 8, it might be worth checking things work, although an opt in/out mechanism is possibly a reasonable substitute.

src/Microsoft.DotNet.Wpf/src/WpfGfx/core/hw/hwbitmapcolorsource.cpp
Kuldeep-MS: When the D3DFormat is X8R8G8B8, we cannot obtain the HResult as S_OK. Therefore, we must use the SUCCEEDED macro to determine whether the result is a pass or fail by passing the HResult to it.

We are using an if-else block to prevent users who do not wish to opt-in for HW Acceleration for RDP to follow the new code flow from doing so.

src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Media/MediaSystem.cs
Kuldeep-MS: Thanks for pointing it out I will make the necessary changes.

src/Microsoft.DotNet.Wpf/src/WpfGfx/core/hw/hwbitmapcolorsource.cpp
miloush: What non S_OK result do you get? This might be worth mentioned in a comment in the code.

miloush: Also to clarify, the risk here is that we would allow WPF applications to run in more scenarios than before? Why would we want to prevent that?

pchaurasia14:
@miloush - Our rationale for making this change opt-in vs opt-out is the lack of appropriate test coverage in this area. There could be use-cases where such feature may start breaking applications post upgrade. Hence we are inclined to keep this opt-in for now and work on improving the tests around this area first. 

Once we have that we can decide to turn it on by default for all users (in future .NET releases).

src/Microsoft.DotNet.Wpf/src/WpfGfx/core/common/renderoptions.cpp
```diff
@@ -23,10 +23,17 @@ RenderOptions_IsSoftwareRenderingForcedForProcess()
     return RenderOptions::IsSoftwareRenderingForcedForProcess();
 }
 
+void WINAPI
+RenderOptions_SetGraphicHWAccelerationForRdp(BOOL value)
```
JeremyKuhne, line 27: More descriptive API names would be helpful. `BOOL enable` or `BOOL fEnable`. And perhaps the method name too: `RenderOptions_EnableGraphicsHardwareAccelerationForRdp`

src/Microsoft.DotNet.Wpf/src/WpfGfx/core/common/renderoptions.cpp
Kuldeep-MS: see [comment](https://github.com/dotnet/wpf/pull/7684#issuecomment-1628405496)

src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Media/MediaSystem.cs
Kuldeep-MS: done

src/Microsoft.DotNet.Wpf/src/WpfGfx/core/common/renderoptions.cpp
Kuldeep-MS: thank you for your feedback. I think this make sense.

src/Microsoft.DotNet.Wpf/src/WpfGfx/core/hw/hwbitmapcolorsource.cpp
Kuldeep-MS: Yes, this make sense to me to add appropriate comment about non S_OK result, I will take care of that.
And due to the lack of appropriate test coverage in this area we are not enabling it for everyone.

miloush:
@pchaurasia14 While I understand the concern, I do not find an opt-in justified or beneficial enough in this case. You also cannot be adding compatibility flags for every change, it adds maintainability costs to the future.

The DirectX documentation states that creating the surface under RDP is now always supported. You have as much testing as you have for any other feature by running the DTS in a remote desktop session.

There already exist several opt out of hardware acceleration options for users:
1. Global, using [registry setting](https://learn.microsoft.com/en-us/previous-versions/dotnet/netframework-4.0/aa970912(v=vs.100)?redirectedfrom=MSDN#disable-hardware-acceleration-option)
2. Process-wide using [ProcessRenderMode](https://learn.microsoft.com/en-us/dotnet/api/system.windows.media.renderoptions.processrendermode) which also documents these 3 options
3. Per window using [RenderMode](https://learn.microsoft.com/en-us/dotnet/api/system.windows.interop.hwndtarget.rendermode)

Hardware acceleration should be the normal state of things, and what we hope to have going forward. You are looking for more testing, but I am worried you will not get enough coverage by real-world applications if this is an opt in.

As indicated in the original issue, this change has been analysed several times in the past by @vatsan-madhavan and you are not going to get anything better than that. He also suggested running regression tests in RDP for verification and no reason for a flag in .NET (though he supported one for .NET Framework).

If you are not comfortable making this change without a flag, it sounds to me like an opt out rather than opt in would be a good compromise and with alignment of the other options to opt out of hw rendering.

In general, how do you decide whether to go for opt in or opt out, are there any criteria?

pchaurasia14:
@miloush - Fair points. Currently, our criteria is the higher number of unknowns in the code paths and the approaching timeline for .NET8 release. The change in the current PR is at a fairly low level which creates higher risk of applications failing to start if we have missed something. All the changes in the current PR were made on the issues we encountered in sample application (that uses multiple animations). With the limited coverage, we don't know if the code paths in question are being executed for any other use cases.

Also, as we have more (and better) tests around this, we will revisit this and turn on the flag for next releases (.NET9). We'll create a tracker issue for this. Having an opt-in gives us the room for experiment without causing disruption to people who would be migrating to newer .NET version.

src/Microsoft.DotNet.Wpf/src/WpfGfx/core/hw/hwbitmapcolorsource.cpp
Kuldeep-MS: As far as I understand, D3DFormat refers to the format of the adapter in which it is displaying. In my opinion, it is possible that the apps you mentioned may not be running on certain devices, but switching to the SUCCEEDED macro could potentially resolve this issue and allow the apps to run on those devices. While this improvement will certainly be beneficial, but as mentioned [here](https://github.com/dotnet/wpf/pull/7684#issuecomment-1630680002) by Pankaj we can revisit this and turn on the flag for next releases.

miloush: I don't think my comment was addressed. Whether there is a flag or not for enabling hw rendering is separate issue to this bit of code, which I would think is actually a bug. There is no working scenario that using `SUCCEEDED` can possibly break.

src/Microsoft.DotNet.Wpf/src/WpfGfx/core/common/d3dutils.cpp
```diff
@@ -232,10 +232,21 @@ GetMinimalTextureDesc(
         }
 
         // If format is ok, restore HR from width/height check
-        if (hr == S_OK)
+        if (RenderOptions::IsGraphicHWAccelerationForRdpEnabled())
         {
-            hr = hrWH;
+            if(SUCCEEDED(hr))
```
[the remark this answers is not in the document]
JeremyKuhne: It would be great to be explicit here. You could also simplify the logic:

```C++
if (hr == S_OK || (hr == D3DOK_NOAUTOGEN && RenderOptions::IsGraphicHWAccelerationForRdpEnabled()))
{
    hr = hrWH;
}
```

I'd recommend doing the same sort of thing in the other blocks.

JeremyKuhne:
On the topic of opt-in vs. opt-out, in WinForms we lean towards opt-out much as described by @vatsan-madhavan
 https://github.com/dotnet/wpf/pull/7684#issuecomment-1630856727. Everything is case-by-case, of course, but we want the better experience to be the default for apps on .NET (core). This is particularly true for new apps.

.NET Framework deals with this by typically flipping `AppContext` flags on by default for the target framework that they are introduced in. That's effectively analogous to the behavior of having them on by default in the version they were introduced in in .NET (core).

vatsan-madhavan:
Please take a look at this to see how AppContext flags are plumbed down into WpfGfx. You'll have to further search the codebase for where `MIL_RT_IS_DISABLE_MULTIMON_DISPLAY_CLIPPING_VALID` is defined to see how the overall plumbing works.

https://github.com/dotnet/wpf/blob/f1987cd1ce5fa536f2f2e020ca5d0c15fe535e29/src/Microsoft.DotNet.Wpf/src/WpfGfx/include/exports.cs#L2536-L2553

vatsan-madhavan:
> @batzen - Yes, you're correct. While the registry solution may not be suitable for all users, we have discovered that several switches in the unmanaged code rely on this approach. [Link](https://github.com/dotnet/wpf/blob/639e7807f186f4dd6c996c967d99413f29a1d560/src/Microsoft.DotNet.Wpf/src/WpfGfx/core/common/display.cpp#L88-L95)
> 
>  
> 
> Although using AppContextSwitch might have been a better option, we need to investigate its viability in the unmanaged part of WPF.

See my previous comment about how AppContext switches work in WpfGfx.

vatsan-madhavan:
>System.Windows.Media.isHardwareAccelerationEnabledForRDP

In this context, `isHardwareAccelerationEnabledForRDP` seems fine, but I'd suggest something slighly clearer like
- `EnableHardwareAccelerationInRDP` (if shipping disabled by default, and supplying an opt-in quirk) or
- `DisableHardwareAccelerationInRDP` (if shipping enabled by default and supplying an opt-out quirk)

In either case, the idea is that the quirks should be structured like this:
- `false` by default when TFM is current framework/runtime version
  - Can be explicitly opted-in by applications targeting the current framework/runtime version by setting it to `true`
- `true` by default when TFM is v.Next framework/runtime version

Kuldeep-MS:
> Please take a look at this to see how AppContext flags are plumbed down into WpfGfx. You'll have to further search the codebase for where `MIL_RT_IS_DISABLE_MULTIMON_DISPLAY_CLIPPING_VALID` is defined to see how the overall plumbing works.
> 
> https://github.com/dotnet/wpf/blob/f1987cd1ce5fa536f2f2e020ca5d0c15fe535e29/src/Microsoft.DotNet.Wpf/src/WpfGfx/include/exports.cs#L2536-L2553
> 
> Given that [`MilRTInitializationFlags`](https://github.com/dotnet/wpf/blob/f1987cd1ce5fa536f2f2e020ca5d0c15fe535e29/src/Microsoft.DotNet.Wpf/src/Common/Graphics/wgx_core_types.cs#L65C19-L65C43) has room for expansion, this is probably an apt way to communicate DX=LatestSupported vs. DX=v9 information down to MIL.
> 
> This information should be plumbed down at the very start of WpfGfx thread, and known early.
> 
> I don't understand whether you evaluated this existing mechanism and decided to do something else out of (unspecified) necessity, or whether you were unaware of this mechanism existing in WpfGfx. If the latter, please rework the changes (and if former, it would help reviewers to understand rationale).

Certainly! To provide more detail, I noticed that the flag I am sending is similar in nature to the flag used for forced software rendering in wpfgfx. Therefore, I decided to replicate the mechanism already in place for passing down that flag to wpfgfx, and use it for my new flag as well. This way, I can ensure that my new flag is handled in a consistent and reliable manner, just like the existing flag.

> Please take a look at this to see how AppContext flags are plumbed down into WpfGfx. You'll have to further search the codebase for where `MIL_RT_IS_DISABLE_MULTIMON_DISPLAY_CLIPPING_VALID` is defined to see how the overall plumbing works.
> 
> https://github.com/dotnet/wpf/blob/f1987cd1ce5fa536f2f2e020ca5d0c15fe535e29/src/Microsoft.DotNet.Wpf/src/WpfGfx/include/exports.cs#L2536-L2553
> 
> Given that [`MilRTInitializationFlags`](https://github.com/dotnet/wpf/blob/f1987cd1ce5fa536f2f2e020ca5d0c15fe535e29/src/Microsoft.DotNet.Wpf/src/Common/Graphics/wgx_core_types.cs#L65C19-L65C43) has room for expansion, this is probably an apt way to communicate DX=LatestSupported vs. DX=v9 information down to MIL.
> 
> This information should be plumbed down at the very start of WpfGfx thread, and known early.
> 
> I don't understand whether you evaluated this existing mechanism and decided to do something else out of (unspecified) necessity, or whether you were unaware of this mechanism existing in WpfGfx. If the latter, please rework the changes (and if former, it would help reviewers to understand rationale).

Certainly! To provide more detail, I noticed that the flag I am sending is similar in nature to the flag used for forced software rendering in wpfgfx. Therefore, I decided to replicate the mechanism already in place for passing down that flag to wpfgfx, and use it for my new flag as well, just like the existing flag.

Kuldeep-MS:
> Please take a look at this to see how AppContext flags are plumbed down into WpfGfx. You'll have to further search the codebase for where `MIL_RT_IS_DISABLE_MULTIMON_DISPLAY_CLIPPING_VALID` is defined to see how the overall plumbing works.
> 
> https://github.com/dotnet/wpf/blob/f1987cd1ce5fa536f2f2e020ca5d0c15fe535e29/src/Microsoft.DotNet.Wpf/src/WpfGfx/include/exports.cs#L2536-L2553
> 
> Given that [`MilRTInitializationFlags`](https://github.com/dotnet/wpf/blob/f1987cd1ce5fa536f2f2e020ca5d0c15fe535e29/src/Microsoft.DotNet.Wpf/src/Common/Graphics/wgx_core_types.cs#L65C19-L65C43) has room for expansion, this is probably an apt way to communicate DX=LatestSupported vs. DX=v9 information down to MIL.
> 
> This information should be plumbed down at the very start of WpfGfx thread, and known early.
> 
> I don't understand whether you evaluated this existing mechanism and decided to do something else out of (unspecified) necessity, or whether you were unaware of this mechanism existing in WpfGfx. If the latter, please rework the changes (and if former, it would help reviewers to understand rationale).

Certainly! To provide more detail, I noticed that the flag I am sending is similar in nature to the flag used for forced software rendering in wpfgfx. Therefore, I decided to replicate the mechanism already in place for passing down that flag to wpfgfx, and use it for my new flag as well, just like the existing flag.

src/Microsoft.DotNet.Wpf/src/WpfGfx/core/hw/hwbitmapcolorsource.cpp
Kuldeep-MS: @miloush Thank you for your suggestion. After thorough discussion with our internal team, we have decided to go with your suggestion and instead creating two separate branches we will be using only SUCCEEDED macro.

pchaurasia14 (approved):

PR #7729, opened 2023-04-17
Unblock AltGr+Oem2/5 typing inside ListBox

### LINKED ISSUE #7641: French Canadian Keyboard Layout "}" key being swallowed by WPF ListView
### Description

When a TextBox is inside a ListView in WPF, the "}" can't be used if the Keyboard Layout is French Canadian. This key is on Right-Alt (or Alt Gr)+< (on a US Keyboard it is the backslash/pipe key). Other key combinations on Right-Alt work fine. 

### Reproduction Steps

1) Insert the following XAML code into MainWindow.xaml:

```
<Grid>
    <Grid.RowDefinitions>
        <RowDefinition Height="Auto"/>
        <RowDefinition Height="Auto"/>
        <RowDefinition/>
        <RowDefinition/>
    </Grid.RowDefinitions>
    <ListView Grid.Row="0">
        <ListView.Items>
            <TextBox Width="300"></TextBox>
        </ListView.Items>    
    </ListView>
    <TextBox Grid.Row="1" HorizontalAlignment="Left" Width="300"></TextBox>
</Grid>
```
2) Build and run.
3) Switch the built in Keyboard settings in Windows to the French Canadian layout
4) Try typing "}" into the TextBox outside of the ListView.
5) Try typing "}" into the TextBox inside of the ListView. 
6) Switch Keyboard layouts to the US layout.
7) Try typing "}" into the TextBox inside of the ListView. 

### Expected behavior

The "}" key should be usable in a Textbox whether that TextBox is inside of a ListView or not, regardless of the keyboard layout.

### Actual behavior

When typing "}" into the TextBox outside of the ListView, the Textbox updates appropriately. 
When typing "}" into the TextBox inside of the ListView, nothing happens if the Keyboard layout is the French Canadian layout.
When typing  "}" into the TextBox inside of the ListView, it updates appropriately if the Keyboard layout is the US layout.

### Regression?

Unknown.

### Known Workarounds

For End Users, switch keyboard layouts.
For a code workaround, maybe handling the PreviewKeyDown event and directly updating the Textbox before it gets to the ListView.

### Impact

It only impacts people who use a French Canadian keyboard layout when typing a specific character into a Textbox, but it may affect others with extended characters on Right-Alt in their keyboard layouts.

### Configuration

Which version of .NET is the code running on? .NET 7.0
What OS and version, and what distro if applicable? Windows 11
What is the architecture (x64, x86, ARM, ARM64)? x64
Do you know whether it is specific to that configuration? I don't believe so.

### Other information

ListView appears to be marking the KeyPress event as handled, therefore it never gets back to TextBox as a TextInput event.

miloush:
Fixes #7641

## Description

`ListBox` and classes derived from it (notably `ListView`) are expected to _select all_ on <kbd>Ctrl</kbd>+<kbd>/</kbd> and _select item with focus_ on <kbd>Ctrl</kbd>+<kbd>\\</kbd>. However, the implementation only checks the status of the <kbd>Ctrl</kbd> modifier, not the others. As a result, this shortcut hijacks any other combination of keys, most importantly <kbd>Ctrl</kbd>+<kbd>Alt</kbd> which is used for typing on [around 60 keyboard layouts](http://kbdlayout.info/kbdus/shortcut/VK_OEM_2+VK_RMENU) that Windows ships.

This PR fixes the problem by limiting the shortcuts to what has been documented in the code comments as intent, i.e. they will work only when <kbd>Ctrl</kbd> is pressed and nothing else.

## Customer Impact

Not taking this fix prevents users from typing characters using these two keys on their keyboards into a `TextBox` or similar control inside `ListBox` or `ListView`.

## Regression

No.

## Testing

Built and verified using the repro in #7641 and Czech keyboard layout that typing is not possible without the fix and works with the fix (.NET 8.0.100-preview.3.23178.7)

## Risk

Selecting all and selecting the focused item is now limited to <kbd>Ctrl</kbd> only, while it used to work with any modifiers combination involving <kbd>Ctrl</kbd>.

However, the current situation is serious - users cannot type. The former behavior is relatively easily achievable by deriving from the control and overriding `OnKeyDown`.

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Controls/ListBox.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Controls/ListBox.cs
@@ -319,7 +319,7 @@ namespace System.Windows.Controls
                 case Key.Divide:
                 case Key.Oem2:
                     // Ctrl-Fowardslash = Select All
-                    if (((Keyboard.Modifiers & ModifierKeys.Control) == (ModifierKeys.Control)) && (SelectionMode == SelectionMode.Extended))
+                    if (((Keyboard.Modifiers) == (ModifierKeys.Control)) && (SelectionMode == SelectionMode.Extended))
                     {
                         SelectAll();
                     }
@@ -332,7 +332,7 @@ namespace System.Windows.Controls
 
                 case Key.Oem5:
                     // Ctrl-Backslash = Select the item with focus.
-                    if (((Keyboard.Modifiers & ModifierKeys.Control) == (ModifierKeys.Control)) && (SelectionMode == SelectionMode.Extended))
+                    if (((Keyboard.Modifiers) == (ModifierKeys.Control)) && (SelectionMode == SelectionMode.Extended))
                     {
                         ListBoxItem focusedItemUI = (FocusedInfo != null) ? FocusedInfo.Container as ListBoxItem : null;
                         if (focusedItemUI != null)


pchaurasia14:
@miloush - #7614 points to an arcade PR. Is that the correct issue?

miloush:
Sorry @pchaurasia14 fixed!

dipeshmsft:
@miloush Thanks for your continuous efforts in fixing issues via PRs and triaging.

PR #7730, opened 2023-04-17
Use TextAlignment for TextBox.GetCharacterIndexFromPoint

### LINKED ISSUE #7651: GetCharacterIndexFromPoint does not work as intended if HorizontalContentAlignment is set to Center
### Description

If the HorizontalContentAlignment of a TextBox is set to "Center" (or to "Right"), [GetCharacterIndexFromPoint](https://learn.microsoft.com/en-us/dotnet/api/system.windows.controls.textbox.getcharacterindexfrompoint?view=windowsdesktop-7.0) does not return the correct character index corresponding to a given point and instead returns -1.

### Reproduction Steps

See the following demo:
[GetCharacterIndexFromPointBugDemo.zip](https://github.com/dotnet/wpf/files/11030050/GetCharacterIndexFromPointBugDemo.zip)

### Expected behavior

GetCharacterIndexFromPoint should return the correct character index from a given point even if the HorizontalContentAlignment is set to Center.

### Actual behavior

GetCharacterIndexFromPoint returns -1 if HorizontalContentAlignment is set to Center/Right

### Regression?

_No response_

### Known Workarounds

_No response_

### Impact

_No response_

### Configuration

.NET 6/7
Windows 10 x64

### Other information

_No response_

miloush:
Fixes #7651 

## Description

When text box (view) uses any text alignment different from left, `TextBox.GetCharacterIndexFromPoint` does not work: wherever the line bounds do not intersect with what would be a left-aligned line bounds, it returns -1.

The PR fixes the bug by taking into account text alignment when deciding whether a point is on a line or not.

## Customer Impact

Customers not taking this fix cannot use `TextBox.GetCharacterIndexFromPoint` for text aligned to the right or centered.

## Regression

No.

## Testing

Built and tested with the repro in #7651, ensuring the characters do not get returned before the fix but correct characters get returned after the fix (using .NET 8.0.0-preview.3.23174.8).

## Risk

Low. Existing working behavior is not changed (the `GetContentOffset` returns 0). Slight performance hit for the extra calls and calculations, but only once - in the specific case of when `snapToText` is disabled and the relevant line is being tested.

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/PresentationFramework/MS/Internal/documents/TextBoxView.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationFramework/MS/Internal/documents/TextBoxView.cs
@@ -1680,10 +1680,13 @@ namespace System.Windows.Controls
                 }
                 else
                 {
-                    if (!snapToText &&
-                        (point.X < 0 || point.X >= record.Width))
+                    if (!snapToText)
                     {
-                        index = -1;
+                        double alignmentOffset = GetContentOffset(record.Width, CalculatedTextAlignment);
+                        if (point.X < alignmentOffset || point.X >= record.Width + alignmentOffset)
+                        {
+                            index = -1;
+                        }
                     }
                     break;
                 }


rampaa:
@miloush thank you very much for this PR! Will this PR be backported to .NET 6, or will it be serviced only through .NET 8?

dipeshmsft:
@miloush, will there be a difference in the case of RTL?

dipeshmsft:
@miloush, thanks for your quick response on the issue. 
Regarding the previous comment, yes I understand that alignment does not know about the direction, however, I just wanted to be sure about other scenarios.

rampaa:
@dipeshmsft will this fix be backported to .NET 6?

PR #7734, opened 2023-04-17
Avoid crashing if temporary file cannot be deleted

### LINKED ISSUE #3138: No permission to delete temporary files leads to crash on load
* .NET Core Version: 3.1.400-preview-015151
* Windows version: 10.0.19619.1000

 **Problem description:**

When WPF app does not have delete permission to the temp directory, loading XAML with Cursor property set crashes the app on load.
 
 **Actual behavior:** 

```
System.Windows.Markup.XamlParseException: Provide value on 'System.Windows.Baml2006.TypeConverterMarkupExtension' threw an exception. ---> System.UnauthorizedAccessException: Access to the path 'C:\Users\User\AppData\Local\Temp\WPF\o4fqaecx.33n' is denied.
   at System.IO.__Error.WinIOError(Int32 errorCode, String maybeFullPath)
   at System.IO.File.InternalDelete(String path, Boolean checkHost)
   at System.IO.FileHelper.DeleteTemporaryFile(String filePath)
   at System.Windows.Input.Cursor.LoadFromStream(Stream cursorStream)
   at System.Windows.Input.CursorConverter.ConvertFrom(ITypeDescriptorContext context, CultureInfo culture, Object value)
   at MS.Internal.Xaml.Runtime.ClrObjectRuntime.CallProvideValue(MarkupExtension me, IServiceProvider serviceProvider)
   at System.Windows.Markup.XamlReader.RewrapException(Exception e, IXamlLineInfo lineInfo, Uri baseUri)
   at System.Windows.Markup.WpfXamlLoader.Load(XamlReader xamlReader, IXamlObjectWriterFactory writerFactory, Boolean skipJournaledProperties, Object rootObject, XamlObjectWriterSettings settings, Uri baseUri)
   at System.Windows.ResourceDictionary.CreateObject(KeyRecord key)
   at System.Windows.ResourceDictionary.OnGettingValue(Object key, Object& value, Boolean& canCache)
   at System.Windows.ResourceDictionary.OnGettingValuePrivate(Object key, Object& value, Boolean& canCache)
   at System.Windows.ResourceDictionary.GetValueWithoutLock(Object key, Boolean& canCache)
   at System.Windows.ResourceDictionary.GetValue(Object key, Boolean& canCache)
   at System.Windows.SystemResources.LookupResourceInDictionary(ResourceDictionary dictionary, Object key, Boolean allowDeferredResourceReference, Boolean mustReturnDeferredResourceReference, Boolean& canCache)
   at System.Windows.SystemResources.FindDictionaryResource(Object key, Type typeKey, ResourceKey resourceKey, Boolean isTraceEnabled, Boolean allowDeferredResourceReference, Boolean mustReturnDeferredResourceReference, Boolean& canCache)
   at System.Windows.SystemResources.FindResourceInternal(Object key, Boolean allowDeferredResourceReference, Boolean mustReturnDeferredResourceReference)
   at System.Windows.StyleHelper.GetThemeStyle(FrameworkElement fe, FrameworkContentElement fce)
   at System.Windows.FrameworkElement.UpdateThemeStyleProperty()
   at System.Windows.FrameworkElement.OnInitialized(EventArgs e)
   at System.Windows.FrameworkElement.TryFireInitialized()
   at System.Windows.FrameworkElement.OnVisualParentChanged(DependencyObject oldParent)
   at System.Windows.Media.Visual.FireOnVisualParentChanged(DependencyObject oldParent)
   at System.Windows.Media.Visual.AddVisualChild(Visual child)
   at System.Windows.Media.VisualCollection.ConnectChild(Int32 index, Visual value)
   at System.Windows.Media.VisualCollection.Add(Visual visual)
   at System.Windows.Controls.UIElementCollection.AddInternal(UIElement element)
   at System.Windows.Controls.Panel.GenerateChildren()
   at System.Windows.Controls.Panel.EnsureGenerator()
   at System.Windows.Controls.Panel.get_InternalChildren()
   at System.Windows.Controls.StackPanel.StackMeasureHelper(IStackMeasure measureElement, IStackMeasureScrollData scrollData, Size constraint)
 ...
   at System.Windows.UIElement.Measure(Size availableSize)
   at System.Windows.Interop.HwndSource.SetLayoutSize()
   at System.Windows.Interop.HwndSource.set_RootVisualInternal(Visual value)
   at System.Windows.Window.SetRootVisual()
   at System.Windows.Window.SetRootVisualAndUpdateSTC()
   at System.Windows.Window.SetupInitialState(Double requestedTop, Double requestedLeft, Double requestedWidth, Double requestedHeight)
   at System.Windows.Window.CreateSourceWindow(Boolean duringShow)
```
 
 **Expected behavior:**

Leave the file behind with no crash, like in the case when it is being used by another process, i.e. `System.IO.FileHelper.DeleteTemporaryFile` should not only catch `IOException` but also `UnauthorizedAccessException`.

miloush:
Fixes #3138 

## Description

When a process does not have rights to delete a temporary file, it crashes, see #3138. The code already contains a catch for the case when the file is in use, so this PR adds another catch for when there are insufficient permissions for the deletion.

## Customer Impact

Customers not taking this fix may experience crashes if they do not have permission to delete temporary files. An example of codepath that uses temporary files is `Cursor.LoadFromStream` or WinRT spell checker.

This affects Visual Studio as well.

## Regression

WinRT spell checker codepath has a regression from .NET Framework in that it used to check for file permissions and skip the delete if they were not granted. This PR restores the .NET Framework behavior in this case.

## Testing

Built using 8.0.0-preview.3.23174.8 and verified that a WPF application crashes when denied access to the temporary folder before the fix but not after the fix.

## Risk

Low. Temporary files will already stay behind in the case of crash, and failing to delete temporary files should not make an application to crash.

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/Shared/System/IO/FileHelper.cs
+++ b/src/Microsoft.DotNet.Wpf/src/Shared/System/IO/FileHelper.cs
@@ -168,6 +168,11 @@ namespace System.IO
                     // We may not be able to delete the file if it's being used by some other process (e.g. Anti-virus check).
                     // There's nothing we can do in that case, so just eat the exception and leave the file behind
                 }
+                catch(System.UnauthorizedAccessException)
+                {
+                    // We may not be able to delete the file if we do not have rights to delete from the file folder.
+                    // There's nothing we can do in that case, so just eat the exception and leave the file behind
+                }
             }
         }
         #pragma warning restore 56502


src/Microsoft.DotNet.Wpf/src/Shared/System/IO/FileHelper.cs
```diff
@@ -168,6 +168,11 @@ static internal void DeleteTemporaryFile(string filePath)
                     // We may not be able to delete the file if it's being used by some other process (e.g. Anti-virus check).
                     // There's nothing we can do in that case, so just eat the exception and leave the file behind
                 }
+                catch(System.UnauthorizedAccessException)
```
Kuldeep-MS, line 171: Adding a new catch here can possibly change the code flow at https://github.com/dotnet/wpf/blob/main/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Documents/Speller.cs#L380. 

Although I don't have an example to substantiate my claim, but I would suggest checking this code flow for its expected behavior with these changes.

miloush: Yes that is correct, mentioned in the Customer Impact. However, if you look into the .NET Framework code, it first demands `FileIOPermissionAccess.AllAccess`. If that fails, the `SecurityException` is silently caught and the execution continues.

Removing this check in .NET Core possibly introduced a regression in that not having rights to delete the file now crashes the process. This PR fixes would fix such regression. I will update the PR text accordingly. Thanks!

Kuldeep-MS: Ok... I think in this case we can modify `UnloadDictionary` function to behave similar like in .NET Framework so that the code flow at https://github.com/dotnet/wpf/blob/main/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Documents/Speller.cs#L380 remains as it is.

Let me know if I am missing any point here or my suggestion won't work for you.

miloush: Sorry I am not quite sure I understand - what change to the `UnloadDictionary` are you proposing?

Kuldeep-MS: I am suggesting to make `UnloadDictionary` function same as it is in .NET Farmework i.e. wrapping the `FileHelper.DeleteTemporaryFile(filePath);` inside a try-catch block and suppressing the exception there instead of suppressing the exception here in `DeleteTemporaryFile`

miloush: That would only fix the case for the dictionary though, the point is to avoid crashes when any temporary file cannot be deleted (and the motivation was crash due to cursor files). What is the issue with catching it inside `DeleteTemporaryFile`?

rchauhan18:
@miloush, thanks for your contribution

PR #7743, opened 2023-04-20
Allow right click in system menu

### LINKED ISSUE #7739: Right clicking on a system menu item (Move, Size, Minimize ....) does not work for a window with custom WindowChrome
### Description

Here is the repo - a simple illustration of the issue https://github.com/nesakobi/window_chrome_wpf
There are 2 windows - one of them is the default Window without any styling and property changes. When the right mouse button is clicked on title bar of the window, the system menu becomes active and it is possible to click on menu items both left and right mouse clicks. Both types of clicks work in the same way. The second is a window with property "WindowStyle = None" and WindowChrome configured. Right clicking at the top of the window activates the system menu. Left-clicking on a menu item performs an action, while right clicking on an item does nothing. When there is a DataGrid and DataGrid has context menu,  right-clicking on the system menu items activates the DataGrid context menu.

context_menu

### Reproduction Steps

1. Create a style for window
2. Set custom chrome and set the property WindowStyle = None
3. Try opening the system menu by clicking on the window icon
4. Right-click any menu item - Move, Size, Minimize, etc...

### Expected behavior

System menu - the right clicking on menu item should work the same as for window with WindowStyle = SingleBorderWindow

### Actual behavior

System menu - menu item does not work on right click.

Also it activates PreviewMouseRightButtonUp and MouseRightButtonUp for controls inside the window content (annoying behavior)

### Regression?

_No response_

### Known Workarounds

_No response_

### Impact

_No response_

### Configuration

Tried both with .Net framework 4.8.1 and .Net 7.0
OS: Microsoft Windows [Version 10.0.19044.2846]

### Other information

_No response_

miloush:
Fixes #7739 

## Description

The `SystemCommands.ShowSystemMenu` command shows the system menu in a way that it does not accept right click, although the menu does accept right click when shown by the Windows system.

This PR updates the flag passed to the native API to allow right click to work.

## Customer Impact

Customers not taking the fix will not be able to use right click in system menu shown explicitly, as is for example the case for `WindowChrome`, introducing inconsistencies between the WPF and Windows system.

## Regression

No.

## Testing

Built using .NET 8.0.100-preview.3.23178.7 and tested with the repro in #7739, verified that the _Minimize_ command gets ignored before the fix and works after the fix.

## Risk

Low. Existing behavior not changed. Allowed right click to not go through the system menu when invoked programmatically.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/wpf/pull/7743)

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/SystemCommands.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/SystemCommands.cs
@@ -76,6 +76,7 @@ namespace System.Windows
         {
             const uint TPM_RETURNCMD = 0x0100;
             const uint TPM_LEFTBUTTON = 0x0;
+            const uint TPM_RIGHTBUTTON = 0x2;
 
             Verify.IsNotNull(window, "window");
             IntPtr hwnd = new WindowInteropHelper(window).Handle;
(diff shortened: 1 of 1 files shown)


dipeshmsft:
Thanks @miloush for fixing this issue.

PR #7833, opened 2023-05-17
Replace String.CompareOrdinal to string.Equals Part 3

lindexi:
## Description

We can use `string.Equals(string1, string2, StringComparison.Ordinal)` to check for equality, see https://learn.microsoft.com/en-us/dotnet/csharp/how-to/compare-strings and https://learn.microsoft.com/en-us/dotnet/standard/base-types/best-practices-strings

Reference: https://github.com/dotnet/wpf/issues/7834

## Customer Impact

## Regression

None.

## Testing

Just CI.

## Risk

Low.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/wpf/pull/7833)


dipeshmsft:
Thank you @lindexi for your contributions.

lindexi:
Thank you @dipeshmsft

PR #7836, opened 2023-05-18
Modernize the empty string checking in ContentType

lindexi:
## Description

Modernize the empty string checking in ContentType.cs

## Customer Impact

None.

## Regression

None.

## Testing

Just CI.

## Risk

Low.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/wpf/pull/7836)

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/WindowsBase/MS/Internal/ContentType.cs
+++ b/src/Microsoft.DotNet.Wpf/src/WindowsBase/MS/Internal/ContentType.cs
@@ -85,8 +85,8 @@ namespace MS.Internal
         {
             if (contentType == null)
                 throw new ArgumentNullException("contentType");
-                       
-            if (String.CompareOrdinal(contentType, String.Empty) == 0)
+
+            if (string.IsNullOrEmpty(contentType))
             {
                 _contentType = String.Empty;             
             }
@@ -278,8 +278,8 @@ namespace MS.Internal
                 if (!_isInitialized)
                     return String.Empty;
 
-                Debug.Assert(String.CompareOrdinal(_type, String.Empty) != 0
-                   || String.CompareOrdinal(_subType, String.Empty) != 0);
+                Debug.Assert(!string.IsNullOrEmpty(_type)
+                             || !string.IsNullOrEmpty(_subType));
             
                 StringBuilder stringBuilder = new StringBuilder(_type);
                 stringBuilder.Append('/');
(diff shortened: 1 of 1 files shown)


src/Microsoft.DotNet.Wpf/src/WindowsBase/MS/Internal/ContentType.cs
```diff
@@ -85,8 +85,8 @@ internal ContentType(string contentType)
         {
             if (contentType == null)
                 throw new ArgumentNullException("contentType");
-                       
-            if (String.CompareOrdinal(contentType, String.Empty) == 0)
+
+            if (string.IsNullOrEmpty(contentType))
```
kronic, line 89: contentType.Length == 0

lindexi: Thank you @kronic and I fixed it.

src/Microsoft.DotNet.Wpf/src/WindowsBase/MS/Internal/ContentType.cs
```diff
@@ -278,8 +278,8 @@ public override string ToString()
                 if (!_isInitialized)
                     return String.Empty;
 
-                Debug.Assert(String.CompareOrdinal(_type, String.Empty) != 0
-                   || String.CompareOrdinal(_subType, String.Empty) != 0);
+                Debug.Assert(!string.IsNullOrEmpty(_type)
+                             || !string.IsNullOrEmpty(_subType));
```
dipeshmsft, line 282: @lindexi, after going through the code, it looks like _type will never be null, so how about using `_type.Length == 0` and `_subType.Length == 0` here ?

What are your thoughts about this ?

lindexi: @dipeshmsft I think the code on debugging can be more flexible. When I wrote this code, I didn’t consider this at the time. Thank you for pointing it out.

lindexi: @dipeshmsft I think that Debug.Assert code is used to intercept unexpected logic during the development process. Since the judgment of the _type field is not null does not occur in the ToString method, in other words, it is possible that developers will break the convention of _type field not being null in subsequent development processes. Therefore, I think that retaining the use of the string.IsNullOrEmpty method can better intercept problems during the development process.

dipeshmsft:
Thanks @lindexi, looking forward to more contributions from you.

PR #7910, opened 2023-06-09
Fix InputEventArgs.Timestamp field

### LINKED ISSUE #7887: InputEventArgs.Timestamp field is static and shared by all instances
### Description

`InputEventArgs.Timestamp` field description is "Gets the time when this event occurred": https://learn.microsoft.com/en-us/dotnet/api/system.windows.input.inputeventargs.timestamp?view=windowsdesktop-8.0

An `InputEventArgs` is created using the constructor `InputEventArgs(InputDevice device, int timestamp)`:
https://github.com/dotnet/wpf/blob/32d08ffa22a0e42730e14efd27d4b2424c401d5c/src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Input/InputEventArgs.cs#L30-L36

The `InputEventArgs.Timestamp` property is implemented as an accessor to an underlying field
https://github.com/dotnet/wpf/blob/32d08ffa22a0e42730e14efd27d4b2424c401d5c/src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Input/InputEventArgs.cs#L51-L54

However, the field `_timestamp` is `static`: https://github.com/dotnet/wpf/blob/32d08ffa22a0e42730e14efd27d4b2424c401d5c/src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Input/InputEventArgs.cs#L74

This means that it is shared between Timestamp instances. As in the test below, when you create a new `InputEventArgs`, it overrides the `Timestamp` property of all other previously created instances.

This can't be correct behaviour. The fix would be to make the field non `static`.

I can submit a fix if approved.

### Reproduction Steps

Consider the following test:
```cs
[StaFact]
public void Timestamp_GetDifferentInstances_ReturnsExpected()
{
    var e1 = new InputEventArgs(Mouse.PrimaryDevice, 0);
    Assert.Equal(0, e1.Timestamp);

    var e2 = new InputEventArgs(Mouse.PrimaryDevice, 1);
    Assert.Equal(0, e1.Timestamp); // !FAILS HERE!
    Assert.Equal(1, e2.Timestamp);
}
```

### Expected behavior

The test passes. `e1.Timestamp != e2.Timestamp`

### Actual behavior

The test fails:
```
  Failed System.Windows.Input.Tests.InputEventArgsTests.Timestamp_GetDifferentInstances_ReturnsExpected [7 ms]
  Error Message:
   Assert.Equal() Failure
Expected: 0
Actual:   1
     at System.Windows.Input.Tests.InputEventArgsTests.Timestamp_GetDifferentInstances_ReturnsExpected() in InputEventArgsTests.cs:line 33
```

### Regression?

_No response_

### Known Workarounds

_No response_

### Impact

_No response_

### Configuration

_No response_

### Other information

_No response_

hughbe:
Fixes #7887

`InputEventArgs.Timestamp` field should not be static.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/wpf/pull/7910)

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Input/InputEventArgs.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Input/InputEventArgs.cs
@@ -71,7 +71,7 @@ namespace System.Windows.Input
         }
 
         private InputDevice _inputDevice;
-        private static int _timestamp;
+        private int _timestamp;
      }
 }
 


lindexi (approved):
Looks good to me

miloush:
I would think it is a good practice to keep the changes to minimum. This could be a PR of one word deleted, but it is full of changes unrelated to the issue. It renames private fields which was not discussed. Not sure why comments need to be deleted.

pchaurasia14:
@hughbe - Could you please address the review comments mentioned by @miloush. 

We intend to include this change in our upcoming test pass.

hughbe:
Have updated this PR, apologies for the delay

pchaurasia14 (approved):

dipeshmsft:
@hughbe thanks for bringing the issue and for providing the fix as well. Thanks for your continued support to WPF.

PR #7929, opened 2023-06-15
CsWin32 usage sample

JeremyKuhne:
DO NOT MERGE.
(I am currently working on an updated PR for this that will share some implementation with WinForms.)

This is a draft to drive the conversation around utilizing CsWin32 in WPF. In it I've converted SplashScreen to use CsWin32 and directly utilize WIC pointers.

This is informed by WinForms transition and utilizes some of the code, notably ComScope and IID. There is more support code that WinForms uses that eases handling of COM situations, including providing ComWrapper based "CCWs" for managed classes.

WinForms also contains other patterns for scoping handles, handling GC keep-alive scenarios, and managing agiile COM pointers and VARIANT interop.

As CsWin32 types Win32 typedefs (such as HWND, and HRESULT) it can get very awkward to exchange these if you use CsWin32 directly in multiple assemblies in the assembly graph. As such I created a new shared assembly for the interop code.

I added a batch file to simplify Visual Studio in the right context (copied from WinForms).

Choices here were expedient and made with the goal of illustrating how to do this. I have not tested anything outside of ensuring compilation works.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/wpf/pull/7929)


JeremyKuhne:
If the interop assembly can be built for x86, x64, and arm64 that avoids an AnyCpu issue with CsWin32 that we're working on. I don't really have a handle on how things are packaged/deployed so I don't know how easy that is to accomplish.

The basic issue is that some Win32 APIs are not technically AnyCPU friendly. More specifically this has to do with some struct definitions. The fields in the struct lay out the same way on x86/x64, but packing would cause the structs to not be aligned correctly when contained in other structs. In internal usage this doesn't really come up- we've always been able to write PInvokes in WinForms for AnyCPU. There is an ask on CsWin32 to be able to make an "unsafe" ask for a given platform projection for these cases.

None of this hits this initial PR. The notable place this comes up is in some of the shell APIs which define with a PACK=1.

NuGet.config
```diff
@@ -10,6 +10,9 @@
     <add key="dotnet7-transport" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet7-transport/nuget/v3/index.json" />
     <add key="dotnet8" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet8/nuget/v3/index.json" />
     <add key="dotnet8-transport" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet8-transport/nuget/v3/index.json" />
+    <!-- CsWin32 dailies -->
+    <add key="winsdk" value="https://pkgs.dev.azure.com/azure-public/winsdk/_packaging/CI/nuget/v3/index.json" />
+    <add key="nuget.org" value="https://api.nuget.org/v3/index.json" />
```
AArnott, line 15: I'm guessing nuget.org here isn't going to fly.

AArnott (commented):
Love it.

NuGet.config
JeremyKuhne: Looking into getting Microsoft.Windows.WDK.Win32Metadata added to the right feeds.

src/Microsoft.DotNet.Wpf/src/System.Windows.Primitives/System.Windows.Primitives.csproj
```diff
@@ -0,0 +1,35 @@
...
+      We align casing and naming with Win32 API. As such some types have all lower case names, which
+      in theory may conflict with new C# keywords in the future. Our types here are internal so end
+      users won't be impacted. If some name becomes difficult to adapt to with the @ symbol we can
+      cross that bridge when we hit it (if ever).
+    -->
+    <NoWarn>$(NoWarn);CS8981;CS3016</NoWarn>
+    <!--
+      We don't care about CLS compliance since everything here is internal and we want to match native types.
+    -->
+    <NoWarn>$(NoWarn);CS3016</NoWarn>
```
sharwell, line 20: 💡 Applying `[assembly: CLSCompliant(false)]` will disable this warning _and_ disable CLS compliance analysis which significantly improves compiler performance.

src/Microsoft.DotNet.Wpf/src/System.Windows.Primitives/System.Windows.Primitives.csproj
```diff
@@ -0,0 +1,35 @@
...
+    <ProduceReferenceAssembly>true</ProduceReferenceAssembly>
+    <UsePublicApiAnalyzers>true</UsePublicApiAnalyzers>
+    <RootNamespace />
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Microsoft.Windows.CsWin32" Version="0.3.5-beta">
+      <PrivateAssets>all</PrivateAssets>
+      <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
+    </PackageReference>
```
sharwell, line 31:
```suggestion
    <PackageReference Include="Microsoft.Windows.CsWin32" Version="0.3.5-beta" PrivateAssets="all" />
```

start-vs.cmd
```diff
@@ -0,0 +1,44 @@
+@echo off
+setlocal enabledelayedexpansion
+
+:: This command launches a Visual Studio solution with environment variables required to use a local version of the .NET Core SDK.
```
sharwell, line 4: 💡 I recommend removing this file. If Visual Studio does not support opening this solution directly, the solution should be corrected to comply with Visual Studio's practices.

sharwell: This workaround misleads developers working on internal versions of Visual Studio into thinking the product is working. We _need_ this to be an acute pain point until such time as we deliver a working product.

src/Microsoft.DotNet.Wpf/src/WindowsBase/System/Windows/SplashScreen.cs
```diff
@@ -62,25 +62,20 @@ public void Show(bool autoClose)
...
                     {
                         umemStream.Seek(0, SeekOrigin.Begin); // ensure stream position
-                        IntPtr pImageSrcBuffer;
-                        unsafe
-                        {
-                            pImageSrcBuffer = new IntPtr(umemStream.PositionPointer);
-                        }
 
-                        if (CreateLayeredWindowFromImgBuffer(pImageSrcBuffer, umemStream.Length, topMost) && autoClose == true)
+                        if (CreateLayeredWindowFromImgBuffer(new(umemStream.PositionPointer, (int)umemStream.Length), topMost) && autoClose == true)
```
sharwell, line 78: 💡 Explicit type would help here

src/Microsoft.DotNet.Wpf/src/WindowsBase/System/Windows/SplashScreen.cs
```diff
@@ -285,144 +293,145 @@ private void Fadeout_Tick(object unused, EventArgs args)
             }
         }
 
-        private void DestroyResources()
+        ~SplashScreen() => DestroyResources(finalizer: true);
```
sharwell, line 296: ❔ Can we wrap unmanaged resources in the correct handle type and remove this finalizer?

src/Microsoft.DotNet.Wpf/src/WindowsBase/System/Windows/SplashScreen.cs
```diff
@@ -285,144 +293,145 @@ private void Fadeout_Tick(object unused, EventArgs args)
...
+
+        private unsafe void DestroyResources(bool finalizer = false)
```
sharwell, line 298: 📝 this parameter appears to be inverted relative to standard practices.

src/Microsoft.DotNet.Wpf/src/System.Windows.Primitives/Windows/Win32/Foundation/ComScope.cs
```diff
@@ -0,0 +1,102 @@
...
+///   This struct has implicit conversions to T** and void** so it can be passed directly to out methods.
+///   For example:
+///  </para>
+///  <code>
+///   using ComScope&lt;IUnknown&gt; unknown = new(null);
+///   comObject-&gt;QueryInterface(&amp;iid, unknown);
+///  </code>
+///  <para>
+///   Take care to NOT make copies of the struct to avoid accidental over-release.
+///  </para>
```
sharwell, line 25: 💡 Roslyn's `[NonCopyable]` can be used for this

src/Microsoft.DotNet.Wpf/src/System.Windows.Primitives/Windows/Win32/Foundation/ComScope.cs
```diff
@@ -0,0 +1,102 @@
...
+/// </remarks>
+/// <typeparam name="T">
+///  This should be one of the struct COM definitions as generated by CsWin32. Ideally we'd constrain to
+///  <see cref="IUnknown.Interface"/> or some other interface tag to enforce that this is being used around
+///  a struct that is actually a COM wrapper.
+/// </typeparam>
+internal readonly unsafe ref struct ComScope<T> where T : unmanaged, IComIID
```
sharwell, line 32: 💡 This appears to be using `ref struct` where the intent is closer to `[NonCopyable]`. I recommend changing this to just be `struct` and applying `[NonCopyable]` along with the corresponding analyzer.

src/Microsoft.DotNet.Wpf/src/System.Windows.Primitives/Windows/Win32/Foundation/ComScope.cs
```diff
@@ -0,0 +1,102 @@
...
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System.Runtime.CompilerServices;
+using Windows.Win32.System.Com;
+
+namespace Windows.Win32.Foundation;
+
+/// <summary>
+///  Lifetime management struct for a native COM pointer. Meant to be utilized in a <see langword="using"/> statement
```
sharwell, line 11: 💡 Consider adding an analyzer for this type which enforces that instances only appear in `using` statements.

src/Microsoft.DotNet.Wpf/src/System.Windows.Primitives/Windows/Win32/Foundation/ComScope.cs
```diff
@@ -0,0 +1,102 @@
...
+///   comObject-&gt;QueryInterface(&amp;iid, unknown);
+///  </code>
+///  <para>
+///   Take care to NOT make copies of the struct to avoid accidental over-release.
+///  </para>
+/// </remarks>
+/// <typeparam name="T">
+///  This should be one of the struct COM definitions as generated by CsWin32. Ideally we'd constrain to
+///  <see cref="IUnknown.Interface"/> or some other interface tag to enforce that this is being used around
+///  a struct that is actually a COM wrapper.
```
sharwell, line 30: 💡 An analyzer could probably enforce this as well

src/Microsoft.DotNet.Wpf/src/System.Windows.Primitives/Windows/Win32/Foundation/ComScope.cs
```diff
@@ -0,0 +1,102 @@
...
+        return scope;
+    }
+
+    public void Dispose()
+    {
+        IUnknown* unknown = (IUnknown*)_value;
+
+        // Really want this to be null after disposal to avoid double releases, but we also want
+        // to maintain the readonly state of the struct to allow passing as `in` without creating implicit
+        // copies (which would break the T** and void** operators).
```
sharwell, line 95: > but we also want to maintain the readonly state of the struct to allow passing as `in` without creating implicit copies

😕 `in` will still work without making implicit copies, even if the type is not readonly. The defensive copy for a parameter passed as `in` would occur at the time a non-readonly member is accessed (of which this type does not have any), except for `Dispose()` which would not be allowed.

src/Microsoft.DotNet.Wpf/src/System.Windows.Primitives/Windows/Win32/Foundation/ComScope.cs
AArnott: Can you provide more information on where this attribute is defined and which analyzer package needs to be installed to leverage it?

start-vs.cmd
JeremyKuhne: @sharwell if you know how to get VS to use the local .NET copy please share. If not, and you think that VS should support this scenario, then please open an issue and I'll be happy to chime in.

Making WinForms/WPF devs daily life difficult is _not_ the answer to making VS better.

src/Microsoft.DotNet.Wpf/src/WindowsBase/System/Windows/SplashScreen.cs
JeremyKuhne: I don't think it has significant value here, but if the WPF team does that's fine.

src/Microsoft.DotNet.Wpf/src/WindowsBase/System/Windows/SplashScreen.cs
JeremyKuhne: Deliberately not doing this. There is a non-trivial cost to using SafeHandle that will have measurable impact on WPF. In most cases things will be scoped to the stack so taking the hit of SafeHandle for API definitions is counterproductive in an interop heavy environment.

src/Microsoft.DotNet.Wpf/src/WindowsBase/System/Windows/SplashScreen.cs
JeremyKuhne: The standard practice is to have a dispose. This shipped so that would require an API change. I deliberately was trying to demonstrate how to use CsWin32 the way WinForms does so that we can align for better productivity. I tried to avoid other refactoring to make the code cleaner.

src/Microsoft.DotNet.Wpf/src/System.Windows.Primitives/Windows/Win32/Foundation/ComScope.cs
JeremyKuhne: I'll look into it, thanks.

src/Microsoft.DotNet.Wpf/src/System.Windows.Primitives/Windows/Win32/Foundation/ComScope.cs
JeremyKuhne: This is intended to be stack bound so I don't see the value in making this non-ref. There is another pattern for things that aren't that I have not shown yet (`AgileComPointer`).

src/Microsoft.DotNet.Wpf/src/System.Windows.Primitives/Windows/Win32/Foundation/ComScope.cs
JeremyKuhne: Would love to, just need time to.

src/Microsoft.DotNet.Wpf/src/System.Windows.Primitives/Windows/Win32/Foundation/ComScope.cs
JeremyKuhne: Again, need cycles. :)

src/Microsoft.DotNet.Wpf/src/System.Windows.Primitives/Windows/Win32/Foundation/ComScope.cs
JeremyKuhne: We're deliberately clearing `_value` here which is very much readonly.

JeremyKuhne:
Replaced by #10047. Closing this one.

Issue #7834, opened 2023-05-17
Replace String.CompareOrdinal to string.Equals


lindexi:
We can use `string.Equals(string1, string2, StringComparison.Ordinal)` to check for equality, see https://learn.microsoft.com/en-us/dotnet/csharp/how-to/compare-strings and https://learn.microsoft.com/en-us/dotnet/standard/base-types/best-practices-strings

> Use the [String.Compare](https://learn.microsoft.com/en-us/dotnet/api/system.string.compare) and [String.CompareTo](https://learn.microsoft.com/en-us/dotnet/api/system.string.compareto) methods to sort strings, not to check for equality.

- https://github.com/dotnet/wpf/pull/7500
- https://github.com/dotnet/wpf/pull/7680
- https://github.com/dotnet/wpf/pull/7833
- https://github.com/dotnet/wpf/pull/8042

lindexi:
Finish

PR #8042, opened 2023-07-21
Replace String.CompareOrdinal to string.Equals Final Part

lindexi:
## Description

We can use string.Equals(string1, string2, StringComparison.Ordinal) to check for equality, see https://learn.microsoft.com/en-us/dotnet/csharp/how-to/compare-strings and https://learn.microsoft.com/en-us/dotnet/standard/base-types/best-practices-strings

Reference: https://github.com/dotnet/wpf/issues/7834

## Customer Impact

## Regression

None.

## Testing

Just CI.

## Risk

Low.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/wpf/pull/8042)

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/WindowsBase/System/Security/RightsManagement/User.cs
+++ b/src/Microsoft.DotNet.Wpf/src/WindowsBase/System/Security/RightsManagement/User.cs
@@ -135,7 +135,7 @@ namespace System.Security.RightsManagement
 
             ContentUser userObj = (ContentUser)obj;
 
-            return (string.Equals(_name.ToUpperInvariant(), userObj._name.ToUpperInvariant(), StringComparison.Ordinal))
+            return (string.Equals(_name, userObj._name, StringComparison.OrdinalIgnoreCase))
                         &&
                             _authenticationType.Equals(userObj._authenticationType);
         }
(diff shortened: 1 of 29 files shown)


src/Microsoft.DotNet.Wpf/src/WindowsBase/System/Security/RightsManagement/User.cs
```diff
@@ -135,7 +135,7 @@ public override bool Equals(object obj)
 
             ContentUser userObj = (ContentUser)obj;
 
-            return (String.CompareOrdinal(_name.ToUpperInvariant(), userObj._name.ToUpperInvariant()) == 0)
```
rchauhan18, line 138: @lindexi Did you intentionally left "ToUpperInvariant()" here?
If no, then we can update it.

lindexi: Thank you @rchauhan18 and I will fix it.

rchauhan18:
@lindexi, thanks for your contribution

lindexi:
Thank you @rchauhan18

PR #8151, opened 2023-08-29
Fixing exception thrown in GenerateTemporaryTargetAssembly when project name has GB characters in it

singhashish-wpf:
Fixes 
 Exception thrown in GenerateTemporaryTargetAssembly when project name has GB characters in it
Main PR 
None
## Description
Bugfix for GB chars, exception thrown when using Load(CurrentProject), when project name has GB characters in it.
Using a filestream instead of using string path to avoid the need to properly compose Uri (which is another way of fixing - but more complicated).
           

## Customer Impact
Users will face build error, when building a project which has GB Characters in the project name.

## Regression
None

## Testing
Local

## Risk
Low

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/wpf/pull/8151)

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/PresentationBuildTasks/Microsoft/Build/Tasks/Windows/GenerateTemporaryTargetAssembly.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationBuildTasks/Microsoft/Build/Tasks/Windows/GenerateTemporaryTargetAssembly.cs
@@ -122,8 +122,12 @@ namespace Microsoft.Build.Tasks.Windows
                 XmlDocument xmlProjectDoc = null;
 
                 xmlProjectDoc = new XmlDocument( );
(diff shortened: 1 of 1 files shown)


singhashish-wpf:
Sample name for project, that can repro the exception.
CoreWPF_𬴂鿗龵苘爫麤υ㏑䡫.csproj

rainersigwald (approved):

miloush:
Please note that using path/url when loading a XML document uses the URL as a base URL for resolving and reading DTDs referenced from the XML file. You lose that by switching to a stream, potentially changing the meaning of the XML content.

That said, I do not expect any csproj files to be using DTD so we are probably OK here.

Kuldeep-MS (approved):
Changes Look good to me

PR #8174, opened 2023-09-01
Closing the bitmapstream for UnknownBitMapDecoder

### LINKED ISSUE #2881: BitmapDecoder remains locked after reading a non-supported file.
* .NET Core Version: 3.1.3 (3.1.201)
* Windows version: 10.0.18363.752
 
 **Problem description:**

Users cannot write or delete files while the application is running because BitmapDecoder will remain locked after reading a non-supported file.

If an exception is thrown during the creation of BitmapDecoder, the FileStream used internally does not seem to be released.
 
 **Actual behavior:** 

The file is locked and cannot be operated until the application is terminated.
 
 **Expected behavior:**

The file does not remain locked even if a non-supported file is given.
 
 **Minimal repro:**

```csharp
{
    public MainWindow()
    {
        InitializeComponent();
    }

    private void Window_Loaded(object sender, RoutedEventArgs e)
    {
        try
        {
            // Always throw `NotSupportedException`
            var decoder = BitmapDecoder.Create(new Uri(@"C:\Users\shibayan\Documents\dummy.md"), BitmapCreateOptions.DelayCreation, BitmapCacheOption.OnDemand);
        }
        catch (NotSupportedException ex)
        {
        }
    }
}
```

singhashish-wpf:
Fixes #2881

Main PR 
None
## Description
If an exception is thrown during the creation of BitmapDecoder, the FileStream used internally does not seem to be released.
BitmapStream wasn't closed and due to which the file wasn't being able to be modified or deleted.

## Customer Impact
The file will be locked for editing/deletion.

## Regression
None

## Testing
Local.

## Risk
Low. 

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/wpf/pull/8174)

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Media/Imaging/BitmapDecoder.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Media/Imaging/BitmapDecoder.cs
@@ -1202,6 +1202,7 @@ namespace System.Windows.Media.Imaging
             }
             catch
             {
+                bitmapStream.Close();
                 #pragma warning disable 6500
 
                 decoderHandle = null;
@@ -1212,7 +1213,6 @@ namespace System.Windows.Media.Imaging
             finally
             {
                 UnsafeNativeMethods.MILUnknown.ReleaseInterface(ref comStream);
-                bitmapStream.Close();
             }
 
             string decoderMimeTypes;
--- a/src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Media/Imaging/BitmapDecoder.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Media/Imaging/BitmapDecoder.cs
@@ -1212,6 +1212,7 @@ namespace System.Windows.Media.Imaging
             finally
             {
                 UnsafeNativeMethods.MILUnknown.ReleaseInterface(ref comStream);
+                bitmapStream.Close();
             }
 
             string decoderMimeTypes;


miloush:
https://github.com/dotnet/wpf/blob/28f8f455c01c4c51965d78617ec8ff66cf8f1aaf/src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Media/Imaging/BitmapDecoder.cs#L1104-L1108

https://github.com/dotnet/wpf/blob/28f8f455c01c4c51965d78617ec8ff66cf8f1aaf/src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Media/Imaging/BitmapDecoder.cs#L1115-L1116

If `GetSeekableStream` replaces the stream with a new one, the original one will still not be closed.

singhashish-wpf:
@miloush The original stream as I understand is bitmapStream which is being closed in the `finally` block. Could you please elaborate if you are referring to `stream` as original stream or some other stream?

miloush:
My point was that by the time you get to `finally`, there could be two streams, one is the `bitmapStream` and another one is `MemoryStream` created if `bitmapStream` turns out to be not seekable (inside the `GetSeekableStream`). You are correct that since you are closing the original `bitmapStream` and not `stream` which could be either of those, you are good.

However I am now slightly concerned about a different thing. The `bitmapStream` is stored in the `uriStream` out parameter:

https://github.com/dotnet/wpf/blob/56dd121f1f0dc59ad3621cf916c0230c50e53258/src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Media/Imaging/BitmapDecoder.cs#L1112

So unless I missed something, even if things go OK, you will be newly returning a disposed stream out of the method and that is not cool. `BitmapDecoder.CreateFromUriOrStream` passes it around to decoders. Maybe put it in `catch` instead?

Kuldeep-MS (approved):
Looks Good to me

PR #138, opened 2018-12-10
Add Windows.Markup tests

hughbe:
Bunch of tests for attributes, ValueSerializer, event args, converters etc in the System.Windows.Markup namespace

Assistance required: wrote these on my mac, they pass :) but not sure how to integrate into CI

Planning on writing more for the remaining stuff soon!

/cc @karelz

PR #8203, opened 2023-09-12
Add more System.Xaml tests

hughbe:
Had these lying around from 5 years ago. Can now send in after #138 

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/wpf/pull/8203)

### DIFF
--- a/src/Microsoft.DotNet.Wpf/tests/UnitTests/System.Xaml.Tests/Common/CustomAssembly.cs
+++ b/src/Microsoft.DotNet.Wpf/tests/UnitTests/System.Xaml.Tests/Common/CustomAssembly.cs
@@ -1,0 +1,53 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System.Collections.Generic;
+using System.Reflection;
+
+namespace System.Xaml.Tests.Common;
+
+public class CustomAssembly : Assembly
+{
+    protected Assembly DelegatingAssembly { get; }
+
+    public CustomAssembly(Assembly delegatingAssembly)
+    {
+        DelegatingAssembly = delegatingAssembly;
+    }
+
+    public Optional<string?> FullNameResult { get; set; }
+    public override string? FullName => FullNameResult.Or(DelegatingAssembly.FullName);
+
+    public Optional<bool> IsDynamicResult { get; set; }
+    public override bool IsDynamic => IsDynamicResult.Or(DelegatingAssembly.IsDynamic);
+
+    public Optional<bool> ReflectionOnlyResult { get; set; }
+    public override bool ReflectionOnly => ReflectionOnlyResult.Or(DelegatingAssembly.ReflectionOnly);
+
+    public Optional<IList<CustomAttributeData>> GetCustomAttributesDataResult { get; set; }
+    public override IList<CustomAttributeData> GetCustomAttributesData()
+    {
+        return GetCustomAttributesDataResult.Or(DelegatingAssembly.GetCustomAttributesData);
+    }
+
+    public Optional<Dictionary<Type, object?[]?>> GetCustomAttributesMap { get; set; }
+
+    public override object[] GetCustomAttributes(Type attributeType, bool inherit)
+    {
+        if (GetCustomAttributesMap.HasValue)
+        {
+            object?[]? result;
+            if (GetCustomAttributesMap.Value.TryGetValue(attributeType, out result))
+            {
+                return result!;
+            }
+        }
+
+        return DelegatingAssembly.GetCustomAttributes(attributeType, inherit);
+    }
+
+    public override bool Equals(object? o) => DelegatingAssembly.Equals(o);
+
+    public override int GetHashCode() => DelegatingAssembly.GetHashCode();
+}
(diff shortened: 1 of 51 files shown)


lindexi:
Awesome

src/Microsoft.DotNet.Wpf/tests/UnitTests/System.Xaml.Tests/System/Xaml/Permissions/XamlAccessLevelTests.cs
```diff
@@ -0,0 +1,152 @@
+#if NETFRAMEWORK
```
harshit7962, line 1: @hughbe Do we need to run these tests on .NET Core, since they are specifically designed for .NET Framework? Additionally, can we omit this flag altogether to run these tests on .NET Core?

harshit7962: @hughbe any comments?

harshit7962: @hughbe can we remove this test as of now to get the remaining one's added? We can get it unblocked later if needed.

hughbe: Sorry! Will remove tonight

hughbe:
Some tests failing due to the change to return `[]` instead of `null` here: https://github.com/dotnet/wpf/blob/main/src/Microsoft.DotNet.Wpf/src/System.Xaml/System/Xaml/Replacements/TypeListConverter.cs#L41

https://github.com/dotnet/wpf/pull/203#discussion_r1327064164

harshit7962:
@hughbe the `ConvertFrom` method should throw a `NotSupportedException` when the `value` is `null`. This was the case previously. Isn't the previous state, checking for the exception, the more correct one?

src/Microsoft.DotNet.Wpf/src/System.Xaml/System/Xaml/InfosetObjects/XamlNodes.cs
```diff
@@ -64,18 +64,9 @@ public XamlNode(XamlNodeType nodeType, object data)
 #if DEBUG
             switch(nodeType)
             {
+            case XamlNodeType.NamespaceDeclaration:
             case XamlNodeType.StartObject:
-                Debug.Assert(data is XamlType, "XamlNode ctor, StartObject data is not a XamlType");
-                break;
-
             case XamlNodeType.StartMember:
-                Debug.Assert(data is XamlMember, "XamlNode ctor, StartMember data is not a XamlMember");
```
rchauhan18, line 72: @hughbe Can we leave the ```Debug.Assert``` statement for now?
We do need them for local debugging. 
We will plan to remove them in future.

src/Microsoft.DotNet.Wpf/src/System.Xaml/System/Xaml/Schema/TypeReflector.cs
```diff
@@ -450,7 +450,6 @@ internal EventInfo LookupEvent(string name)
             out ICollection<EventInfo> newEvents, out List<XamlMember> knownMembers)
         {
             Debug.Assert(UnderlyingType != null, "Caller should check for UnderlyingType == null");
-            Debug.Assert(_nonAttachableMemberCache != null, "Members property should have been invoked before this");
```
rchauhan18, line 453: Same as above

hughbe: Same response as above.

rchauhan18 (approved):
LGTM.
@hughbe 
Thanks for your contribution!
Are you planning to add more test?
If not, I can continue to merge it.

hughbe:
I'm not planning to add any more tests :)

miloush:
Disclaimer: I haven't gone through the PR. However, I am worried about the asserts being removed, presumably they were added for a reason and express an expectation of the code that follows. Either the tests should be fixed to not hit them or if we think they are unreasonable, the code needs to be checked that it can deal with the conditions being broken.

ThomasGoulet73:
@hughbe It seems like the only remaining point of discussion in this PR is about the removal of asserts, would it be easy to extract the tests that requires the removal of asserts and open a separate PR with them ? If so, I think it would be a good idea to do so to be able to merge this PR as soon as possible and the changes in this PR would be less controversial since there would be no non-test changes.

hughbe:
Apologies for the delay. I have just fixed all the IDE errors and failing tests

dipeshmsft:
Thanks @hughbe, I was just getting started with using the NoAssertContext mentioned in the other pull request. Thanks for fixing this though.

dipeshmsft (approved):

dipeshmsft:
Thanks a lot @hughbe for the PR, and apologies for the delays from our side. I will move on to reviewing the other PR.

hughbe:
Thank you! Really glad this was merged in :)

PR #8259, opened 2023-09-28
Fix for tab grid error

### LINKED ISSUE #2662: TabControl bug with sorted DataGrid
* .NET Core Version: 3.1
* Windows version: Windows 10.0.17763

I think the TabControl has a bug-like behavior with sorted DataGrids.
If you have a DataTemplate that has DataGrids in it, and you sort a DataGrid column and try to edit a cell in that column, and then try to change tabs in the TabControl the program crashes with a framework exception that seems impossible to catch beforehand.

I don't believe that this is a desired behavior. The DataGrid should just stop the sorting when tabs are changed, I think the easiest solution is to just remove the throw in the source code.

I have created a minimal repro here:
https://github.com/Markz878/TabDataGridError

If anyone knows another solution to this problem then please advice.

anjali-wpf:
Fixes #2662

## Description

Fix for program crashing with a framework exception when we have a TabControl using a DataTemplate with DataGrids in it, and first we sort a DataGrid column and then try to edit a cell in that column, and then try to change tabs in the TabControl

## Customer Impact

Program crashing with a framework exception

## Regression

None

## Testing

Done, no new failures encountered. 

## Risk

None

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/wpf/pull/8259)

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Data/ListCollectionView.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Data/ListCollectionView.cs
@@ -2930,7 +2930,7 @@ namespace System.Windows.Data
         // SortDescription was added/removed, refresh CollectionView
         private void SortDescriptionsChanged(object sender, NotifyCollectionChangedEventArgs e)
         {
-            if (IsAddingNew || IsEditingItem)
+            if ((IsAddingNew || IsEditingItem) && e.Action != NotifyCollectionChangedAction.Reset)
                 throw new InvalidOperationException(SR.Format(SR.MemberNotAllowedDuringAddOrEdit, "Sorting"));
 
             // adding to SortDescriptions overrides custom sort


src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Data/ListCollectionView.cs
```diff
@@ -2930,7 +2930,7 @@ private void SetSortDescriptions(SortDescriptionCollection descriptions)
         // SortDescription was added/removed, refresh CollectionView
         private void SortDescriptionsChanged(object sender, NotifyCollectionChangedEventArgs e)
         {
-            if (IsAddingNew || IsEditingItem)
+            if ((IsAddingNew || IsEditingItem) && e.Action != NotifyCollectionChangedAction.Reset)
```
Kuldeep-MS, line 2933: Could you please provide a descriptive comment about the condition being checked here?

miloush:
Thix fix seems to avoid throwing exception but not fixing the underlying issue. If it is valid to proceed during adding/editing when the collection is reset, why not in other cases? If not, a legitimate cases which should throw will no longer throw.

singhashish-wpf:
Closing as of now, because more investigation is required.

PR #8338, opened 2023-10-20
Adding Stable Branding .NET 8 RTM

harshit7962:
## Description
Adding StabilizePackageVersion tag to enable stable branding.

## Testing
Local build pass

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/wpf/pull/8338)

### DIFF
--- a/eng/Versions.props
+++ b/eng/Versions.props
@@ -19,6 +19,7 @@
             dotnet/winforms is handling versions for the analyzers.
       -->
     <AssemblyVersion>$(MajorVersion).$(MinorVersion).0.0</AssemblyVersion>
+    <StabilizePackageVersion Condition="'$(StabilizePackageVersion)' == ''">true</StabilizePackageVersion>
   </PropertyGroup>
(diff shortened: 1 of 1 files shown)


Kuldeep-MS:
@harshit7962 - can you please add more details, why the change is required also attach any issue if it exists.

Kuldeep-MS (approved):
As discussed, changes look good to me.

dkurepa:
I don't think this change is needed, since WPF doesn't have any shipping packages.
Has something changed, do we construct some shipping packages in wpf now?

dkurepa (changes requested):
I don't think this change is needed, since WPF doesn't have any shipping packages.
Has something changed, do we construct some shipping packages in wpf now?

singhashish-wpf:
We don't produced shipping packages. This happened before confirmation. Closing the PR.

PR #8389, opened 2023-11-03
Update FixedDocument.cs

wstaelens:
Fixes # 

Main PR

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Documents/FixedDocument.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Documents/FixedDocument.cs
@@ -140,9 +140,7 @@ namespace System.Windows.Documents
 
 //             Dispatcher.VerifyAccess();
 
-            PageContent fp = value as PageContent;
-
-            if (fp == null)
+            if (value is not PageContent fp)
             {
                 throw new ArgumentException(SR.Format(SR.UnexpectedParameterType, value.GetType(), typeof(PageContent)), "value");
             }
@@ -409,8 +407,7 @@ namespace System.Windows.Documents
 
             ArgumentNullException.ThrowIfNull(contentPosition);
 
-            FixedTextPointer fixedTextPointer = contentPosition as FixedTextPointer;
-            if (fixedTextPointer == null)
+            if (contentPosition is not FixedTextPointer fixedTextPoiner)
             {
                 throw new ArgumentException(SR.IDPInvalidContentPosition);
             }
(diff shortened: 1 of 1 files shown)


src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Documents/FixedDocument.cs
```diff
@@ -409,8 +407,7 @@ internal int GetPageNumber(ContentPosition contentPosition)
 
             ArgumentNullException.ThrowIfNull(contentPosition);
 
-            FixedTextPointer fixedTextPointer = contentPosition as FixedTextPointer;
-            if (fixedTextPointer == null)
+            if (contentPosition is not FixedTextPointer fixedTextPoiner)
```
halgab, line 410:
```suggestion
            if (contentPosition is not FixedTextPointer fixedTextPointer)
```

src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Documents/FixedDocument.cs
```diff
@@ -836,7 +829,7 @@ private void _Init()
         {
             _paginator = new FixedDocumentPaginator(this);
             _pages = new PageContentCollection(this);
-            _highlights = new Dictionary<FixedPage, ArrayList>();
+            _highlights = new Dictionary<FixedPage, ArrayList>(); // todo: get rid of ArrayLists
```
halgab, line 832: What's the point of adding this todo? Isn't it more useful to open an issue?

Symbai: Docs say that [ArrayList aren't recommended and shouldn't be used anymore](https://learn.microsoft.com/en-us/dotnet/api/system.collections.arraylist?view=net-7.0#remarks). But I agree that it makes more sense to track this in a separate issue and leave comments like this out of code as they only confuses people.

src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Markup/XamlReader.cs
```diff
@@ -374,14 +367,12 @@ private object LoadAsync(XmlReader reader, ParserContext parserContext, bool use
...
-                        UIElement uiElement = args.Instance as UIElement;
-                        if (uiElement != null)
+                        if (args.Instance is UIElement uiElement)
                         {
                             uiElement.SetPersistId(_persistId++);
                         }
 
-                        DependencyObject dObject = args.Instance as DependencyObject;
-                        if (dObject != null && _stack.CurrentFrame.XmlnsDictionary != null)
+                        if (args.Instance is DependencyObject dObject && && _stack.CurrentFrame.XmlnsDictionary != null)
```
halgab, line 375:
```suggestion
                        if (args.Instance is DependencyObject dObject && _stack.CurrentFrame.XmlnsDictionary != null)
```

wstaelens:
(@halgab doing some quick wins via github browser edit, hoping to improve little bits.)

himgoyalmicro:
Hey @wstaelens,
Can you please rebase it to the main branch? Else is it okay if I make the changes and take it forward ?

PR #8403, opened 2023-11-08
Use ArgumentOutOfRangeException.ThrowIf* across codebase

halgab:
## Description

Use `ArgumentOutOfRangeException.ThrowIf*` across codebase.  A lot of potential use cases are not covered yet because CA1512 cannot automatically take care of them though

## Regression

No

## Testing

CI

## Risk

Low. We use fixer CA1512

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/wpf/pull/8403)

PR #8408, opened 2023-11-08
Use ArgumentOutOfRangeException.ThrowIf* to replace custom error messages

halgab:
This is a followup to #8403.

## Description

Many error messages can be replaced with messages from the BCL using the new `ArgumentOutOfRangeException` throw helpers.
Here are the messages I removed, each in its own commit:

- ParameterMustBeGreaterThanZero
- ParameterCannotBeNegative
- ParameterCannotBeGreaterThan
- ParameterCannotBeLessThan
- ParameterValueMustBeGreaterThanZero
- ParameterValueCannotBeNegative
- ShouldBeZero

ParameterMustBeBetween would be a good candidate, but this would slightly change the information conveyed by the error message. I still think it's worth it though.
ParameterValueCannotBeInfinity and ParameterValueCannotBeNaN also come to mind.

## Customer Impact

Slight performance improvement, and a bit fewer resources.

## Regression

No

## Testing

CI

## Risk

Low. I manually searched and removed the error messages, and let CA1512 finish the work

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/wpf/pull/8408)

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/PresentationCore/MS/internal/TextFormatting/FullTextLine.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationCore/MS/internal/TextFormatting/FullTextLine.cs
@@ -1498,12 +1498,9 @@ namespace MS.Internal.TextFormatting
                     throw new ObjectDisposedException(SR.TextLineHasBeenDisposed);
                 }
 
-                if(textLength == 0)
-                {
-                    throw new ArgumentOutOfRangeException("textLength", SR.ParameterMustBeGreaterThanZero);
-                }
+                ArgumentOutOfRangeException.ThrowIfZero(textLength);
 
-                if(textLength < 0)
+                if (textLength < 0)
                 {
                     firstTextSourceCharacterIndex += textLength;
                     textLength = -textLength;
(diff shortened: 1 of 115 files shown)


rchauhan18:
Hey @halgab , I think we also need to update these changes to the files below:

https://github.com/dotnet/wpf/blob/main/src/Microsoft.DotNet.Wpf/src/UIAutomation/UIAutomationClientSideProviders/MS/Internal/UnsupportedAutomationProxies/windowspager.cs
Line: 53-56

https://github.com/dotnet/wpf/blob/main/src/Microsoft.DotNet.Wpf/src/UIAutomation/UIAutomationClientSideProviders/MS/Internal/UnsupportedAutomationProxies/WindowsDateTimePicker.cs
Line: 59-63

halgab:
I kept those because of the asserts. But I'd be happy to make the change!

halgab:
> I kept those because of the asserts. But I'd be happy to make the change!

Actually that's not true. Those fell off my radar because they're not part of the solution. I'll make the change.

rchauhan18 (approved):

rchauhan18:
Thanks @halgab for contributing!

PR #8478, opened 2023-11-29
Adding a Null check for data in SetData function

rchauhan18:
## Description
Adding a Null check for Data in ```SetData``` function of ```DataObject ``` class. 
A null check has been overlooked within one of the ```SetData``` functions. Even when we use the constructor to set the data with the same attributes, it still verifies whether the data is null or not.

## Regression

None 

## Testing

Build and local testing passed.

## Risk

Low
It may cause few applications to crash with Null exception, if we try to drag and drop a null object.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/wpf/pull/8478)

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/dataobject.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/dataobject.cs
@@ -339,6 +339,8 @@ namespace System.Windows
         {
             ArgumentNullException.ThrowIfNull(format);
 
+            ArgumentNullException.ThrowIfNull(data);
+
             if (format == string.Empty)
             {
                 throw new ArgumentException(SR.DataObject_EmptyFormatNotAllowed);
--- a/src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/dataobject.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/dataobject.cs
@@ -331,20 +331,17 @@ namespace System.Windows
         /// to specify whether the
         /// data can be converted to another format.
         /// </summary>
-        /// <remarks>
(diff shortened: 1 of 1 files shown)


Kuldeep-MS:
@rchauhan18 -A more detailed explanation of why the changes is necessary would be beneficial.

Kuldeep-MS (approved):
It appears that we may have overlooked a null check in this particular instance of the overloaded `setData` function. While the changes seem straightforward and acceptable, there is a possibility that they could cause issues with the client's code. Therefore, I recommend thoroughly analyzing the function's usage before pushing the changes.

Overall, LGTM

src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/dataobject.cs
```diff
@@ -339,6 +339,8 @@ public void SetData(string format, Object data, bool autoConvert)
         {
             ArgumentNullException.ThrowIfNull(format);
 
+            ArgumentNullException.ThrowIfNull(data);
```
lindexi, line 342: Digression: should we remove the comment from the remarks:

https://github.com/dotnet/wpf/blob/821aa2440005f3b8911d6beb19d798155c3bd931/src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/dataobject.cs#L334-L336

Recent versions of .NET do not honor Code Access Security (CAS).

pchaurasia14: Also, I think we need to maintain parity between the implementations of the `SetData` functions. 
The check for `null` can be done after the `format` empty check to align with the same behavior.

lindexi (approved):
LGTM

singhashish-wpf (approved):

PR #8532, opened 2023-12-13
Replace binary formatter

rchauhan18:
## Description

The Binary Formatter was initially developed prior to the comprehensive understanding of deserialization vulnerabilities as a significant threat.  As a result, the code doesn’t follow the modern practices. Due to the security vulnerability Binary formatter is considered obsolete. In .NET 9.0, the entire Binary formatter infrastructure will be removed from the product. 

Binary Reader and Writer can be used to read and write the primitive data types into binary data and does not have any security vulnerability. We are creating a safe handle ‘BinaryFormatWriter’ which handles all primitives, Enums, strings, arrays and lists of primitives, and other items before falling into the process of serialization/deserialization. For serialization/deserialization we will use Binary writer. This approach will allow us to disable Binary formatter for the primitive and some other cases. 
 To convert the object to binary data we call TryWriteFrameworkObject with a memory stream and the object.  
BinaryFormattedObject class is added to parse a binary formatted object from a stream. It used Binary reader to read from a stream.
This solution is similar as in WinForms
PR: https://github.com/dotnet/winforms/pull/9088
CC
@JeremyKuhne @lonitra 

image

## Customer Impact
 In .NET 9.0, the entire Binary formatter infrastructure will be removed from the product, and it will become obsolete. 
It has additional benefits of making applications more friendly to linker trimming.

## Regression

N/A

## Testing

Internal testing done.

## Risk

Low
There might be a few instances where this solution won't be effective. Currently, this PR includes all the primitive types and few more data types. The plan is to handle additional cases as feedback comes in.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/wpf/pull/8532)


src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/dataobject.cs
```diff
@@ -1677,19 +1677,13 @@ private int SaveObjectToHandle(IntPtr handle, Object data, bool doNotReallocate)
...
                 using (binaryWriter = new BinaryWriter(stream))
                 {
                     binaryWriter.Write(_serializedObjectID);
-
-                    formatter = new BinaryFormatter();
-
-                    #pragma warning disable SYSLIB0011 // BinaryFormatter is obsolete 
-                    formatter.Serialize(stream, data);
-                    #pragma warning restore SYSLIB0011 // BinaryFormatter is obsolete
+                    BinaryFormatWriter.TryWriteFrameworkObject(stream,data);
```
JeremyKuhne, line 1686: The idea here is to try to use the `BinaryFormatWritter` first, then fall back to the `BinaryFormatter` if you cannot for some reason. See our [`DataObject` implementation](https://github.com/dotnet/winforms/blob/dbc7ce32e8bf24466e69c1907af2977f3aee68dc/src/System.Windows.Forms/src/System/Windows/Forms/OLE/DataObject.cs#L653-L671).

You shouldn't remove the `BinaryFormatter` as a fall back unless you know that there are no user scenarios that might need it. While it is depreciated and will likely be disabled by default for WPF/WinForms soon, there will always be a way for users to enable it if they need tol

src/Microsoft.DotNet.Wpf/src/PresentationFramework/MS/Internal/DataStreams.cs
```diff
@@ -126,9 +126,7 @@ private ArrayList SaveSubStreams(UIElement element)
                         {
                             // Convert the value of the DP into a byte array
                             MemoryStream byteStream = new MemoryStream();
-                            #pragma warning disable SYSLIB0011 // BinaryFormatter is obsolete 
-                            this.Formatter.Serialize(byteStream, currentValue);
-                            #pragma warning restore SYSLIB0011 // BinaryFormatter is obsolete 
+                            BinaryFormatWriter.TryWriteFrameworkObject(byteStream,currentValue);
```
JeremyKuhne, line 129: Do you know that this works in all cases? Why are you not checking for success here?

src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/dataobject.cs
wjk: > there will always be a way for users to enable it

That isn’t true. If you use the `BinaryFormatter` anywhere, for anything, you will see member-does-not-exist errors that stop every WPF build come .NET 9. `BinaryFormatter` must be completely removed.

JeremyKuhne:
To be clear, `BinaryFormatter` APIs will never be removed. Setting a switch in .NET 9 to enable the `BinaryFormatter` will likely become required for WPF and WinForms (as it is already for all other .NET projects in .NET 8). Enabling the functionality will possibly require additional steps in the future, but the surface area should always be there.

wjk:
> To be clear, `BinaryFormatter` APIs will never be removed.

Then it is my mistake for misunderstanding the “Customer Impact” section of the OP. Sorry to bother you.

src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/dataobject.cs
rchauhan18: @JeremyKuhne Thanks for the review.
I have added ```BinaryFormatter``` as a fall back.

src/Microsoft.DotNet.Wpf/src/PresentationFramework/MS/Internal/DataStreams.cs
rchauhan18: I have added a flag ```success```, on getting false it will use BinaryFormatter.

michmela44:
> To be clear, BinaryFormatter APIs will never be removed.

While the APIs may never be removed, according to [BinaryFormatter infrastructure removed from .NET (.NET 9)](https://github.com/dotnet/designs/blob/main/accepted/2020/better-obsoletion/binaryformatter-obsoletion.md#binaryformatter-infrastructure-removed-from-net-net-9), "_The internal BinaryFormatter implementation will no longer exist in .NET in any form and cannot be reenabled through any compat switch._"

src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/dataobject.cs
JeremyKuhne: > > there will always be a way for users to enable it
> 
> That isn’t true. If you use the `BinaryFormatter` anywhere, for anything, you will see member-does-not-exist errors that stop every WPF build come .NET 9. `BinaryFormatter` must be completely removed.

Can you please source where you're getting your information? I'm positive the APIs are not being removed.

JeremyKuhne:
> > To be clear, BinaryFormatter APIs will never be removed.
> 
> While the APIs may never be removed, according to [BinaryFormatter infrastructure removed from .NET (.NET 9)](https://github.com/dotnet/designs/blob/main/accepted/2020/better-obsoletion/binaryformatter-obsoletion.md#binaryformatter-infrastructure-removed-from-net-net-9), "_The internal BinaryFormatter implementation will no longer exist in .NET in any form and cannot be reenabled through any compat switch._"

The hope is to make you jump through more hoops by putting the actual implementation in another package, but there is no way the API will be removed or there will be no way to re-enable for quite some time. It is fundamental to many existing applications, particularly WinForms.

To be clear, one should avoid the `BinaryFormatter` if at all possible, but as a platform we will not completely abandon existing users of this functionality.

JeremyKuhne (dismissed):
This looks good. You should consider ensuring you have test coverage with and without the `BinaryFormatter` being enabled. See https://github.com/dotnet/winforms/blob/main/src/Common/tests/TestUtilities/BinaryFormatterScope.cs for how we do this in WinForms.

src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/dataobject.cs
michmela44: I suspect the source is the same place I linked in the dotnet github repo design document

[BinaryFormatter Obsoletion Strategy](https://github.com/dotnet/designs/blob/main/accepted/2020/better-obsoletion/binaryformatter-obsoletion.md#binaryformatter-infrastructure-removed-from-net-net-9)

While it does say it **_may_** get moved to a 3rd party package, it does not guarantee it.

> "All code bases which target net9.0 or above must be fully migrated away from BinaryFormatter at this time. This applies to dotnet org first-party code, partner team code, and third-party user code."

> "The implementation of all such APIs will be to throw PlatformNotSupportedException. The internal BinaryFormatter implementation will no longer exist in .NET in any form and cannot be reenabled through any compat switch."

If this information is false, then perhaps the dotnet team should update the design document to more accurately represent the future plans.

eng/copy-wpf.ps1
```diff
@@ -30,13 +30,15 @@ function Print-Usage()
 Write-Host "*** Copy WPF files procedure ***"
 
 $RepoRoot = Resolve-Path (Join-Path $PSScriptRoot "..")
+Write-host "Repo root = " $RepoRoot
```
jonwis, line 33: It looks like some of these are debugging prints. Consider `Write-Debug` - https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.utility/write-debug?view=powershell-7.4

src/Microsoft.DotNet.Wpf/src/PresentationCore/OtherAssemblyAttrs.cs
```diff
@@ -18,7 +18,7 @@
 [assembly:InternalsVisibleTo(BuildInfo.PresentationFrameworkSystemDrawing)]
 [assembly:InternalsVisibleTo(BuildInfo.SystemWindowsControlsRibbon)]
 [assembly:InternalsVisibleTo(BuildInfo.WindowsFormsIntegration)]
-
+[assembly:InternalsVisibleTo("PresentationCore.Tests,PublicKey=0024000004800000940000000602000000240000525341310004000001000100b5fc90e7027f67871e773a8fde8938c81dd402ba65b9201d60593e96c492651e889cc13f1415ebb53fac1131ae0bd333c5ee6021672d9718ea31a8aebd0da0072f25d87dba6fc90ffd598ed4da35e44c398c454307e8e33b8426143daec9f596836f97c8f74750e5975c64e2189f45def46b2a2b1247adc3652bf5c308055da9")]
```
jonwis, line 21: Is this the normal way of making tests work? Is this public key value stable for all people who clone, build, and test?

rchauhan18: @jonwis Yes, the public key value is stable for all other's build.
It is used in a similar manner within the repository.
https://github.com/dotnet/wpf/blob/44e26b44141f81af2004330681de835a36e5d07b/src/Microsoft.DotNet.Wpf/src/DirectWriteForwarder/OtherAssemblyAttrs.cpp

lindexi:
I wonder if this change has changed the binary format. In other word, does this change maintain compatibility with the old content?

Symbai:
> To be clear, `BinaryFormatter` APIs will never be removed. Setting a switch in .NET 9 to enable the `BinaryFormatter` will likely become required for WPF and WinForms

Looks like this is not the case. The switch won't work and any usage will throw an exception, see https://github.com/dotnet/runtime/issues/98245 so either:

a) WPF stays on .NET 8 forever
b) WPF subscribes to the nuget package
c) WPF replaces BinaryFormatter. Any breaking change has to be accepted then (since its a new major version of .NET this isn't as bad as it sounds)

JeremyKuhne:
@Symbai the API surface will still be there, you'll just need the package for the APIs to work.

Symbai:
> @Symbai the API surface will still be there, you'll just need the package for the APIs to work.

Which is exactly what I said for option B.

IrinaPykhova:
What is the current state of it? I'm testing .Net 9 Preview 4 with my .Net 6 control which has serializable objects and uses them for drag&drop. Apparently, DataObject class still uses binary serialization and it only fails on attempt to deserialize with exception about not supported binary formatter. What are your plans about this, should I try to fix it from my side and serialize object manually before creating DataObject or you will try to use other formatter from your side? Is it time to post you bugs about this?
I haven't checked yet, but suppose I'll have the same question about clipboard. We want our controls to be compatible with .Net 9 by November, so I need to know more details in advance.

JeremyKuhne:
Note that the runtime is adding a public type that is based, in part, on the code you've copied from WinForms. `NrbfDecoder` will read binary format records.

https://github.com/dotnet/runtime/pull/103232

Our plan is to move the WinForms code to use this in the next preview.

Fyi: @lonitra and @Tanya-Solyanik are working on adding additional APIs to `DataObject` in WinForms to allow for safer binary format usage.

src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/BinaryFormat/BinaryArrayType.cs
```diff
@@ -0,0 +1,48 @@
...
+
+        /// <summary>
+        ///  A jagged array where the lower bound index is greater than 0.
+        /// </summary>
+        JaggedOffset = 4,
+
+        /// <summary>
+        ///  Multi-dimensional arrays where the lower bound index of at least one of the dimensions is greater than 0.
+        /// </summary>
+        RectangularOffset = 5,
```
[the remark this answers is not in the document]
JeremyKuhne: If we share the code they'll get this, so I wouldn't worry about this sort of alignment.

src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/BinaryFormat/BinaryFormatWriter.cs
```diff
@@ -0,0 +1,625 @@
...
+        ///  Writes a primitive list in binary format.
+        /// </summary>
+        public static void WritePrimitiveList<T>(Stream stream, List<T> list)
+            where T : unmanaged
+        {
+            PrimitiveType primitiveType = TypeInfo.GetPrimitiveType(typeof(T));
+            if (primitiveType == default)
+            {
+                throw new NotSupportedException($"{nameof(T)} is not primitive.");
+            }
```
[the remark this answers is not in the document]
JeremyKuhne: I'm not super concerned as it is pretty unlikely that people would cut/paste nints.

JeremyKuhne (approved):
Looks good, you'll want to make sure to manually test that handled types work without the BinaryFormatter.

singhashish-wpf (approved):

PR #8581, opened 2023-12-21
Adding win11 styles to RichTextBox and Separator

anjali-wpf:
## Description

Adding Win11 Styles to the following controls:

1. RichTextBox 
2. Separator

## Regression

NA

## Testing

Build Pass.
Verified the visual appearance changes of controls in a sample application.

## Risk

NA

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/wpf/pull/8581)


src/Microsoft.DotNet.Wpf/src/Themes/PresentationFramework.Win11/Styles/RichTextBox.xaml
```diff
@@ -0,0 +1,84 @@
...
+    Copyright (c) Microsoft Corporation. All Rights Reserved.
+-->
+
+<ResourceDictionary
+    xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
+    xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
+    xmlns:controls="clr-namespace:System.Windows;assembly=PresentationFramework">
+
+    <Style x:Key="DefaultRichTextBoxStyle" TargetType="{x:Type RichTextBox}">
+        <Setter Property="Foreground" Value="{DynamicResource TextControlForeground}" />
```
lindexi, line 12: Can we use StaticResource here?

src/Microsoft.DotNet.Wpf/src/Themes/PresentationFramework.Win11/Styles/RichTextBox.xaml
```diff
@@ -0,0 +1,84 @@
...
+        </Setter>
+    </Style>
+
+    <Style
+        x:Key="DefaultUiRichTextBoxStyle"
+        BasedOn="{StaticResource DefaultRichTextBoxStyle}"
+        TargetType="{x:Type RichTextBox}">
+
+        <Style.Triggers>
+            <Trigger Property="TextBox.IsTextSelectionEnabled" Value="True">
```
wjk, line 70: This looks backwards to me. I would think you would want to use `Value="False"` here, seeing as the `IsReadOnly` property is then set to `true`.

wjk (changes requested):

src/Microsoft.DotNet.Wpf/src/Themes/PresentationFramework.Win11/Styles/RichTextBox.xaml
anjali-wpf: Resolving this for now as the issue is listed. Will keep things updating there,
Issue available [here](https://github.com/dotnet/wpf/issues/8599)

src/Microsoft.DotNet.Wpf/src/Themes/PresentationFramework.Win11/Styles/RichTextBox.xaml
anjali-wpf: @wjk Thanks for the suggestion, I made changes accordingly. I changed `IsReadOnly` flag to `False`.
If we will set `Value="False"` then it will make `RichTextBox` disabled on load which is not the desired behavior, but by setting `IsReadOnly` to `False` will work as expected.

dipeshmsft (approved):

PR #8593, opened 2023-12-22
Adding Win11 Styles to Combo box

anjali-wpf:
## Description

Adding Win11 Styles to Combo box

## Regression

NA

## Testing

Build Pass.
Verified the visual appearance changes of controls in a sample application.

## Risk

NA

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/wpf/pull/8593)


src/Microsoft.DotNet.Wpf/src/Themes/PresentationFramework.Win11/Styles/ComboBox.xaml
```diff
@@ -0,0 +1,387 @@
...
+                                        Margin="0"
+                                        Padding="0,4,0,6"
+                                        Background="{DynamicResource ComboBoxDropDownBackground}"
+                                        BorderBrush="{DynamicResource ComboBoxDropDownBorderBrush}"
+                                        BorderThickness="1"
+                                        CornerRadius="{DynamicResource PopupCornerRadius}"
+                                        SnapsToDevicePixels="True">
+                                        <Border.RenderTransform>
+                                            <TranslateTransform />
+                                        </Border.RenderTransform>
```
Symbai, line 275: Why the `TranslateTransform`? If this is for some reason required then [MSDN recommends](https://learn.microsoft.com/en-us/dotnet/api/system.windows.media.translatetransform?view=windowsdesktop-8.0#remarks) to freeze it to increase performance.

src/Microsoft.DotNet.Wpf/src/Themes/PresentationFramework.Win11/Styles/ComboBox.xaml
```diff
@@ -0,0 +1,387 @@
...
+                                            Foreground="{TemplateBinding Foreground}"
+                                            IsReadOnly="{TemplateBinding IsReadOnly}"
+                                            Style="{StaticResource DefaultComboBoxTextBoxStyle}" />
+                                    </Grid>
+                                </Grid>
+                                <Popup
+                                    x:Name="Popup"
+                                    MinWidth="{TemplateBinding ActualWidth}"
+                                    VerticalAlignment="Center"
+                                    AllowsTransparency="True"
```
lindexi, line 258: AllowsTransparency be set to true may cause performance issues. After set AllowsTransparency, the popup windows have to copy the window's render content to GDI. See https://docs.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-updatelayeredwindow and https://docs.microsoft.com/en-us/previous-versions/windows/desktop/legacy/ms633557(v=vs.85)

https://github.com/dotnet/wpf/blob/446d4d008fe56097db66d96e2310662b23112da4/src/Microsoft.DotNet.Wpf/src/WpfGfx/core/hw/d3dswapchainwithswdc.cpp#L142-L189

anjali-wpf: I tried my changes on a sample app, if we are using `AllowsTransparency = True`, then the combo box appearance will be like this - 
image

If we will set it to `False` or if we remove it then it is going to change the combo box appearance, it will look like this - 
image

Also I compared the performance, the value difference is less.

src/Microsoft.DotNet.Wpf/src/Themes/PresentationFramework.Win11/Styles/ComboBox.xaml
anjali-wpf: Thanks for the suggestion. We can remove this, I checked the behavior, we can eliminate this.

dipeshmsft:
I think we still need to do some more tests before we can remove this. As of now, we shall let it be here, and make a note / comment in performance issue to look at it when doing performance evaluation.

src/Microsoft.DotNet.Wpf/src/Themes/PresentationFramework.Win11/Styles/ComboBox.xaml
```diff
@@ -270,9 +270,6 @@
                                         BorderThickness="1"
                                         CornerRadius="{DynamicResource PopupCornerRadius}"
                                         SnapsToDevicePixels="True">
-                                        <Border.RenderTransform>
```
dipeshmsft, line 273: This is used for animation and we need to evaluate the effect of removing this properly. We should leave it here for now and get back to this issue when we are doing performance evaluation for dynamic resources.

PR #8519, opened 2023-12-08
Prefer use of Interpolated strings in project Presentation

halgab:
Sorry, this PR is massive. But I'm not sure how to break it down as it contains mostly automated changes.

## Description

We replace as many manual string concatenations and `string.Format` with interpolated strings. This saves allocations in many instances, and makes for a more readable code.

## Customer Impact

Less allocations

## Regression

No

## Testing

CI

## Risk

Low. Most changes are automated using the IDE

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/wpf/pull/8519)

PR #8616, opened 2023-12-30
Prefer use interpolated strings in PresentationCore

halgab:
Follow-up to #8519 

## Description

We replace as many manual string concatenations and string.Format with interpolated strings. This saves allocations in many instances, and makes for a more readable code.

## Customer Impact

Less allocations

## Regression

No

## Testing

CI

## Risk

Low. Most changes are automated using the IDE


src/Microsoft.DotNet.Wpf/src/PresentationCore/MS/internal/AppModel/SiteOfOriginContainer.cs
```diff
@@ -58,9 +58,7 @@ internal static Uri SiteOfOrigin
 #if DEBUG
             if (_traceSwitch.Enabled)
                 System.Diagnostics.Trace.TraceInformation(
-                        DateTime.Now.ToLongTimeString() + " " + DateTime.Now.Millisecond + " " +
-                        Environment.CurrentManagedThreadId + 
-                        ": SiteOfOriginContainer: returning site of origin " + siteOfOrigin);
+                    $"{DateTime.Now:T} {DateTime.Now.Millisecond} {Environment.CurrentManagedThreadId}: SiteOfOriginContainer: returning site of origin {siteOfOrigin}");
```
halgab, line 61: A follow-up to this is to introduce a Trace method with a custom interpolated string handler that encapsulate all this logic. It would clean-up the code quite a lot

rchauhan18: @halgab Having an encapsulated function would be beneficial for maintaining clean code. Could you please provide the function's arguments and structure that you are considering?

We are planning to add this PR for this month test-pass. So, can you resolve the merge conflicts for this PR?

halgab: I've added a draft PR based on this one here: #8702. Its last commit is what I have in mind. It's based on similar changes I contributed in the winforms repo.

src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Media/Color.cs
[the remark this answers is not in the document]
halgab: After having another read, the difference between the current version and mine is that when `nativeColorValue` is empty, the current version still appends a separator value, which is not the case with my version. I feel like my version is more correct.
Tell me if you prefer that I maintain the current behaviour.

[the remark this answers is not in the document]
ThomasGoulet73: Your changes are functionally different. Floats were formatted using `format` previously but are now always formatted  using `R`. Also, the loop in your changes always prepends the separator instead of prepending the separator only when between items: ",A,B,C" vs "A,B,C". Also, there should be a space [here](https://github.com/dotnet/wpf/pull/8616/files/4ea763103e02b739c4990762a927be7052f4e2f0#diff-e0b0996546e517f2040463a25c03d8b23308a1bd2e247d76e390af57fbd1456cR306) between `{uriString}` and `{scRgbColor.a:R}`, it was there before your changes.

I would recommend testing your changes because by looking at the code I don't see how your changes returns the same result as the existing code.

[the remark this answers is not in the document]
halgab: Those are valid points. I'll rollback the functional changes. I'd like to keep the constant format though, as `format` is always set to `R` before this loop, and that's what makes it possible to use interpolated strings. What do you think?

src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Media/PathFigure.cs
```diff
@@ -223,10 +223,7 @@ internal bool CanSerializeToString()
         internal string ConvertToString(string format, IFormatProvider provider)
         {
             PathSegmentCollection segments = Segments;
-            return "M" + 
-                ((IFormattable)StartPoint).ToString(format, provider) + 
-                (segments != null ? segments.ConvertToString(format, provider) : "") +
-                (IsClosed ? "z" : "");
+            return $"M{((IFormattable)StartPoint).ToString(format, provider)}{(segments != null ? segments.ConvertToString(format, provider) : "")}{(IsClosed ? "z" : "")}";
```
dipeshmsft, line 226: Again, this seems hard to be understood. Maybe consider using string.Concat / string.Format here.

src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Media/Color.cs
```diff
@@ -303,14 +303,13 @@ internal string ConvertToString(string format, IFormatProvider provider)
                 String uriString = safeUnescapedUri.GetComponents(UriComponents.SerializationInfoString, UriFormat.UriEscaped);
 
                 var sb = new StringBuilder();
-                sb.AppendFormat(provider, "{0}{1} ", Parsers.s_ContextColor, uriString);
-                sb.AppendFormat(provider,"{1:" + format + "}{0}",separator,scRgbColor.a);
-                for (int i= 0; i< nativeColorValue.GetLength(0); ++i )
+                sb.Append(provider, $"{Parsers.s_ContextColor}{uriString} {scRgbColor.a:R}{separator}");
```
dipeshmsft, line 306: We can split this back in two lines and use the same suggestion as below for the second line.

src/Microsoft.DotNet.Wpf/src/PresentationCore/MS/internal/Ink/StrokeIntersection.cs
```diff
@@ -96,10 +96,7 @@ internal double InEnd
         /// </summary>
         public override string ToString()
         {
-            return "{"  + StrokeFIndices.GetStringRepresentation(_hitSegment.BeginFIndex) + ","
-                        + StrokeFIndices.GetStringRepresentation(_inSegment.BeginFIndex)  + ","
-                        + StrokeFIndices.GetStringRepresentation(_inSegment.EndFIndex)    + ","
-                        + StrokeFIndices.GetStringRepresentation(_hitSegment.EndFIndex)   + "}";
+            return $"{{{StrokeFIndices.GetStringRepresentation(_hitSegment.BeginFIndex)},{StrokeFIndices.GetStringRepresentation(_inSegment.BeginFIndex)},{StrokeFIndices.GetStringRepresentation(_inSegment.EndFIndex)},{StrokeFIndices.GetStringRepresentation(_hitSegment.EndFIndex)}}}";
```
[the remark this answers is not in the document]
h3xds1nz: @dipeshmsft  We can just use Raw String Literals lang feature and combine it with interpolation, no need for string.Format ugliness.

src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Media/PathFigure.cs
h3xds1nz: Same here, we can just use Raw String Literals lang feature and combine it with interpolation, no need for string.Format ugliness.

src/Microsoft.DotNet.Wpf/src/PresentationCore/MS/internal/Ink/StrokeIntersection.cs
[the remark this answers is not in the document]
dipeshmsft: @h3xds1nz,  I was just going through `DefaultInterpolatedStringHandler` in your PR and your suggestion sounds good.

dipeshmsft:
@halgab, can you take a look at the above PR and make the necessary changes ?

halgab:
Hi. I'm sorry, I'm not sure when I'll have time to get back to this...

dipeshmsft:
@halgab, is it okay we take this up ?

halgab:
Yes, of course

src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Media/Color.cs
```diff
@@ -294,14 +294,13 @@ internal string ConvertToString(string format, IFormatProvider provider)
                 String uriString = safeUnescapedUri.GetComponents(UriComponents.SerializationInfoString, UriFormat.UriEscaped);
 
                 var sb = new StringBuilder();
-                sb.AppendFormat(provider, "{0}{1} ", Parsers.s_ContextColor, uriString);
-                sb.AppendFormat(provider,"{1:" + format + "}{0}",separator,scRgbColor.a);
-                for (int i = 0; i < nativeColorValue.Length; ++i )
+                sb.Append(provider, $"{Parsers.s_ContextColor}{uriString} {scRgbColor.a:R}{separator}");
```
himgoyalmicro, line 297: We should use the variable format here.

PR #8622, opened 2024-01-02
Use span for LengthConverter

ThomasGoulet73:
## Description
Use span for LengthConverter and other similar places. I also introduced the internal struct `PixelUnit` to share some of the logic.

I found that LengthConverter.FromString can be called a lot on startup (~3000 times in a sample app).

<details>
  <summary>Benchmark</summary>
  
| Method | s    | Mean      | Error     | StdDev    | Ratio | Gen0   | Code Size | Allocated | Alloc Ratio |
|------- |----- |----------:|----------:|----------:|------:|-------:|----------:|----------:|------------:|
| Old    | 12   | 37.447 ns | 0.0442 ns | 0.0369 ns |  1.00 |      - |   1,146 B |         - |          NA |
| New    | 12   | 33.244 ns | 0.0190 ns | 0.0148 ns |  0.89 |      - |     670 B |         - |          NA |
|        |      |           |           |           |       |        |           |           |             |
| Old    | 12cm | 39.023 ns | 0.0914 ns | 0.0810 ns |  1.00 | 0.0019 |   1,176 B |      32 B |        1.00 |
| New    | 12cm | 33.253 ns | 0.0478 ns | 0.0424 ns |  0.85 |      - |     714 B |         - |        0.00 |
|        |      |           |           |           |       |        |           |           |             |
| Old    | 12CM | 43.755 ns | 0.1675 ns | 0.1567 ns |  1.00 | 0.0038 |   1,158 B |      64 B |        1.00 |
| New    | 12CM | 33.792 ns | 0.0985 ns | 0.0921 ns |  0.77 |      - |     714 B |         - |        0.00 |
|        |      |           |           |           |       |        |           |           |             |
| Old    | 12in | 36.953 ns | 0.0672 ns | 0.0561 ns |  1.00 | 0.0019 |   1,176 B |      32 B |        1.00 |
| New    | 12in | 32.398 ns | 0.0756 ns | 0.0707 ns |  0.88 |      - |     693 B |         - |        0.00 |
|        |      |           |           |           |       |        |           |           |             |
| Old    | 12IN | 43.497 ns | 0.1695 ns | 0.1585 ns |  1.00 | 0.0038 |   1,176 B |      64 B |        1.00 |
| New    | 12IN | 31.147 ns | 0.0807 ns | 0.0755 ns |  0.72 |      - |     664 B |         - |        0.00 |
|        |      |           |           |           |       |        |           |           |             |
| Old    | 12pt | 40.756 ns | 0.0865 ns | 0.0722 ns |  1.00 | 0.0019 |   1,176 B |      32 B |        1.00 |
| New    | 12pt | 33.522 ns | 0.0525 ns | 0.0410 ns |  0.82 |      - |     679 B |         - |        0.00 |
|        |      |           |           |           |       |        |           |           |             |
| Old    | 12PT | 46.214 ns | 0.1658 ns | 0.1551 ns |  1.00 | 0.0038 |   1,176 B |      64 B |        1.00 |
| New    | 12PT | 34.190 ns | 0.0717 ns | 0.0636 ns |  0.74 |      - |     676 B |         - |        0.00 |
|        |      |           |           |           |       |        |           |           |             |
| Old    | 12px | 35.508 ns | 0.1054 ns | 0.0985 ns |  1.00 | 0.0019 |   1,176 B |      32 B |        1.00 |
| New    | 12px | 30.360 ns | 0.0408 ns | 0.0341 ns |  0.85 |      - |     630 B |         - |        0.00 |
|        |      |           |           |           |       |        |           |           |             |
| Old    | 12PX | 40.378 ns | 0.1564 ns | 0.1463 ns |  1.00 | 0.0038 |   1,176 B |      64 B |        1.00 |
| New    | 12PX | 29.259 ns | 0.0727 ns | 0.0644 ns |  0.72 |      - |     630 B |         - |        0.00 |
|        |      |           |           |           |       |        |           |           |             |
| Old    | auto |  5.010 ns | 0.0224 ns | 0.0210 ns |  1.00 |      - |     656 B |         - |          NA |
| New    | auto |  2.149 ns | 0.0287 ns | 0.0268 ns |  0.43 |      - |     509 B |         - |          NA |
  
</details>

## Customer Impact
Better perf and lower memory usage.

## Regression
No.

## Testing
Local testing.

## Risk
Low.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/wpf/pull/8622)


src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/LengthConverter.cs
```diff
@@ -181,60 +181,41 @@ public override bool CanConvertTo(ITypeDescriptorContext typeDescriptorContext,
...
+            ReadOnlySpan<char> valueSpan = s.AsSpan().Trim();
             double unitFactor = 1.0;
 
             //Auto is represented and Double.NaN
             //properties that do not want Auto and NaN to be in their ligit values,
             //should disallow NaN in validation callbacks (same goes for negative values)
-            if (goodString == "auto") return Double.NaN;
-
-            for (int i = 0; i < PixelUnitStrings.Length; i++)
+            if (valueSpan.Equals("auto", StringComparison.OrdinalIgnoreCase))
```
psmulovics, line 190: Should use SequenceEquals with an invariant comparer, supposed to be faster than Equals.

ThomasGoulet73: Can you include what you think this line of code should be ? I believe this call is the most efficient way to compare 2 ReadOnlySpan<char> while using OrdinalIgnoreCase comparison.

psmulovics: Actually, in my micro measurement, even ToLowerInvariant().Equals("auto") was faster than (valueSpan.Equals("auto", StringComparison.OrdinalIgnoreCase)), no idea why.

ThomasGoulet73: In your micro measurement, do you do `"auto".ToLowerInvariant().Equals("auto")` ? Because in that case ToLowerInvariant does not allocate a new string because "auto" is already lowercase and in that case comparing strings might be slightly faster than span because of the need to create the span.

lindexi (approved):
Awesome

src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/LengthConverter.cs
psmulovics: No, I built AuTo from characters to compare it with auto.

psmulovics: We can build a 16 element static rainbow table too that we can statically compare the binary memory bytes for absolute performance gain.

ThomasGoulet73: I tried benchmarking this and I found using span to be much faster.

<details>
  <summary>Benchmark code</summary>
  
  ```csharp
[MemoryDiagnoser]
public class Benchmark1
{
    public IEnumerable<object[]> GetArguments()
    {
        yield return new object[] { "AuTo" };
    }

    [Benchmark(Baseline = true)]
    [ArgumentsSource(nameof(GetArguments))]
    public bool Old(string str1)
    {
        return str1.ToLowerInvariant().Equals("auto");
    }

    [Benchmark]
    [ArgumentsSource(nameof(GetArguments))]
    public bool New(string str1)
    {
        return str1.AsSpan().Equals("auto", StringComparison.OrdinalIgnoreCase);
    }
}
  ```
  
</details>

<details>
  <summary>Benchmark results</summary>
  
| Method | str1 | Mean       | Error     | StdDev    | Ratio | Gen0   | Allocated | Alloc Ratio |
|------- |----- |-----------:|----------:|----------:|------:|-------:|----------:|------------:|
| Old    | AuTo | 16.6884 ns | 0.3390 ns | 0.3005 ns |  1.00 | 0.0051 |      32 B |        1.00 |
| New    | AuTo |  0.9111 ns | 0.0370 ns | 0.0309 ns |  0.05 |      - |         - |        0.00 |
  
</details>

psmulovics: Can you try AsSpan().ToLowerInvariant()?

lindexi: @psmulovics Sorry, I do not think `AsSpan().ToLowerInvariant()` has any advantage.

psmulovics: I am still slightly concerned on the 15% slower 'auto' - as mentioned above, can we do a rainbow table of the 16 possible ways to write auto and compare it bitwise?

ThomasGoulet73: I forgot to mention it but I opened an issue in dotnet/runtime for the regression of "auto": dotnet/runtime#96524. It is a performance regression introduced in .Net 8. I have a workaround that I'll post soon which should remove the slight regression of "auto".

> I am still slightly concerned on the 15% slower 'auto' - as mentioned above, can we do a rainbow table of the 16 possible ways to write auto and compare it bitwise?

String comparisons are already highly optimized and we shouldn't make the code way more complex just to maybe make a ~8 ns case a bit faster.

ThomasGoulet73: I pushed a commit that fixed the perf regression and I updated the benchmark results. Every case improved now, I'm marking this conversation as resolved.

src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/LengthConverter.cs
```diff
@@ -203,16 +203,21 @@ static internal double FromString(string s, CultureInfo cultureInfo)
...
             // to create a Length instance from a string.  This addresses windows bug 968884
             try
             {
-                return double.Parse(valueSpan, cultureInfo) * unitFactor;
+                return double.Parse(span, cultureInfo);
             }
             catch (FormatException)
             {
-                throw new FormatException(SR.Format(SR.LengthFormatError, valueSpan.ToString()));
+                throw new FormatException(SR.Format(SR.LengthFormatError, span.ToString()));
```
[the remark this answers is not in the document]
Rob-Hague: Actually, it's an exceptional case, so it probably doesn't matter...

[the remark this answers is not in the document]
ThomasGoulet73: It is true that we could replace the try-catch with TryParse but I didn't want to do this change in this PR since, as you said, it would only improve an exceptional case.

src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Controls/DataGridLengthConverter.cs
```diff
@@ -196,22 +196,21 @@ internal static string ConvertToString(DataGridLength length, CultureInfo cultur
         /// </remarks>
         private static DataGridLength ConvertFromString(string s, CultureInfo cultureInfo)
         {
-            string goodString = s.Trim().ToLowerInvariant();
+            ReadOnlySpan<char> valueSpan = s.AsSpan().Trim();
```
harshit7962, line 199: s.Trim().AsSpan() better option?

src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/LengthConverter.cs
```diff
@@ -181,60 +181,46 @@ public override object ConvertTo(ITypeDescriptorContext typeDescriptorContext,
         // NOTE - This code is called from FontSizeConverter, so changes will affect both.
         static internal double FromString(string s, CultureInfo cultureInfo)
         {
-            string valueString = s.Trim();
-            string goodString = valueString.ToLowerInvariant();
-            int strLen = goodString.Length;
-            int strLenUnit = 0;
+            ReadOnlySpan<char> valueSpan = s.AsSpan().Trim();
```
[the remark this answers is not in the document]
h3xds1nz: This usage is correct, the other ones are not. We should `Trim` on the `Span` to prevent an allocation.

src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Controls/DataGridLengthConverter.cs
```diff
@@ -267,7 +260,7 @@ private static DataGridLength ConvertFromString(string s, CultureInfo cultureInf
                     (unit == DataGridLengthUnitType.Pixel) || DoubleUtil.AreClose(unitFactor, 1.0),
                     "unitFactor should not be other than 1.0 unless the unit type is Pixel.");
 
-                ReadOnlySpan<char> valueString = goodString.AsSpan(0, strLen - strLenUnit);
+                ReadOnlySpan<char> valueString = valueSpan.Slice(0, valueSpan.Length - strLenUnit);
```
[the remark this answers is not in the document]
ThomasGoulet73: Yes we would now call double.Parse with the non-lowercased value. I chose to unify all converters to do the same because of this comment here which explains why we should double.Parse with the non-lowercased value (Though the comment seems obsolete because I made some tests and I checked the code of double.Parse and the case of the value in double.Parse shouldn't matter): https://github.com/dotnet/wpf/pull/8622/files/8ea913b2f0e29f8d82d2bcd23dcc555996abc7d7#diff-c16c7b22fe582e51d67bbddff73c7ed62887fde21bb575147ee6c62c09c9b6bcL207-L210

src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Markup/XamlGridLengthSerializer.cs
```diff
@@ -187,21 +187,20 @@ static internal void FromString(
...
-            int strLen = goodString.Length;
             int strLenUnit = 0;
             double unitFactor = 1.0;
 
             //  this is where we would handle trailing whitespace on the input string.
             //  peel [unit] off the end of the string
             i = 0;
 
-            if (goodString == UnitStrings[i])
+            if (valueSpan.Equals(UnitStrings[i], StringComparison.OrdinalIgnoreCase))
```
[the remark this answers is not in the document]
ThomasGoulet73: I used Ordinal because the existing comparison uses Ordinal (== uses Ordinal) and I believe the "ignore case" part of OrdinalIgnoreCase is the same as InvariantCultureIgnoreCase so I believe the code is correct.

src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/LengthConverter.cs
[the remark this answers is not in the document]
ThomasGoulet73: Yes the right one is `s.AsSpan().Trim()`, I'll fix the other places.

src/Microsoft.DotNet.Wpf/src/PresentationFramework/PresentationFramework.csproj
```diff
@@ -39,6 +39,7 @@
     <Compile Include="$(WpfCommonDir)src\System\SR.cs">
         <Link>Common\System\SR.cs</Link>
     </Compile>
+    <Compile Include="$(WpfCommonDir)\src\MS\Internal\PixelUnit.cs" Link="MS\Internal\PixelUnit.cs" />
```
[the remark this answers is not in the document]
ThomasGoulet73: Yes, I'll fix it.

src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Markup/XamlGridLengthSerializer.cs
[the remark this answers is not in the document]
lindexi: @ThomasGoulet73 Could you see https://stackoverflow.com/a/72766 ? It says that the OrdinalIgnoreCase  is not same as InvariantCultureIgnoreCase .

ThomasGoulet73:
I updated my branch to main to fix a conflict and pushed a commit with the PR feedback.

src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Markup/XamlGridLengthSerializer.cs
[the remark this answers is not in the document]
miloush: Ordinal[IgnoreCase] is certainly not the same as Invariant[IgnoreCase]. You can see it e.g. on ae and æ. However, I don't think that is a concern for the involved strings here (auto, px, *).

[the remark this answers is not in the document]
lindexi: I think using Ordinal[IgnoreCase] will be better than Invariant[IgnoreCase] in this case.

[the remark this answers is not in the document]
h3xds1nz: Same.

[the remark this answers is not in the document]
ThomasGoulet73: > Ordinal[IgnoreCase] is certainly not the same as Invariant[IgnoreCase]. You can see it e.g. on ae and æ. However, I don't think that is a concern for the involved strings here (auto, px, *).

Sorry if I wasn't clear, what I meant is better explained [here](https://learn.microsoft.com/en-us/dotnet/standard/base-types/best-practices-strings#:~:text=Comparisons%20made%20using%20StringComparison.OrdinalIgnoreCase%20are%20behaviorally%20the%20composition%20of%20two%20calls%3A%20calling%20ToUpperInvariant%20on%20both%20string%20arguments%2C%20and%20doing%20a%20comparison%20using%20StringComparison.Ordinal.).

Basically, it says that `value.ToLowerInvariant().Equals(value2.ToLowerInvariant(), StringComparison.Ordinal)` is supposed to be the same as `value.Equals(value2, StringComparison.OrdinalIgnoreCase)` which is what I meant.

src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Controls/DataGridLengthConverter.cs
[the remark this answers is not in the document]
harshit7962: Yeah, testing English based alphabets, figured out that the case should not have any affect on double.Parse, but wanted to point it out as I wasn't sure due to the involvement of different types of culture here.

harshit7962 (approved):
Overall the changes looks good. Checked performance gain on multiple applications. There is about 4% gain on performance at xaml reading and it reflects out to be around 2% of gain at `LoadComponent` level when averaged out over multiple runs.

h3xds1nz:
Sounds amazing, I can't wait to push a PR benefiting from those changes! 😄

harshit7962:
@ThomasGoulet73 we have encountered a use case where the following changes are causing a regression. Please find the minimal repro [here](https://github.com/harshit7962/CTP_Failures/tree/main/Sep24_Failure). 

A bit of context, and summarize the failure, the Height specified to `Figure` tag here does not take `Content` as a unit with these changes.

Can you please look if we can mitigate this failure. I believe we might need changes in `XamlFigureLengthSerializer` here and might have to possibly check for similar use case in `XamlGridLengthSerializer`.

ThomasGoulet73:
Hi @harshit7962, I think I found the problem. Here it should've been `StringComparison.OrdinalIgnoreCase` instead of `StringComparison.Ordinal`: https://github.com/dotnet/wpf/pull/8622/files#diff-e8d1ce9b0355662f53afd3f319bfcb23d2d2c8b1baa37572c7daacfed58bbf83R213

I looked at my other changes and they all use `StringComparison.OrdinalIgnoreCase` correctly, I'm not sure how I missed this one.

I'll test to make sure it fixes your repro later today.

ThomasGoulet73:
@harshit7962: I was able to reproduce the issue using your minimal repro and I have confirmed that the suspected fix in my previous comment did fix the issue.

Thank you for the simple repro!

src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/LengthConverter.cs
```diff
@@ -181,60 +181,46 @@ public override object ConvertTo(ITypeDescriptorContext typeDescriptorContext,
...
             double unitFactor = 1.0;
 
             //Auto is represented and Double.NaN
             //properties that do not want Auto and NaN to be in their ligit values,
             //should disallow NaN in validation callbacks (same goes for negative values)
-            if (goodString == "auto") return Double.NaN;
-
-            for (int i = 0; i < PixelUnitStrings.Length; i++)
+            if (valueSpan.Equals("auto", StringComparison.OrdinalIgnoreCase))
```
h3xds1nz, line 190: Could we change this to "Auto"? There are no "auto" strings in the assembly after your changes than this one.

harshit7962 (approved):

harshit7962:
@ThomasGoulet73 verified that the edge case is fixed with the change. Thank you for your contributions.

ThomasGoulet73:
Thanks @harshit7962

PR #8643, opened 2024-01-05
Do not count invisible UI elements towards size of/position in set

### LINKED ISSUE #8642: PositionInSet and SizeOfSet do not take into account hidden/collapsed elemnts
### Description

Elements that present a collection of items (such as a Menu with MenuItems or a ListView with ListViewItems) do not take into account non-visible elements when calculating PositionInSet and SizeOfSet. This results in screen readers announcing incorrect information.

### Reproduction Steps

Create a WPF project that contains one such element. For example, consider the following XAML:

```xaml
<Window x:Class="BlankWpf.MainWindow"
        xmlns:local="clr-namespace:BlankWpf"
        mc:Ignorable="d"
    <Grid>
        <ListView Width="500">
            <ListViewItem Content="A" />
            <ListViewItem Content="B" Visibility="Collapsed" />
            <ListViewItem Content="C" />
        </ListView>
    </Grid>
</Window>
```

Compile this project, open Windows Narrator, and select one of the two viisble items in the list. 

### Expected behavior

When selecting the first item, Windows Narrator should announce "1 of 2." When selecting the second item (C), Windows Narrator should announce "2 of 2."

### Actual behavior

When selecting the first item, Windows Narrator announces "1 of 3." When selecting the second item (C), Windows Narrator announces "3 of 3."

### Regression?

_No response_

### Known Workarounds

_No response_

### Impact

This is an accessibility issue, since screen readers announce incorrect information.

### Configuration

_No response_

### Other information

_No response_

mslukebo:
Fixes #8642

## Description

PositionInSet and SizeOfSet for items currently count all items in their container, resulting in potentially incorrect information. This PR updates the logic to exclude non-visible UI elements from these counts.

## Customer Impact

Screen readers may report incorrect information with the current implementation.

## Testing

I tested this patched PresentationFramework.dll against the repro mentioned in #8642 and confirmed Windows Narrator reads the expected values.

## Risk

These changes only effect non-grouping paths. I am not familiar with how the grouped collections calculate set size/position, so those values may still be inaccurate.

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Automation/Peers/ItemAutomationPeer.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Automation/Peers/ItemAutomationPeer.cs
@@ -338,12 +338,29 @@ namespace System.Windows.Automation.Peers
         {
             int position = AutomationProperties.AutomationPositionInSetDefault;
             ItemCollection itemCollection = itemsControl.Items;
-            position = itemCollection.IndexOf(item);
 
             if (itemsControl.IsGrouping)
             {
                 int sizeOfGroup;
                 position = FindPositionInGroup(itemCollection.Groups, position, out sizeOfGroup);
+            }
+            else
+            {
+                position = itemCollection.IndexOf(item);
+
+                // Some items may not be visible, so we don't want to count
+                foreach (var child in (itemCollection.CollectionView))
+                {
+                    if (item == child)
+                    {
+                        break;
+                    }
+
+                    if (child is not UIElement element || element.Visibility != Visibility.Visible)
+                    {
+                        position -= 1;
+                    }
+                }
             }
 
             return position + 1;
@@ -362,6 +379,15 @@ namespace System.Windows.Automation.Peers
             else
             {
                 size = itemCollection.Count;
+
+                // Some items may not be visible, so we don't want to count
+                foreach (var child in (itemCollection.CollectionView))
+                {
+                    if (child is not UIElement element || element.Visibility != Visibility.Visible)
+                    {
+                        size -= 1;
+                    }
+                }
             }
 
             return size;
--- a/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Automation/Peers/ItemAutomationPeer.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Automation/Peers/ItemAutomationPeer.cs
@@ -348,15 +348,15 @@ namespace System.Windows.Automation.Peers
             {
                 position = itemCollection.IndexOf(item);
 
-                // Some items may not be visible, so we don't want to count
-                foreach (var child in (itemCollection.CollectionView))
+                // Some items may not be visible, so we don't want to count them
+                foreach (var child in itemCollection.CollectionView)
                 {
                     if (item == child)
                     {
                         break;
                     }
 
-                    if (child is not UIElement element || element.Visibility != Visibility.Visible)
+                    if (child is UIElement element && element.Visibility != Visibility.Visible)
                     {
                         position -= 1;
                     }
@@ -380,10 +380,10 @@ namespace System.Windows.Automation.Peers
             {
                 size = itemCollection.Count;
 
-                // Some items may not be visible, so we don't want to count
-                foreach (var child in (itemCollection.CollectionView))
+                // Some items may not be visible, so we don't want to count them
+                foreach (var child in itemCollection.CollectionView)
                 {
-                    if (child is not UIElement element || element.Visibility != Visibility.Visible)
+                    if (child is UIElement element && element.Visibility != Visibility.Visible)
                     {
                         size -= 1;
                     }


src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Automation/Peers/ItemAutomationPeer.cs
```diff
@@ -338,13 +338,30 @@ internal static int GetPositionInSetFromItemsControl(ItemsControl itemsControl,
         {
             int position = AutomationProperties.AutomationPositionInSetDefault;
             ItemCollection itemCollection = itemsControl.Items;
-            position = itemCollection.IndexOf(item);
 
             if (itemsControl.IsGrouping)
             {
                 int sizeOfGroup;
                 position = FindPositionInGroup(itemCollection.Groups, position, out sizeOfGroup);
```
halgab, line 345: This line will be affected by the deleted line above unless I'm missing something. If I read your comment, I understand it was not your intention though

mslukebo: Good catch - did not realize this variable was used on this line. Re-added the deleted line.

mslukebo:
I did a quick search in the PresentationFramework library to see if anyone else calls these methods, and found a special case for MenuItems. With the changes from #1977, my original changes would cause menu items with invisible separators to under-report size of/position in menu item sets. I updated the MenuItem automation peer code to only subtract *visible* separators from the count. Please let me know if there's a better way to deal with this case.

src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Automation/Peers/MenuItemAutomationPeer.cs
```diff
@@ -104,7 +104,10 @@ override protected int GetSizeOfSetCore()
 
                 foreach (var item in parent.Items)
                 {
-                    if (item is Separator)
+                    // ItemAutomationPeer.GetSizeOfSetFromItemsControl excludes invisible UI elements
+                    // from its count, so we only exclude separators that are visible. Otherwise, we would
+                    // double count invisible separators, leading to a lower count than expected.
+                    if (item is Separator separator && separator.Visibility == Visibility.Visible)
```
halgab, line 110:
```suggestion
                    if (item is Separator { Visibility: Visibility.Visible })
```
and save a local declaration

src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Automation/Peers/MenuItemAutomationPeer.cs
```diff
@@ -126,21 +129,25 @@ override protected int GetSizeOfSetCore()
...
                     if (item == owner)
                     {
                         break;
                     }
-                    if (item is Separator)
+
+                    // ItemAutomationPeer.GetPositionInSetFromItemsControl excludes invisible UI elements
+                    // from its count, so we only exclude separators that are visible. Otherwise, we would
+                    // double count invisible separators, leading to a lower position than expected.
+                    if (item is Separator separator && separator.Visibility == Visibility.Visible)
```
halgab, line 150:
```suggestion
                    if (item is Separator { Visibility: Visibility.Visible })
```

src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Automation/Peers/ItemAutomationPeer.cs
```diff
@@ -345,6 +345,22 @@ internal static int GetPositionInSetFromItemsControl(ItemsControl itemsControl,
...
+            {
+                // Some items may not be visible, so we don't want to count them
+                foreach (var child in itemCollection.CollectionView)
+                {
+                    if (item == child)
+                    {
+                        break;
+                    }
+
+                    if (child is UIElement element && element.Visibility != Visibility.Visible)
```
halgab, line 358:
```suggestion
                    if (child is UIElement { Visibility: not Visibility.Visible })
```

src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Automation/Peers/ItemAutomationPeer.cs
```diff
@@ -362,6 +378,15 @@ internal static int GetSizeOfSetFromItemsControl(ItemsControl itemsControl, obje
             else
             {
                 size = itemCollection.Count;
+
+                // Some items may not be visible, so we don't want to count them
+                foreach (var child in itemCollection.CollectionView)
+                {
+                    if (child is UIElement element && element.Visibility != Visibility.Visible)
```
halgab, line 385:
```suggestion
                    if (child is UIElement { Visibility: not Visibility.Visible })
```

src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Automation/Peers/MenuItemAutomationPeer.cs
mslukebo: What would the local declaration be used for in this scenario?

mslukebo: Or did you mean taking this suggestion saves the local declaration?

halgab: Exactly

rchauhan18:
@mslukebo Thanks for adding a fix to resolve the issue. 
The current solution will work for the sample mentioned in the issue, where the ```ListItem``` are directly added.

However, in case when we bind the data to the List and use ```IsVisible``` property of an object to make the ListItem visible. The narrator is still calling the wrong sequence and total count.
This is because in this case the ```child``` is an object of a class that is bound to ListItem and will always return false for this condition.

```
if (child is UIElement { Visibility: not Visibility.Visible })
                    {
                        position -= 1;
                    }
 ```
 Here is the sample code:

```
 <ListView Name="myListView" ItemsSource="{Binding YourCollection}">
    <ListView.View>
        <GridView>
            <GridViewColumn Header="Name" DisplayMemberBinding="{Binding Name}" />
        </GridView>
    </ListView.View>
    <ListView.ItemContainerStyle>
        <Style TargetType="ListViewItem">
            <Style.Triggers>
                <DataTrigger Binding="{Binding IsVisible}" Value="False">
                    <Setter Property="Visibility" Value="Collapsed" />
                </DataTrigger>
            </Style.Triggers>
        </Style>
    </ListView.ItemContainerStyle>
</ListView>
```

**Comment by mslukebo** [comment]
> @mslukebo Thanks for adding a fix to resolve the issue. The current solution will work for the sample mentioned in the issue, where the `ListItem` are directly added.
> 
> However, in case when we bind the data to the List and use `IsVisible` property of an object to make the ListItem visible. The narrator is still calling the wrong sequence and total count. This is because in this case the `child` is an object of a class that is bound to ListItem and will always return false for this condition.
> 
> ```
> if (child is UIElement { Visibility: not Visibility.Visible })
>                     {
>                         position -= 1;
>                     }
> ```
> 
> Here is the sample code:
> 
> ```
>  <ListView Name="myListView" ItemsSource="{Binding YourCollection}">
>     <ListView.View>
>         <GridView>
>             <GridViewColumn Header="Name" DisplayMemberBinding="{Binding Name}" />
>         </GridView>
>     </ListView.View>
>     <ListView.ItemContainerStyle>
>         <Style TargetType="ListViewItem">
>             <Style.Triggers>
>                 <DataTrigger Binding="{Binding IsVisible}" Value="False">
>                     <Setter Property="Visibility" Value="Collapsed" />
>                 </DataTrigger>
>             </Style.Triggers>
>         </Style>
>     </ListView.ItemContainerStyle>
> </ListView>
> ```

I see. What is the proper way to handle this case then? Is there a way to get the item container for a child?

**Comment by mslukebo** [comment]
@rchauhan18 I did some testing and can address the case you describe by using code similar to the following:

```cs
// Some items may not be visible, so we don't want to count them
foreach (var child in itemCollection.CollectionView)
{
    if (child is UIElement { Visibility: not Visibility.Visible })
    {
        size -= 1;
        continue;
    }

    var container = itemsControl.ItemContainerGenerator.ContainerFromItem(child);
    if (container == null || container is UIElement { Visibility: not Visibility.Visible })
    {
        size -= 1;
        continue;
    }
}
```

This fixes the "x of y" that narrator says, but doesn't address the grid/table size.

My test case looks like

```cs
{
    public List<MyObject> MyObjects { get; }

    public MainWindow()
    {
        InitializeComponent();
        this.MyObjects = new List<MyObject>()
        {
            new MyObject("A", true),
            new MyObject("B", false),
            new MyObject("C", true),
        };

        this.DataContext = this;
    }

    public record MyObject(string Name, bool IsVisible);
}
```

```xaml
<ListView Width="500" ItemsSource="{Binding MyObjects}">
    <ListView.View>
        <GridView>
            <GridViewColumn Header="Name" DisplayMemberBinding="{Binding Name}" />
        </GridView>
    </ListView.View>
    <ListView.ItemContainerStyle>
        <Style TargetType="ListViewItem">
            <Style.Triggers>
                <DataTrigger Binding="{Binding IsVisible}" Value="False">
                    <Setter Property="Visibility" Value="Collapsed" />
                </DataTrigger>
            </Style.Triggers>
        </Style>
    </ListView.ItemContainerStyle>
</ListView>
```

 When tabbing into the listview, it'll say "Enter table 3 by 1 (...) selected 1 of 2". This is likely because `GridViewAutomationPeer`'s `ITableProvider` implementation doesn't take into account invisible rows when calculating `IGridProvider.RowCount`. There are likely other implementations of this interface that would need to take invisible items into account too, such as `TableAutomationPeer` and `DataGridAutomationPeer` (really anything that implements `IGridProvider`).

Options therefore are
1. Ignore the case mentioned above for this PR
2. Update my PR with the new `itemsControl.ItemContainerGenerator.ContainerFromItem` check to address the sizeof/positionin set issue. Ignore the `IGridProvider.RowCount` issue and solve with a separate PR
3. Try to address the `IGridProvider.RowCount` issue in this PR as well

What would you like me to do?

rchauhan18:
@mslukebo I am adding a sample application which contain 4 cases.
https://github.com/rchauhan18/A11y-TestApplication/tree/master
My suggestion is that the fix should work for all these cases.

### Case 1: ListItem directly added
Your current code is passing for this case.


### Case 2: ListItem with data binding
Your approach of obtaining the item container from the object and checking its visibility should work in this case.
Regarding "Enter table" count, I think so we can add the similar code to ```GridProvider.RowCount``` 
Here is a suggestion: 

```cs
/// <summary>
/// number of rows in the grid
/// </summary>
int IGridProvider.RowCount
{
    get
    {
        int rowcount = _listview.Items.Count;
        foreach(var item in _listview.Items)
        {
            if (item is UIElement { Visibility: not Visibility.Visible })
            {
                rowcount -= 1;
                continue;
            }

            var container = _listview.ItemContainerGenerator.ContainerFromItem(item);
            if (container == null || container is UIElement { Visibility: not Visibility.Visible })
            {
                rowcount -= 1;
                continue;
            }
        }
        return rowcount;
    }
}
```


### Case 3: Using ListBox items
Similar as ListView. It is working properly if we add items directly to the ListBox, but failing when using data binding without the update.


### Case 4: Grouping ListItem 
 In this case the narrator is failing to announce the correct order. 
 This is because in case of grouping, the position and size are returned by ``` FindPositionInGroup```.
 Suggestion: 
We can solve this similarly as above case. We can remove the count of non-visible items.

In WPF when there is a grouping, narrator announces the name and position of group, then announces the position item and size of the group.
Example:
 " [Name of the group] group, [Group Position] of [Total number of groups], Enter Table [Total size of table]  ItemName [position in group] and [size of group] "

mslukebo:
@rchauhan18 thanks for the suggestions. My question is if you want this PR to address
1, Only ListItems/ListBoxItems added directly (i.e. the current state of my PR)
2. ListItems/ListBoxItems added both directly or with data binding (i.e. update with my suggestion in my previous comment)
3. All of the above, plus grouping
4. All of the above, plus fixing `GridProvider.RowCount` implementations

I'm not _too_ familiar with this codebase, so I'd prefer if my PR only addressed 1 and/or 2 and left the rest for people more familiar with the solutions. If you want, I can take a stab fixing all the issues in this PR though.

Kuldeep-MS:
Hi @mslukebo, thank you for your contribution. While we appreciate your suggestion, we prefer to merge pull requests that include a full fix unless the issue is urgent or particularly difficult to resolve. That being said, you don't need to fix all four cases mentioned in your pull request - we'll take care of it from here. 

@rchauhan18, could you please update the pull request to include fixes for those four cases?

mslukebo:
> @mslukebo Will it be okay with you if we add the suggested fix on your branch?

Yes, please feel free to work on top of my changes.

src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Automation/Peers/ItemAutomationPeer.cs
```diff
@@ -396,6 +439,34 @@ private static int FindPositionInGroup(ReadOnlyObservableCollection<object> coll
                         else
                         {
                             newCollection = currentGroup.Items;
+                        }                 
+                        foreach (var child in group.Items)
+                        {
+
```
hiteshwpfmsft, line 445: nit - remove extra newlines.

src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Automation/Peers/MenuItemAutomationPeer.cs
```diff
@@ -104,7 +104,10 @@ override protected int GetSizeOfSetCore()
 
                 foreach (var item in parent.Items)
                 {
-                    if (item is Separator)
+                    // ItemAutomationPeer.GetSizeOfSetFromItemsControl excludes invisible UI elements
+                    // from its count, so we only exclude separators that are visible. Otherwise, we would
+                    // double count invisible separators, leading to a lower count than expected.
+                    if (item is Separator { Visibility: Visibility.Visible })
```
miloush, line 110: [IsVisible](https://learn.microsoft.com/en-us/dotnet/api/system.windows.uielement.isvisible) sounds like better firt for this purpose

miloush: (applies to all the visibiltiy tests)

src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Automation/Peers/GridViewAutomationPeer.cs
```diff
@@ -195,7 +195,26 @@ int IGridProvider.ColumnCount
...
+                int rowcount = _listview.Items.Count;
+                foreach(var item in _listview.Items)
+                {
+                    if (item is UIElement { Visibility: not Visibility.Visible })
+                    {
+                        rowcount -= 1;
+                        continue;
+                    }
+        
+                    var container = _listview.ItemContainerGenerator.ContainerFromItem(item);
```
miloush, line 209: Doesn't this return the item if it is the container? Also the item being UIElement does not necessarily mean it is part of the tree.

miloush: For the record this was is the example showing that checking visibility on items is misleading:

```C#
<ListView>
    <ListView.View>
        <GridView>
            <GridViewColumn Header="Visibility">
                <GridViewColumn.CellTemplate>
                    <DataTemplate>
                        <TextBlock Text="{Binding Visibility}" />
                    </DataTemplate>
                </GridViewColumn.CellTemplate>
            </GridViewColumn>
        </GridView>
    </ListView.View>
    <ListView.Items>
        <TextBlock Visibility="Visible" Text="TB visible" />
        <TextBlock Visibility="Hidden" Text="TB hidden" />
        <TextBlock Visibility="Collapsed" Text="TB collapsed" />
        <TextBlock Visibility="Visible" Text="TB visible" />
    </ListView.Items>
</ListView>
```

This will show all 4 items despite them being UIElements with not Visible Visibility.

miloush: > Doesn't this return the item if it is the container?

That was a performance concern where you would be checking an element twice. However, the first check of an item directly should simply be dropped, so it's fine.

miloush:
Let's make it clear this is always going to be a best effort. Developer can trivially "circumvent" the counting, make the item empty, non-selectable etc.

However, I am not quite sure the original issue is valid. The item _is_ in the collection, whether it's visible or not does not change that fact.

miloush:
See the [documentation](https://learn.microsoft.com/en-us/windows/win32/api/uiautomationcore/nf-uiautomationcore-igridprovider-get_rowcount):

> Hidden rows and columns, depending on the provider implementation, may be loaded in the logical tree and will therefore be reflected in the IGridProvider::RowCount and [IGridProvider::ColumnCount](https://learn.microsoft.com/en-us/windows/desktop/api/uiautomationcore/nf-uiautomationcore-igridprovider-get_columncount) properties. If the hidden rows and columns have not yet been loaded they will not be counted.

src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Automation/Peers/GridViewAutomationPeer.cs
rchauhan18: > For the record this is an example showing that checking visibility on items is misleading:
> 
> ```
> <ListView>
>     <ListView.View>
>         <GridView>
>             <GridViewColumn Header="Visibility">
>                 <GridViewColumn.CellTemplate>
>                     <DataTemplate>
>                         <TextBlock Text="{Binding Visibility}" />
>                     </DataTemplate>
>                 </GridViewColumn.CellTemplate>
>             </GridViewColumn>
>         </GridView>
>     </ListView.View>
>     <ListView.Items>
>         <TextBlock Visibility="Visible" Text="TB visible" />
>         <TextBlock Visibility="Hidden" Text="TB hidden" />
>         <TextBlock Visibility="Collapsed" Text="TB collapsed" />
>         <TextBlock Visibility="Visible" Text="TB visible" />
>     </ListView.Items>
> </ListView>
> ```
> 
> This will show all 4 items despite them being UIElements with not Visible Visibility.

Yes, this is a valid case and our current fix won't work. It will give an incorrect result.
Thanks for bringing this up.

rchauhan18:
> Let's make it clear this is always going to be a best effort. Developer can trivially "circumvent" the counting, make the item empty, non-selectable etc.
> 
> However, I am not quite sure the original issue is valid. The item _is_ in the collection, whether it's visible or not does not change that fact.

@miloush I agree with you that we cannot use the visibility property to update the RowCount. 

According to documentation, RowCount should be the count of all the element loaded in the logical tree. This is as per the design.
Thanks for sharing the document.

rchauhan18:
@mslukebo 
On further investigation, I found that the current fix won't work in WPF as it will return incorrect count in few cases (one is mentioned by @miloush)
In case of Virtualization the current fix will fail as only few items will be loaded as a UI Element and other which are not loaded will be considered as not visible despite their property set to visible and the count will be incorrect.

I have a workaround, 
to update the ```PositionInSet``` and ```SizeOfSet``` in the application itself after loading.

```cs
private void UpdatePositionInSet(object sender, RoutedEventArgs e)
{
    int position = 1;
    int size = 0;
    foreach (var item in MyItemCollection)
    {
        if (item.IsVisible)
        {
            var container = myListView.ItemContainerGenerator.ContainerFromItem(item);
            AutomationProperties.SetPositionInSet(myListView.ItemContainerGenerator.ContainerFromItem(item) as UIElement, position++);
            size++;
        }
    }
    foreach (var item in MyItemCollection)
    {
        AutomationProperties.SetSizeOfSet(myListView.ItemContainerGenerator.ContainerFromItem(item) as UIElement, size);
    }
        
}
```
Sample application: https://github.com/rchauhan18/ListViewSample
Can you confirm, will the workaround work for you?

pchaurasia14:
Closing this PR as this would incur a change in existing design.

PR #8520, opened 2023-12-08
Prepare System.Xaml for nullable reference type annotations

halgab:
Contributes to #7563

## Description

We activate the NRT annotations for System.Xaml.csproj, then disable it in all files so that we can annotate them one by one. The second commit annotates one file to showcase the process.

## Customer Impact

None so far

## Testing

CI

## Risk

Very low. Annotations are not really activated so far, except for an internal class.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/wpf/pull/8520)

PR #8700, opened 2024-01-18
Disable nullability in System.Xaml unit test to fix build

rchauhan18:
Disable nullability in System.Xaml unit test to fix build of https://github.com/dotnet/wpf/pull/8695
Will revert these changes when nullability is completed for System.Xaml

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/wpf/pull/8700)


rchauhan18:
FYI @hughbe

hughbe:
Should we disable nullability at the assembly level?

rchauhan18:
> Should we disable nullability at the assembly level?

@hughbe I disabled it at the file level because we're enabling NRT annotation for System.Xaml.csproj and System.Xaml.Test.csproj. As part of this process, we need to disable it in all files and then annotate them individually.
PR: https://github.com/dotnet/wpf/pull/8520

PR #8739, opened 2024-01-28
Prefer use of interpolated strings in PresentationUI and ReachFramework

halgab:
Follow-up to #8519

## Description

We replace as many manual string concatenations and string.Format with interpolated strings. This saves allocations in many instances, and makes for a more readable code.

## Customer Impact

Less allocations

## Regression

No

## Testing

CI

## Risk

Low. Most changes are automated using the IDE


src/Microsoft.DotNet.Wpf/src/PresentationUI/MS/Internal/Documents/Application/DocumentStream.cs
```diff
@@ -803,8 +803,10 @@ protected override void Dispose(bool disposing)
 #if DEBUG
             Invariant.Assert(
                 ((i != 3) || (temporary != null) || (inSameFolder)),
-                "Unable to create a temp file.\n"
-                + "Unless IE Cache is read-only we have a defect.");
+                """
+                Unable to create a temp file.
```
lindexi, line 807: Will it become `\r\n` ?

src/Microsoft.DotNet.Wpf/src/PresentationUI/MS/Internal/Documents/SignatureResourceHelper.cs
```diff
@@ -89,12 +89,7 @@ internal static class SignatureResourceHelper
                 sigStatus = SignatureStatus.Invalid;
             }
 
-            string resourceName = string.Format(
-                CultureInfo.InvariantCulture,
-                @"{0}_{1}x{2}",
-                sigStatus.ToString(),
-                height,
-                width);
+            string resourceName = $@"{sigStatus}_{height}x{width}";
```
lindexi, line 92: Missing InvariantCulture here.

halgab: In the winforms repo, when doing such refactorings, a choice was made to ignore the invariant culture for int. I am no specialist but apparently it has no impact on integer formatting. I can put it back just for peace of mind

src/Microsoft.DotNet.Wpf/src/PresentationUI/MS/Internal/Documents/Application/DocumentStream.cs
halgab: Yes, but it's an exception message, so I don't think the change is a problem

src/Microsoft.DotNet.Wpf/src/PresentationUI/MS/Internal/Documents/SignatureResourceHelper.cs
miloush: Correct, as long as there is no format requested, the culture is not used (other than for negative sign, which shouldn't be an issue for width/height). 

https://github.com/dotnet/runtime/blob/2f43856e78faebe7347fdd76acc26de7ffaec539/src/libraries/System.Private.CoreLib/src/System/Number.Formatting.cs#L919-L925

src/Microsoft.DotNet.Wpf/src/ReachFramework/PrintConfig/PTProvider.cs
```diff
@@ -238,12 +238,8 @@ public override MemoryStream GetPrintCapabilities(MemoryStream printTicket)
 
                     if (hResult == (uint)NativeErrorCode.E_PRINTTICKET_FORMAT)
                     {
-                        throw new ArgumentException(String.Format(CultureInfo.CurrentCulture,
-                                      "{0} {1} {2}",
-                                      PrintSchemaTags.Framework.PrintTicketRoot,
-                                      PTUtility.GetTextFromResource("FormatException.XMLNotWellFormed"),
-                                      errorMsg),
-                                      "printTicket");
+                        throw new ArgumentException($"{PrintSchemaTags.Framework.PrintTicketRoot} {PTUtility.GetTextFromResource("FormatException.XMLNotWellFormed")} {errorMsg}",
```
dipeshmsft, line 241: I prefer the previous way here. The current change results in too long interpolation. Same suggestion for the change below and `FallbackPTProvider.cs`

src/Microsoft.DotNet.Wpf/src/PresentationUI/MS/Internal/Documents/Application/AddressUtility.cs
```diff
@@ -54,11 +54,7 @@ internal static string GetEmailAddressFromMailtoUri(Uri mailtoUri)
             if (IsMailtoUri(mailtoUri))
             {
                 // Create the address from the URI
-                address = string.Format(
-                    CultureInfo.CurrentCulture,
-                    _addressTemplate,
-                    mailtoUri.GetComponents(UriComponents.UserInfo, UriFormat.Unescaped),
-                    mailtoUri.GetComponents(UriComponents.Host, UriFormat.Unescaped));                                                   
+                address = $"{mailtoUri.GetComponents(UriComponents.UserInfo, UriFormat.Unescaped)}@{mailtoUri.GetComponents(UriComponents.Host, UriFormat.Unescaped)}";
```
dipeshmsft, line 57: Can we revert this back to the original way ? Same reasoning, too long string.

dipeshmsft:
@halgab, can you take a look at the above PR and make the necessary changes ?

halgab:
Hi. I'm sorry, I'm not sure when I'll have time to get back to this...

dipeshmsft:
@halgab is it okay if I take this forward ?

halgab:
Of course, feel free to do so

Issue #8789, opened 2024-02-08
The GetHashCode implementation of the Thickness struct is broken for uniform length.


davidskula:
### Description

For any uniform length instance of Thickness struct GetHashCode returns zero.

### Reproduction Steps

```
var h1 = new Thickness(5).GetHashCode();
var h2 = new Thickness(10).GetHashCode();
```

### Expected behavior

h1 and h2 should have different values

### Actual behavior

h1 and h2 have the same value

lindexi:
The reason is:

https://github.com/dotnet/wpf/blob/19f227f76e8d3789f1fcb762410c56d88584b88f/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Thickness.cs#L96-L99

I'm not sure I can change its behavior.

PR #8822, opened 2024-02-18
Try fix GetHashCode implementation of the Thickness struct is broken for uniform length

lindexi:
Fixes https://github.com/dotnet/wpf/issues/8789

## Description

I change the GetHashCode implementation of the Thickness struct. I use the HashCode to calculate the hash code of Thickness.

I'm not sure I'm doing the right thing.

## Customer Impact

The GetHashCode  of Thickness will be change

## Testing

Just CI.

## Risk

Normal. It's hard to evaluate. Because some logic that depends on the original behavior can break down.

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/wpf/pull/8822)

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Thickness.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Thickness.cs
@@ -95,7 +95,13 @@ namespace System.Windows
         /// <returns>Hash code</returns>
         public override int GetHashCode()
         {
-            return _Left.GetHashCode() ^ _Top.GetHashCode() ^ _Right.GetHashCode() ^ _Bottom.GetHashCode();
+            var hashCode = new System.HashCode();
+            hashCode.Add(_Left);
+            hashCode.Add(_Top);
+            hashCode.Add(_Right);
+            hashCode.Add(_Bottom);
+            var code = hashCode.ToHashCode();
+            return code;
         }
 
         /// <summary>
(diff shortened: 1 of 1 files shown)


miloush:
Hash codes are explicitly not guaranteed to remain the same not only across framework versions but even across runs of the same application.

This might be a better way to do this: 
https://github.com/dotnet/winforms/blob/2441a36314a2312e219278e0459ba0aef8a2b7f6/src/System.Windows.Forms.Primitives/src/System/Windows/Forms/Padding.cs#L167

lindexi:
Thank you @miloush and I will use your code.

halgab:
Apparently there are many other places that could switch to such a pattern. Should the scope of this PR be broadened? Or perhaps we could open a follow-up issue to review the `GetHashCode` implementations across the repo?

lindexi:
@halgab At first, I must sure I do the right thing. I do not know change the GetHashCode behavior is the right thing.

dipeshmsft:
@lindexi, I am taking in this PR for CTP, I think the changes are fine, although there may be some issues but overall it seems fine.

h3xds1nz:
It is the recommended way, the only issue is performance but given the fact `Thickness` is a mutable struct (ew), I think it is fine too and fixes the issue.

dipeshmsft:
>  the only issue is performance but given the fact Thickness is a mutable struct (ew), I think it is fine too and fixes the issue.

There is a 2ns increase ( 3ns -> 5ns ) with the change. I think it is okay in that way.

h3xds1nz:
Yeah, I think we will survive. Luckily it's not used a key anywhere.

dipeshmsft (approved):

PR #8834, opened 2024-02-23
Update `MS.Internal.IO.Packaging` classes

onkrot:
Fixes #

Main PR

## Description
Code improvements.
Removed unneeded casts.
Removed ArrayList and HybridDictionary

## Customer Impact

Small performance improvement

## Regression

None

## Testing

None

## Risk

Low risk

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/PresentationCore/MS/internal/IO/Packaging/PreloadedPackages.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationCore/MS/internal/IO/Packaging/PreloadedPackages.cs
@@ -10,16 +10,10 @@
 //
 
 using System;
-using System.Security;
-using System.Collections;
-using System.Collections.Specialized;
-using System.Diagnostics;
-using System.Globalization;
-using System.IO;
+using System.Collections.Generic;
 using System.IO.Packaging;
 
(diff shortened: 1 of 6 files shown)


src/Microsoft.DotNet.Wpf/src/PresentationCore/MS/internal/IO/Packaging/PreloadedPackages.cs
```diff
@@ -237,8 +213,8 @@ internal bool ThreadSafe
         // ListDictionary is the best fit for this scenarios; otherwise we should be using
         // Hashtable. HybridDictionary already has functionality of switching between
         //  ListDictionary and Hashtable depending on the size of the collection
-        static private HybridDictionary _packagePairs;
-        static private readonly Object  _globalLock;
+        static private Dictionary<Uri, PackageThreadSafePair> _packagePairs;
```
lindexi, line 216: Sorry, I do not think replace the HybridDictionary with `Dictionary<Uri, PackageThreadSafePair>` is a good idea. Please see the comment:

> ListDictionary is the best fit for this scenarios; otherwise we should be using Hashtable. HybridDictionary already has functionality of switching between ListDictionary and Hashtable depending on the size of the collection

Issue #6198, opened 2022-03-03
Non-existing StaticResources are cached the second time, no exception is thrown


KirillOsenkov:
[wpf.zip](https://github.com/dotnet/wpf/files/8175381/wpf.zip)

If you build and run the attached WPF project, it will seemingly work fine with no exceptions. However if you open App.xaml and comment out one of the two ResourceDictionary references, you'll get an exception here:
https://github.com/dotnet/wpf/blob/255cdad0b0e637b28305843940585ade55545186/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/StaticResourceExtension.cs#L104

```
Cannot find resource named 'wpf.Test'. Resource names are case sensitive.
```

However if you keep two ResourceDictionary references, there is no exception.

This is because the first time the value is set to `UnsetValue`, and the exception throws.
However for the second dictionary, the value is found in `FindCachedResource` and set back to `null` thanks to the code here:
https://github.com/dotnet/wpf/blob/255cdad0b0e637b28305843940585ade55545186/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/SystemResources.cs#L313-L314

I think this looks like a bug - if it throws with one resource dictionary, it should still throw with two.

However my worry is that this might be a breaking change if we suddenly start throwing where we were silently not throwing before.

I'm just filing this issue for completeness and historic record.

miloush:
Sounds like #2543

lindexi:
I think this looks like a bug too. But if we fix this, there will be a change in behavior.

batzen:
IMHO, it's clearly a bug and should be fixed. Even if the broken default behavior changes.

ThomasGoulet73:
This bug should be fixed by dotnet/wpf#8853. For now, you can workaround this issue by adding a dummy item in `<Application.Resources>`.

Issue #4538, opened 2021-05-19
In.NET 5.0 WPF, a window in a container cannot call a style resource in app.xaml 


lifuhua123:
In.NET 5.0 WPF, a window in a container cannot call a style resource in app.xaml ,
as shown in the figure :
![1](https://user-images.githubusercontent.com/33051107/118790163-16c65a00-b8c8-11eb-99e9-3a187dc03ceb.png)
![2](https://user-images.githubusercontent.com/33051107/118790174-18901d80-b8c8-11eb-92c0-389a1806287c.png)
![3](https://user-images.githubusercontent.com/33051107/118790181-1a59e100-b8c8-11eb-943e-8f5a596d7d8a.png)
![4](https://user-images.githubusercontent.com/33051107/118790193-1b8b0e00-b8c8-11eb-8ac0-993a53bd0ea8.png)

miloush:
This reproduces in 4.8.

This seems to be a bug in the compiler actually. When `Application.StartupUri` is not in _App.xaml_, the compiler does not generate and include app.baml in the assembly resources.

miloush:
...which seems to be a intentional optimization:

https://github.com/dotnet/wpf/blob/89d172db0b7a192de720c6cfba5e28a1e7d46123/src/Microsoft.DotNet.Wpf/src/PresentationBuildTasks/MS/Internal/MarkupCompiler/ParserExtension.cs#L163-L180

miloush:
@lifuhua123 and this code also hints of a workaround: set any property on the `Application` element, for example explicitly include `ShutdownMode="OnLastWindowClose"`.

ThomasGoulet73:
I think this issue is a duplicate of this one: https://github.com/dotnet/wpf/issues/4457. As I said in this comment: https://github.com/dotnet/wpf/issues/4457#issuecomment-827078499, the issue is caused by PresentationBuildTasks.dll so it will only apply to SDK-style csproj (.Net Framework or .Net Core) but this issue does not reproduce when using the old csproj style of .Net Framework.

miloush:
This one reproduces for me with the old project file. That said, I agree this looks caused by the same underlying issue.

ThomasGoulet73:
@miloush You're right, just tested it again and I was able to reproduce this issue and #4457 using the old project style. I think I just forgot to remove StartupUri in my test.

miloush:
[...]

Issue #4457, opened 2021-04-25
Application styles are only applied if there is more than one style when StartupUri is not used


jnm2:
* .NET Core Version: 5.0.202
* Windows version: Windows 10 20H2
* Does the bug reproduce also in WPF for .NET Framework 4.8?: Yes

The application style has no effect, but it should have an effect. The style starts taking effect as soon as a second unrelated style is added before or after it.

This causes a long search for ways to fix the problem and is not beginner-friendly.

![image](https://user-images.githubusercontent.com/8040367/116013499-85343700-a5fe-11eb-83d8-da5d2a0cca52.png)

## Full repro created from VS template

Download: [Repro.zip](https://github.com/dotnet/wpf/files/6373451/Repro.zip)

```xaml
<Application x:Class="WpfRepro.App"
             xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
             xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml">
    <Application.Resources>
        <Style TargetType="TextBox">
            <Setter Property="Background" Value="Red"/>
        </Style>
        <!-- Uncomment to see the style above start taking effect -->
        <!-- <Style x:Key="Workaround" /> -->
    </Application.Resources>
</Application>
```

```xaml
<Window x:Class="WpfRepro.MainWindow"
        xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
        xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
        xmlns:d="http://schemas.microsoft.com/expression/blend/2008"
        xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006"
        xmlns:local="clr-namespace:WpfRepro"
        mc:Ignorable="d"
        Title="MainWindow" Height="450" Width="800">
    <TextBox />
</Window>
```

```cs
using System.Windows;

namespace WpfRepro
{
    public partial class App : Application
    {
        protected override void OnStartup(StartupEventArgs e)
        {
            new MainWindow(startupUriIsNotAnOption: 42).Show();
        }
    }
}
```

```xml
<Project Sdk="Microsoft.NET.Sdk">

  <PropertyGroup>
    <OutputType>WinExe</OutputType>
    <TargetFramework>net5.0-windows</TargetFramework>
    <UseWPF>true</UseWPF>
  </PropertyGroup>

</Project>
```

daisyTian0517:
[...]

Issue #2926, opened 2020-04-23
App.xaml does not generate initialize method when only resource dictionary applied using MVVM locator


marik91:
* .NET Core Version: 3.1 
* Visual Studio version: VS 16.5.4
* Windows version: 18363.778
* Does the bug reproduce also in WPF for .NET Framework 4.8?:  Not tested
 
 **Problem description:**
 when using MVVM pattern StartupUri property is deleted from app.xaml, instead in application.resource resource dictionary is added and static resource for view model locator is added. then resolving view and trying to call show on it will throw runtime exception "Cannot find resource named 'Locator'. Resource names are case sensitive.". Turns out when only single static resource is added InitializeComponent method is not generated. Work around inside resource dictionary add any mapping to merged resource dictionary then InitializeComponent method is generated and application works as expected

 **Expected behavior:**
When added locator resource, InitializeComponent method should be generated 

 **Minimal repro:**
Was using Host class to enable core IoC and Mvvmlight
// App.xaml: not generating InitializeComponent
```
<Application x:Class="WpfApp1.App"
             xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
             xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
             xmlns:local="clr-namespace:WpfApp1"
             >
    <Application.Resources>
         <ResourceDictionary>
             <local:ViewModelLocator x:Key="Locator" />
         </ResourceDictionary>
    </Application.Resources>
</Application>
```

// App.xaml generating InitializeComponent
```
<Application x:Class="WpfApp1.App"
             xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
             xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
             xmlns:local="clr-namespace:WpfApp1"
             >
    <Application.Resources>
         <ResourceDictionary>
             <local:ViewModelLocator x:Key="Locator" />
             <ResourceDictionary.MergedDictionaries>
                <ResourceDictionary Source="styles.xaml" /> <!--empty resource dictionary -->
            </ResourceDictionary.MergedDictionaries>
         </ResourceDictionary>
    </Application.Resources>
</Application>
```

// MainWindow.xaml
```
<Window x:Class="WpfApp1.MainWindow"
        xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
        xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
[...]

PR #8853, opened 2024-02-29
Fix Xaml compilation generating an empty InitializeComponent method when App contains a single item

ThomasGoulet73:
Fixes dotnet/wpf#2543
Fixes dotnet/wpf#2926
Fixes dotnet/wpf#4457
Fixes dotnet/wpf#4538
Fixes dotnet/wpf#6198

## Description
There were many bug reports of different type that were all caused by the same bug. Xaml compilation generating an empty InitializeComponent method when App contains a single item.

Some context:
When generating App.InitializeComponent for a Xaml where there is only the StartupUri property, it will skip loading Baml in App.InitializeComponent and only set the StartupUri property in the generated code for performance reasons. There is currently a bug when detecting if the StartupUri property is the only property that would be loaded from the Baml.

See https://github.com/dotnet/wpf/pull/8853#discussion_r1507014370 for an explanation of the changes in this PR.

## Customer Impact
Should fix issues linked at the top (Might be more since the bug does not always surface the same way)

## Regression
Yes and no, the bug was not present in non-SDK style .Net Framework projects.

## Testing
Manual testing using the samples provided in the issues and manual testing with my own samples with multiple Xaml combination.

## Risk
Low to medium. Worst case scenario this PR could cause existing project that were not loading Baml to now load Baml which could cause slight perf regression but I wasn't able to cause this in my testing.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/wpf/pull/8853)

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/PresentationBuildTasks/MS/Internal/MarkupCompiler/ParserExtension.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationBuildTasks/MS/Internal/MarkupCompiler/ParserExtension.cs
@@ -517,7 +517,7 @@ namespace MS.Internal
                         xmlReader.MoveToAttribute(attrName);
                 }
             }
-            else if (!_compiler.IsBamlNeeded && !_compiler.ProcessingRootContext && _compiler.IsCompilingEntryPointClass && xmlReader.Depth > 0)
+            else if (!_compiler.IsBamlNeeded && _compiler.IsCompilingEntryPointClass && xmlReader.Depth > 0)
             {
                 if ((!localName.Equals(MarkupCompiler.CODETAG) &&
                      !localName.Equals($"{MarkupCompiler.CODETAG}Extension")) ||


src/Microsoft.DotNet.Wpf/src/PresentationBuildTasks/MS/Internal/MarkupCompiler/ParserExtension.cs
```diff
@@ -517,7 +517,7 @@ public override void WriteElementEnd(XamlElementEndNode xamlEndObjectNode)
                         xmlReader.MoveToAttribute(attrName);
                 }
             }
-            else if (!_compiler.IsBamlNeeded && !_compiler.ProcessingRootContext && _compiler.IsCompilingEntryPointClass && xmlReader.Depth > 0)
```
ThomasGoulet73, line 520: This code is called too early to call `_compiler.ProcessingRootContext`.

Basically, there are two steps involved here:
1. Reading a Xaml element.
2. Parsing the Xaml element.

`_compiler.ProcessingRootContext` will only be set in step 2 but this code here is ran in step 1. This means that we get its value before it was set. Since both steps are ran for each Xaml element, when we have 2 or more element we execute this code here with `_compiler.ProcessingRootContext` initialized.

We can remove the check to `_compiler.ProcessingRootContext` because it was already superfluous because the check here `ProcessedRootElement`:
https://github.com/dotnet/wpf/blob/c9a4c6b4a3974df86ffac8d21af1b488c63c630a/src/Microsoft.DotNet.Wpf/src/PresentationBuildTasks/MS/Internal/MarkupCompiler/ParserExtension.cs#L414
basically does the same check as `_compiler.ProcessingRootContext` but `ProcessedRootElement` is set in step 1 instead of step 2.

I hope my explanation was clear enough to explain my reasoning. Feel free to ask question and I'll try to answer though the Xaml compilation process is very complex and I might not be able to properly explain it.  It makes sense in my head 😄.

anjali-wpf: Thanks for your contribution @ThomasGoulet73 , this change simplifies the code by removing a redundant check to _compiler.ProcessingRootContext, which is called too early in the XAML compilation process to have its value set. The existing check to ProcessedRootElement already serves the same purpose and is set earlier in the process, making the _compiler.ProcessingRootContext check unnecessary. This improves code clarity and avoids potential confusion.
Possibly you can add this in the description as well for better understanding.

ThomasGoulet73: @anjali-wpf I found it easier to comment on the changes on this specific line. I added a link to this comment in the description to improve clarity.

anjali-wpf (approved):

hiteshwpfmsft (approved):

src/Microsoft.DotNet.Wpf/src/PresentationBuildTasks/MS/Internal/MarkupCompiler/ParserExtension.cs
dipeshmsft: @ThomasGoulet73, can you give me a general idea of where this function is being called in both the stages ?

ThomasGoulet73: Sure, I just need to refresh my memory a bit and I'll get back to you.

dipeshmsft: @ThomasGoulet73, I understood what you meant. 
From my understanding, when TokenReder.Read method is called in XamlReaderHelper, we get to this point where the ProcessingRootContext is not yet set, and when we do _compiler.StartElement call from WriteElementStart in ParserExtension, then this variable get's set. That's why when we have two nodes, it is processed correctly.

LGTM.

dipeshmsft (approved):

src/Microsoft.DotNet.Wpf/src/PresentationBuildTasks/MS/Internal/MarkupCompiler/ParserExtension.cs
ThomasGoulet73: Thanks for merging. Sorry I didn't have time to reply. I started looking into it but hadn't finished yet but from what I saw my understanding matched what you said.

Issue #8842, opened 2024-02-25
Xaml Unsupported MouseAction.None value


MauNguyenVan:
### Description

Bug in parsing Xaml for `MouseAction .None`
NotSupportedException: Unsupported MouseAction 'NONE'.

I use Win 10, VS 2022 17.9.1
SDK : .NET 8.0

### Reproduction Steps

- Create new Wpf App.
- in Xaml Editer, add ` <MouseBinding MouseAction="None" Command="{Binding cmd}"/>` in grid.
```xml
<Window x:Class="WpfApp.MainWindow"
        xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
        xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
        xmlns:d="http://schemas.microsoft.com/expression/blend/2008"
        xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006"
        xmlns:local="clr-namespace:WpfApp"
        mc:Ignorable="d"
        Title="MainWindow" Height="450" Width="800" >

    <Grid>
        <Grid.InputBindings>
            <MouseBinding MouseAction="None" Command="{Binding cmd}"/>//Bug here when use MouseAction="None", in API docs, MouseAction.None is a valid value, but in xaml can't convert and emit an error , =>when try run app , it still run  but throw an Ex
        </Grid.InputBindings>
    </Grid>
</Window>
```

### Expected behavior

No Error. None is a valid value. Compile and run normal
ref docs: https://learn.microsoft.com/en-us/dotnet/api/system.windows.input.mouseaction?view=windowsdesktop-8.0

### Actual behavior

Error	XLS0431	Invalid value for property 'MouseAction': 'None'

![image](https://github.com/dotnet/wpf/assets/48544354/272f221c-8cde-454c-825f-b31f17f5b2e1)

then run the app, it will occur an exception.  

![image](https://github.com/dotnet/wpf/assets/48544354/8f659120-a557-43b4-a270-3a37a6f5146b)

miloush:
Technically the documentation says that `None` is a valid enum value. It does not say whether it is valid in this context (which it happens to be), neither it says how the value should be represented as string or in XAML. If you look at the `MouseActionConverter`, you can see that the corresponding string for the `None` value is an empty string.

It probably wouldn't hurt to make it recognize `"None"`, but it would be a breaking change to produce `"None"` when converting to string.

[...]

PR #8871, opened 2024-03-05
Supported MouseAction.None in xaml #8842 

MauNguyenVan:
Fixes Xaml Unsupported MouseAction.None value #8842 

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/wpf/pull/8871)

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Input/Command/MouseActionConverter.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Input/Command/MouseActionConverter.cs
@@ -86,6 +86,7 @@ namespace System.Windows.Input
                 MouseAction mouseAction = MouseAction.None;
                 switch (mouseActionToken)
                 {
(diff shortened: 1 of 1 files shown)


dipeshmsft:
@MauNguyenVan, the team is currently occupied with Win11 theming changes. Once we have bandwidth, we will get to your PR.

Kuldeep-MS:
ccing @miloush

miloush:
Yeah I don't have anything against the code change, the use case seems a bit weak because the gesture will never match anything and the desired outcome can already be achieved by omitting the attribute or setting it to an empty string.

A slight design concern is that the conversion does not roundtrip. `None` will produce empty string (for compatibility reasons), but converting empty string back throws. Perhaps an empty string should be added as another case yielding `None`. 

If we don't know the reason why the `None` value was given such a special treatment, then we don't have a good reason not to accept this change.

anjali-wpf (approved):
LGTM.

anjali-wpf:
@MauNguyenVan - Thanks for your contribution.

PR #8927, opened 2024-03-19
Clean-up public and internal pipelines

rchauhan18:
### Description
We have separated the public and internal pipelines, so we can remove the conditions throughout that are checking internal vs. public to make things more readable.

### DIFF
--- a/azure-pipelines-pr.yml
+++ b/azure-pipelines-pr.yml
@@ -19,9 +19,6 @@ variables:
   - ${{ else }}:
     - name: PostBuildSign
       value: true
-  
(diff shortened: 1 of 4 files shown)


singhashish-wpf (approved):

mmitche (changes requested):
WPF runs internal PRs (not yet for .NET 9, but eventually), and at that point many of these changes will be incorrect. Essentially, it won't be correct to assume that PR == public.

RussKie:
@mmitche clearly has more context than I do so I defer to his judgement. Apologies for the troubles.

Issue #109, opened 2018-12-06
Hyphen based ligatures don't work in WPF


clairernovotny:
See https://github.com/tonsky/FiraCode/issues/259 for the full issue description.

Would be good to get this fixed for .NET Core 3.

DrPizza:
This makes Visual Studio very sad. -> is a very common symbol in C++, and I don't get pretty arrows for it.

Fell:
I've tracked this bug through various different platforms now, and it would be really great if it could get fixed. It's a small thing, but it would make everyone's programming life a little more fun.

limbusdev:
Problem persists in Visual Studio 2019.1

HolyChen:
in vs2017 too, but not in vs code

be5invis:
I suspect there's something wrong in TextItemizer that it breaks out hyphens (`-`) as individual run.

weltkante:
I understand this originally was probably by design, but considering that hyphen based ligatures are a desired feature in some fonts, that original "by design" is not appropriate anymore and needs to be revisited. I would have contributed by researching implementation and proposing a fix, but considering the text engine is going to stay closed source this really needs to be done by MS.

AmadeusW:
Windows Terminal, VS Code and JetBrains all render the text better than Visual Studio, which uses the WPF renderer. Text rendered by WPF is less crisp than other technologies; horizontal lines are thinner, and vertical lines are thicker.

![image](https://user-images.githubusercontent.com/1673956/73796707-d12d5e80-4762-11ea-896f-25f26e3a7d0c.png)

I heard that WPF used by Visual Studio uses ~2010 version of DWrite, and there was no demand to update the version of WPF used in .NET Framework. I would like this comment to serve as proof of demand for a fix on behalf of Visual Studio Editor's users and developers.

Some of the user feedback:
https://www.reddit.com/r/VisualStudio/comments/ev5b2u/is_there_an_editor_text_readability_fix_already/
https://developercommunity.visualstudio.com/content/problem/44061/some-programming-ligatures-do-not-work.html
https://developercommunity.visualstudio.com/content/problem/899361/editor-readability.html?childToView=904574#comment-904574

AmadeusW:
It seems that WPF devs use #2025 to track this bug.

weltkante:
Just to be clear, the last two comments are off-topic and have nothing to do with hyphen-based ligatures (which is what this issue is about). The text layouting is unrelated to the font rendering.

[...]

PR #8990, opened 2024-04-04
Adding AppContext Switch for disabling special character ligatures

Kuldeep-MS:
Contributes to #109 

## Description

Special character ligatures will be included in .NET 9 preview 4. This PR is introducing an AppContext switch to disable the feature.

## Customer Impact

Without this changes customer won't be able to disable the special character ligature.

## Regression

NO

## Testing

DRT + Microsuites + Feature Tests

## Risk

Low. After upgrading to .NET 9 preview 4, WPF applications which have editing features that use special character ligatures may notice some unforeseen issues. For such cases, an AppContext switch (`Switch.System.Windows.DisableSpecialCharacterLigature`) can be set to `true` which will disable the feature.

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/PresentationCore/MS/internal/CoreAppContextSwitches.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationCore/MS/internal/CoreAppContextSwitches.cs
@@ -395,6 +395,21 @@ namespace MS.Internal
 
         #endregion
 
+        #region DisableSpecialCharacterLigature
+
+        internal const string DisableSpecialCharacterLigatureSwitchName = "Switch.System.Windows.DisableSpecialCharacterLigature";
+        private static int _disableSpecialCharacterLigature;
+        public static bool DisableSpecialCharacterLigature
+        {
+            [MethodImpl(MethodImplOptions.AggressiveInlining)]
+            get
+            {
+                return LocalAppContext.GetCachedSwitchValue(DisableSpecialCharacterLigatureSwitchName, ref _disableSpecialCharacterLigature);
+            }
+        }
+
+        #endregion
+
     }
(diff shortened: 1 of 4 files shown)


pchaurasia14 (approved):

miloush:
I will just note as I already stated elsewhere that it is not sustainable long-term to gate every change to WPF by an AppContext switch. Other avenues need to be considered, or a plan for removing them in the future should be in place (e.g. every other LTS release).

The risk of associated with the changes should play a role. Compared to .NET Framework, there is an abundance of releases and if a change breaks someone, they can easily stay on a previous release and raise the issue, and switch can be added as a part of maintenance update if justified.

For this one, like in the case of hardware acceleration for RDP, the cost of the introduced complexity, maintaining and documenting the switch is not justified by the low risk of the change in my opinion.

Kuldeep-MS:
> I will just note as I already stated elsewhere that it is not sustainable long-term to gate every change to WPF by an AppContext switch. Other avenues need to be considered, or a plan for removing them in the future should be in place (e.g. every other LTS release).
> 
> The risk associated with the changes should play a role. Compared to .NET Framework, there is an abundance of releases and if a change breaks someone, they can easily stay on a previous release and raise the issue, and switch can be added as a part of maintenance update if justified.
> 
> For this one, like in the case of hardware acceleration for RDP, the cost of the introduced complexity, maintaining and documenting the switch is not justified by the low risk of the change in my opinion.

I completely understand and appreciate your suggestion. However, for this particular change, which is a low level change, we have planned initially, we would like to incorporate an AppContext switch to mitigate any unforeseen issues. As we have faced similar situations in the past, we want to take extra precautions by implementing this switch.
 
We will discuss more on this internally and if there is a need to revisit to remove switches, we will definitely do it.

PR #9153, opened 2024-05-23
Add IsClosing property to the Window class

### LINKED ISSUE #9152: How I can check is window closing?
Hi! I need to check is window closing, but I can't do it right.
Now I can make some crutch: invoke window.Show() and catch exception if _isClosing == true, but it's now the best solution

loyvsc:
IsClosing property has been added to the Window class

Fixes #9152

## Description

I add IsClosing property to the Window class

## Regression

None

## Testing

Local Build Pass
Sample Application Testing

## Risk

None
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/wpf/pull/9153)

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Window.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Window.cs
@@ -7989,6 +7989,18 @@ namespace System.Windows
     }
     #endregion Enums
 
+    #region Public Fields
+    
+    /// <summary>
+    ///     Is the window closing
+    /// </summary>
+    public bool IsClosing
+    {
+        get=>_isClosing;
+    }
+
+    #endregion
+
     internal class SingleChildEnumerator : IEnumerator
     {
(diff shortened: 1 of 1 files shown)


miloush:
Changes to public API need to go through API review process first. Can you tell us a bit about the issue are you trying to solve and what you cannot do without this property?

loyvsc:
Yes, of course. I save information about the last open windows, as well as dynamic information in the ViewModel of these windows. The last windows are loaded using the for loop for the RecentWindows collection: a window is created, a ViewModel is loaded, dynamic information is loaded, and the window is subscribed to the OnClosing event. When the OnClosing method is triggered, the main window is checked or any other window is closed - if any other, then it is deleted from the RecentWindows collection.

miloush:
In this scenario, what do you need IsClosing for? Sounds like you could just delete from RecentWindows on the Closed event.

lindexi:
@miloush Subjectively, I don't like this API.

PR #9174, opened 2024-05-29
Disable CA2022 error

rchauhan18:
Fix build failure: https://github.com/dotnet/wpf/pull/9163
Replacing System.IO.Stream.Read with System.IO.Stream.ReadExactly, to resolve the CA2022 error.

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/PresentationCore/MS/internal/Ink/InkSerializedFormat/DrawingAttributeSerializer.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationCore/MS/internal/Ink/InkSerializedFormat/DrawingAttributeSerializer.cs
@@ -189,22 +189,7 @@ namespace MS.Internal.Ink.InkSerializedFormat
                     }
 
                     byte[] data = new byte[ropSize];
-#if NET7_0_OR_GREATER
-                stream.ReadExactly(data, 0, (int)ropSize);
-#else
-                int readCount =  (int)ropSize;
-                int totalRead = 0;
-                while(totalRead < readCount)
-                {
-                    int read = stream.Read(data, totalRead, readCount - totalRead);
-                    if (read <= 0)
-                    {
-                        throw new EndOfStreamException();
-                    }
-                    totalRead += read;
-                }
-                
-#endif
+                    stream.ReadExactly(data);
 
                     if (data != null && data.Length > 0)
                     {
--- a/src/Microsoft.DotNet.Wpf/src/PresentationCore/MS/internal/Ink/InkSerializedFormat/DrawingAttributeSerializer.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationCore/MS/internal/Ink/InkSerializedFormat/DrawingAttributeSerializer.cs
@@ -189,8 +189,9 @@ namespace MS.Internal.Ink.InkSerializedFormat
                     }
 
                     byte[] data = new byte[ropSize];
+                    #pragma warning disable CA2022
                     stream.Read (data, 0, (int)ropSize);
-
+                    #pragma warning restore CA2022
                     if (data != null && data.Length > 0)
                     {
                         //data[0] holds the allowable values of 0-255
--- a/src/Microsoft.DotNet.Wpf/src/PresentationCore/MS/internal/Ink/InkSerializedFormat/DrawingAttributeSerializer.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationCore/MS/internal/Ink/InkSerializedFormat/DrawingAttributeSerializer.cs
@@ -189,9 +189,23 @@ namespace MS.Internal.Ink.InkSerializedFormat
                     }
 
                     byte[] data = new byte[ropSize];
-                    #pragma warning disable CA2022
-                    stream.Read (data, 0, (int)ropSize);
-                    #pragma warning restore CA2022
+#if NET7_0_OR_GREATER
+                stream.ReadExactly(data, 0, (int)ropSize);
+#else
+                int readCount =  (int)ropSize;
+                int totalRead = 0;
+                while(totalRead < readCount)
+                {
+                    int read = stream.Read(data, totalRead, readCount - totalRead);
+                    if (read <= 0)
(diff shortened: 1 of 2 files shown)


singhashish-wpf (dismissed):
LGTM, newly added CA2022, disabling for now.

rchauhan18:
@stephentoub 
I have added changes to fix CA2022 warning in WPF. Can you please review the changes?

src/Microsoft.DotNet.Wpf/src/PresentationCore/MS/internal/Ink/InkSerializedFormat/DrawingAttributeSerializer.cs
```diff
@@ -189,7 +189,22 @@ internal static uint DecodeAsISF(Stream stream, GuidList guidList, uint maximumS
                     }
 
                     byte[] data = new byte[ropSize];
-                    stream.Read (data, 0, (int)ropSize);
+#if NET7_0_OR_GREATER
```
ThomasGoulet73, line 192: Why did you add `#if` ? This project should always build with NET7_0_OR_GREATER set to true so I don't understand the value of having the `#else`.

stephentoub: I'm curious as well... doesn't this only build for net9.0?
https://github.com/dotnet/wpf/blob/47efe4600a00ac059aeca109459e9a4e8132d907/Directory.Build.props#L5-L6

src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/dataobject.cs
```diff
@@ -1586,7 +1586,22 @@ private int GetDataIntoOleStructsByTypeMedimIStream(string format, object data,
 
                         byte[] buffer = new byte[NativeMethods.IntPtrToInt32(size)];
                         inkStream.Position = 0;
-                        inkStream.Read(buffer, 0, NativeMethods.IntPtrToInt32(size));
+#if NET7_0_OR_GREATER
+                inkStream.ReadExactly(buffer, 0, NativeMethods.IntPtrToInt32(size));
```
stephentoub, line 1590: What kind of stream can inkStream be? If it can be an arbitrary stream, are there tests (or can we add tests) validating what happens when the stream provides back fewer bytes than requested?

stephentoub, line 1590:
```suggestion
                inkStream.ReadExactly(buffer, 0, buffer.Length);
```

src/Microsoft.DotNet.Wpf/src/PresentationCore/MS/internal/Ink/InkSerializedFormat/DrawingAttributeSerializer.cs
```diff
@@ -189,7 +189,22 @@ internal static uint DecodeAsISF(Stream stream, GuidList guidList, uint maximumS
...
+                stream.ReadExactly(data, 0, (int)ropSize);
```
stephentoub, line 193:
```suggestion
                stream.ReadExactly(data, 0, data.Length);
```

src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/dataobject.cs
```diff
@@ -1725,7 +1740,22 @@ private int SaveStreamToHandle(IntPtr handle, Stream stream, bool doNotReallocat
 
                 bytes = new byte[NativeMethods.IntPtrToInt32(size)];
                 stream.Position = 0;
-                stream.Read(bytes, 0, NativeMethods.IntPtrToInt32(size));
+#if NET7_0_OR_GREATER
+                stream.ReadExactly(bytes, 0, NativeMethods.IntPtrToInt32(size));
```
stephentoub, line 1744:
```suggestion
                stream.ReadExactly(bytes);
```

src/Microsoft.DotNet.Wpf/src/PresentationCore/MS/internal/Ink/InkSerializedFormat/DrawingAttributeSerializer.cs
ashishkumarsinghh: +1, we will remove it. Our first try was to just disable it, then we checked the runtime PR and added same changes to call `ReadExactly`

singhashish-wpf: +1, We will remove it. Our first try was to disable it, then we checked on the runtime changes and added the same changes.

src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/dataobject.cs
stephentoub: I typically do this by writing a type that derives from MemoryStream and overrides the Read methods to return back a maximum of one byte, e.g.
```C#
internal sealed class TrickleStream : MemoryStream
{
    public override int Read(byte[] buffer, int offset, int count) =>
        base.Read(buffer, offset, Math.Min(count, 1));
    ...
}
```

miloush: Oh this particular one is actually from OLE, so in most cases it might be limited to IStream, although within application you can still store any Stream in the DataStore. This particular code doesn't have many checks around, unlike the deserialization. If Read returns 0, this just writes 0 to the medium and calls it a day. I can see how if we start throwing here, applications that did not use to crash can start crashing. Still probably a reasonably thing to do.

rchauhan18:
Thanks @stephentoub for the suggestions.
I have updated the PR.

stephentoub (approved):
Thanks.

singhashish-wpf (approved):

PR #9179, opened 2024-05-30
Add TargetName to Button's Trigger

harshit7962:
## Description
Due to missing `TargetName` property in various triggers of default button styles in Fluent window, there were visible discrepancies in UI when background property was set. The following changes adds the property and applies the `Background` and `BorderBrush` property specifically to the border enclosing the content presenter.

## Customer Impact
It fixes the accessibility issues with hover and pressed states of a button, when `Background` or `BorderBrush` property is set at the application level.

## Testing
Local Build Pass
Sample Application Testing

## Risk
_None_

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/wpf/pull/9179)


anjali-wpf (approved):
LGTM.

PR #9180, opened 2024-05-30
Fixed styles for editable ComboBox chevron icon styles

anjali-wpf:
## Description

Fixed styles for editable ComboBox chevron icon styles

## Regression

None

## Testing

Local Build Pass
Sample Application Testing

## Risk

None

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/wpf/pull/9180)


src/Microsoft.DotNet.Wpf/src/Themes/PresentationFramework.Fluent/Styles/ComboBox.xaml
```diff
@@ -208,20 +209,27 @@
...
-                                            Text ="{StaticResource ComboBoxChevronDownGlyph}">
-                                            <TextBlock.RenderTransform>
-                                                <RotateTransform Angle="0" />
-                                            </TextBlock.RenderTransform>
-                                        </TextBlock>
+                                    <Grid Grid.Column="1" 
+                                        VerticalAlignment="Center" 
+                                        HorizontalAlignment="Center"
+                                        Margin="{StaticResource ComboBoxChevronMargin}">
+                                        <Border
```
dipeshmsft, line 216: Why not use a button with textblock, I think that will be sufficient and we won't need to add any new resources.

anjali-wpf: Button styles are not going to align with chevron icon style. Even my first approach was to achieve this was through button.

dipeshmsft:
This issue fixed by this PR is already handled by another fix to Button style, hence closing the PR.

dipeshmsft:
The modifications in this PR, brings an unwanted behavior in non-editable combo box. When the ComboBox is not editable, hovering over the button should not have any different effect, however after this PR, when we hover over the ComboBox, even in that case, the background of the button changes. 

Hence closing this PR.

PR #9182, opened 2024-05-30
Fix ListView/GridView scroll performance when automation peers are created

### LINKED ISSUE #9181: ListView/GridView scroll performance is O(n) when automation peers are created
## Description

I was running some tests with ListView (using GridView) and when I loaded `1 000 000` items (yes, an obscure use case but that's why we got virtualization, no?), the scrolling performance would degrade and keyboard scrolling felt almost unusable.

During a quick rundown, I've discovered the culprit is in override of `GetChildrenCore` of `GridViewItemAutomationPeer`, which performs IndexOf (linear) search in the `Items` collection of the underlying `ItemsSource` to find the respective item's Row. 

## Reproduction Steps

The easiest step to reproduce is to add a ListView with GridView plus a ComboBox, after opening the ComboBox's dropdown (clicking it), `Accessibility.dll` gets loaded and the party gets started. You can find a demo repository on [Github - here](https://github.com/h3xds1nz/ListViewPerformance).

## Behavior

The scrolling performance becomes worse to unusable with increasing number of items.

## Known Workarounds

The "workaround" is to subclass `GridView`, overriding `GetAutomationPeer` to return NULL. Not really a solution though.

## Other information

I have created a pull request at the same time, fixing the respective issue. #9182

h3xds1nz:
Fixes #9181 

## Description

I was running some tests with ListView (using GridView) and when I loaded `1 000 000` items (yes, an obscure use case but that's why we got virtualization, no?), the scrolling performance would degrade and keyboard scrolling felt almost unusable.

During a quick rundown, I've discovered the culprit is in override of `GetChildrenCore` of `GridViewItemAutomationPeer`, which performs IndexOf (linear) search in the `Items` collection of the underlying `ItemsSource` to find the respective item's Row. 

## Reproduction Steps

The easiest step to reproduce is to add a ListView with GridView plus a ComboBox, after opening the ComboBox's dropdown (clicking it), `Accessibility.dll` gets loaded and the party gets started. You can find a demo repository on [Github - here](https://github.com/h3xds1nz/ListViewPerformance).

## Behavior

The scrolling performance becomes worse to unusable with increasing number of items.

## Known Workarounds

The "workaround" is to subclass `GridView`, overriding `GetAutomationPeer` to return NULL. Not really a solution though.

## Risk

None.

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/wpf/pull/9182)

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Automation/Peers/GridViewItemAutomationPeer.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Automation/Peers/GridViewItemAutomationPeer.cs
@@ -62,7 +62,7 @@ namespace System.Windows.Automation.Peers
                     _dataChildren = new Hashtable(rowPresenter.ActualCells.Count);
 
                     List<AutomationPeer> list = new List<AutomationPeer>();
-                    int row = listview.Items.IndexOf(item);
+                    int row = listview.ItemContainerGenerator.IndexFromContainer(lvi);
                     int column = 0;
 
                     foreach (UIElement ele in rowPresenter.ActualCells)


h3xds1nz:
@dotnet-policy-service agree

miloush:
Tested?

Code changes look good to me. Would it make sense to have a reference to something holding the cached row number so that it doesn't have to be searched for each cell in a row?

h3xds1nz:
@miloush Yeah, tested, it works as expected.

It would definitely make more sense to cache the row index for all the cells, however I don't see an easy way how to do it here, given that we would load the index lazily.

h3xds1nz:
@miloush I have actually thought of a much simpler solution that didn't occur to me before, we can just retrieve the index from the realized item in ItemGenerator. That way the change is pretty much a one-liner.

I have also thought about your caching proposal, and my original solution does actually ruin the UAI performance with multiple columns indeed. This solution on the other hand is rather smooth.

miloush:
For one liner I think it would appreciated if the PR didn't contain all the other stylistic changes.

miloush:
Doesn't the IndexFromContainer do the same linear search?

h3xds1nz:
Agreed, I'll push a commit with only the one-liner.

No, ItemGenerator contains always just the realized items and it does a linear search but only in the bunch, so in case of virtualization that's only the handful few that are currently realized. I use it a lot in other projects to draw over controls so I'm aware of the performance. I ran it with the build of the new PresentationFramework just to be sure before comitting.

Wil-fried:
Hello, i'm following this conversation since the start because i'm running into this issue.
I was very glad when @h3xds1nz came up with a solution and seeing the progress on the reviews.
Now the downside, there is not much activity since 30th may. Is there a way upvote / make the owner aware of this issue please?
When "Accessibility.dll" load it greatly impact the datagrid /listview negatively.

pchaurasia14:
Thanks @h3xds1nz.
We will include this PR in our upcoming test pass.

h3xds1nz:
@pchaurasia14 Thank you.

singhashish-wpf:
@h3xds1nz I was trying to check your sample repro by replacing the locally built PresentationFramework.dll and could experience similar scrolling performance. Couldn't observe something noticeable. I might be missing a step here, Could you please share the exact steps to verify?

h3xds1nz:
@singhashish-wpf Thanks for checking, I've included sample videos for you with the repro that I've originally recorded. First I'm regularly scrolling when peers are not being created.

Notice what happens in the first/original video when `Accessibility` gets loaded (VS output) - e.g. clicking dropdown, and automation peers start being created. The scrolling up that takes like 250ms for each item (because for each item in the view it does linear search (`IndexOf`) in the entire `Items` collection) is me actually holding `Up` arrow key. 

Same goes for the scrollbar being laggy after I reach the bottom.

By the way, the fact that automation peers are created after interacting in with dropdowns, popups, even though there's no client subbed, that's a long-standing issue since the beginning, I believe there's a tracking for this as well in this repo but that's just a side note.

https://github.com/user-attachments/assets/8fd7aa62-798d-4c86-babe-8660019e38ab

https://github.com/user-attachments/assets/080f0016-c124-4214-8e07-515e7603efef

h3xds1nz:
I have rebased and cleaned up the commit history.

harshit7962:

h3xds1nz:
@harshit7962 Thanks, finally can use UIA again in the upcoming project (with .NET 10 I presume?)

PR #9199, opened 2024-06-04
Replace ArrayList in DataFormats with List<DataFormat> and refactor the class

h3xds1nz:
## Description
`DataFormats` is one of those classes that still use `ArrayList` instead of generic `List<DataFormat>`.

There are no functional changes besides removing the `Ordinal` -> `OrdinalIgnoreCase` compare and just using case-insensitive one by default as in .NET 8 this path is vectorized and I believe decreasing the code-size outweights the now immeasurable benefit of two subsequent compares (not to mention, should such a case occur, this solution will come on top) as this isn't a hot path regularly.

While refactoring, I've also removed obsolete CAS remarks and fixed some grammar mistakes.

As a last step, I removed the unnecessary double lock on `InternalGetDataFormat`/`GetDataFormat` methods by creating an internal static class, running the initialization of predefined formats in the static constructor, saving all the unnecessary overhead (the internal static class and the formats are not initialized if only the fields are used, as `beforefieldinit` is set, so there's no regression in that sense as well.)

### Benchmark for id `16` and `"locale"` string

| Method          | Mean [ns] | Error [ns] | StdDev [ns] | Code Size [B] | Allocated [B] |
|---------------- |----------:|-----------:|------------:|--------------:|--------------:|
| Original_int    |   33.11 ns |   0.230 ns |    0.204 ns |         684 B |             - |
| Original_string |   91.79 ns |   0.635 ns |    0.594 ns |       1,064 B |             - |
| PR_EDIT_int     | 12.31 ns |   0.089 ns |    0.079 ns |         465 B |             - |
| PR_EDIT_string  |   26.76 ns |   0.171 ns |    0.151 ns |         617 B |             - |

## Customer Impact
This will increase performance 3x and almost 4x times increase can be observed should you query data formats with wrong casing input (which shall not be the case of WPF itself but if used externally, it might).

## Regression
No.

## Risk

I believe there are none as the static collection is used internally as a storage and access is done via defined `GetDataFormat` overloads.

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/wpf/pull/9199)

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/DataFormats.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/DataFormats.cs
@@ -14,22 +14,12 @@ using MS.Win32;
 
 namespace System.Windows
 {
-    #region DataFormats class
-
     /// <summary>
     /// Translates between Windows Design text-based formats and
     /// 32-bit signed integer-based clipboard formats.
     /// </summary>
     public static class DataFormats
     {
-        //------------------------------------------------------
-        //
-        //  Public Methods
-        //
-        //------------------------------------------------------
-
-        #region Public Methods
-
         /// <summary>
         /// Gets the data format with the Windows Clipboard numeric ID and name for the specified ID.
         /// </summary>
@@ -39,16 +29,6 @@ namespace System.Windows
         /// Gets the data format with the Windows Clipboard numeric ID and name for the specified data format.
         /// </summary>
         public static DataFormat GetDataFormat(string format) => DataFormatsImpl.GetDataFormat(format);
-
-        #endregion Public Methods
-
-        //------------------------------------------------------
-        //
-        //  Public Fields
-        //
-        //------------------------------------------------------
-
-        #region Public Fields
 
         /// <summary>
         /// Specifies the standard ANSI text format. This field is read-only.
@@ -185,15 +165,7 @@ namespace System.Windows
         /// Specifies a data format as Xaml Package. This field is read-only.
         /// </summary>
         public static readonly string XamlPackage = "XamlPackage";
-        #endregion Public Fields
 
-        //------------------------------------------------------
-        //
-        //  Internal Fields
-        //
-        //------------------------------------------------------
-
-        #region Internal Fields
         /// <summary>
         /// Specifies a data format as ApplicationTrust which is used to block
         /// paste from partial trust to full trust applications. The intent of this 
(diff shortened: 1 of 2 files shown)


miloush:
@singhashish-wpf there is a lot of code style changes departing from the current codebase. Perhaps we should have some rules about what is expected going forward.

@h3xds1nz Can you provide any numbers demonstrating the performance/size impact? How did you test the changes?

miloush:
The last change with static constructor would penalize anyone just using the static properties which I would think are the most common usage of this class.

lindexi:
If the change doesn't involve a behavioral change, that's a particularly good thing.

h3xds1nz:
@miloush @lindexi As per your very good points, I've changed the implementation to behave the same way as it did before with the last commit while keeping the benefits of all the changes. The internal list/class won't be initialized unless it is actually meant to be used (with one of the Get*) methods.

The call methods are likely to get inlined by the JIT and with tiered compilation this can all go smooth like butter.

Also the lock now belongs to the internal class.

ThomasGoulet73:
I'm personally not a big fan of moving code to static constructor just for performance, especially when the static constructor runs arbitrary code (Like UnsafeNativeMethods.RegisterClipboardFormat that could have side effects depending on when a static constructor ran).

You can probably improve the performance of the method EnsurePredefined by also checking if _formatList is null before locking (It would be: If _formatList is null -> Lock -> If _formatList is still null -> Initialize). This is a very common pattern and a cheap way to avoid locking for no reason.

Other than that, the other changes LGTM.

miloush:
> I'm actually testing waters to be entirely honest. As for code style, I'm happy to adapt if we get proper guidelines.

I am not saying these cannot or should not change, but indeed guidelines should be in place.

I have to admit I am a bit puzzled by the original code, I am having difficult time imagining a scenario where case sensitive search through data formats would provide a performance improvement worth the complexity but not a hardcoded switch. The only situation I can think of this needs to be done quickly is if it is executed on every mouse move during dragging.

I will also note that in general, this change would cause different format to match if they differ by case only, however RegisterClipboardFormat is documented as case-insensitive, so this seems like a fair change.

h3xds1nz:
Earlier in the days the CI search essentially meant two different compares where CI was way more expensive as you couldn't just load a register and OR the two masks. However I think the main intention was to optimize for the internal library use where the formats are pre-defined. Now I find this irrelevant and as you've perfectly noted, formats are CI on their own.

As per the static constructor, I understand that the PInvoke call might be of concern (was originally mine as well) however we do not depend on any state of the call there, either it registers or it doesn't. Even though 'popular', I've always found the double checking a bit of an anti-pattern, where static constructor is clear and expressive. It also removes `_formatList` from being `readonly`, which means that currently the null check itself can never be optimized out unless I've missed something.

amarinov-msft:
@h3xds1nz Can you resolve the merge conflicts in this PR when you have the opportunity?

h3xds1nz:
@amarinov-msft The relevance of this PR has pretty much vanished after #10544 since it was chilling here for a year.

I'll check later on what's left there to be done on this PR, if anything, but right now I'd recommend spending time on different PRs that are more impactful.

amarinov-msft:
OK, thank you!

PR #9211, opened 2024-06-06
Use Environment.ProcessId instead of calling a Win32 function

ThomasGoulet73:
## Description
Use Environment.ProcessId instead of calling a Win32 function. Environment.ProcessId uses the same [Win32 function](https://learn.microsoft.com/en-us/windows/win32/api/processthreadsapi/nf-processthreadsapi-getcurrentprocessid) that we were currently using but the difference is that Environment.ProcessId caches the process id instead of doing a PInvoke every time it is accessed (The Win32 function is documented as not changing during the lifetime of the process, see [here](https://learn.microsoft.com/en-us/windows/win32/api/processthreadsapi/nf-processthreadsapi-getcurrentprocessid#remarks)). 

Here's the current implementation of Environment.ProcessId on Windows: https://github.com/dotnet/runtime/blob/60edbf9371ff00349e9224fb0920c8bca51f1a4f/src/libraries/System.Private.CoreLib/src/System/Environment.Windows.cs#L124.

The goal of this PR is to improve performance by using Environment.ProcessId which caches the value and reduce the complexity of the WPF codebase by using a property from the BCL instead of a Win32 function.

## Customer Impact
Better performance.

## Regression
No.

## Testing
Local testing.

## Risk
Low.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/wpf/pull/9211)

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/PresentationCore/MS/internal/DpiUtil/DpiUtil+ProcessDpiAwarenessHelper.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationCore/MS/internal/DpiUtil/DpiUtil+ProcessDpiAwarenessHelper.cs
@@ -99,7 +99,7 @@ namespace MS.Internal
                 {
                     // If a valid window is not specified, then query the current process instead of the process
                     // associated with the window
-                    windowThreadProcessId = SafeNativeMethods.GetCurrentProcessId();
+                    windowThreadProcessId = Environment.ProcessId;
                 }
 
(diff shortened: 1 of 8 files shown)


siagupta0202 (approved):

harshit7962:
@ThomasGoulet73 Thank you for your contributions.

ThomasGoulet73:
Thanks @harshit7962

PR #9215, opened 2024-06-06
Replace ArrayList with List<string> in DataObject, reduce some allocations

h3xds1nz:
## Description

Replaces `ArrayList`  instances with `List<string>`  and reduces some unnecessary copies/allocations between internal List array and string arrays by modifying `GetDistinctStrings` to accept `List<string>` instead of  `string[]`.

I couldn't justify swapping List<string> for HashSet<string> in `GetDistinctStrings`, the number of items is too low generally and allocates too much.

## Customer Impact

Improved performance, less memory consumption.

## Regression

None.

## Testing

Verifying basic method functionality, the changes are minimal.

## Risk

None.

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/wpf/pull/9215)


miloush:
In both cases `List<string> distinct = [];` could use initial capacity of the incoming list to avoid reallocations,

In general it would make reviews easier if purely stylistic changes were in a separate commit. Note that in the past, not doing stylistic changes as a part of PR made it easier to merge.

h3xds1nz:
I was thinking of introducing it there but I've retracted from it in the end. But if I'm not the only one on that, I'll add the default capacity to match, it is mostly gonna be filled anyways, good point, thank you.

As per the stylistic changes, I'm trying but it is a force of strong habit. Hopefully we get those guidelines if I keep going.

src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/dataobject.cs
```diff
@@ -1263,27 +1264,21 @@ private int OleGetDataUnrestricted(ref FORMATETC formatetc, ref STGMEDIUM medium
         /// <summary>
         /// Retrieves a list of distinct strings from the array.
         /// </summary>
-        private static string[] GetDistinctStrings(string[] formats)
+        private static string[] GetDistinctStrings(List<string> formats)
```
lindexi, line 1267: How about `formats.Distinct().ToArray()` ? Which is faster?

src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/dataobject.cs
```diff
@@ -2467,41 +2462,29 @@ public void SetData(Object data)
 
             public string[] GetFormats(bool autoConvert)
             {
+                IEnumFORMATETC enumFORMATETC = EnumFormatEtcInner(DATADIR.DATADIR_GET);
+                List<string> formats = [];
```
lindexi, line 2466: How about re-define formats with `HashSet<string>`, which can remove the `GetDistinctStrings` method and be faster?

miloush: @h3xds1nz already discussed this in the description, the number of items is too low to justify the overhead of hashset.

src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/dataobject.cs
h3xds1nz: @lindexi Distinct will use internally HashSet<T>, which as I've mentioned, don't think is suitable for this usage. It usually comes out on top with 14 or more objects but even then will allocate double the memory just to build it.

src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/dataobject.cs
h3xds1nz: @lindexi  I further checked if using a HastSet out of the box and thus saving ourselves the uniqueness check (+method) could prove itself useful but I couldn't come up with a use-case where the overhead is justified even this way. Thanks for the inquiry though, love it.

lindexi: Thank you for your contribution.

dipeshmsft:
@h3xds1nz, I agree with @miloush here that making stylistic changes in a separate commit does help while we are reviewing the PRs. If the changes per file are small then it's okay, but if there are a lot of changes please consider making a separate commit in the new PRs from here onwards. Just a suggestion.

dipeshmsft (approved):

dipeshmsft (approved):

himgoyalmicro (approved):

PR #9222, opened 2024-06-06
Replace ArrayList in CursorConverter with preallocated array and reduce allocations

h3xds1nz:
## Description

Optimizes `GetStandardValues` function from TypeConverter by replacing `ArrayList` with preallocated array as we know the size beforehand, removes unnecessary allocations caused by copying the list and re-allocating the internal array.

Removes the private helper function `GetProperties` that's only used in 1 place, inlining it instead.

| Method                    | Mean     | Error    | StdDev   | Median   | Gen0   | Allocated |
|-------------------------- |---------:|---------:|---------:|---------:|-------:|----------:|
| GetStandardValues         | 560.4 ns | 11.12 ns | 15.21 ns | 555.2 ns | 0.0677 |    1136 B |
| GetStandardValues_PR      | 455.9 ns |  9.11 ns | 16.88 ns | 447.2 ns | 0.0315 |     528 B |

## Customer Impact

Improved performance, reduced allocations.

## Regression

No.

## Testing

Method functionality test, build.

## Risk

None.

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/wpf/pull/9222)

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Input/CursorConverter.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Input/CursorConverter.cs
@@ -4,7 +4,6 @@
 
 using System.ComponentModel;
 using System.ComponentModel.Design.Serialization;
-using System.Collections;
 using System.Globalization;
 using System.Reflection;
 using System.Windows.Media; // TypeConverterHelper, UriHolder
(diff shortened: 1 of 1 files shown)


src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Input/CursorConverter.cs
```diff
@@ -53,35 +52,29 @@ public override bool CanConvertTo(ITypeDescriptorContext context, Type destinati
...
         /// <returns>TypeConverter.StandardValuesCollection</returns>
         public override TypeConverter.StandardValuesCollection GetStandardValues(ITypeDescriptorContext context)
         {
-            if(this._standardValues == null)
+            if(_standardValues is null)
             {
-                ArrayList list1 = new ArrayList();
-                PropertyInfo[] infoArray1 = this.GetProperties();
-                for(int num1 = 0; num1 < infoArray1.Length; num1++)
+                PropertyInfo[] properties = typeof(Cursors).GetProperties(BindingFlags.Public | BindingFlags.Static);
```
lindexi, line 64: Personally, I prefer to keep the old method. Thank you.

h3xds1nz: @lindexi I'm sorry but I do not see a reason for the method. I actually had to go and find out what the method does originally, it is only used in one place and introduces imo unnecessary code complexity for a one-liner.

miloush: Does it not get inlined? Not everything is about code complexity, there is also legibility and maintainability. This one did raise an eyebrow but I don't feel that strong about it.

h3xds1nz: @miloush It will get inlined, it is rather small and can even be made static iirc. But my point is that since this is not a function on Cursor class but in the converter, called as "this.GetProperties()", I had to go and check what this function does previously, everyone will going through the code. That is what I meant when I said code complexity. Not the actual codegen — that will be identical.

Even though you see it returns PropertyInfo array, you don't know **which flags** and **which properties** as in reflection context it can mean anything. Does it filter it afterwards? Questions. So that is my reason for putting this one inside.

siagupta0202 (approved):

harshit7962:

h3xds1nz:
@harshit7962 Thanks for including them.

PR #9223, opened 2024-06-06
Replace boxing ArrayList in LassoSelectionBehavior, remove unnecessary array alloc

h3xds1nz:
## Description

Replaces boxing `ArrayList` with generic `List<Point>` with preallocated size of 4 as that's always gonna be filled in the least, and removes array allocation through usage of `Span<Point>` as a new return type. 

In `LassoHelper` the function to match points finally runs an efficient `foreach` (as it was iterating over an array before).

Note that we could return `ROS<Point>` instead and iterate via value but this solution offers an additional performance without really introducing any real risks.

In total, this will run about 3x faster along with less than 50% of originally introduced allocations.

## Customer Impact

## Regression

None.

## Testing

Build and benchmark only but there are no functional changes.

## Risk

None.

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/wpf/pull/9223)

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/PresentationFramework/MS/Internal/Ink/LassoHelper.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationFramework/MS/Internal/Ink/LassoHelper.cs
@@ -167,9 +167,8 @@ namespace MS.Internal.Ink
 
         #region ArePointsInLasso
         /// <summary>Copy-pasted Platform's Lasso.Contains(...)</summary>
-        public bool ArePointsInLasso(Point[] points, int percentIntersect)
+        public bool ArePointsInLasso(Span<Point> points, int percentIntersect)
         {
-            System.Diagnostics.Debug.Assert(null != points);
             System.Diagnostics.Debug.Assert((0 <= percentIntersect) && (100 >= percentIntersect));
 
             // Find out how many of the points need to be inside the lasso to satisfy the percentIntersect.
@@ -185,9 +184,9 @@ namespace MS.Internal.Ink
             // If the no of such segments is odd then the point is within the lasso otherwise not.
             int countPointsInLasso = 0;
 
-            foreach (Point point in points)
+            foreach (ref Point point in points)
             {
-                if (true == Contains(point))
+                if (true == Contains(in point))
                 {
                     countPointsInLasso++;
                     if (countPointsInLasso == marginCount)
@@ -198,8 +197,8 @@ namespace MS.Internal.Ink
             return (countPointsInLasso == marginCount);
         }
 
-        /// <summary>TBS</summary>
-        private bool Contains(Point point)
+        /// <summary>Checks whether supplied point is inside.</summary>
+        private bool Contains(ref readonly Point point)
         {
             if (false == _boundingBox.Contains(point))
             {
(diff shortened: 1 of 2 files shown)


src/Microsoft.DotNet.Wpf/src/PresentationFramework/MS/Internal/Ink/LassoSelectionBehavior.cs
```diff
@@ -426,16 +427,14 @@ private static ElementCornerPoints GetTransformedElementCornerPoints(InkCanvasIn
         /// Private helper that will generate a grid of points 5 px apart given the elements bounding points
         /// this works with any affline transformed points
         /// </summary>
-        private Point[] GeneratePointGrid(ElementCornerPoints elementPoints)
+        private Span<Point> GeneratePointGrid(ElementCornerPoints elementPoints)
```
ThomasGoulet73, line 430: Could you return `ReadOnlySpan<Point>` instead of `Span<Point>` ? There is no perf difference and it just makes it clear that it shouldn't be mutated.

src/Microsoft.DotNet.Wpf/src/PresentationFramework/MS/Internal/Ink/LassoHelper.cs
```diff
@@ -185,9 +184,9 @@ public bool ArePointsInLasso(Point[] points, int percentIntersect)
             // If the no of such segments is odd then the point is within the lasso otherwise not.
             int countPointsInLasso = 0;
 
-            foreach (Point point in points)
+            foreach (ref Point point in points)
```
ThomasGoulet73, line 187: Could show the performance improvement of only adding these keywords here because I ran a reduced benchmark for this and I didn't see any improvements.

lindexi: `Point` is a very small structure. Parameters may be passed using registers.

h3xds1nz: @ThomasGoulet73 the assembly JIT generates is very slightly different, you trade off copy of 16 bytes for a mem ref, which is when it is actually worth it already on x64 if it can save you a copy on stack. It's about 2ns (27ns vs 25ns via ref) from using it like this but as I wrote in the original description, I have no issues with changing it because I see the other side :)

lindexi: @h3xds1nz Thank you. But I'm confused about the data from your test results. Because I do not think it costs 2ns, 2ns is too long.

h3xds1nz: @lindexi I can't get a proper benchmark that's conclusive, but it's always between 1 and 2ns diff for 20/20 in favour of ref. I know from the codegen it spills and it saves non-volatile registers as it uses them as for copying in case of copies. Eitherway I'll just do the regular code, lemme push the commit with ROS. No point wasting time on that, it's just my years of coding in assembly.

src/Microsoft.DotNet.Wpf/src/PresentationFramework/MS/Internal/Ink/LassoSelectionBehavior.cs
h3xds1nz: As per other discussion and original description, ROS it is.

siagupta0202 (approved):

siagupta0202:
@h3xds1nz Thank you so much for your contributions!

h3xds1nz:
@siagupta0202 Happy to! :)

PR #9230, opened 2024-06-07
Remove bunch of allocations caused by static arrays by favoring ReadOnlySpan<T>

h3xds1nz:
## Description

Fixes unnecessary allocations on `static readonly` arrays by refactoring them as `ReadOnlySpans<T>` of primitive types.
This under the hood allows references in runtime to the data directly as stored in the mapped file, not requiring any runtime allocs. See [RuntimeHelpers.CreateSpan&lt;T&gt;](https://learn.microsoft.com/en-us/dotnet/api/system.runtime.compilerservices.runtimehelpers.createspan?view=net-8.0) intrinsic for more info. In most cases the value is then inlined directly into the method, saving memory references as well.

In other cases, such as the Geometry classes, it allows us to reference data directly without any pinning (`fixed` is then merely used to retrieve the pointer that's passed to the unsafe method.

This also removes some unoptimized foreaches as replacement for `IndexOf`/`Contains` (guess that's coming from very early in the days).

In the end, it also removes inefficient `foreach` and `IndexOf` replacements for `Contains` methods with `ROS<char>.Contains(value)`.

| Method         | Mean      | Error     | StdDev    | Allocated |
|--------------- |----------:|----------:|----------:|----------:|
| ReadOnlySpan_Contains |  1.118 ns | 0.0111 ns | 0.0098 ns |         - |
| Original_Foreach_Contains | 10.343 ns | 0.0543 ns | 0.0508 ns |         - |

## Customer Impact

## Regression

None.

## Testing

Build and standard use. Should be easy to CR.

## Risk

Low.

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/wpf/pull/9230)


src/Microsoft.DotNet.Wpf/src/PresentationBuildTasks/Microsoft/Build/Tasks/Windows/GenerateTemporaryTargetAssembly.cs
```diff
@@ -952,7 +952,7 @@ static XmlNode CreateImportProjectSdkNode(XmlDocument xmlProjectDoc, string proj
 
         private const string WPFTMP = "wpftmp";
 
-        private static readonly char[] _semicolonChar = new char[] { ';' };
+        private const char SemicolonChar = ';';
```
miloush, line 955: If this is going into `string.Split(params char[])` then this will make it allocating array in every call, or did I miss something? Same with `ForwardSlashSeparator` below.

h3xds1nz: Thanks, it will be marginally better this way.

h3xds1nz:
I had to remove the optimization in `PresentationBuildTasks` as it still targets `.NET Framework 4.7.2` and the build then fails.

anjali-wpf (approved):
Thanks for your contribution.

h3xds1nz:
Thanks @anjali-wpf for including it.

src/Microsoft.DotNet.Wpf/src/PresentationCore/MS/internal/Ink/InkSerializedFormat/InkSerializer.cs
```diff
@@ -281,14 +281,13 @@ private void ExamineStreamHeader(Stream inkdata, out bool fBase64, out bool fGif
...
-                                                        (byte)'6',
-                                                        (byte)'4',
-                                                        (byte)':'};
+        private static ReadOnlySpan<byte> Base64HeaderBytes => [(byte)'b',
+                                                                (byte)'a',
+                                                                (byte)'s',
+                                                                (byte)'e',
+                                                                (byte)'6',
+                                                                (byte)'4',
+                                                                (byte)':'];
```
bgrainger, line 290: UTF-8 string literals can be used to make this (IMO) easier to read:

```csharp
private static ReadOnlySpan<byte> Base64HeaderBytes => "base64"u8;
```

h3xds1nz: @bgrainger This doesn't make use of [RuntimeHelpers.CreateSpan&lt;T&gt;](https://learn.microsoft.com/en-us/dotnet/api/system.runtime.compilerservices.runtimehelpers.createspan?view=net-8.0) yet, maybe once this is improved in the future.

ThomasGoulet73: I believe `RuntimeHelpers.CreateSpan` is only used when endianness matters (When sizeof(T) > sizeof(byte)). In this case, RuntimeHelpers.CreateSpan is useless so Roslyn does not use it.

It does basically the same thing:
https://sharplab.io/#v2:CYLg1APgAgTAjAWAFBQMwAJboMLoN7LLrHoAOATgJYBuAhgC4CmmcAbOgEqO3ADyAdgBsAngGVStfgB4ARsKYA+dACFaAZ0asALAAluwRuWXzGauOgC8SgNoAKOUwCUAchnOANERLefvv/4DAoPsTF1oPLyCo6JiokKdnNQikWNS01PjGF0Zk9Lz8v0yXVlyCsvyi5y1S8trYypBnAF0Abkj0dooaBmYoNk59ARFxSVkTJVUNbT0eQ2MmNRhLJQAiGXVNLRAVgFcADjaUkjRMLXQAWVtHfHaAX2RboA=

h3xds1nz: Ah, yeah, I've just checked the generated assembly, it is still an RVA in the module. And the endianess thing is correct, forgot about that.

Another time then; more readable code.

PR #9250, opened 2024-06-14
Release font resources to fix memory leak

### LINKED ISSUE #6296: XPS huge memory issues
* .NET Core Version: 5, 6, ... every version. 
* Windows version: Windows 10 21H2 Windows 10 21H1, Windows 11 21H2
* Does the bug reproduce also in WPF for .NET Framework 4.8?: Yes, I guess so.

 **Problem description:**
 
Opening a large XPS document requires a huge amount of memory.
Same situation when merging several XPS documents into one XPS documents (the total amount of merged pages is: 500 - 1.000 - 5.000 or even 10.000).

1.000 pages can easily run up to 15 or 20GB of memory. Depending on the document content and fonts being used.

There are several problems:
- XPS documents need to be fully loaded into memory when working with them. When for example converting XPS documents, you loop over every page. Once a page has been loaded in memory it is being kept there. There is no way to only load a Visual per Visual (e.g. FixedPage per FixedPage (page per page)) as there is no signal on when the Visual/FixedPage is not needed anymore in order to unload it from memory. 
(Besides this, is it even possible to unload a single Visual/FixedPage from memory?)

- When processing a document (e.g. adding content, printing, converting, ...) the memory can increase very rapidly to several GB's of memory. Especially when the source documents contain TrueType CID fonts, as when converting these to XPS (e.g. converting PDF files and draw them to XPS) the font and character information is being converted to polygons (a lot of doubles) and huge strings because the XML of the XPS is fully loaded into memory.

- XPS uses doubles (EMF uses floats), because the nature of XPS that everything is kept in memory, the doubles take also a huge amount of memory. Same for strings as XPS = XML file = strings.

 **Actual behavior:** 
 
Unable to process and/or merge thousands of XPS documents. Even 500 pages can be problematic. Running out of memory.

MicrosoftTeams-image (12)
This screenshot is a view of the memory usage over a short time where several XPS documents of the XPS printer are being processed when being combined together. The huge string is because of the XML content being loaded all into memory (because documents need to be merged, meaning everything should be loaded into memory.) Once the document has been merged, the source documents can be unloaded from memory. 

situation: 1000 documents of 2 (or 3 pages) need to be merged into a big document .
The pages to be merged, can be "random", according to the logic of the configured product. Meaning that e.g. from document1 we can use page 2 and 3, from document 2 we can use page 1, from document 3 we can use all pages, from document 4 we can use the last page, etc...
Pseudo-code:

```
foreach (doc in documents) { mergeIntoBigFile(doc); }
```

It is also possible that we first merge e.g. document 3, document 1, document 999, again document 1, document 2, document 36, document 1000. 
The final code that does the merging, doesn't need any knowlegde of the original document(s). 

 **Expected behavior:**
Being able to at least process and/or merge several hundreds and thousands of XPS documents. 500 pages shouldn't be a problem. Memory should be better under control.

In case it is possible to process/merge/whatever big XPS documents without having huge memory issues or unload pages, please show code samples as it seems to be impossible.

 **Minimal repro:**
/ (unable to repro, if someone of msft PM's me I can help configuring a product to demonstrate).

**Question**
One of the biggest questions:
Is there a way to unload a FixedPage/Visual from memory?

**General XPS issues and :shit: that nobody seems to care about**

- https://github.com/dotnet/runtime/issues/66756
- https://github.com/dotnet/runtime/issues/51930
- https://github.com/dotnet/wpf/issues/3546
- https://github.com/dotnet/runtime/issues/51929#issuecomment-835808702
- https://github.com/dotnet/wpf/issues/4000
- https://github.com/dotnet/wpf/issues/4466 
- https://github.com/dotnet/wpf/issues/2106 
- https://github.com/dotnet/wpf/issues/2085 

( related: https://stackoverflow.com/questions/2342623/how-to-generate-and-print-large-xps-documents-in-wpf  and https://social.msdn.microsoft.com/Forums/en-US/f3eeaf1e-e397-44b8-819a-8a79d1e5baa5/batch-printing-of-visuals-to-xps-leads-to-outofmemoryexception )

hiteshwpfmsft:
Fixes #6296

## Description
Printing a large size of document ends up consuming a lot of memory. The change in the PR targets to address this issue.
This change releases the font resources after every four pages. We see huge memory consumption difference as the number of pages grow.
Also, with the change we make sure that all resources gests released once printing is completed.

image

## Regression
Don't know.

## Testing
Made sure existing functionality remains same. Validated this by comparing the file sizes and the content of it.

## Risk
Low.

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/ReachFramework/Serialization/RCW/IXpsOMPartUriCollection.cs
+++ b/src/Microsoft.DotNet.Wpf/src/ReachFramework/Serialization/RCW/IXpsOMPartUriCollection.cs
@@ -23,16 +23,16 @@ namespace System.Windows.Xps.Serialization.RCW
     [ComImport]
     internal interface IXpsOMPartUriCollection
     {
-        void Append([In] IOpcPartUri partUri);
+        uint GetCount();
 
         IOpcPartUri GetAt([In] uint index);
-
-        uint GetCount();
 
         void InsertAt([In] uint index, [In] IOpcPartUri partUri);
 
         void RemoveAt([In] uint index);
 
         void SetAt([In] uint index, [In] IOpcPartUri partUri);
+        
+        void Append([In] IOpcPartUri partUri);
     }
 }
--- a/src/Microsoft.DotNet.Wpf/src/ReachFramework/Serialization/RCW/IXpsOMFontResource.cs
+++ b/src/Microsoft.DotNet.Wpf/src/ReachFramework/Serialization/RCW/IXpsOMFontResource.cs
@@ -35,5 +35,9 @@ namespace System.Windows.Xps.Serialization.RCW
         [MethodImpl(MethodImplOptions.InternalCall)]
         [return: ComAliasName("System.Windows.Xps.Serialization.RCW.XPS_FONT_EMBEDDING")]
         XPS_FONT_EMBEDDING GetEmbeddingOption();
+
+        [MethodImpl(MethodImplOptions.InternalCall)]
+        [return: MarshalAs(UnmanagedType.Interface)]
+        IOpcPartUri GetPartUri();
     }
 }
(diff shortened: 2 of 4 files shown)


src/Microsoft.DotNet.Wpf/src/ReachFramework/Serialization/RCW/IXpsOMPartUriCollection.cs
```diff
@@ -23,16 +23,16 @@ namespace System.Windows.Xps.Serialization.RCW
     [ComImport]
     internal interface IXpsOMPartUriCollection
     {
-        void Append([In] IOpcPartUri partUri);
+        uint GetCount();
```
lindexi, line 26: After reading xpsobjectmodel.h, I think your changes are correct.

dipeshmsft (approved):

PR #9294, opened 2024-06-26
Avoid unlocking ComLockableWrapper when server object is already null

dipeshmsft:
Adds a check in ComLockableWrapper to verify the availability of the m_serverobject before calling Release on the object.

Mitigates [CVE-2024-21409](https://www.cve.org/CVERecord?id=CVE-2024-21409)
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/wpf/pull/9294)

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/PenImc/dll/ComLockableWrapper.hpp
+++ b/src/Microsoft.DotNet.Wpf/src/PenImc/dll/ComLockableWrapper.hpp
@@ -40,6 +40,11 @@ namespace ComUtils
         // The apartment is verified during this call.
         HRESULT Unlock();
 
+        // Unlocking a wrapper permanently nulls out the server object pointer, so a
+        // wrapper contains a non-null server object pointer if and only if it is
+        // bound to a server object which has never been unlocked.
+        bool HasNotBeenUnlocked() { return (m_serverObject != nullptr); }
+
     private:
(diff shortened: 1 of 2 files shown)


src/Microsoft.DotNet.Wpf/src/PenImc/dll/ComLockableWrapper.hpp
```diff
@@ -40,6 +40,11 @@ namespace ComUtils
         // The apartment is verified during this call.
         HRESULT Unlock();
 
+        // Unlocking a wrapper permanently nulls out the server object pointer, so a
+        // wrapper contains a non-null server object pointer if and only if it is
+        // bound to a server object which has never been unlocked.
+        bool HasNotBeenUnlocked() { return (m_serverObject != nullptr); }
```
miloush, line 46: That is a double-negative. Wouldn't `IsLocked` work better?

dipeshmsft: I kept it how it is in .NET Framework and rest of the fixes, and this particularly checks the scenario where we not yet called Unlock on the object.

pchaurasia14 (approved):

PR #9330, opened 2024-06-30
Remove Marshalling code during Color Translation, reduce allocations

h3xds1nz:
## Description

Improves slightly performance and subsequent allocations when subtracting/multiplying or just generally working with `Color` using a custom profile / `ColorContext` by removing `Marshal` code/allocs and simply using `stackalloc` for the buffer, replacing some dead code (It is obvious `Translate` method has been rewritten previously) and removing the `try`/`finally` block used for unmanaged memory deallocation.

Redundant array inits to `0` have been removed in `ICM2Color` and also some heap-allocated arrays have been replaced with stack-allocated spans in the call chain.

Last two commits are not included in benchmarks but are pretty straightforward improvements.

### Color Subtraction with custom profile

| Method     | Mean     | Error     | StdDev    | Gen0   | Allocated |
|----------- |---------:|----------:|----------:|-------:|----------:|
| Marshalled | 2.919 ms | 0.0460 ms | 0.0408 ms | 3.9063 |  79.03 KB |
| StackAlloc | 2.887 ms | 0.0555 ms | 0.0519 ms | 3.9063 |  78.82 KB |

```c#
public Uri sRGB = new("C:\\Windows\\System32\\spool\\drivers\\color\\sRGB Color Space Profile.icm", UriKind.Absolute);

[Benchmark]
public Color Benchmark()
{
    Color x = Color.FromAValues(1.0f, [0.0f, 1.0f, 1.0f], sRGB);
    Color y = Color.FromAValues(0.5f, [0.0f, 0.5f, 1.0f], sRGB);

    return x - y;
}
```

</details>

### Multiplication by factor with custom profile
 
| Method     | Mean     | Error     | StdDev    | Allocated |
|----------- |---------:|----------:|----------:|----------:|
| Marshalled | 1.918 ms | 0.0116 ms | 0.0097 ms |  47.67 KB |
| StackAlloc | 1.915 ms | 0.0153 ms | 0.0144 ms |  47.53 KB |

```c#
public Uri sRGB = new("C:\\Windows\\System32\\spool\\drivers\\color\\sRGB Color Space Profile.icm", UriKind.Absolute);

[Benchmark]
public Color Benchmark()
{
    Color x = Color.FromAValues(1.0f, [0.0f, 1.0f, 1.0f], sRGB);

    return x * 0.1f;
}
```

</details>

### Raw Translate method call chain

| Method     | Mean      | Error    | StdDev   | Gen0   | Allocated |
|----------- |----------:|---------:|---------:|-------:|----------:|
| Marshalled | 118.36 ns | 1.284 ns | 1.201 ns | 0.0086 |     144 B |
| StackAlloc |  22.80 ns | 0.430 ns | 0.402 ns | 0.0043 |      72 B |

## Customer Impact

## Regression

None.

## Testing

Build, benchmark, using the code and comparing outputs (addition, subtraction, multiplication).

## Risk

Low.

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/wpf/pull/9330)

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Media/ColorTransform.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Media/ColorTransform.cs
@@ -106,52 +106,30 @@ namespace System.Windows.Media
 
         #region Internal Methods
 
-        internal void Translate(float[] srcValue, float[] dstValue)
+        internal unsafe void Translate(float[] srcValue, float[] dstValue)
         {
-            // 3. create Win32 unmanaged profile handle from memory source profile using OpenColorProfileW
-            IntPtr[] pProfiles = new IntPtr[2];
+            // Transform colors using TranslateColors
+            const UInt32 NumColors = 1;
 
-            IntPtr paInputColors = IntPtr.Zero;
-            IntPtr paOutputColors = IntPtr.Zero;
+            // There's no SkipLocalsInit, will be all zeroes
+            long* paInputColors = stackalloc long[64 / sizeof(long)];
+            long* paOutputColors = stackalloc long[64 / sizeof(long)];
 
-            try
+            long inputColor = ICM2Color(srcValue);
+            paInputColors[0] = inputColor;
+
+            _colorTransformHelper.TranslateColors((nint)paInputColors, NumColors, _inputColorType, (nint)paOutputColors, _outputColorType);
+
+            long outputColor = paOutputColors[0];
+            for (int i = 0; i < dstValue.Length; i++)
             {
-                // 6. transform colors using TranslateColors
-                UInt32 numColors = 1;
-                long inputColor = ICM2Color(srcValue);
-                paInputColors = Marshal.AllocHGlobal(64);
+                UInt32 result = 0x0000ffff & (UInt32)(outputColor >> (16 * i));
+                float a = (result & 0x7fffffff) / (float)0x10000;
 
-                Marshal.WriteInt64(paInputColors, inputColor);
-
-                paOutputColors = Marshal.AllocHGlobal(64);
-                long outputColor = 0;
-
-                Marshal.WriteInt64(paOutputColors, outputColor);
-
-                _colorTransformHelper.TranslateColors(
-                    (IntPtr)paInputColors,
-                    numColors,
-                    _inputColorType,
-                    (IntPtr)paOutputColors,
-                    _outputColorType
-                    );
-
-                outputColor = Marshal.ReadInt64(paOutputColors);
-                for (int i = 0; i < dstValue.GetLength(0); i++)
-                {
-                    UInt32 result = 0x0000ffff & (UInt32)(outputColor >> (16 * i));
-                    float a = (result & 0x7fffffff) / (float)(0x10000);
-
-                    if (result < 0)
-                        dstValue[i] = -a;
-                    else
-                        dstValue[i] = a;
-                }
-            }
-            finally
-            {
-                Marshal.FreeHGlobal(paInputColors);
-                Marshal.FreeHGlobal(paOutputColors);
+                if (result < 0)
+                    dstValue[i] = -a;
+                else
+                    dstValue[i] = a;
             }
         }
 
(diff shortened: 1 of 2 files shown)


miloush:
LGTM

src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Media/ColorTransform.cs
```diff
@@ -204,11 +181,9 @@ private long ICM2Color(float[] srcValue)
             }
             else
             {
-                byte[] channel = new byte[8];
+                Span<byte> channel = stackalloc byte[8];
```
hiteshwpfmsft, line 185: I think we should call clear here, just to make sure all values are initialized to zero.
channel.Clear();

src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Media/ColorTransform.cs
```diff
@@ -170,21 +148,20 @@ private void InitializeICM()
...
+            if (srcValue.Length < 3 || srcValue.Length > 8)
             {
                 throw new NotSupportedException(); // Only support color spaces with 3,4,5,6,7,8 channels
             }
 
-            if (srcValue.GetLength(0) <= 4)
+            if (srcValue.Length <= 4)
             {
-                UInt16[] channel = new UInt16[4];
+                Span<UInt16> channel = stackalloc UInt16[4];
```
hiteshwpfmsft, line 163: Same I would suggest here - 
channel.Clear();

miloush: The `src.Value` is gauranteed to be 4 here. All `channel` values are assigned below.

src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Media/ColorTransform.cs
miloush: This one seems like a valid point, since the `srcValue.Length` can be less than 8, and all the 8 values are then used below.

miloush:
> using the code and comparing outputs (addition, subtraction, multiplication).

Did you test it on multichannel color contexts?

src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Media/ColorTransform.cs
hiteshwpfmsft: But here the check says that it could be less that 4 too.
if (srcValue.Length <= 4)

miloush: Oh sorry it can also be 3, my bad, you are correct!

src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Media/ColorTransform.cs
h3xds1nz: @hiteshwpfmsft @miloush I do not see the reason for that, same in the other case. If .localsinit is emitted, which is by default unless [SkipLocalsInit] attribute is specified (it is not here), all stackalloc memory is zero-init by default (besides that one-time when Roslyn didn't do it, which was a bug and was fixed).

miloush: Hm are you suggesting this documentation is outdated?

> The content of the newly allocated memory is undefined.

https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/operators/stackalloc

h3xds1nz: Very technically speaking, the documentation is correct. The `stackalloc` keyword doesn't gurantee the zero-init of the "allocated" memory. However, it is guranteed that Roslyn will emit .`localsinit` if `stackalloc` (`localloc` in IL) is present and the `SkipLocalsInit` attribute has not been specified.

And with that, it is guranteed that JIT will generate efficient stack zero-init loop. Anything else is a bug.

miloush: More relevant documentation here: https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/proposals/csharp-9.0/skip-localsinit

Thank you @h3xds1nz, I am convinced this gets zeroed.

hiteshwpfmsft (approved):

src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Media/ColorTransform.cs
h3xds1nz: Thanks guys for the discussion.

harshit7962:

h3xds1nz:
@harshit7962 Thank you, happy to see this one finally get in.

PR #9333, opened 2024-07-01
Simplify constructor in CachedBitmap by removing duplicate type checks

h3xds1nz:
## Description

Removes redundant type checks second-time around when constructing `CachedBitmap` (this constructor is used by `BitmapSource.Create()`. There's minimal performance benefit, the main goal here is code simplification, that has been made possible with the introduction of `GetArrayDataReference` few releases ago.

Method has been tested with all array types for correct functionality when constructing in-memory Bitmap.

## Customer Impact

Smaller code-size of PresentationCore.

## Regression

None.

## Testing

Build, constructing Bitmap from memory source.

## Risk

Low.

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/wpf/pull/9333)

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Media/Imaging/CachedBitmap.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Media/Imaging/CachedBitmap.cs
@@ -1,9 +1,9 @@
 // Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the MIT license.
-// See the LICENSE file in the project root for more information.
 
-using MS.Internal;
+using System.Runtime.InteropServices;
 using MS.Win32.PresentationCore;
+using MS.Internal;
 
 namespace System.Windows.Media.Imaging
 {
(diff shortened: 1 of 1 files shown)


h3xds1nz:
@siagupta0202 The auto merge results in a missing using for `System.Runtime.InteropServices`, I'll add it later today.

By the way, #9395 is almost the same if you wanna have a look.

src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Media/Imaging/CachedBitmap.cs
```diff
@@ -1,9 +1,9 @@
 // Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the MIT license.
-// See the LICENSE file in the project root for more information.
```
siagupta0202, line 3: Is this change intentional?

h3xds1nz: So that being said, it would be intentional.

siagupta0202 (approved):

siagupta0202:
@h3xds1nz Thank you for your contribution!

h3xds1nz:
@siagupta0202 Thank you :)

PR #9352, opened 2024-07-04
Fix InvalidOperationException in AvTrace and improve general thread-safety

### LINKED ISSUE #9347: Sporadic crash in  `AvTrace.IsWpfTracingEnabledInRegistry()` when calling `PresentationTraceSources.Refresh()` from non-UI thread
### Description

Calling `PresentationTraceSources.Refresh()` from a thread other than the UI thread may sporadically lead to `System.InvalidOperationException: Nullable object must have a value.` in `AvTrace.IsWpfTracingEnabledInRegistry()` (see two possible stack traces below).

The documentation of `PresentationTraceSources.Refresh()` does not mention anything about having to be called on a specific thread. And looking at [the source code implicated by the stack traces](https://github.com/dotnet/wpf/blob/f2a578639ada4f5d9b06d701e657667b4ec9c634/src/Microsoft.DotNet.Wpf/src/WindowsBase/MS/Internal/AvTrace.cs#L217-L241), it seems like the class is intended to be thread-safe, in particular because of the comment in `IsWpfTracingEnabledInRegistry()`:

```csharp
if( _enabledInRegistry == null )
{
	[...]

	// Update the static.  Doing this last protects us from threading problems; worse case, multiple
	// threads will set the same value into it.
	_enabledInRegistry = enabled;
}

return (bool) _enabledInRegistry;
```

However turns out this comment isn't quiet correct: Doing it twice isn't the worst case. A thread other than the UI thread may call `PresentationTraceSources.Refresh()`, which in turn calls `AvTrace.Refresh()`. If this happens between the assignment (`_enabledInRegistry = enabled`) and the return statement then the `Nullable<bool> _enabledInRegistry` will not have a value and the cast to `bool` throws an exception.

```csharp
public void Refresh()
{
	// Cause the registry to be re-read in case it's changed.
	_enabledInRegistry = null;

	// Re-initialize everything
	Initialize();
}
```
<details>
  <summary>Stack traces</summary>

```
Unhandled exception. System.Windows.Markup.XamlParseException: 'Initialization of 'System.Windows.Controls.TextBlock' threw an exception.' Line number '9' and line position '76'.
 ---> System.Xaml.XamlObjectWriterException: Set property 'System.Windows.ResourceDictionary.DeferrableContent' threw an exception.
 ---> System.TypeInitializationException: The type initializer for 'MS.Internal.TraceResourceDictionary' threw an exception.
 ---> System.InvalidOperationException: Nullable object must have a value.
   at System.Nullable`1.get_Value()
   at MS.Internal.AvTrace.IsWpfTracingEnabledInRegistry()
   at MS.Internal.AvTrace.ShouldCreateTraceSources()
   at MS.Internal.AvTrace.Initialize()
   at MS.Internal.TraceResourceDictionary..cctor()
   at MS.Internal.TraceResourceDictionary.get_IsEnabled()
   at System.Windows.ResourceDictionary.SetKeys(IList`1 keyCollection, IServiceProvider serviceProvider)
   at MS.Internal.Xaml.Runtime.ClrObjectRuntime.SetValue(Object inst, XamlMember property, Object value)
   at MS.Internal.Xaml.Runtime.ClrObjectRuntime.SetValue(Object inst, XamlMember property, Object value)
   at MS.Internal.Xaml.Runtime.PartialTrustTolerantRuntime.SetValue(Object obj, XamlMember property, Object value)
   at System.Xaml.XamlObjectWriter.Logic_ApplyPropertyValue(ObjectWriterContext ctx, XamlMember prop, Object value, Boolean onParent)
   at System.Xaml.XamlObjectWriter.Logic_DoAssignmentToParentProperty(ObjectWriterContext ctx)
   at System.Xaml.XamlObjectWriter.WriteEndMember()
   at System.Xaml.XamlServices.Transform(XamlReader xamlReader, XamlWriter xamlWriter, Boolean closeWriter)
   at System.Windows.SystemResources.ResourceDictionaries.LoadDictionary(Assembly assembly, String assemblyName, String resourceName, Boolean isTraceEnabled, Uri& dictionarySourceUri)
   at System.Windows.SystemResources.ResourceDictionaries.LoadThemedDictionary(Boolean isTraceEnabled)
   at System.Windows.SystemResources.FindDictionaryResource(Object key, Type typeKey, ResourceKey resourceKey, Boolean isTraceEnabled, Boolean allowDeferredResourceReference, Boolean mustReturnDeferredResourceReference, Boolean& canCache)
   at System.Windows.SystemResources.FindResourceInternal(Object key, Boolean allowDeferredResourceReference, Boolean mustReturnDeferredResourceReference)
   at System.Windows.StyleHelper.GetThemeStyle(FrameworkElement fe, FrameworkContentElement fce)
   at System.Windows.FrameworkElement.UpdateThemeStyleProperty()
   at System.Windows.FrameworkElement.OnInitialized(EventArgs e)
   at MS.Internal.Xaml.Runtime.ClrObjectRuntime.InitializationGuard(XamlType xamlType, Object obj, Boolean begin)
   at System.Windows.Markup.XamlReader.RewrapException(Exception e, IXamlLineInfo lineInfo, Uri baseUri)
   at System.Windows.Markup.WpfXamlLoader.Load(XamlReader xamlReader, IXamlObjectWriterFactory writerFactory, Boolean skipJournaledProperties, Object rootObject, XamlObjectWriterSettings settings, Uri baseUri)
   at System.Windows.Markup.WpfXamlLoader.LoadBaml(XamlReader xamlReader, Boolean skipJournaledProperties, Object rootObject, XamlAccessLevel accessLevel, Uri baseUri)
   at System.Windows.Markup.XamlReader.LoadBaml(Stream stream, ParserContext parserContext, Object parent, Boolean closeStream)
   at System.Windows.Application.LoadBamlStreamWithSyncInfo(Stream stream, ParserContext pc)
   at System.Windows.Application.DoStartup()
   at System.Windows.Application.<.ctor>b__1_0(Object unused)
   at System.Windows.Threading.DispatcherOperation.InvokeImpl()
   at MS.Internal.CulturePreservingExecutionContext.CallbackWrapper(Object obj)
   at System.Threading.ExecutionContext.RunInternal(ExecutionContext executionContext, ContextCallback callback, Object state)
--- End of stack trace from previous location ---
   at System.Threading.ExecutionContext.RunInternal(ExecutionContext executionContext, ContextCallback callback, Object state)
   at MS.Internal.CulturePreservingExecutionContext.Run(CulturePreservingExecutionContext executionContext, ContextCallback callback, Object state)
   at System.Windows.Threading.DispatcherOperation.Invoke()
   at System.Windows.Threading.Dispatcher.ProcessQueue()
   at System.Windows.Threading.Dispatcher.WndProcHook(IntPtr hwnd, Int32 msg, IntPtr wParam, IntPtr lParam, Boolean& handled)
   at System.Windows.Application.RunDispatcher(Object ignore)
   at System.Windows.Application.RunInternal(Window window)
   at WpfApp1.Program.Main()
```

```
Unhandled exception. System.TypeInitializationException: The type initializer for 'MS.Internal.TraceRoutedEvent' threw an exception.
 ---> System.InvalidOperationException: Nullable object must have a value.
   at System.Nullable`1.get_Value()
   at MS.Internal.AvTrace.IsWpfTracingEnabledInRegistry()
   at MS.Internal.AvTrace.ShouldCreateTraceSources()
   at MS.Internal.AvTrace.Initialize()
   at MS.Internal.TraceRoutedEvent..cctor()
   at MS.Internal.TraceRoutedEvent.get_IsEnabled()
   at System.Windows.FrameworkElement.OnRenderSizeChanged(SizeChangedInfo sizeInfo)
   at System.Windows.ContextLayoutManager.fireSizeChangedEvents()
   at System.Windows.ContextLayoutManager.UpdateLayout()
   at System.Windows.Interop.HwndSource.SetLayoutSize()
   at System.Windows.Interop.HwndSource.set_RootVisualInternal(Visual value)
   at System.Windows.Window.SetRootVisualAndUpdateSTC()
   at System.Windows.Window.SetupInitialState(Double requestedTop, Double requestedLeft, Double requestedWidth, Double requestedHeight)
   at System.Windows.Window.CreateSourceWindow(Boolean duringShow)
   at System.Windows.Threading.DispatcherOperation.InvokeImpl()
   at MS.Internal.CulturePreservingExecutionContext.CallbackWrapper(Object obj)
   at System.Threading.ExecutionContext.RunInternal(ExecutionContext executionContext, ContextCallback callback, Object state)
--- End of stack trace from previous location ---
   at System.Threading.ExecutionContext.RunInternal(ExecutionContext executionContext, ContextCallback callback, Object state)
   at MS.Internal.CulturePreservingExecutionContext.Run(CulturePreservingExecutionContext executionContext, ContextCallback callback, Object state)
   at System.Windows.Threading.DispatcherOperation.Invoke()
   at System.Windows.Threading.Dispatcher.ProcessQueue()
   at System.Windows.Threading.Dispatcher.WndProcHook(IntPtr hwnd, Int32 msg, IntPtr wParam, IntPtr lParam, Boolean& handled)
   at System.Windows.Application.RunDispatcher(Object ignore)
   at System.Windows.Application.RunInternal(Window window)
   at WpfApp1.Program.Main()
```
</details>

### Reproduction Steps

We are experiencing this as a very rare occurrence on some machines. I have created a [minimal reproduction case that forces the issue to happen more frequently](https://github.com/ssrmm/AvTraceCrashRepro).

Simply starting this application a few times produces the crash pretty easily on my machine.

### Expected behavior

Preferred:
Crash is avoided by making `AvTrace` properly thread-safe.

Alternatively:
Requirement of calling `PresentationTraceSources.Refresh()` on the UI thread is documented and ideally enforced by always throwing an exception when called on the wrong thread.

### Actual behavior

A crash occurs very rarely and only on specific systems, making this problem hard to find during testing.

### Regression?

Looking at the Git history, a regression seems unlikely. The code in question seems to have been unchanged for a long time.

### Known Workarounds

Perform calls to `PresentationTraceSources.Refresh()` on the UI thread only.

### Impact

Rare application startup crash that may only happen on customer machines and is difficult to find during development/testing.

### Configuration

.NET 6/.NET 8 x64
Windows 10/Windows 11

### Other information

_No response_

h3xds1nz:
Fixes #9347

## Description

As I believe `AvTrace` should be close to thread-safe class and it seems the original intention was for it to be from the comments, especially since even running during debug will have another thread initializing a source, based on the report in issue #9347, I've decided to create a little improvement.

- We introduce a static constructor that loads the setting from registry with true/false output (false by default) in a thread-safe manner (so that internal methods that call might call into `IsWpfTracingEnabledInRegistry` get what they came for)
- Each new instance initializes based on the latest fetched/saved in static `s_enabledInRegistry` result
- Upon calling Refresh, we reload the setting from registry and update the value in `s_enabledInRegistry` accordingly
- We then use this fetched variable down the callstack to ensure the current operation has the latest value
- `IsWpfTracingEnabledInRegistry` now merely retrieves the latest value for static `s_enabledInRegistry`
- `s_enabledInRegistry` doesn't necessarily need to be volatile but I felt like the memory fence here is worth it for consistency.
- This should improve overall performance (very minor) and stability of the `AvTrace` instances/refreshes in general.
- Also fixes a possible but unlikely race-condition in `Initialize` function between `ShouldCreateTraceSources` where a different value could have been obtained from `IsWpfTracingEnabledInRegistry` afterwards if changed by a different thread meanwhile; if the setting in registry was changed.

It is not perfect, but it prevents the exception problem and improves the behaviour overall.

## Customer Impact

Ocassionally, `InvalidOperationException` might be thrown when setting up WPF trace sources, note that the class is used by multiple threads by default without forcing it; e.g. when debugging.

## Regression

No.

## Testing

Build, local testing, verifying setting retrieval and stability.

## Risk

Low.

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/WindowsBase/MS/Internal/AvTrace.cs
+++ b/src/Microsoft.DotNet.Wpf/src/WindowsBase/MS/Internal/AvTrace.cs
@@ -59,13 +59,13 @@ namespace MS.Internal
         /// </summary>
         static AvTrace() => LoadWpfTracingSettings();
 
-        //
-        //  Refresh this trace source -- see if it needs to be enabled
-        //  because registry setting has changed or debugger is attached.
-        //
-        //  To re-read the config file, call System.Diagnostics.Trace.Refresh() .
-        //
 
+        /// <summary>
+        /// Refreshes this trace source. Checks again if it has been enabled
+        /// either because registry setting has changed or debugger has been attached.
+        /// <br/><br/>
+        ///  To re-read the config file, call <see cref="System.Diagnostics.Trace.Refresh"/>
+        /// </summary>
         public void Refresh()
         {
             // Re-read current WPF Trace settings from Registry in case it has changed.
@@ -165,7 +165,7 @@ namespace MS.Internal
         //
         // Internal initialization
         //
-       private void Initialize(bool isEnabledInRegistry)
+        private void Initialize(bool isEnabledInRegistry)
         {
             // Decide if we should actually create a TraceSource instance (doing so isn't free,
             // so we don't want to do it if we can avoid it).
@@ -198,17 +198,9 @@ namespace MS.Internal
         //  the TraceSource.)
         //
 
-        static private bool ShouldCreateTraceSources(bool isEnabledInRegistry)
+        private static bool ShouldCreateTraceSources(bool isEnabledInRegistry)
         {
-            if( isEnabledInRegistry
-                || IsDebuggerAttached()
-                || _hasBeenRefreshed
-              )
-            {
-                return true;
-            }
-
-            return false;
+            return isEnabledInRegistry || IsDebuggerAttached() || _hasBeenRefreshed;
         }
 
         /// <summary> Initializes <see cref="s_enabledInRegistry"/> variable from Registry and returns the value. </summary>
@@ -231,10 +223,10 @@ namespace MS.Internal
             return enabled;
         }
 
-        /// <summary> Retrieves cached value to see whether WPF Tracing is enabled. </summary>
+        /// <summary> Retrieves cached value from <see cref=" s_enabledInRegistry"/> to see whether WPF Tracing is enabled. </summary>
         /// <returns> A boolean value specifying whether WPF Tracing is enabled. </returns>
         [FriendAccessAllowed]
-        static internal bool IsWpfTracingEnabledInRegistry() => s_enabledInRegistry;
+        internal static bool IsWpfTracingEnabledInRegistry() => s_enabledInRegistry;
 
         //
         // Check for an attached debugger.
--- a/src/Microsoft.DotNet.Wpf/src/WindowsBase/MS/Internal/AvTrace.cs
+++ b/src/Microsoft.DotNet.Wpf/src/WindowsBase/MS/Internal/AvTrace.cs
@@ -51,9 +51,13 @@ namespace MS.Internal
             PresentationTraceSources.TraceRefresh += new TraceRefreshEventHandler(Refresh);
 
             // Fetch and cache the TraceSource from PresentationTraceSources
-            Initialize();
+            Initialize(IsWpfTracingEnabledInRegistry());
         }
 
+        /// <summary>
+        /// Performs thread-safe initialization of <see cref="s_enabledInRegistry"/> value.
+        /// </summary>
+        static AvTrace() => LoadWpfTracingSettings();
 
         //
         //  Refresh this trace source -- see if it needs to be enabled
@@ -64,11 +68,11 @@ namespace MS.Internal
 
         public void Refresh()
         {
-            // Cause the registry to be re-read in case it's changed.
-            _enabledInRegistry = null;
+            // Re-read current WPF Trace settings from Registry in case it has changed.
+            bool isEnabledInRegistry = LoadWpfTracingSettings();
 
             // Re-initialize everything
-            Initialize();
+            Initialize(isEnabledInRegistry);
         }
 
         //
@@ -161,12 +165,12 @@ namespace MS.Internal
         //
         // Internal initialization
         //
-        void Initialize( )
+       private void Initialize(bool isEnabledInRegistry)
         {
             // Decide if we should actually create a TraceSource instance (doing so isn't free,
             // so we don't want to do it if we can avoid it).
 
-            if( ShouldCreateTraceSources() )
+            if(ShouldCreateTraceSources(isEnabledInRegistry))
             {
                 // Get TraceSource from the PresentationTraceSources
                 // (this call will indirectly create the TraceSource if one doesn't already exist)
@@ -175,7 +179,7 @@ namespace MS.Internal
                 // We go enabled if tracing is enabled in the registry, if
                 // PresentationTraceSources.Refresh has been called, or if we're in the debugger
                 // and the debugger is supposed to enable tracing.
-                _isEnabled = IsWpfTracingEnabledInRegistry() || _hasBeenRefreshed || _enabledByDebugger ;
+                _isEnabled = isEnabledInRegistry || _hasBeenRefreshed || _enabledByDebugger ;
             }
             else
             {
@@ -194,9 +198,9 @@ namespace MS.Internal
         //  the TraceSource.)
         //
 
-        static private bool ShouldCreateTraceSources()
+        static private bool ShouldCreateTraceSources(bool isEnabledInRegistry)
         {
-            if( IsWpfTracingEnabledInRegistry()
+            if( isEnabledInRegistry
                 || IsDebuggerAttached()
                 || _hasBeenRefreshed
               )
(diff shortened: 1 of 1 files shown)


h3xds1nz:
While digging into this; there's one race-condition I've found along the way after fixing the linked issue, which ain't a WPF issue but imperfection/bug in the thread-safety of `TraceSource` (which is documented to be thread-safe).

When calling `PresentationTraceSources.Refresh()` right at launch, after `AvTrace`, it calls the `System.Diagnostics.Trace.Refresh()` -> `TraceSource.RefreshAll()` -> this method locks on the static list of TraceSource references (`s_tracesources`), which calls `Refresh` on each respective `TraceSource`, that goes either into  `Initialize` -> `NoConfigInit_BeforeEvent`, or directly ends up in `OnInitializing` method after init, which enumerates through its `Listeners` collection, and at that point you'll end up with either `InvalidOperationException` or `NullReferenceException` from the `Listeners` list as the access to this collection is not synchronized under the same lock. It seems to have been introduced in .NET 7, I've created an issue on [dotnet/runtime - #104457](https://github.com/dotnet/runtime/issues/104457).

The easiest workaround in that case is to just call first `Trace.Refresh()` like second after launch (or `PresentationTraceSources.Refresh()` in this case).

More simply speaking, using your the attached with my PR (or without, with multiple threads but harder to catch as it usually dies on `AvTrace` previously), if you call `Refresh` right after launch and Visual Studio subscribes to the trace source or it all happens too fast from at the same time, you'll get `InvalidOperationException` due to `TraceListenerCollection` version while enumerating or `NullReferenceException` from different thread that still sees `null`.

src/Microsoft.DotNet.Wpf/src/WindowsBase/MS/Internal/AvTrace.cs
```diff
@@ -194,53 +198,35 @@ void Initialize( )
...
 
-            return (bool) _enabledInRegistry;
+            return enabled;
         }
 
-
+        /// <summary> Retrieves cached value from <see cref=" s_enabledInRegistry"/> to see whether WPF Tracing is enabled. </summary>
+        /// <returns> A boolean value specifying whether WPF Tracing is enabled. </returns>
+        [FriendAccessAllowed]
+        internal static bool IsWpfTracingEnabledInRegistry() => s_enabledInRegistry;
```
lindexi, line 229: Can it sure already call the `LoadWpfTracingSettings` method to init the `s_enabledInRegistry` field?

h3xds1nz: @lindexi It is initialized by the static constructor, line 60. So the field will have the correct value once this is called.

h3xds1nz:
The dotnet/runtime issue has been now fixed for .NET 9, so this now works flawlessly.

h3xds1nz:
Resolved merge conflicts 05

PR #9364, opened 2024-07-07
Remove allocations on all base converters, improve TokenizerHelper

h3xds1nz:
## Description

This is probably one of the impactful PRs I've posted; optimizes conversion/parsing of all the base-generated types but also `ThicknessConverter` (once more), `CornerRadiusConverter`, `KeySplineConverter` and `VirtualizationCacheLengthConverter`.

Completely removes any allocations caused by the original Tokenizer by using `ReadOnlySpan<char>` instead of substrings.
I have also introduced a `ref struct` variant that can be used in any code that doesn't compile for `net472`.

MilCodeGen has also been adjusted to support this.

### Single Thickness parsing (4 units)

| Method     |  Mean [ns] | Error [ns] | StdDev [ns] | Code Size [B] | Gen0   | Allocated [B] |
|----------- |-----------:|-----------:|------------:|--------------:|-------:|--------------:|
| Original   |   237.4 ns |    4.64 ns |     4.34 ns |         555 B | 0.0105 |         176 B |
| PR_EDITREF |    191.6 ns |    1.62 ns |     1.51 ns |         761 B |      - |             - |

### Parsing Thickness 10 times (4 units)

| Method     |  Mean [ns]  | Error [ns] | StdDev [ns] | Code Size [B] | Gen0   | Allocated [B] |
|----------- |------------:|-----------:|------------:|--------------:|-------:|--------------:|
| Original   |  2,434.0 ns |   17.02 ns |    14.08 ns |         608 B | 0.1030 |        1760 B |
| PR_EDITREF |  1,885.1 ns |    5.68 ns |     4.44 ns |         814 B |      - |             - |

### Creation of 10x Matrix3D via Parse

| Original |   7,643.3 ns |   26.76 ns |    20.89 ns | 0.3052 |       6,876 B |        5200 B |
| PR_EDIT  |  5,885.4 ns |   67.98 ns |    60.27 ns |      - |       7,177 B |             - |

## Customer Impact

Improved performance by around 20% on base types, zeroed out allocations.

## Regression

No.

## Testing

Local build, basic testing of the tokenizer with span and some of the converters.

## Risk

Low, the scope of the PR is big but with a simple change, hence any mistakes shall be easy to figure out on a DRT run.

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/wpf/pull/9364)


src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/LengthConverter.cs
```diff
@@ -181,59 +174,61 @@ public override bool CanConvertTo(ITypeDescriptorContext typeDescriptorContext,
         // NOTE - This code is called from FontSizeConverter, so changes will affect both.
         static internal double FromString(string s, CultureInfo cultureInfo)
```
ThomasGoulet73, line 175: Could you revert your changes in this method ? I already made other changes in dotnet/wpf#8622 which I think would be slightly faster and a bit cleaner (I centralized other similar pixel unit).

src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/LengthConverter.cs
```diff
@@ -143,23 +137,22 @@ public override bool CanConvertTo(ITypeDescriptorContext typeDescriptorContext,
...
-                &&  value is double )
+            if (value is double doubleValue)
             {
-                double l = (double)value;
                 if (destinationType == typeof(string)) 
                 { 
-                    if(double.IsNaN(l)) 
-                        return "Auto";
+                    if(double.IsNaN(doubleValue)) 
+                        return NaNValue.ToString();
```
ThomasGoulet73, line 145: You should keep the string literal here because I think your changes here will regress because it now allocates a string.

src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/LengthConverter.cs
```diff
@@ -181,59 +174,61 @@ public override bool CanConvertTo(ITypeDescriptorContext typeDescriptorContext,
...
+        private static ReadOnlySpan<double> PixelUnitFactors =>
+        [ 
+            1.0,              // px - Pixel itself
+            96.0,             // in - Pixels per Inch
+            96.0 / 2.54,      // cm - Pixels per Centimeter
+            96.0 / 72.0,      // pt - Pixels per Point
+        ];
+
+        /// <summary> Holds the "Auto" string representation for <see cref="double"/> conversion. </summary>
+        private static ReadOnlySpan<char> NaNValue => ['A', 'u', 't', 'o'];
```
ThomasGoulet73, line 231: Same question as https://github.com/dotnet/wpf/pull/9363#discussion_r1669514041

src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/LengthConverter.cs
h3xds1nz: Haha, you're totally right, I just copied and overlooked afterwards, this would be a major regression. Thank you!

src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/LengthConverter.cs
h3xds1nz: @ThomasGoulet73 Those look like great changes, I'll go through them. Totally missed this PR, sorry.

src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/LengthConverter.cs
h3xds1nz: We don't need this, could be a literal, I'll fix that. I just went through different iterations previously and this is a leftover.

Thank you.

src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/LengthConverter.cs
dipeshmsft: @h3xds1nz, have you reverted the changes as mentioned above ? We are reviewing this PR for the next CTP that we will start.

h3xds1nz: @dipeshmsft I'll fix it within 30m.

h3xds1nz: @dipeshmsft Since the other for changes LengthConverter were included in #9363 instead, I propose I put this PR into draft state right now and I'll do the additional changes I have planned once Thomas's PR is merged, it will be easier.

dipeshmsft: > I propose I put this PR into draft state right now and I'll do the additional changes I have planned once Thomas' PR is merged, it will be easier.

Sure, that works for us. Infact, for us as well this was the initial plan, get Thomas's PR merged and then take a look at this PR once that is done. I will ping once the current CTP is done.

h3xds1nz:
Resolved merge conflicts 12, there are better ways to spent an hour and a half.

src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/LengthConverter.cs
```diff
@@ -191,22 +176,21 @@ static internal void FormatLengthAsString(double value, ref DefaultInterpolatedS
...
 
-            //Auto is represented and Double.NaN
-            //properties that do not want Auto and NaN to be in their ligit values,
-            //should disallow NaN in validation callbacks (same goes for negative values)
-            if (valueSpan.Equals("auto", StringComparison.OrdinalIgnoreCase))
-                return Double.NaN;
-
-            PixelUnit pixelUnit;
-            if (PixelUnit.TryParsePixel(valueSpan, out pixelUnit)
-                || PixelUnit.TryParsePixelPerInch(valueSpan, out pixelUnit)
```
ThomasGoulet73, line 207: Is the `||` or `&&` at the start of the line instead of at the end of the line wrong according to the coding style ? Both seems to be used a lot and if one isn't better than the other I don't see a reason to change it.

src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/LengthConverter.cs
```diff
@@ -191,22 +176,21 @@ static internal void FormatLengthAsString(double value, ref DefaultInterpolatedS
...
+            ReadOnlySpan<char> valueSpan = value.Trim();
             double unitFactor = 1.0;
 
-            //Auto is represented and Double.NaN
-            //properties that do not want Auto and NaN to be in their ligit values,
-            //should disallow NaN in validation callbacks (same goes for negative values)
-            if (valueSpan.Equals("auto", StringComparison.OrdinalIgnoreCase))
-                return Double.NaN;
-
-            PixelUnit pixelUnit;
```
ThomasGoulet73, line 205: When I wrote this I put it on its own line on purpose because I think it's cleaner when the variable is used multiple times. Is there a reason you changed it ? I know there's an IDE0018 suggestion to change it but projects like winforms or runtime keep it as a suggestion and reason might be that sometimes it's cleaner not to do it.

src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Media/Animation/KeySplineConverter.cs
```diff
@@ -2,140 +2,126 @@
...
-//
-
-// Allow suppression of certain presharp messages
-#pragma warning disable 1634, 1691
-
-using MS.Internal;
-using System;
-using System.ComponentModel;
 using System.ComponentModel.Design.Serialization;
+using System.Windows.Media.Animation;
```
ThomasGoulet73, line 6: Could you double check your using changes in this PR ? There's a couple places where the changes don't make sense to me, like this one where you change unsorts the usings (It might be caused by a bad merge conflict resolution).

ThomasGoulet73 (commented):
Side note: @h3xds1nz when there's a review on a PR I recommend adding new commits and doing merges instead of rebasing. It makes it much easier to track what changed since the last review.

src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Media/Animation/KeySplineConverter.cs
h3xds1nz: The usings are a mess thanks to that other PR, I will have to take a look onto the files once this PR is closer to review. I was hoping it could be before P1 but maybe P2 or P3 😀

src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/LengthConverter.cs
h3xds1nz: I will revert the stylistic changes to your code before merge if you think it's cleaner, as it is your code. I just auto formatted here (which does for me a tad more than pure .editorconfig).

src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Media/Animation/KeySplineConverter.cs
```diff
@@ -2,140 +2,126 @@
...
+
+            handler.AppendFormatted(keySpline.ControlPoint1.Y);
+            handler.AppendLiteral(listSeparator);
+
+            handler.AppendFormatted(keySpline.ControlPoint2.X);
+            handler.AppendLiteral(listSeparator);
+
+            handler.AppendFormatted(keySpline.ControlPoint2.Y);
+
+            return handler.ToStringAndClear();
```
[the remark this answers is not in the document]
ThomasGoulet73: But the fact remains that `DefaultInterpolatedStringHandler` is not supposed to be called directly, it is supposed to be called by the compiler (See the remarks [here](https://learn.microsoft.com/en-us/dotnet/api/system.runtime.compilerservices.defaultinterpolatedstringhandler.-ctor?view=net-9.0#system-runtime-compilerservices-defaultinterpolatedstringhandler-ctor(system-int32-system-int32-system-iformatprovider-system-span((system-char))))). We should avoid using it where possible, even if the syntax is less pretty. There are cases which are "fine" IMO the format specifier is not a constant or where we want to do custom formatting to the value (Like you did here: dotnet/wpf#9363).

> It would be wiser to introduce a ValueStringBuilder variant but that's for another PR; right now I'd prefer to keep it as it is and change it later along with other places its already used.

If you already plan on changing it in the future I think it would be best to use `DefaultInterpolatedStringHandler` how it's intended (By letting the compiler do it) for now and to change it in the future.

h3xds1nz:
@himgoyalmicro I've noticed this has been cut, can you share details about the failed test(s)?

This is really important PR and preferrably it should be in .NET 10.

h3xds1nz:
@himgoyalmicro Right, that was silly. Fixed.

I had to include #10683 for now to be able to run MilCodeGen, that's a PR that doesn't really need tests, could be just approved.

src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Media/Animation/KeySplineConverter.cs
```diff
@@ -1,129 +1,124 @@
...
 
-            TokenizerHelper th = new TokenizerHelper(stringValue, cultureInfo);
+            ValueTokenizerHelper tokenizer = new(stringValue, cultureInfo);
 
-            return new KeySpline(
-                Convert.ToDouble(th.NextTokenRequired(), cultureInfo),
-                Convert.ToDouble(th.NextTokenRequired(), cultureInfo),
-                Convert.ToDouble(th.NextTokenRequired(), cultureInfo),
-                Convert.ToDouble(th.NextTokenRequired(), cultureInfo));
+            return new KeySpline(double.Parse(tokenizer.NextTokenRequired(), cultureInfo),
```
himgoyalmicro, line 61: Nit: New Line

src/Microsoft.DotNet.Wpf/src/Shared/MS/Internal/TokenizerHelper.cs
```diff
@@ -91,15 +67,31 @@ internal string GetCurrentToken()
...
             return _str.Substring(_currentTokenIndex,_currentTokenLength);
         }
-        
+
+        /// <summary>
+        /// Sibling to <see cref="GetCurrentToken"/> with a minor difference; if there's no token, it will return <see cref="ReadOnlySpan{char}.Empty"/>
+        /// instead of <see langword="null"/>. However, if used with <see cref="NextToken()"/>, this edge-case is never hit.
+        /// </summary>
+        /// <returns></returns>
+        internal ReadOnlySpan<char> GetCurrentTokenAsSpan()
```
himgoyalmicro, line 76: Can we remove this and `NextTokenRequiredAsSpan` as you have implemented `ValueTokenizerHelper` for the same purpose?

h3xds1nz: I think I had additional plans here with those two regarding PBT, I'll investigate.

src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Media/Animation/KeySplineConverter.cs
h3xds1nz: Will fix, thanks.

src/Microsoft.DotNet.Wpf/src/Shared/MS/Internal/TokenizerHelper.cs
h3xds1nz: Yeah, no, they're not needed even for the future. Removed, good catch :)

src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Media/Animation/KeySplineConverter.cs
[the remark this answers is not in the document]
h3xds1nz: I've decided I will not change this, besides the fact that the remarks in the documentation are strictly because `DefaultInterpolatedStringHandler` lacks the necessary input checks that people might expect on otherwise public API (as explained directly in the source), moreover, further usability APIs have been exposed few weeks ago here (110505 - runtime) to promote the use in cases where it is needed. I'd recommend watching the API review for that if you want the stance of the team.

Also, just yesterday **DefaultInterpolatedStringHandler** was approved for a public API in HybridCache (112866 - runtime), therefore I just see no reason why to complicate things or reduce readibility.

PR #9393, opened 2024-07-13
Fixing issues with DynamicResource usage

batzen:
This PR fixes the issues introduced by my other PR #5610.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/wpf/pull/9393)

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/ResourceReferenceExpression.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/ResourceReferenceExpression.cs
@@ -278,7 +278,10 @@ namespace System.Windows
                     }
                 }
 
-                deferredResourceReference.RemoveFromDictionary();
+                // This will inflate the deferred reference, causing it
+                // to be removed from the list.  The list may also be
+                // purged of dead references.
+                deferredResourceReference.GetValue(BaseValueSourceInternal.Unknown);
             }
 
             StopListeningForFreezableChanges(resource);
@@ -341,8 +344,8 @@ namespace System.Windows
         internal void InvalidateExpressionValue(object sender, EventArgs e)
         {
             // VS has a scenario where a TreeWalk invalidates all reference expressions on a DependencyObject.
-            // If there is a dependency between RRE's, 
-            // invalidating one RRE could cause _targetObject to be null on the other RRE. Hence this check. 
+            // If there is a dependency between RRE's,
+            // invalidating one RRE could cause _targetObject to be null on the other RRE. Hence this check.
             if (_targetObject == null)
             {
                 return;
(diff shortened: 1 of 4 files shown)


batzen:
@harshit7962 This fixes the issues introduced by my PR.

src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/ResourceReferenceExpression.cs
```diff
@@ -278,7 +278,10 @@ private void InvalidateCacheValue()
                     }
                 }
 
-                deferredResourceReference.RemoveFromDictionary();
+                // This will inflate the deferred reference, causing it
+                // to be removed from the list.  The list may also be
+                // purged of dead references.
+                deferredResourceReference.GetValue(BaseValueSourceInternal.Unknown);
```
dipeshmsft, line 284: Why are we doing this changer here? On calling GetValue, the value will again get stored in _keyOrValue.

batzen: Because the instances of the deferred resource are now shared.
When the instance is simply removed from the dictionary there is no way for other consumers to ever get the value.
That issue appeared after i fixed the initial issue.

singhashish-wpf: @batzen wanted to understand if canCache be false in any case, thereby we will not be making a call to the RemoveFromDictionary in that scenario, which might break the code equivalence. Do we have a guarantee that canCache will always be true with the changes?

batzen: I don't see any canCache here.

harshit7962: We are talking about the canCache [here](https://github.com/dotnet/wpf/blob/main/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/SystemResources.cs#L1744). Just to keep things in equivalence, we wanted to make sure that `canCache` in this case will always be `True` so that `RemoveFromDictionary` is always called.

batzen: There is no guarantee that canCache is always true and it doesn't need to.
We now should only have one deferred resource per key per dictionary. If there are no references to a deferred resource left it gets purged after a while anyway.
Before my changes, including the other PR, every request for a resource might have created a new deferred resource so it was quite important to remove it as soon as possible as we would end up with thousands of instances which, even with the call to remove, was the case.
And because we should only have one shared instance, instead of thousands, now it's important to not simply call remove as that would render the instance unusable for all other consumers.

singhashish-wpf (approved):

Issue #9295, opened 2024-06-26
[API Proposal] Introducing AccentColor as SystemColors


dipeshmsft:
## Background and Motivation

Since Windows 10 onwards, Accent color has been one of the key compoenents of visual styles in Windows controls. However, there is no easy way to get the system's accent color. Accent colors can be used in custom styles to align the styling of controls with what the OS is showing.

In earlier versions of Windows, visual styles were defined using system colors and brushes, and corresponding to that we had `SystemColors` static class that contained getters to all the system colors, brushes and corresponding resource keys for using the colors and brushes in XAML.

Since, Accent color and it's variations are also system wide defined colors, we propose to expose them using the `SystemColors` class.

## API Proposal

```csharp

    public static class SystemColors
    {
        // Accent Colors
        public static Color AccentColor;
        public static Color AccentColorLight1;
        public static Color AccentColorLight2;
        public static Color AccentColorLight3;
        public static Color AccentColorDark1;
        public static Color AccentColorDark2;
        public static Color AccentColorDark3;

        // Accent Color Keys
        public static ResourceKey AccentColorKey;
        public static ResourceKey AccentColorLight1Key;
        public static ResourceKey AccentColorLight2Key;
        public static ResourceKey AccentColorLight3Key;
        public static ResourceKey AccentColorDark1Key;
        public static ResourceKey AccentColorDark2Key;
        public static ResourceKey AccentColorDark3Key;
        
        // Accent Color Brushes
        public static SolidColorBrush AccentColorBrush;
        public static SolidColorBrush AccentColorLight1Brush;
        public static SolidColorBrush AccentColorLight2Brush;
        public static SolidColorBrush AccentColorLight3Brush;
        public static SolidColorBrush AccentColorDark1Brush;
        public static SolidColorBrush AccentColorDark2Brush;
        public static SolidColorBrush AccentColorDark3Brush;

        // Accent Color Brush Keys
        public static ResourceKey AccentColorBrushKey;
        public static ResourceKey AccentColorLight1BrushKey;
        public static ResourceKey AccentColorLight2BrushKey;
        public static ResourceKey AccentColorLight3BrushKey;
        public static ResourceKey AccentColorDark1BrushKey;
        public static ResourceKey AccentColorDark2BrushKey;
        public static ResourceKey AccentColorDark3BrushKey;
    }
```

[...]

PR #9400, opened 2024-07-15
Implemented SystemColors.AccentColor API

dipeshmsft:
Fixes # 
#9295 
Main PR 

## Customer Impact
WPF Developers will now have an API to access the window AccentColors

## Regression
N/A

## Testing
Local ad-hoc testing, using the new API in sample apps.


src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/AccentColorHelper.cs
```diff
@@ -0,0 +1,64 @@
...
+
+            return _uiSettings;
+        }
+    }
+
+    #endregion
+
+    #region Private Fields
+
+    private static Color _defaultAccentColor = Color.FromArgb(0xff, 0x00, 0x78, 0xd4);
```
h3xds1nz, line 59: This could be readonly, right?

src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/AccentColorHelper.cs
```diff
@@ -0,0 +1,64 @@
...
+    internal static Color GetAccentColor(UISettingsRCW.UIColorType uiColorType = UISettingsRCW.UIColorType.Accent)
+    {
+        Color color = _defaultAccentColor;
+
+        if (_UISettings.TryGetColorValue(uiColorType, out color))
+        {
+            return color;
+        }
+
+        return color;
```
miloush, line 27: I believe color would be overwritten here in both cases, and the initialization is not necessary in the success case. I suggest removing line 20, doing `out Color color` and 27 just `return _defaultAccentColor`

h3xds1nz: Yeah, agreed.

singhashish-wpf (approved):

PR #9404, opened 2024-07-15
Replace accent color resources to use SystemColors.AccentColor

dipeshmsft:
## Description
In this PR, I have modified the resources in Fluent color dictionaries to use SystemColors.AccentColor. With this change, we don't need to have DWMColorization to manage the AccentColor resources seperately, hence removed DWMColorization as well.

This PR is based on the implementation of AccentColors in SystemColors PR ( #9400 )

## Regression
N/A

## Testing
Local build. Validated using the WPF Gallery application.

###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/wpf/pull/9404)


src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/ThemeManager.cs
```diff
@@ -232,6 +227,31 @@ private static Uri GetFluentWindowThemeColorResourceUri(string systemTheme, bool
...
     }
 
+    private static void FindFluentThemeAndColorDictionary(out ResourceDictionary fluentThemeDictionary, out ResourceDictionary fluentColorDictionary)
+    {
+        fluentThemeDictionary = null;
+        fluentColorDictionary = null;
+
+        if (Application.Current == null) return;
+
+        foreach (ResourceDictionary mergedDictionary in Application.Current.Resources.MergedDictionaries)
```
singhashish-wpf, line 237: Is there a possibility to optimize this?

singhashish-wpf (dismissed):

src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/ThemeManager.cs
miloush: One option would be to always put the dictionary as first (which probably we should be doing anyway) and then only check the first one or two.

dipeshmsft: I will go ahead, resolve this thread and raise subsequent PRs where I have changed the method entirely. It will be the second PR from here on.

hiteshwpfmsft (approved):

anjali-wpf (approved):

PR #9407, opened 2024-07-16
Enabling Fluent theme in Windows 10

harshit7962:
## Description
The current implementation of `Fluent` themes does not work in Windows 10. The following changes makes the control styles and theme effects affective for windows 10 as well. 

## Customer Impact
Customers will be able to use Fluent themes in Windows 10 by including the Fluent resource dictionary as they could have in windows 11.

## Regression
_None_

## Testing
Local Build Pass
Sample Application Testing

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/wpf/pull/9407)

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/ThemeManager.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/ThemeManager.cs
@@ -132,7 +132,7 @@ internal static class ThemeManager
     /// <returns></returns>
     private static bool SetImmersiveDarkMode(Window window, bool useDarkMode)
     {
-        if (window == null)
+        if (window == null || !Utility.IsOSWindows11OrNewer)
         {
             return false;
         }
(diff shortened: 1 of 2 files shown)


src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/ThemeManager.cs
```diff
@@ -116,7 +116,10 @@ internal static void ApplySystemTheme(IEnumerable windows = null, bool forceUpda
                 }
                 
                 SetImmersiveDarkMode(window, !requestedUseLightMode);
```
dipeshmsft, line 118: Can you check if the call made in SetImmersiveDarkMode is supported on Windows10 or not ?

dipeshmsft: Yeah, the attribute ( DWMWA_USE_IMMERSIVE_DARK_MOD ) that we are using here for SetImmersiveDarkMode, is available in Win11 only : https://learn.microsoft.com/en-us/windows/win32/api/dwmapi/ne-dwmapi-dwmwindowattribute#constants

So, we should move SetImmersiveDarkMode call inside the Win11 check

harshit7962: Addressed.

dipeshmsft (dismissed):

dipeshmsft (approved):

PR #9441, opened 2024-07-22
Debug::Assert(false to Debug::Fail

kasperk81:
https://github.com/dotnet/sdk/pull/42262#issuecomment-2243885156
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/wpf/pull/9441)

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/System.Printing/CPP/src/GDIExporter/gdibitmap.cpp
+++ b/src/Microsoft.DotNet.Wpf/src/System.Printing/CPP/src/GDIExporter/gdibitmap.cpp
@@ -188,7 +188,7 @@ void CGDIBitmap::SetupPalette(interior_ptr<BITMAPINFO> bmi, int bitCount)
     }
     else
     {
-        Debug::Assert(false, "Unsupported format");
+        Debug::Fail("Unsupported format");
     }
 
     bmi->bmiHeader.biClrUsed      = count;
@@ -242,7 +242,7 @@ void CGDIBitmap::SetBits(interior_ptr<BITMAPINFO> bmi)
     }
     else
     {
-        Debug::Assert(false, "Unexpected bitcount");
+        Debug::Fail("Unexpected bitcount");
         m_Buffer = nullptr;
     }
 }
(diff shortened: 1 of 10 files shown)


src/Microsoft.DotNet.Wpf/src/System.Printing/CPP/src/GDIExporter/gdibitmap.cpp
```diff
@@ -188,7 +188,7 @@ void CGDIBitmap::SetupPalette(interior_ptr<BITMAPINFO> bmi, int bitCount)
     }
     else
     {
-        Debug::Assert(false, "Unsupported format");
+        Debug::Fail("Unsupported format");
```
333fred, line 191: For these, did the compilation process just fail before it got here, and that's why we only saw 4 errors? Or is there something else going on?

333fred: Yes. Stephen pulled out 4 error messages from the build, but there are 15 changes here. I'm asking if the other 11 are just from further down in the build tree and we didn't attempt to build them, so we didn't see 15 errors, or if something more complex is going on.

stephentoub: I think this is a search/replace for all Debug.Assert(false. I'm concerned the sdk build ended after those first four, and that even this set of 15 won't fix it, as I see lots of two-arg asserts with non "false" codinitions throughout the C++/CLI code.  But I also don't know what the overload resolution problem is yet and so don't know how to reason about it.

stephentoub (approved):

akoeplinger:
I ran a test build of this change on https://github.com/dotnet/dotnet/pull/101 and it fixes the issue so this is ready to be merged.

anjali-wpf (approved):
Thanks @kasperk81 and @akoeplinger

PR #9460, opened 2024-07-25
remove allocation in eventroute

kasperk81:
https://github.com/dotnet/wpf/issues/9467?notification_referrer_id=NT_kwDOBPO9d7QxMTQ4NTM4MjA3NTo4MzA4MjYxNQ#issuecomment-2249972593
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/wpf/pull/9460)


h3xds1nz:
Imho @lindexi solution would fix the immediate problem with releases in a more safe manner.

This whole thing with `ROS<object>` and stackallocs should be done in a more complex that fixes up the call chain from top to bottom with the additional allocations in the functions as you will miss plenty this way.

Do note also that you would want to use with C#13 `params ReadOnlySpan<object>` already.

PR #9361, opened 2024-07-05
Spanify AntiFormat function in AvTrace, decrease allocations, improve perf

h3xds1nz:
## Description

Improves performance of the `AntiFormat` function in `AvTrace`, reducing allocations and improving the overall performance up to 2 times in some cases. Also changes `FormatChars` from `char[]` to `ReadOnlySpan<char>`, occurrence I've missed in #9230.

Second commit features added documentation for the function. I've included some basic asserts at the **Testing** section.

### Small strings within `StringBuilder` default capacity

| Method     | Mean [ns] | Error [ns] | StdDev [ns] | Gen0   | Allocated [B] |
|----------- |----------:|-----------:|------------:|-------:|--------------:|
| PR__Edit |  236.1 ns |    4.75 ns |     5.84 ns | 0.0730 |        1224 B |
| Original |  400.7 ns |    8.04 ns |    15.29 ns | 0.1040 |        1744 B |

<details><summary>Benchmark code</summary>

```c#
AntiFormat2("{aaaa}");
AntiFormat2("{aaaa}}");
AntiFormat2("{{aaaa}}");
AntiFormat2("{aa{{aa}}");
AntiFormat2("{aa{{aa}}ab}c}d");
AntiFormat2("aaaa}");
AntiFormat2("aaaa");
AntiFormat2("a{{{aaa");
AntiFormat2("a{{{aaa}}}");

```

</details>

### Large strings, mostly without any replacements

| Method     | Mean [ns] | Error [ns] | StdDev [ns] | Gen0   | Allocated [B] |
|----------- |----------:|-----------:|------------:|-------:|--------------:|
| PR__Edit|  127.5 ns |    1.29 ns |     1.08 ns | 0.0880 |        1472 B |
| Original |  242.9 ns |    4.87 ns |     5.00 ns | 0.1094 |        1832 B |

<details><summary>Benchmark code</summary>

```c#
AntiFormat("aaaaxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx");
AntiFormat("aaaaxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx");
AntiFormat("aaaxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxa");
[...]

PR #9468, opened 2024-07-25
Modify AvTrace call chain to use params ReadOnlySpan<object> instead of an array

### LINKED ISSUE #9467: WPF control is not reclaimed on NET9 by GC
### Description

GC does not reclaim control memory in following scenario. The control is placed in the visual tree and then removed from it. Conditions must be met:
- PresentationTraceSources.Refresh() call
- there is a subscriber to the Unloaded event of control

### Reproduction Steps

Run test `TestLeak` in project targeting net9.0-windows. NUnit version is 3.12.0.

```csharp
using System;
using System.Diagnostics;
using System.Windows;
using System.Windows.Controls;
using System.Windows.Threading;
using NUnit.Framework;

namespace Sample {
    [TestFixture]
    public class MemoryTests {
        [Test]
        public void TestLeak() {
            PresentationTraceSources.Refresh();
            var window = new Window();
            var wr = AddAndRemoveImage(window);
            GC.Collect();
            GC.WaitForPendingFinalizers();
            Assert.IsFalse(wr.IsAlive);
        }

        static WeakReference AddAndRemoveImage(Window window) {
            var img = new Image();
            var wr = new WeakReference(img);
            var holder = new ImageHolder(img);
            window.Content = img;
            window.Show();
            DoEvents();
            window.Content = null;
            DoEvents();
            return wr;
        }

        static void DoEvents() {
            var frame = new DispatcherFrame();
            Dispatcher.CurrentDispatcher.InvokeAsync(() => frame.Continue = false, DispatcherPriority.Background);
            Dispatcher.PushFrame(frame);
        }

        class ImageHolder {
            readonly Image image;
            public ImageHolder(Image image) {
                this.image = image;
                this.image.Loaded += OnLoaded;
                this.image.Unloaded += OnUnloaded;
            }

            void OnLoaded(object sender, RoutedEventArgs e) {
                Debug.WriteLine("OnLoaded");
            }

            void OnUnloaded(object sender, RoutedEventArgs e) {
                Debug.WriteLine("OnUnloaded");
            }
        }
    }
}
```

### Expected behavior

The test passes.

### Actual behavior

The test fails.

### Regression?

This works in net8.0 and net462.

### Known Workarounds

_No response_

### Configuration

.NET: 9.0.0-preview.6.24327.6
OS: Windows 10, Version 22H2 (OS Build 19045.4651)
Architecture: x64

### Other information

_No response_

h3xds1nz:
Fixes #9467 Reference #9463 

## Description

More of a complex fix, modifies the call chain to make use of the latest .NET features.
- ~~I've used `Span<object>` instead of `ReadOnlySpan<object>` as the `parameters` array has always been mutated by the trace class, hence I didn't want to currently change the behaviour.~~
- Builds on the original intention of #6700 
- The fixes should be compatible with my other PR too, #9361 and #9352.
- `ArrayList` replaced with pre-allocated array and used interpolation for building the trace data.
- Modifies the `AvTraceMessages.tt` for completion.
- The overall cost of `AvTrace` / `AvTraceMessages` will be greatly reduced.

## Customer Impact

Improved performance when tracing is active, decreased array allocations per each executed trace, smaller assemblies due to the function removal.

## Regression

Fixes the ref held by `_traceArguments` from #6700 

## Testing 

Local build, app run with tracing, comparing outputs of release/PR.

## Risk

Low, the actual change is small and was reviewed/tested.

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/wpf/pull/9468)

### DIFF
--- a/eng/WpfArcadeSdk/tools/CodeGen/AvTrace/AvTraceMessages.tt
+++ b/eng/WpfArcadeSdk/tools/CodeGen/AvTrace/AvTraceMessages.tt
@@ -70,51 +70,27 @@ namespace MS.Internal
 <#		} #>
 
         /// <summary> Send a single trace output </summary>
-        public static void Trace( TraceEventType type, AvTraceDetails traceDetails, params object[] parameters )
+        public static void Trace(TraceEventType type, AvTraceDetails traceDetails, params Span<object> parameters)
         {
-            _avTrace.Trace( type, traceDetails.Id, traceDetails.Message, traceDetails.Labels, parameters );
+            _avTrace.Trace(type, traceDetails.Id, traceDetails.Message, traceDetails.Labels, parameters);
         }
 
         /// <summary> These help delay allocation of object array </summary>
-        public static void Trace( TraceEventType type, AvTraceDetails traceDetails )
+        public static void Trace(TraceEventType type, AvTraceDetails traceDetails)
         {
-            _avTrace.Trace( type, traceDetails.Id, traceDetails.Message, traceDetails.Labels, Array.Empty<object>() );
-        }
-        public static void Trace( TraceEventType type, AvTraceDetails traceDetails, object p1 )
-        {
-            _avTrace.Trace( type, traceDetails.Id, traceDetails.Message, traceDetails.Labels, new object[] { p1 } );
-        }
-        public static void Trace( TraceEventType type, AvTraceDetails traceDetails, object p1, object p2 )
-        {
-            _avTrace.Trace( type, traceDetails.Id, traceDetails.Message, traceDetails.Labels, new object[] { p1, p2 } );
-        }
-        public static void Trace( TraceEventType type, AvTraceDetails traceDetails, object p1, object p2, object p3 )
-        {
-            _avTrace.Trace( type, traceDetails.Id, traceDetails.Message, traceDetails.Labels, new object[] { p1, p2, p3 } );
+            _avTrace.Trace(type, traceDetails.Id, traceDetails.Message, traceDetails.Labels, Array.Empty<object>());
         }
 
         /// <summary> Send a singleton "activity" trace (really, this sends the same trace as both a Start and a Stop) </summary>
-        public static void TraceActivityItem( AvTraceDetails traceDetails, params Object[] parameters )
+        public static void TraceActivityItem(AvTraceDetails traceDetails, params Span<object> parameters)
         {
-            _avTrace.TraceStartStop( traceDetails.Id, traceDetails.Message, traceDetails.Labels, parameters );
+            _avTrace.TraceStartStop(traceDetails.Id, traceDetails.Message, traceDetails.Labels, parameters);
         }
 
         /// <summary> These help delay allocation of object array </summary>
-        public static void TraceActivityItem( AvTraceDetails traceDetails )
+        public static void TraceActivityItem(AvTraceDetails traceDetails)
         {
-            _avTrace.TraceStartStop( traceDetails.Id, traceDetails.Message, traceDetails.Labels, Array.Empty<object>() );
-        }
-        public static void TraceActivityItem( AvTraceDetails traceDetails, object p1 )
-        {
-            _avTrace.TraceStartStop( traceDetails.Id, traceDetails.Message, traceDetails.Labels, new object[] { p1 } );
-        }
-        public static void TraceActivityItem( AvTraceDetails traceDetails, object p1, object p2 )
-        {
-            _avTrace.TraceStartStop( traceDetails.Id, traceDetails.Message, traceDetails.Labels, new object[] { p1, p2 } );
-        }
-        public static void TraceActivityItem( AvTraceDetails traceDetails, object p1, object p2, object p3 )
-        {
-            _avTrace.TraceStartStop( traceDetails.Id, traceDetails.Message, traceDetails.Labels, new object[] { p1, p2, p3 } );
+            _avTrace.TraceStartStop(traceDetails.Id, traceDetails.Message, traceDetails.Labels, Array.Empty<object>());
         }
 
         public static bool IsEnabled
--- a/eng/WpfArcadeSdk/tools/CodeGen/AvTrace/AvTraceMessages.tt
+++ b/eng/WpfArcadeSdk/tools/CodeGen/AvTrace/AvTraceMessages.tt
@@ -70,13 +70,13 @@ namespace MS.Internal
 <#		} #>
 
         /// <summary> Send a single trace output </summary>
-        public static void Trace(TraceEventType type, AvTraceDetails traceDetails, params Span<object> parameters)
+        public static void Trace(TraceEventType type, AvTraceDetails traceDetails, params ReadOnlySpan<object> parameters)
         {
             _avTrace.Trace(type, traceDetails.Id, traceDetails.Message, traceDetails.Labels, parameters);
         }
 
         /// <summary> Send a singleton "activity" trace (really, this sends the same trace as both a Start and a Stop) </summary>
-        public static void TraceActivityItem(AvTraceDetails traceDetails, params Span<object> parameters)
+        public static void TraceActivityItem(AvTraceDetails traceDetails, params ReadOnlySpan<object> parameters)
         {
             _avTrace.TraceStartStop(traceDetails.Id, traceDetails.Message, traceDetails.Labels, parameters);
         }
--- a/eng/WpfArcadeSdk/tools/CodeGen/AvTrace/AvTraceMessages.tt
+++ b/eng/WpfArcadeSdk/tools/CodeGen/AvTrace/AvTraceMessages.tt
@@ -81,7 +81,7 @@ namespace MS.Internal
             _avTrace.TraceStartStop(traceDetails.Id, traceDetails.Message, traceDetails.Labels, parameters);
         }
 
-        static public bool IsEnabled
+        public static bool IsEnabled
         {
             get { return _avTrace != null && _avTrace.IsEnabled; }
         }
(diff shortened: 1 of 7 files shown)


src/Microsoft.DotNet.Wpf/src/WindowsBase/MS/Internal/AvTrace.cs
```diff
@@ -258,73 +259,66 @@ internal static bool IsDebuggerAttached()
...
             {
                 int i = 1, j = 0;
-                for( ; i < labels.Length && j < parameters.Length; i++, j++ )
+                for (; i < labels.Length && j < parameters.Length; i++, j++)
                 {
                     // Append to the format string a "; {0} = '{1}'", where the index increments (e.g. the second iteration will
                     // produce {2} & {3}).
 
-                    traceBuilder.Append("; {" + (formatIndex++).ToString() + "}='{" + (formatIndex++).ToString() + "}'" );
+                    traceBuilder.Append($"; {{{++formatIndex}}}='{{{++formatIndex}}}'");
```
bgrainger, line 284: Changing `formatIndex++` to `++formatIndex` will change the output string from `; {0}='{1}'` to `; {1}='{2}`. This seems like an unintentionally-introduced bug.

h3xds1nz: Yeah, thanks, don't know how I've managed to get that there. Fixed.

lindexi: Good catch

src/Microsoft.DotNet.Wpf/src/WindowsBase/MS/Internal/AvTrace.cs
```diff
@@ -258,73 +259,66 @@ internal static bool IsDebuggerAttached()
...
-                    traceBuilder.Append("; {" + (formatIndex++).ToString() + "}='{" + (formatIndex++).ToString() + "}'" );
+                    traceBuilder.Append($"; {{{formatIndex++}}}='{{{formatIndex++}}}'");
 
                     // If this parameter is null, convert to "<null>"; otherwise, when a string.format is ultimately called
                     // it produces bad results.
 
-                    if( parameters[j] == null )
+                    if(parameters[j] == null)
                     {
                         parameters[j] = "<null>";
```
ThomasGoulet73, line 291: Could you check if any caller depend on the modification of `parameters` ? Because this place seems like the only place were we modify `parameters` so if no caller depend on it we could put `parameters[j]` in a variable instead (Which would also be cleaner IMO) and mutate it which would allow changing all the `Span<object>` to `ReadOnlySpan<object>` which would be more efficient and cleaner.

src/Microsoft.DotNet.Wpf/src/WindowsBase/MS/Internal/AvTrace.cs
```diff
@@ -258,73 +259,66 @@ internal static bool IsDebuggerAttached()
...
 
                     // Otherwise, if this is an interesting object, add the hash code and type to
                     // the format string explicitely.
 
-                    else if( !SuppressGeneratedParameters
-                             && parameters[j].GetType() != typeof(string)
-                             && !(parameters[j] is ValueType)
-                             && !(parameters[j] is Type)
-                             && !(parameters[j] is DependencyProperty) )
+                    else if(SuppressGeneratedParameters == false
```
ThomasGoulet73, line 297: Could you revert this change ? It doesn't seem necessary and it looks worse IMO (It might also be against the coding style).

src/Microsoft.DotNet.Wpf/src/WindowsBase/MS/Internal/AvTrace.cs
```diff
@@ -258,73 +259,66 @@ internal static bool IsDebuggerAttached()
...
                     // produce {2} & {3}).
 
-                    traceBuilder.Append("; {" + (formatIndex++).ToString() + "}='{" + (formatIndex++).ToString() + "}'" );
+                    traceBuilder.Append($"; {{{formatIndex++}}}='{{{formatIndex++}}}'");
 
                     // If this parameter is null, convert to "<null>"; otherwise, when a string.format is ultimately called
                     // it produces bad results.
 
-                    if( parameters[j] == null )
+                    if(parameters[j] == null)
```
ThomasGoulet73, line 289: Since you're already changing the whitespaces of this line, could you also add a space between the `if` and `(`.

There's a couple other place in your changes where you could do the same thing.

src/Microsoft.DotNet.Wpf/src/WindowsBase/MS/Internal/AvTrace.cs
h3xds1nz: Yeah, I originally didn't want to change this but all your points are true; if you're of the opinion we should do it, I'll investigate it and change accordingly.

src/Microsoft.DotNet.Wpf/src/WindowsBase/MS/Internal/AvTrace.cs
h3xds1nz: Totally, I'll go through the changes again.

src/Microsoft.DotNet.Wpf/src/WindowsBase/MS/Internal/AvTrace.cs
h3xds1nz: I couldn't find anything against the coding style here, plus given a base search it is kinda used throughout the repo.

src/Microsoft.DotNet.Wpf/src/WindowsBase/MS/Internal/AvTrace.cs
h3xds1nz: I couldn't find any call where the array was explicitly defined, so it shouldn't be an issue.

Also the underlying `TraceExtraMessages` should only get non-mutated slice of the array, so it shall be without issues.

src/Microsoft.DotNet.Wpf/src/WindowsBase/MS/Internal/AvTrace.cs
```diff
@@ -336,32 +334,28 @@ public string Trace( TraceEventType type, int eventId, string message, string[]
                 type,
                 eventId,
                 traceMessage,
-                arrayList.ToArray() );
+                combinedList.ToArray()); //Cannot avoid the alloc here, no ROS<object> overload
```
GerardSmit, line 337: You could change the `List<object> combinedList` to an `object[]` with length `parameters.Length * 2`; then you won't need to create a list, which creates an underlying array (`_items`) and creates an array in `.ToArray()`.

bgrainger: It would need to be `Math.Min(labels.Length - 1, parameters.Length) * 2` (unless we can assume those are always equal to the same value).

h3xds1nz: As @bgrainger mentions, that is why I didn't do this change and put comment instead. 

Theoretically I could use `Max(...) * 2` instead and just allocate one bigger array, it would work out for the better. `TraceEvent` uses `string.Format` under the hood, which doesn't care how many params in the array, as long as it is not less.

h3xds1nz: Done, I've also delayed the alloc, according to my tests of all possible trace variations, all the trace outputs work as expected, and are same.

https://github.com/dotnet/runtime/blob/99c9f5b95a68a5e864436f9ec0726eed759350bd/src/libraries/System.Diagnostics.TraceSource/src/System/Diagnostics/TraceListener.cs#L374-L377 ref for null arrays.

Alexgoon:
Hi there! Could you please update us on the status of this pull request? It's important for DevExpress components, and we're eager to know if you plan to merge it.

h3xds1nz:
Hey @Alexgoon, I'm gonna CC a few guys from the wpf team just to be sure it doesn't get lost.

cc @pchaurasia14 @dipeshmsft

Alexgoon:
Hi again,
I see the merging process has started - thanks for that! However, it looks like it’s stalled. Could you let me know if you plan to include this PR in the upcoming .NET 9 release?

pchaurasia14:
@Alexgoon  - We will be including this PR for upcoming test cycles, and it will go in for .NET 10.

h3xds1nz:
@siagupta0202 Any news on this one?

siagupta0202:
> @siagupta0202 Any news on this one?

Hey @h3xds1nz, I was involved with some other tasks so couldn't take a look at this PR. I will review this soon.

h3xds1nz:
@siagupta0202 Rebased so we can get this done finally.

siagupta0202:
LGTM, @h3xds1nz can you resolve the merge conflicts?

h3xds1nz:
@siagupta0202 Done, sorry for the delay :)

eng/WpfArcadeSdk/tools/CodeGen/AvTrace/AvTraceMessages.tt
```diff
@@ -70,54 +70,18 @@ namespace MS.Internal
...
-        }
-        public static void TraceActivityItem( AvTraceDetails traceDetails, object p1, object p2, object p3 )
+        public static void TraceActivityItem(AvTraceDetails traceDetails, params ReadOnlySpan<object> parameters)
         {
-            _avTrace.TraceStartStop( traceDetails.Id, traceDetails.Message, traceDetails.Labels, new object[] { p1, p2, p3 } );
+            _avTrace.TraceStartStop(traceDetails.Id, traceDetails.Message, traceDetails.Labels, parameters);
         }
 
-        public static bool IsEnabled
+        static public bool IsEnabled
```
ThomasGoulet73, line 84: Looks like a conflict resolution mistake:
```suggestion
        public static bool IsEnabled
```

h3xds1nz: Thanks :)

siagupta0202 (approved):

siagupta0202:
@h3xds1nz Thank you so much for your contribution!

PR #9470, opened 2024-07-25
Fix WriteableBitmap's Write/CopyPixels (Clone) to support up to 4GB backbuffers

### LINKED ISSUE #9438: System.OverflowException in WriteableBitmap.WritePixels
### Description

`WriteableBitmap` can not exceed 4 GB I think (`WriteableBitmap` constructor fails) but even for sizes below 4 GB (but above 2GB) there are overflow exceptions in `WriteableBitmap.WritePixels`, even when only writing a tiny amount of data.

### Reproduction Steps

```
                int size = 30000;
                int tileSize = 100;
                int channels = 4;
                var pix = new byte[tileSize * tileSize * channels];
                var bitmap = new WriteableBitmap(size, size, 96, 96, PixelFormats.Pbgra32, null);
                bitmap.WritePixels(new Int32Rect(0, 0, tileSize, tileSize),
                      pix, tileSize * channels, size-tileSize, size-tileSize);
```

### Expected behavior

No exception. (Or if >2GB is considered unsupported, the constructor should already fail?)

### Actual behavior

Exception:
```
System.OverflowException
  HResult=0x80131516
  Message=Arithmetic operation resulted in an overflow.
  Source=PresentationCore
  StackTrace:
   at System.Windows.Media.Imaging.WriteableBitmap.WritePixelsImpl(Int32Rect sourceRect, IntPtr sourceBuffer, Int32 sourceBufferSize, Int32 sourceBufferStride, Int32 destinationX, Int32 destinationY, Boolean backwardsCompat) in System.Windows.Media.Imaging\WriteableBitmap.cs:line 583
   at System.Windows.Media.Imaging.WriteableBitmap.WritePixels(Int32Rect sourceRect, Array sourceBuffer, Int32 sourceBufferStride, Int32 destinationX, Int32 destinationY) in System.Windows.Media.Imaging\WriteableBitmap.cs:line 271
   at Program.Main() in Program.cs:line 15
```

### Regression?

I don't think so.

### Known Workarounds

None

### Impact

Unknown

### Configuration

.NET 8, Windows 11, x64. I don't think this is specific.

### Other information

None. 

Maybe uint should be used somewhere instead of int in the internal offset calculations?

h3xds1nz:
Fixes #9438 

## Description

Fixes working with `WriteableBitmap` to (`WritePixels`/`Clone`) operations on large buffers (over `Int32.MaxValue`) as those are supported by `WpfGfx`/`Milcore` as the underlying layer supports creation and working with such buffers.

Previously you were just able to create such WriteableBitmap but unable to `WritePixels` to the bottom-right corner or `Clone` it.

This departs from `BitmapSource` behaviour which uses `WIC` and there's an ongoing disparity between `IWICBitmapSource_CopyPixels_Proxy` and `IWICImagingFactory_CreateBitmapFromMemory_Proxy` where the former works with up to 4GB buffers but creating such bitmap fails with `0x80070057` even though both bufferSize params are documented as `uint`.

Base code that would fail with `OverflowException` before on `WritePixels` (creation would succeed).

```csharp
//Base vars
int width = 30000;
int height = 30000;
int tileSize = 1000;
int channels = 4;
int stride = tileSize * channels;
byte[] smallRect = new byte[tileSize * tileSize * channels];
WriteableBitmap bitmap = new WriteableBitmap(width, height, 96, 96, PixelFormats.Pbgra32, null);

//Set teal color
for (int row = 0; row < tileSize; row++)
{
    for (int col = 0; col < tileSize; col++)
    {
        // BGRA
        smallRect[row * stride + col * 4 + 0] = 255;
        smallRect[row * stride + col * 4 + 1] = 230;
        smallRect[row * stride + col * 4 + 2] = 0;
        smallRect[row * stride + col * 4 + 3] = 255;
    }
}

//Top-Left
bitmap.WritePixels(new Int32Rect(0, 0, tileSize, tileSize),
                    smallRect, tileSize * channels, 0, 0);

//Top-Right
bitmap.WritePixels(new Int32Rect(0, 0, tileSize, tileSize),
                    smallRect, tileSize * channels, width - tileSize, 0);
                    
//Middle Rect
bitmap.WritePixels(new Int32Rect(0, 0, tileSize, tileSize),
                    smallRect, tileSize * channels, (width - tileSize) / 2, (height - tileSize) / 2);

//Bottom-Left
bitmap.WritePixels(new Int32Rect(0, 0, tileSize, tileSize),
                    smallRect, tileSize * channels, 0, height - tileSize);

//Bottom-Right
bitmap.WritePixels(new Int32Rect(0, 0, tileSize, tileSize),
                    smallRect, tileSize * channels, width - tileSize, height - tileSize);

//Save (Test CriticalPixelsCopy)
BitmapSource bitmapCopy = bitmap.Clone();
```

## Customer Impact

Ability to write pixels on bitmaps over 23xxx x 23xxxx bitmaps with 4 channels, etc. (using backbuffers over 2GB).

## Regression

No.

## Testing

Local build, working with `WriteableBitmap`.

## Risk

Low.

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/wpf/pull/9470)

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Media/Imaging/WriteableBitmap.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Media/Imaging/WriteableBitmap.cs
@@ -752,7 +752,7 @@ namespace System.Windows.Media.Imaging
                 try
                 {
                     Int32Rect rcFull = new Int32Rect(0, 0, _pixelWidth, _pixelHeight);
-                    int bufferSize = checked(_backBufferStride * source.PixelHeight);
+                    uint bufferSize = checked((uint)_backBufferStride * (uint)source.PixelHeight);
                     source.CriticalCopyPixels(rcFull, _backBuffer, bufferSize, _backBufferStride);
                     AddDirtyRect(rcFull);
                 }
@@ -886,7 +886,7 @@ namespace System.Windows.Media.Imaging
                 //
                 unsafe
                 {
-                    uint destOffset = (uint)(destinationY * _backBufferStride) + destXbyteOffset;
+                    uint destOffset = ((uint)destinationY * (uint)_backBufferStride) + destXbyteOffset;
                     byte* pDest = (byte*)_backBuffer.ToPointer();
                     pDest += destOffset;
                     uint outputBufferSize = _backBufferSize - destOffset;
--- a/src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Media/Imaging/WriteableBitmap.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Media/Imaging/WriteableBitmap.cs
@@ -353,7 +353,7 @@ namespace System.Windows.Media.Imaging
 
             WritePixelsImpl(sourceRect,
                             sourceBuffer,
-                            sourceBufferSize,
+                            (uint)sourceBufferSize,
                             sourceBufferStride, 
                             destinationX,
                             destinationY,
@@ -381,14 +381,11 @@ namespace System.Windows.Media.Imaging
         {
             WritePreamble();
 
-            int elementSize;
-            int sourceBufferSize;
-            Type elementType;
             ValidateArrayAndGetInfo(sourceBuffer,
                                     backwardsCompat: false,
-                                    out elementSize,
-                                    out sourceBufferSize,
-                                    out elementType);
+                                    out _,
+                                    out uint sourceBufferSize,
+                                    out Type elementType);
 
             // We accept arrays of arbitrary value types - but not reference types.
             if (elementType == null || !elementType.IsValueType)
@@ -426,7 +423,7 @@ namespace System.Windows.Media.Imaging
         {
             WritePreamble();
 
-            ArgumentOutOfRangeException.ThrowIfNegativeOrZero(bufferSize);
+            ArgumentOutOfRangeException.ThrowIfZero(bufferSize);
             ArgumentOutOfRangeException.ThrowIfNegativeOrZero(stride);
 
             if (sourceRect.IsEmpty || sourceRect.Width <= 0 || sourceRect.Height <= 0)
@@ -449,7 +446,7 @@ namespace System.Windows.Media.Imaging
 
             WritePixelsImpl(sourceRect, 
                             buffer,
-                            bufferSize,
+                            (uint)bufferSize,
                             stride,
                             destinationX,
                             destinationY,
@@ -477,14 +474,11 @@ namespace System.Windows.Media.Imaging
                 return;
             }
 
-            int elementSize;
-            int sourceBufferSize;
-            Type elementType;
             ValidateArrayAndGetInfo(pixels,
                                     backwardsCompat: true,
-                                    out elementSize,
-                                    out sourceBufferSize, 
-                                    out elementType);
+                                    out int elementSize,
+                                    out uint sourceBufferSize, 
+                                    out Type elementType);
 
             ArgumentOutOfRangeException.ThrowIfNegativeOrZero(stride);
             ArgumentOutOfRangeException.ThrowIfNegative(offset);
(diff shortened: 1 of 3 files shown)


miloush:
LGTM

h3xds1nz:
@miloush I had to add one more fix, for when you use f.e. `uint[]` array which is over 2GB in size (e.g. you wanna write from a source buffer of 3GB, etc.) Looks like I've missed that one during my tests and it triggered me now.

h3xds1nz:
I have fixed the merge conflicts and rebased.

h3xds1nz:
Resolved merge conflicts 07

farosch:
Any plans on implementing this PR?

dipeshmsft (dismissed):

src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Media/Imaging/WriteableBitmap.cs
```diff
@@ -451,7 +448,7 @@ int stride
         {
             WritePreamble();
 
-            ArgumentOutOfRangeException.ThrowIfNegativeOrZero(bufferSize);
+            ArgumentOutOfRangeException.ThrowIfZero(bufferSize);
```
dipeshmsft, line 451: I think we should revert this check to what it was as the bufferSize is still an int. Just thinking that if we add more code around this and miss this detail.

dipeshmsft:
@farosch @h3xds1nz I have reviewed this PR, I will ask the team for queuing this for testing. Hopefully we will be able to close this within this month.

src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Media/Imaging/WriteableBitmap.cs
h3xds1nz: @dipeshmsft Well that makes it impossible to write a 3GB buffer at a given pointer; it would create a different behaviour between the "safe" and "unsafe" public API. And between copying and writing.

The issue is that the unsafe API shape was released with `int` in the first place but I guess it was just an attempt to keep it CLS-compliant or nobody in 2005 when designing it wanted to create 30k x 30k bitmaps.

The only true fix is to change the API shape. I could add a comment in the code as to why we check only zero and negative values are fine.

h3xds1nz:
@dipeshmsft Yeah, I'll put something up over the weekend.

src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Media/Imaging/WriteableBitmap.cs
dipeshmsft: > The only true fix is to change the API shape.
I got this idea while reviewing the PR. Anyhow, adding a comment would suffice for now.

h3xds1nz:
Rebased, tests will go into a different PR.

h3xds1nz:
@dipeshmsft So it seems to work on CI but I guess the VM is sweating given the log from running unit tests:

```
##[warning]Free memory is lower than 5%; Currently used: 98.05%
##[warning]Free memory is lower than 5%; Currently used: 98.05%
##[warning]Free memory is lower than 5%; Currently used: 98.05%
##[warning]Free memory is lower than 5%; Currently used: 98.05%
```

It's no wonder, this will easily get the testhost to consume 70GB+ on my machine (since the runtime can allow to expand, I guess that CI VM is running with 16 GB or less).
It also takes about 50s on CI to run, which is not ideal, I'm happy with 8s on my machine.

I assume we don't have the ability to run different types of tests on-demand with the CI so we could exclude this from regular runs.

dipeshmsft:
> I assume we don't have the ability to run different types of tests on-demand with the CI so we could exclude this from regular runs.

Not yet, I think it would be best to disable this for now.

h3xds1nz:
@dipeshmsft The tests for large bitmaps are now skipped.

dipeshmsft (approved):

dipeshmsft:
Thanks @h3xds1nz for this PR.

h3xds1nz:
@dipeshmsft Thanks :)

Issue #9486, opened 2024-07-30
[dotnet-sdk-9.0.100-preview.7.24375.12] Fail to load the interface controls for youtubedlwpf.


Junjun-zhao:
### Description

When run the 3rd party application youtubedlwpf with the latest .NET 9 build, it failed to load interface controls without any errors.  After we tried to debug following the steps , it doesn't show the error either. It seems related to the fix for resource.
 Could you please help look into it? Please refer to repro information from [Internal bug](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/2174443)  

### Reproduction Steps

**Debug Repro steps:**

Please get the repro information from [Internal bug](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/2174443)
1.Open "C:\ReproAppSource\youtube-dl-wpf\youtube-dl-wpf.sln" with Visual Studio.
2. Build project YoutubeDl.Wpf.
3. Update ...\bin\Debug\net7.0-windows10.0.22621.0\youtube-dl-wpf.runtimeconfig.json to run against with dotnet-sdk-9.0.100-preview.7.24375.12:
```
"frameworks": [
      {
        "name": "Microsoft.NETCore.App",
        "version": "9.0.0-preview.7.24374.12"
      },
      {
        "name": "Microsoft.WindowsDesktop.App",
        "version": "9.0.0-preview.7.24375.4"
      }
```
5. Run the app.
6. Click "Setting" icon.
7. Click "Dark" in Color theme.
8. Close the app and restart it.

Expected Result:

### Expected behavior

App launch successfully.
![image](https://github.com/user-attachments/assets/a179d080-24fc-4d8a-9d73-c35dee95302d)

### Actual behavior

App interface loading failed without any errors.

![image](https://github.com/user-attachments/assets/e7d87637-b39c-4be8-a83a-592f338f8765)

### Regression?

Yes
Verify Scenarios:
1). Windows 10 21H2 AMD64 + dotnet-sdk-8.0.300: Pass
2). Windows 10 21H2 AMD64 + dotnet-sdk-9.0.100-preview.7.24371.4: Pass
3). Windows 10 21H2 AMD64 + dotnet-sdk-9.0.100-preview.7.24375.12: Fail

### Configuration

**Application Name:** youtubedlwpf
**OS**： Windows 10 21H2
**CPU:** X64
**.NET Build Number:** dotnet-sdk-9.0.100-preview.7.24375.12
[...]

PR #9512, opened 2024-08-02
Fix : Invalidate application resources once while syncing deffered thememode and resources

dipeshmsft:
Fixes : #9486 

## Description

To allow syncing between theme mode and resources, we have deffered app resource invalidation till the first window initializes. During which we see the current state of the app resources and sync both the properties.

However, currently this method checks if thememode needs to be changed or not, if yes then we apply the theme mode and that results in another invalidation call which takes care of all the previous invalidation calls.

However, if there is nothing to be set in thememode, we don't respond to any of the previous invalidation request and that results in wrong colors in the UI.

I have added that case in the SyncDeferredThemeModeAndResources method

## Customer Impact
Applications that update application resources while initialization will get the correct UI.

## Regression
Yes

## Testing
Local testing

###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/wpf/pull/9512)

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/ThemeManager.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/ThemeManager.cs
@@ -318,7 +318,7 @@ internal static class ThemeManager
 
     #region Internal Properties
 
-    internal static bool DeferSyncingThemeModeAndResources { get; set; } = true;
+    internal static bool DeferSyncingThemeModeAndResources { get; set; } = Standard.Utility.IsOSWindows10OrNewer ? true : false;
 
     internal static bool IsFluentThemeEnabled
     {
--- a/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/ThemeManager.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/ThemeManager.cs
@@ -176,6 +176,13 @@ internal static class ThemeManager
         {
             Application.Current.ThemeMode = themeMode;
         }
+        else
+        {
+            // In case ThemeMode was not set and we did not include the 
+            // Fluent theme resources, we need to invalidate the resources,
+            // that were earlier deferred.
+            Application.Current.InvalidateResourceReferences(ResourcesChangeInfo.CatastrophicDictionaryChangeInfo);
+        }
     }
 
     internal static void ApplyStyleOnWindow(Window window)


src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/ThemeManager.cs
```diff
@@ -318,7 +318,7 @@ private static void ApplyStyleOnWindow(Window window, bool useLightColors)
 
     #region Internal Properties
 
-    internal static bool DeferSyncingThemeModeAndResources { get; set; } = true;
+    internal static bool DeferSyncingThemeModeAndResources { get; set; } = Standard.Utility.IsOSWindows10OrNewer ? true : false;
```
h3xds1nz, line 321: Could just be `internal static bool DeferSyncingThemeModeAndResources { get; set; } = Standard.Utility.IsOSWindows10OrNewer`.

But do we need this at all? .NET 8 already doesn't run on less than Win8: https://github.com/dotnet/core/blob/main/release-notes/8.0/supported-os.md

dipeshmsft: Since the call that makes this flag `False` occurs from inside an `if` block with this OS version check, I didn't want to leave  this `True` for other cases.

h3xds1nz: I understand but is there any reason we should be checking if the Windows build is less than `10240` altogether? The runtime shouldn't even install in that case and is certainly unsupported, the lowest supported is `14393` (1607).

And since less than Win10 is not supported by the runtime, then I see no reason for the checks altogether.
I understand this was previously Win11 check and it was just modified but do we need it? Not in my opinion.

h3xds1nz: Yes, exactly, two different things.

If you think we should keep those checks around for unsupported system when it's a new function, how crazy should we get? Do we implement every new feature with checks for unsupported systems and then try to figure out some sort of fallback if it ain't supported? I'm obviously exaggerating but you get my point.

Even JIT has now removed tons of Win95 compat code for .NET 9, but I can still select compat mode for Win95 on latest Win10, guess I might have some issues when I'm lying in `PEB` to `RtlGetVersion`.

miloush: I don't know the answer. I have never heard from anyone in .NET about how supported running under compatibility mode is or what the expectations are. Just pointing out that user can relatively easily run the application in a way that the check makes a difference.

dipeshmsft: I agree with @miloush on this one, we have seen customers use applications in ways that I didn't expect before. Moreover, if the time to purge has come, then we will do it as a whole.

dipeshmsft:
#9527 presents an alternative less behavior changing approach to fix the above issue. @miloush @h3xds1nz what are  your views on the other PR.

src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/ThemeManager.cs
h3xds1nz: Fine by me, was just really looking for opinions on this.

dipeshmsft: @h3xds1nz , I agree with you that there are definitely checks in the codebase which won't be triggered, but at the same time, I also think that we should deal with them at once.

h3xds1nz: Yeah, agreed, this can be looked at later with a complex PR. Thank you for the opinion.

PR #9517, opened 2024-08-02
Added System.Formatters.Nrbf to Version.Details.xml

dipeshmsft:
###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/wpf/pull/9517)

### DIFF
--- a/eng/Version.Details.xml
+++ b/eng/Version.Details.xml
@@ -95,9 +95,9 @@
       <Uri>https://github.com/dotnet/runtime</Uri>
       <Sha>7f2b222f92d323b7ca522a06c7116f1ec25d12ca</Sha>
     </Dependency>
-    <Dependency Name="System.Formats.Nrbf" Version="9.0.0-preview.7.24401.8" CoherentParentDependency="Microsoft.Private.Winforms">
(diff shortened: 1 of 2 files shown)


eng/Version.Details.xml
```diff
@@ -95,6 +95,10 @@
       <Uri>https://github.com/dotnet/runtime</Uri>
       <Sha>7f2b222f92d323b7ca522a06c7116f1ec25d12ca</Sha>
     </Dependency>
+    <Dependency Name="System.Formats.Nrbf" Version="9.0.0-preview.7.24401.8" CoherentParentDependency="Microsoft.Private.Winforms">
```
anjali-wpf, line 98: Are we going to add reference for this in versions.props as well?

pchaurasia14: Version.props already has the reference to this package. We missed adding it here.

pchaurasia14, line 98: @dipeshmsft - Check the version number. Should be rc1 version.

ericstj: Correct, this needs to match the other entries here.  It should also match what's in Versions.props.

pchaurasia14 (approved):

harshit7962 (approved):

PR #9521, opened 2024-08-03
Replace SecurityHelper.AreStringTypesEqual with string.Equals for code quality

h3xds1nz:
## Description

One of the few leftover remnants from CAS times is the `SecurityHelper.AreStringTypesEqual` that just calls case-insensitive `string.Equals`. There is no reason for this method as it has even lost the original comment throughout the removals and just makes newcomers wonder why is it there.

## Customer Impact

Few bytes less in size of all libraries.

## Regression

No.

## Testing

Local build.

## Risk

None.

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/wpf/pull/9521)

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/Shared/MS/Utility/BindUriHelper.cs
+++ b/src/Microsoft.DotNet.Wpf/src/Shared/MS/Utility/BindUriHelper.cs
@@ -16,22 +16,21 @@ using System.Windows.Navigation; // BaseUriHelper
 using MS.Win32;
 #endif
 
-using System.Security;
 // The functionality in this class is shared across framework and core. The functionality in core
 // is a subset of the functionality in framework, so rather than create a dependency from core to
 // framework we have choses to duplicate this chunk of  code.
 #if PRESENTATION_CORE
 namespace MS.Internal.PresentationCore
 #elif PRESENTATIONFRAMEWORK
-using MS.Internal.PresentationFramework; // SecurityHelper
+using MS.Internal.PresentationFramework;
 
 namespace MS.Internal.Utility
 #elif PBTCOMPILER
-using MS.Internal.PresentationBuildTasks // SecurityHelper
+using MS.Internal.PresentationBuildTasks
 
 namespace MS.Internal.Utility
 #elif REACHFRAMEWORK
-using MS.Internal.ReachFramework; // SecurityHelper
+using MS.Internal.ReachFramework;
 
 namespace MS.Internal.Utility
 #else
@@ -88,7 +87,7 @@ namespace MS.Internal.Utility
         static internal bool DoSchemeAndHostMatch(Uri first, Uri second)
         {
             // Check that both the scheme and the host match. 
-           return (SecurityHelper.AreStringTypesEqual(first.Scheme, second.Scheme) && first.Host.Equals(second.Host) == true);
+            return string.Equals(first.Scheme, second.Scheme, StringComparison.OrdinalIgnoreCase) && string.Equals(first.Host, second.Host);
         }
 
         static internal Uri GetResolvedUri(Uri baseUri, Uri orgUri)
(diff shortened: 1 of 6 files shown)


lindexi:
I think the `SecurityHelper.AreStringTypesEqual` will be inline

h3xds1nz:
Yes, it would be inlined, of course. The reason for this is purely code quality. There should be no `SecurityHelper` whatsoever because there is no CAS anymore. This was planned to be removed by the team but after handover it has never happened., you may find the discussion in older issues/PRs regarding CAS.

It only complicates the code readibility with another level of indirection. Oh, look, "AreStringTypesEqual"... but it is CI compare!

lindexi:
> there is no CAS anymore

Yeah, I fully agree.

src/Microsoft.DotNet.Wpf/src/Shared/MS/Utility/BindUriHelper.cs
```diff
@@ -88,7 +87,7 @@ static internal Uri BaseUri
         static internal bool DoSchemeAndHostMatch(Uri first, Uri second)
         {
             // Check that both the scheme and the host match. 
-           return (SecurityHelper.AreStringTypesEqual(first.Scheme, second.Scheme) && first.Host.Equals(second.Host) == true);
+            return string.Equals(first.Scheme, second.Scheme, StringComparison.OrdinalIgnoreCase) && string.Equals(first.Host, second.Host);
```
ThomasGoulet73, line 90: Could you revert the change of `String.Equals(String)` to `String.Equals(String, String)`. Before your change, if `first.Host` was null it would throw a NullReferenceException and with your change it would not throw. This is a change in behavior which IMO does not provide any value.

```suggestion
            return string.Equals(first.Scheme, second.Scheme, StringComparison.OrdinalIgnoreCase) && first.Host.Equals(second.Host);
```

h3xds1nz: @ThomasGoulet73 How would I get `Host` from Uri to be null?

ThomasGoulet73: I agree I don't think it is possible for `Uri.Host` to be null. I'm still not a fan of unrelated changes that **could** change the behavior, it makes it harder to review. I don't mind if you want to keep your PR as is.

h3xds1nz: @ThomasGoulet73 No worries, I'm always glad for your reviews.

I'll keep the change as I've made it with checking this exact fact on purpose, as it imho signifies the meaning of the compare.

dipeshmsft (approved):

src/Microsoft.DotNet.Wpf/src/Shared/MS/Utility/BindUriHelper.cs
dipeshmsft: @ThomasGoulet73  @h3xds1nz 
This code here : https://github.com/dotnet/runtime/blob/5535e31a712343a63f5d7d796cd874e563e5ac14/src/libraries/System.Private.Uri/src/System/Uri.cs#L343-L360 doesn't it indicate that `Host` can be `null` ?

h3xds1nz: @dipeshmsft `UriInfo` != `Uri`, while `UriInfo.Host` is nullable because it can be, `Uri.Host` is not.

One of the benefits of NRT annotations! Anyways in the exact function you've linked, it goes right into [CreateHostString](https://source.dot.net/System.Private.Uri/R/f4f0ffc0f7c2a51d.html) that does the necessary fixes.

And as for the return; https://source.dot.net/#System.Private.Uri/System/UriExt.cs,941 -> https://source.dot.net/#System.Private.Uri/System/UriExt.cs,966 (You first call `true`, then `false`, second time it hops into `CreateHostString`)

dipeshmsft: My bad, never mind it. Although there is a change in behaviour, it looks good to me.

h3xds1nz: Thanks, happy to see every bit of `SecurityHelper` gone.

PR #9527, opened 2024-08-06
[Fix] Don't ignore application resource invalidations

dipeshmsft:
Fixes : #9486 

## Description

This is another variation of the fix for above issue. 

Currently, when we get an Application Resource invalidation request, we check if there is a need to update the ThemeMode property due to the change in the Application.Resources . In the case, when ThemeMode property needs to be updated we returned true after sync, and when we get true, we did not do further invalidation of the resources for the original request.

Ignoring the original request results in incoherent UI, as the values changed are not getting properly reflected in the visual tree.

So, the core fix here is that I have removed the `return` statement and let all the invalidation's take place. Apart from this, I have also renamed the flags and methods to make it a bit easier to understand. Moreover, as an optimization, I have added a check to avoid syncing when the `info.IsResourceAddOperation == true` as this indicates that a single value is changed and not the merged dictionaries, where we are loading our fluent theme dictionaries. 

## Customer Impact
Applications that update application resources while initialization will get the correct UI.

## Regression
Yes

## Testing
Local testing

###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/wpf/pull/9527)

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/ThemeManager.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/ThemeManager.cs
@@ -20,7 +20,7 @@ internal static class ThemeManager
     {
         if(IsFluentThemeEnabled)
         {
-            IgnoreAppResourcesChange = true;
+            SkipAppThemeModeSyncing = true;
 
             try
             {
@@ -53,7 +53,7 @@ internal static class ThemeManager
             }
             finally
             {
-                IgnoreAppResourcesChange = false;
+                SkipAppThemeModeSyncing = false;
             }
 
         }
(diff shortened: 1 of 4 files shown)


h3xds1nz:
This seems cleaner to me to be honest than the previous PR.

src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/ThemeManager.cs
```diff
@@ -140,10 +137,11 @@ internal static bool SyncThemeModeAndResources()
...
     }
 
-    internal static void SyncDeferredThemeModeAndResources()
+    internal static void SyncThemeModeAndResources()
     {
-        if (Application.Current == null)
-            return;
+        // Since, this is called from window there is a possiblity that the application
+        // instance is null. Hence, we need to check for null.
+        if(Application.Current == null) return;
```
h3xds1nz, line 144: Can we revert the change back to the original?

```csharp
if (Application.Current == null)
    return;
```

harshit7962 (approved):
Verified that the issue is fixed and changes LGTM.

PR #9531, opened 2024-08-06
Remove CheckedPointer struct and replace its use with Span<T> instances

h3xds1nz:
## Description

Removes `CheckedPointer` which was used in quite some places back in early WPF implementations in .NET Framework but now the only thing using it is `TrueTypeFontDriver` and replaces it with `ReadOnlySpan<T>`.

- The `Probe` checks are the same as implemented by `ReadOnlySpan<T>` when slicing.
- Removes now useless `Marshal.Copy` with simple `.ToArray()` call from `ReadOnlySpan<T>` in `ComputeFontSubset`.
- I've used `Slice` and non-Try methods from `BinaryPrimitives` to be sure the exception behavior stays the same.
- Offers quite nice performance improvement over the original version.

### `SetFace` method
| Original |  94.60 ns |   1.523 ns |    1.564 ns | 0.0358 |       1,325 B |         600 B |
| PR__EDIT |  53.56 ns |   1.039 ns |    0.921 ns | 0.0358 |         512 B |         600 B |

## Customer Impact

Smaller code-size of PresentationCore, improved performance in this particular scenario.
For developers it introduces more familiar types in internal classes than custom implementations for a simple thing.

## Regression

No.

## Testing

Tested base font file parsing and class (`TrueTypeFontDriver`) functionality after `CheckedPointer` removal.

## Risk

Low.


h3xds1nz:
Resolved merge conflicts 10

PR #9532, opened 2024-08-06
Replace boxing Hashtable in ManagedWndProcTracker with Dictionary<K, V>

h3xds1nz:
## Description

Replaces boxing `Hashtable` with generic `Dictionary<HwndSubclass, IntPtr>` and improves overall performance.
- Since all operations are under a lock, there are no thread-safety concerns with replacing `Hashtable`.
- I've also added some more comments for clarity on the class.
- I've felt like the memory fence on `s_exiting` can help because the reads can come from multiple threads.
- Usings for `SecurityHelper` were removed as those are not needed since CAS removal.

## Customer Impact

Decreased allocations, improved performance.

## Regression

No.

## Testing

Local build/sample app run.

## Risk

Low, the swap is safe.

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/wpf/pull/9532)

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/Shared/MS/Win32/ManagedWndProcTracker.cs
+++ b/src/Microsoft.DotNet.Wpf/src/Shared/MS/Win32/ManagedWndProcTracker.cs
@@ -39,7 +39,7 @@ namespace MS.Win32
         {
             // if exiting the AppDomain, ignore this call.  This avoids changing
             // the list during the loop in OnAppDomainProcessExit
-            if (_exiting)
+            if (s_exiting)
                 return;
 
             lock (s_hwndList)
@@ -57,7 +57,7 @@ namespace MS.Win32
             // the DefaultWindowProc.
             //DbgUserBreakPoint();
 
-            _exiting = true;
+            s_exiting = true;
 
             lock (s_hwndList)
             {
@@ -104,7 +104,7 @@ namespace MS.Win32
             LogFinishHWND(hwnd, "Core HookUpDWP");
 #endif
 
-            IntPtr result = IntPtr.Zero ;
+            IntPtr result = IntPtr.Zero;
 
             // We've already cleaned up, return immediately.
             if (hwnd == IntPtr.Zero)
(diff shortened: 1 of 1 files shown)


src/Microsoft.DotNet.Wpf/src/Shared/MS/Win32/ManagedWndProcTracker.cs
```diff
@@ -105,16 +93,16 @@ private static void OnAppDomainProcessExit()
                         // of Hwnd subclasses.
 
                         UnsafeNativeMethods.SendMessage(hwnd, HwndSubclass.DetachMessage,
-                                                            IntPtr.Zero /* wildcard */,
-                                                            (IntPtr) 2 /* force and forward */);
+                                                        IntPtr.Zero /* wildcard */,
+                                                        2 /* force and forward */);
```
dipeshmsft, line 97: @h3xds1nz  Is int implicilty cast into IntPtr ?

h3xds1nz: @dipeshmsft Yes, otherwise it wouldn't compile at all if it wasn't.

See how it gets lowered: https://sharplab.io/#v2:EYLgxg9gTgpgtADwGwBYA+ABADAAgwRgDoAlAVwDsAXASwFsZCBhCWgB2oBsYoBlbgN2pgYAZwDcAWABQGAEz5pGAMx5ZORjgDeOaTj369yvChwAVUZVkAKAJRbdB/X3IATALKiRAQwDmMACIwtBBWAPQAVDh0rBxC1JRRVDhwAHw45NRJkOT83CLUEOQ44aE4sjaSUgYAvg56dTgA2h6UABYQLgCSbBxWLe1dPQDyrDSFIoQAchCd5LEZ5D42ALoNRhgmzu6evgFBIQsJXjYNmg0GDbVS1UA===

miloush: Wellll while your answer is true, it is a bit dismissive of a fair question, because that is not the case in .NET Framework, and the documentation for `IntPtr` does not mention any implicit operators.

miloush: It makes sense for the `nint` idea and it is captured here https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/builtin-types/numeric-conversions, together with the note on IntPtr page that _In C# starting from version 11 and when targeting the .NET 7 or later runtime, nint is an alias for IntPtr_ So it is a fairly recent development.

dipeshmsft: Thanks @miloush , @h3xds1nz for the information on this.

dipeshmsft (approved):

PR #9546, opened 2024-08-08
Fixed DatePicker AccentBorder style issue

dipeshmsft:
Fixes # 

Main PR 

###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/wpf/pull/9546)

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/Themes/PresentationFramework.Fluent/Themes/Fluent.HC.xaml
+++ b/src/Microsoft.DotNet.Wpf/src/Themes/PresentationFramework.Fluent/Themes/Fluent.HC.xaml
@@ -2093,6 +2093,7 @@
     <Setter Property="Border.CornerRadius" Value="{DynamicResource ControlCornerRadius}" />
     <Setter Property="SnapsToDevicePixels" Value="True" />
     <Setter Property="OverridesDefaultStyle" Value="True" />
+    <Setter Property="KeyboardNavigation.TabNavigation" Value="Local" />
     <Setter Property="Template">
       <Setter.Value>
         <ControlTemplate TargetType="{x:Type DatePicker}">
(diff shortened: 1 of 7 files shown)


src/Microsoft.DotNet.Wpf/src/Themes/PresentationFramework.Fluent/Themes/Fluent.HC.xaml
```diff
@@ -2093,6 +2093,7 @@
     <Setter Property="Border.CornerRadius" Value="{DynamicResource ControlCornerRadius}" />
     <Setter Property="SnapsToDevicePixels" Value="True" />
     <Setter Property="OverridesDefaultStyle" Value="True" />
+    <Setter Property="KeyboardNavigation.TabNavigation" Value="Local" />
```
hiteshwpfmsft, line 2096: Just for my learning purposes - 
How did this fix the accessibility issue?

dipeshmsft: Earlier this was set to Once by default, due to which only one child of the DatePicker control would get the navigation and then skip to other control. Now it will go to each of the siblings. Since there are two, textbox and the button, it goes to each one of them.

dipeshmsft: Here's the docuement for your reference : https://learn.microsoft.com/en-us/dotnet/api/system.windows.input.keyboardnavigationmode?view=windowsdesktop-8.0

hiteshwpfmsft (approved):

PR #9548, opened 2024-08-09
Fixing initial load of Fluent Dictionaries

dipeshmsft:
Fixes # 

Main PR 

###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/wpf/pull/9548)

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Application.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Application.cs
@@ -932,14 +932,21 @@ namespace System.Windows
                     oldValue.RemoveOwner(this);
                 }
 
-                if(ThemeManager.DeferredAppThemeLoading && !_resourcesInitialized)
+                if(_reloadFluentDictionary && !_resourcesInitialized)
                 {
+                    // We are here only when ThemeMode is set before Application.Resources is set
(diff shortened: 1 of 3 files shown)


src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Application.cs
```diff
@@ -990,10 +997,17 @@ public ThemeMode ThemeMode
...
-                    // If the resources are not initializd, 
-                    // fluent dictionary included will be reset.
-                    // Hence, deferring the step.
-                    ThemeManager.DeferredAppThemeLoading = true;
+                    // We do this because, Resources may not be present in App.xaml
+                    // so the setter will never be called, and we will not get a chance to 
+                    // load the fluent dictionary before parsing Window, where Fluent resources
+                    // may be used.
+                    ThemeManager.OnApplicationThemeChanged(oldValue, value);
+                    _resourcesInitialized = false;
```
harshit7962, line 1005: Is this needed? We are here only if `_resourcesInitialized` is already `false`.

dipeshmsft: Yes, because OnApplicatonThemeChanged will set the _resourcesInitialized to true

PR #9552, opened 2024-08-11
Replace boxing Hashtable in AdornerPresentationContext, reduce allocations

h3xds1nz:
## Description

Replaces two boxing `Hashtable` collections with generic `Dictionary<K, V>`, improving performance and decreasing allocations.
- There are no locks but I do not believe there are any thread-safety concerns as the collections are only set once per my code inspection and testing it with few sample apps, plus if it was a concern, even `Hashtable` would need lock for proper functionality in this scenario.
- Worst case, we could swap it to `ConcurrentDictionary<K, V>` which would still work out very well over boxing `Hashtable` in read-only context.
- ZRange is now a `readonly struct`, further decreasing memory footprint of the underlying collection.
- As `AdornerPresentationContext` is a derived class, I've sealed it.

### Base difference between boxing `Hashtable` and `Dictionary<K, V>`

| Original |  339.5 ns |    6.56 ns |     5.81 ns | 0.1044 |       2,756 B |        1752 B |
| PR__EDIT |  202.3 ns |    4.02 ns |     6.38 ns | 0.0534 |       1,301 B |         896 B |

### More of a real use-case in `AdornerPresentationContext`

| Original | 1,269.0 ns |   20.86 ns |    18.49 ns | 0.3757 |       2,775 B |        6312 B |
| PR__EDIT |   773.8 ns |    5.49 ns |     4.58 ns | 0.0534 |       1,283 B |         896 B |

```csharp
[Benchmark]
public void Original()
{
    Hashtable ZRanges = new Hashtable();

    for (int i = 0; i < 10; i++)
        BenchV1(i, ZRanges);

    //200 is used for the second benchmark instead of ZRanges!.Count
    for (int i = 0; i < ZRanges!.Count; i++)
    {
        ZRange range = (ZRange)ZRanges[i];
        if (range != null)
            EmptyMethod(range);
    }
}

[Benchmark]
public void PR__EDIT()
{
    Dictionary<int, ZRangeStruct> ZRanges = new();

    for (int i = 0; i < 10; i++)
        BenchV2(i, ZRanges);

    //200 is used for the second benchmark instead of ZRanges!.Count
    for (int i = 0; i < ZRanges!.Count; i++)
    {
        if (ZRanges.TryGetValue(i, out ZRangeStruct range))
            EmptyMethod2(ref range);
    }
}

private static int EmptyMethod(ZRange zRange) => zRange.Min;

//We use ref to prevent copy that doesn't happen in actual code
private static int EmptyMethod2(ref ZRangeStruct zRange) => zRange.Min;

[MethodImpl(MethodImplOptions.AggressiveInlining)]
private static void BenchV1(int level, Hashtable ZRanges)
{
    if (ZRanges[level] == null)
        ZRanges.Add(level, new ZRange(1, 2));
}

[MethodImpl(MethodImplOptions.AggressiveInlining)]
private static void BenchV2(int level, Dictionary<int, ZRangeStruct> ZRanges)
{
    if (!ZRanges.ContainsKey(level))
        ZRanges.Add(level, new ZRangeStruct(1, 2));
}
```
</details>

## Customer Impact

## Regression

No.

## Testing

Local build, sample apps run.

## Risk

There is a small concern as per my first point, however again, very unlikely given how the code is written.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/wpf/pull/9552)

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/PresentationFramework/MS/Internal/Annotations/Component/AdornerPresentationContext.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationFramework/MS/Internal/Annotations/Component/AdornerPresentationContext.cs
@@ -4,7 +4,7 @@
 
 //
 // Description:
-//      AdornerPresentationContext knows that annotation comonents are wrapped 
+//      AdornerPresentationContext knows that annotation components are wrapped 
 //      in an AnnotationAdorner and hosted in the AdornerLayer. Note, implementation-wise 
 //      a new PresentationContext is created for every annotation component. Executing 
 //      operations on a presentation context for a different annotation component 
@@ -24,7 +24,7 @@ using System.Collections;
 namespace MS.Internal.Annotations.Component
 {
     /// <summary>
-    /// AdornerPresentationContext knows that annotation comonents are wrapped in an AnnotationAdorner and hosted in the AdornerLayer.
+    /// AdornerPresentationContext knows that annotation components are wrapped in an AnnotationAdorner and hosted in the AdornerLayer.
     /// Note,  implementation-wise a new PresentationContext is created for every annotation component. Executing operations on a presentation context
     /// for a different annotation component (located in the same adorner layer) works, but is slower than using the presentation context stored in the
     /// annotation component.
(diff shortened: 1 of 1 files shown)


siagupta0202 (approved):

harshit7962:

h3xds1nz:
@harshit7962 Thank you for including them.

PR #9559, opened 2024-08-12
Replace boxing IList in HighlightChangedEventArgs and seal derived classes

edwardneal:
## Description

The `Ranges` property on the internal class `HighlightChangedEventArgs` was declared as an `IList`, but was always backed by an instance of `ReadOnlyList<TextSegment>`. Retrieving a list of text segments would thus always box an instance of the `TextSegment` struct. I've changed this property to declare it as an `IList<TextSegment>`.

I also swapped an `ArrayList` with a `List<int>` in FixedDocument.cs, which stored a list of page numbers.

Finally, I marked all the five classes derived from `HighlightChangedEventArgs` as sealed. All fives of these classes were private, so there's no API surface change here.

## Customer Impact

Reduced memory usage/GC pressure.

## Regression

No.

## Testing

Local build and tests, sample application runs.

## Risk

Low. The `Ranges` property is only read by `DocumentSequenceTextContainer`, `FixedDocument` and `FlowDocument`. In all three locations, it's indexed into and the result was explicitly cast to a `TextSegment`.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/wpf/pull/9559)


dipeshmsft:
LGTM, @edwardneal can you resolve the above merge-conflicts ?

edwardneal:
Thanks @dipeshmsft, I've just resolved them.

PR #9568, opened 2024-08-13
Support range notifications

### LINKED ISSUE #52: Improved INotifyCollectionChanged Handling
This is a carry over from .NET Framework. 

If INotifyCollectionChanged.CollectionChanged -> NotifyCollectionChangedEventArgs.NewItems has more than one item, most WPF controls will throw an exception. 

The main reason for this original design is that ObservableCollection doesn't have an AddRange method and the controls assum that nothing else will support INotifyCollectionChanged.

Mrxx99:
Fixes #52 

## Description

This adds support for range notifications coming from INotifyCollectionChanged for Add/Remove/Replace.
I hope I have covered all scenarios where it need to be adjusted. It works with and without `AllowsCrossThreadChanges`.

## Customer Impact

Customers can finally use ObservableCollections that implement range operations. Would also help unblock this: https://github.com/dotnet/runtime/issues/18087

## Testing

A test app that hopefully covers all scenarios, see here: https://github.com/Mrxx99/ObservableRangeWpfTestApp
Will look into DRTs.

## Risk

I am not aware of any, except of maybe regression bugs.

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Controls/ItemContainerGenerator.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Controls/ItemContainerGenerator.cs
@@ -2400,25 +2400,25 @@ namespace System.Windows.Controls
             switch (args.Action)
             {
                 case NotifyCollectionChangedAction.Add:
-                    if (args.NewItems.Count != 1)
+                    if (args.NewItems.Count < 1)
                         throw new NotSupportedException(SR.RangeActionsNotSupported);
-                    OnItemAdded(args.NewItems[0], args.NewStartingIndex);
+                    OnItemsAdded(args.NewItems, args.NewStartingIndex);
                     break;
 
                 case NotifyCollectionChangedAction.Remove:
-                    if (args.OldItems.Count != 1)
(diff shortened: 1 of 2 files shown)


h3xds1nz:
This unfortunately requires way more effort, just one out of many things that need to be addressed before the runtime PR would be unblocked in a way where it doesn't implement a new type / or using a switch:

For example, the changes you've made in `ListCollectionView`, this requires way more work for LiveShaping to work properly (Grouping, Filtering, etc.)

Then there's all the other views that would have to be fixed too.

Mrxx99:
Was thinking about that and had already had some more changes e.g. in the BindingListCollectionView, but I am not sure if all of these really are required. At least I thought that e.g. when using filtering you can't be using a normal ObservalbleCollection, but you would use a different CollectionView(Source). But now I guess it still would be handled as the CollectionViewSource will forward and adjust the change events?

I already thought it was not complete, that's why I created it as a draft and to also start a discussion to find everything that has to be done. So thanks for your help. Will take a look at filtering next. Never heard about the LiveShapingList (besides seeing the name in the code). I will have to investigate about this.

If you or someone else wants to help, the work could also be maybe be split into multiple pull requests. I will also accept pull requests on my branch and my test app to cover more scenarios.

PR #9643, opened 2024-08-24
Replace StringBuilder in DateTimeValueSerializer with format constants

h3xds1nz:
## Description

Replaces `StringBuilder` with simple on-stack build as we easily know the maximum length of the format string and don't need to therefore have two heap allocations when we can just have one.

Also replaces the awful virtual call to `GetFormat` for `DateTimeFormatInfo` on `InvariantCulture` and a sad nullability annotation with the very simple, static property of `DateTimeFormatInfo.InvariantInfo`.

### Basic method run

| Method   | value               | Mean [ns] | Error [ns] | StdDev [ns] | Gen0   | Code Size [B] | Allocated [B] |
|--------- |-------------------- |----------:|-----------:|------------:|-------:|--------------:|--------------:|
| Original | 10/10/1999 00:01:07 | 110.30 ns |   1.214 ns |    1.013 ns | 0.0210 |       3,211 B |         352 B |
| PR__EDIT | 10/10/1999 00:01:07 |  70.24 ns |   0.380 ns |    0.355 ns | 0.0038 |       1,739 B |          64 B |

### Benchmark source param

```csharp
public static IEnumerable<DateTime> RandomTime
{
    get
    {
        yield return new DateTime(1999, 10, 10, 0, 1, 7, 123, 456, DateTimeKind.Utc);
    }
}
```

## Customer Impact

## Regression

No.

## Testing

Local build, benchmark assert test.

## Risk

Low.

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/wpf/pull/9643)

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/System.Xaml/System/Windows/Markup/DateTimeValueSerializer.cs
+++ b/src/Microsoft.DotNet.Wpf/src/System.Xaml/System/Windows/Markup/DateTimeValueSerializer.cs
@@ -1,11 +1,10 @@
-// Licensed to the .NET Foundation under one or more agreements.
+﻿// Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
 using System.Runtime.CompilerServices;
 using System.ComponentModel;
 using System.Globalization;
-using System.Diagnostics;
 
 namespace System.Windows.Markup
 {
--- a/src/Microsoft.DotNet.Wpf/src/System.Xaml/System/Windows/Markup/DateTimeValueSerializer.cs
+++ b/src/Microsoft.DotNet.Wpf/src/System.Xaml/System/Windows/Markup/DateTimeValueSerializer.cs
@@ -5,7 +5,7 @@
 using System.Runtime.CompilerServices;
 using System.ComponentModel;
 using System.Globalization;
-using System.Xaml;
+using System.Diagnostics;
 
 namespace System.Windows.Markup
 {
@@ -66,9 +66,13 @@ namespace System.Windows.Markup
                 throw GetConvertToException(value, typeof(string));
             }
 
-            // Build up the format string to be used in DateTime.ToString()
-            DefaultInterpolatedStringHandler formatSpan = new(5, 0, CultureInfo.CurrentCulture, stackalloc char[38]);
-            formatSpan.AppendLiteral("yyyy-MM-dd");
+            // Build up the format string to be used in DateTime.TryParse()
+            Span<char> dateTimeSpan = stackalloc char[28];
+            Span<char> formatSpan = stackalloc char[36];
+            int formatLength = 0;
+
+            "yyyy-MM-dd".CopyTo(formatSpan);
+            formatLength += 10;
             if (dateTime.TimeOfDay == TimeSpan.Zero)
             {
                 // The time portion of this DateTime is exactly at midnight.
@@ -77,7 +81,8 @@ namespace System.Windows.Markup
                 // we'll have to include the time.
                 if (dateTime.Kind != DateTimeKind.Unspecified)
                 {
-                    formatSpan.AppendLiteral("'T'HH':'mm");
+                    "'T'HH':'mm".CopyTo(formatSpan.Slice(formatLength));
+                    formatLength += 10;
                 }
             }
             else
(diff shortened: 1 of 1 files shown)


Rob-Hague:
You mentioned TryFormat already but if this directly used spans it could avoid the InterpolatedStringHandler trickery (["This is intended to be called only by compiler-generated code"](https://learn.microsoft.com/dotnet/api/system.runtime.compilerservices.defaultinterpolatedstringhandler.-ctor?view=net-8.0#system-runtime-compilerservices-defaultinterpolatedstringhandler-ctor(system-int32-system-int32))) and avoid the format string allocation:

```c#
Span<char> formatSpan = stackalloc char[38];
int formatLength = 0;
Span<char> dateTimeSpan = stackalloc char[38];

"yyyy-MM-dd".CopyTo(formatSpan); formatLength += 10;

// etc

dateTimeSpan = dateTimeSpan.Slice(0, formatLength);
dateTime.TryFormat(dateTimeSpan, out int charsWritten, formatSpan.Slice(0, formatLength), DateTimeFormatInfo.InvariantInfo);

Debug.Assert(charsWritten == formatLength);

return new string(dateTimeSpan);
```

h3xds1nz:
Trickery, haha. But yeah could build it this way, I'll address it later, thanks :)

src/Microsoft.DotNet.Wpf/src/System.Xaml/System/Windows/Markup/DateTimeValueSerializer.cs
```diff
@@ -50,29 +47,28 @@ public override object ConvertFromString(string value, IValueSerializerContext?
...
         {
-            if (value == null || !(value is DateTime dateTime))
+            if (value is not DateTime dateTime)
             {
                 throw GetConvertToException(value, typeof(string));
             }
 
             // Build up the format string to be used in DateTime.ToString()
-            var formatString = new StringBuilder("yyyy-MM-dd");
+            DefaultInterpolatedStringHandler formatSpan = new(5, 0, CultureInfo.CurrentCulture, stackalloc char[38]);
```
lindexi, line 70: Can I know why the number of char is 38?

h3xds1nz: It should have been 36 actually.

Rob-Hague:
Perhaps "misuse" is the better term 😉. In reality I think this function could probably be just a number of dateTime.ToString calls with literal format arguments

ThomasGoulet73:
Since there is only 4 possible formats, I think it would be cleaner and faster to just define those 4 formats as constants and use those. Something like this (I haven't tested it and it can probably be refactored):
```
// Every format contains the "K" format specifier that indicates we want the DateTimeKind to be
// included in the output formulation -- UTC gets written out with a "Z",
// and Local gets written out with e.g. "-08:00" for Pacific Standard Time.
const string Format = "yyyy-MM-ddK";
const string FormatWithHoursAndMinutes = $"yyyy-MM-dd'T'HH':'mmK";
const string FormatWithHoursMinutesAndSeconds = $"yyyy-MM-dd'T'HH':'mm':'ssK";
const string FormatWithHoursMinutesSecondsAndDigits = $"yyyy-MM-dd'T'HH':'mm':'ss'.'FFFFFFFK";

string formatString = Format;
if (dateTime.TimeOfDay == TimeSpan.Zero)
{
    // The time portion of this DateTime is exactly at midnight.
    // We don't include the time component if the Kind is unspecified.
    // Otherwise, we're going to be including the time zone info, so'll
    // we'll have to include the time.
    if (dateTime.Kind != DateTimeKind.Unspecified)
    {
        formatString = FormatWithHoursAndMinutes;
    }
}
else
{
    long digitsAfterSecond = dateTime.Ticks % 10000000;
    int second = dateTime.Second;
    // We're going to write out at least the hours/minutes
    formatString = FormatWithHoursAndMinutes;
    if (second != 0 || digitsAfterSecond != 0)
    {
        // need to write out seconds
        formatString = FormatWithHoursMinutesAndSeconds;
        if (digitsAfterSecond != 0)
        {
            // need to write out digits after seconds
            formatString = FormatWithHoursMinutesSecondsAndDigits;
        }
    }
}
```

h3xds1nz:
Yeah, might be the best solution after all. Will look into it later.

himgoyalmicro (approved):

h3xds1nz:
@himgoyalmicro Done :)

himgoyalmicro (approved):

himgoyalmicro:
Thank you @h3xds1nz for your continuous contribution! 😄

h3xds1nz:
@himgoyalmicro Thank you for continuously merging them! 😀

PR #9683, opened 2024-08-30
Optimize conversion of ModifierKeys enum from/to string, reduce allocations

h3xds1nz:
Fixes #7502

## Description

Parsing optimization for `ModifierKeys` and its conversion from/to string. Conversion **from** is now alloc-free, conversion to is either free or reduced by a big margin (see benchmarks).

- `ConvertFrom` is now fully flying on `ReadOnlySpan<char>` instead of multiple substrings and trims on string instances.
- `ConvertTo` is now within 2 procs, fast-path and slow-path. The only allocated string instance is with multiple modifiers.
- Strings in `ConvertTo` will now be interned with the rest since we use `StringComparison.OrdinalIgnoreCase`.
- `ModifierKeysFlag` static field is now a compile-time constant, saving memory and providing speed-up.
- The JITTed codegen takes up less memory as we removed few things.
- Adds/completes code documentation for the whole class.

### `ConvertFrom` Benchmark for all possible modifiers

| Method   | source               | Mean [ns] | Error [ns] | StdDev [ns] | Gen0   | Code Size | Allocated |
|--------- |--------------------- |----------:|-----------:|------------:|-------:|----------:|----------:|
| Original | CTRL (...)+ ALT [24] | 101.55 ns |   1.790 ns |    1.587 ns | 0.0224 |       2,662 B |         376 B |
| PR__EDIT | CTRL (...)+ ALT [24] |  34.68 ns |   0.218 ns |    0.182 ns |      - |       1,891 B |             - |
| Original | Ctrl (...)+ Alt [28] | 147.59 ns |   1.616 ns |    1.512 ns | 0.0319 |       2,640 B |         536 B |
| PR__EDIT | Ctrl (...)+ Alt [28] |  35.19 ns |   0.483 ns |    0.451 ns |      - |       1,891 B |             - |

### `ConvertTo` benchmark for different scenarios
| Method    | value | Mean [ns] | Error [ns] | StdDev [ns] | Gen0   | Code Size [B] | Allocated [B] |
|---------- |------ |----------:|-----------:|------------:|-------:|--------------:|--------------:|
| Original  | None  | 1.0984 ns |  0.0108 ns |   0.0090 ns |      - |         685 B |             - |
| PR__EDIT  | None  | 0.8665 ns |  0.0078 ns |   0.0066 ns |      - |         238 B |             - |
| Original  | Shift | 1.1785 ns |  0.0149 ns |   0.0132 ns |      - |         780 B |             - |
[...]

PR #9697, opened 2024-09-01
Optimize conversion of Key enum from/to string, reduce allocations

h3xds1nz:
## Description

Parsing optimization for `Key` and its conversion from/to string. Conversion from is now alloc-free, conversion to is either untouched or free.

- `ConvertFrom` is now allocation free, optimizing performance from x4 for normal keys to x20 for F1-F12 keys.
   - The code-size of the method has increased but I think it is an acceptable compromise.
- `ConvertTo` is pretty much unchanged however F10-F12 were added to switch as those are interned.
- Adds/completes code documentation for the whole class.

### ConvertFrom benchmarks (code size is not real here 'cause depth is set to 1)
| Method      | source | Mean [ns]  | Error [ns] | StdDev [ns] | Gen0   | Code Size [B] | Allocated [B] |
|------------ |------- |-----------:|-----------:|------------:|-------:|--------------:|--------------:|
| Original | \x20F5\x20 | 170.052 ns |  1.7088 ns |   1.5984 ns | 0.0048 |         291 B |          80 B |
| PR__EDIT    | \x20F5\x20 |   9.761 ns |  0.0319 ns |   0.0298 ns |      - |       4,689 B |             - |
| Original | F5     | 184.430 ns |  3.5985 ns |   3.3661 ns | 0.0029 |         291 B |          48 B |
| PR__EDIT    | F5     |   3.748 ns |  0.0134 ns |   0.0119 ns |      - |       4,315 B |             - |
| Original | \x20A\x20  | 22.271 ns |  0.2355 ns |   0.1967 ns | 0.0043 |         291 B |          72 B |
| PR__EDIT    | \x20A\x20  | 10.832 ns |  0.0123 ns |   0.0109 ns |      - |       4,792 B |             - |
| Original | Play   | 23.951 ns |  0.3515 ns |   0.3288 ns | 0.0048 |         291 B |          80 B |
| PR__EDIT    | Play   |  4.095 ns |  0.0124 ns |   0.0110 ns |      - |       4,171 B |             - |
| Original | A      | 16.815 ns |  0.1958 ns |   0.1832 ns | 0.0029 |         291 B |          48 B |
| PR__EDIT    | A      |  4.148 ns |  0.0086 ns |   0.0077 ns |      - |       4,402 B |             - |

### ConvertTo benchmarks
| Method            | source | Mean [ns] | Error [ns] | StdDev [ns] | Gen0   | Code Size [B] | Allocated [B] |
|------------------ |------- |----------:|-----------:|------------:|-------:|--------------:|--------------:|
| Original| Back   |  2.264 ns |  0.0123 ns |   0.0109 ns |      - |         465 B |             - |
| PR__EDIT| Back   |  2.252 ns |  0.0123 ns |   0.0109 ns |      - |         545 B |             - |
| Original| F10    | 14.026 ns |  0.2135 ns |   0.1892 ns | 0.0014 |         473 B |          24 B |
| PR__EDIT| F10    |  2.460 ns |  0.0082 ns |   0.0068 ns |      - |         567 B |             - |
| Original| A      |  5.766 ns |  0.1288 ns |   0.1204 ns | 0.0014 |         381 B |          24 B |
| PR__EDIT| A      |  5.257 ns |  0.1345 ns |   0.1439 ns | 0.0014 |         553 B |          24 B |

## Customer Impact

Increased performance, zero/reduced allocations.

## Regression

No.

## Testing

Local build, CI, assert tests:
```csharp
// ConvertFrom Tests
Key x1 = (Key)oldKeyConverter.ConvertFrom(null, null, "F5");
Key x2 = (Key)newKeyConverter.ConvertFrom(null, null, "F5");
Assert.AreEqual(x1, x2);
x1 = (Key)oldKeyConverter.ConvertFrom(null, null, " BS");
x2 = (Key)newKeyConverter.ConvertFrom(null, null, " BS");
Assert.AreEqual(x1, x2);
x1 = (Key)oldKeyConverter.ConvertFrom(null, null, "A ");
x2 = (Key)newKeyConverter.ConvertFrom(null, null, "A ");
Assert.AreEqual(x1, x2);
x1 = (Key)oldKeyConverter.ConvertFrom(null, null, " Question ");
x2 = (Key)newKeyConverter.ConvertFrom(null, null, " Question ");
Assert.AreEqual(x1, x2);
x1 = (Key)oldKeyConverter.ConvertFrom(null, null, " OemQUEstion ");
x2 = (Key)newKeyConverter.ConvertFrom(null, null, " OemQUEstion ");
Assert.AreEqual(x1, x2);
x1 = (Key)oldKeyConverter.ConvertFrom(null, null, "  ");
x2 = (Key)newKeyConverter.ConvertFrom(null, null, "  ");
Assert.AreEqual(x1, x2);
// Invalid single character specified that is not a key
Assert.ThrowsException<ArgumentException>(() => oldKeyConverter.ConvertFrom(null, null, " á "));
Assert.ThrowsException<ArgumentException>(() => newKeyConverter.ConvertFrom(null, null, " á "));
// This is a case where 
// "NotSupportedException(SR.Format(SR.Unsupported_Key, fullName))"
// was expected but it is a dead code
Assert.ThrowsException<ArgumentException>(() => oldKeyConverter.ConvertFrom(null, null, " áb "));
Assert.ThrowsException<ArgumentException>(() => newKeyConverter.ConvertFrom(null, null, " áb "));

// CanConvertTo Tests
StandardContextImpl context = new StandardContextImpl();
context.Instance = (int)1;
bool x6 = oldKeyConverter.CanConvertTo(context, typeof(string));
bool x7 = newKeyConverter.CanConvertTo(context, typeof(string));
Assert.AreEqual(x6, x7);
context.Instance = (Key)1;
x6 = oldKeyConverter.CanConvertTo(context, typeof(string));
x7 = newKeyConverter.CanConvertTo(context, typeof(string));
Assert.AreEqual(x6, x7);
context.Instance = (byte)1;
Assert.ThrowsException<InvalidCastException>(() => oldKeyConverter.CanConvertTo(context, typeof(string)));
Assert.ThrowsException<InvalidCastException>(() => newKeyConverter.CanConvertTo(context, typeof(string)));

// ConvertTo Tests
string x3 = (string)oldKeyConverter.ConvertTo(null, null, Key.A, typeof(string));
string x4 = (string)newKeyConverter.ConvertTo(null, null, Key.A, typeof(string));
Assert.AreEqual(x3, x4);

x3 = (string)oldKeyConverter.ConvertTo(null, null, Key.F10, typeof(string));
x4 = (string)newKeyConverter.ConvertTo(null, null, Key.F10, typeof(string));
Assert.AreEqual(x3, x4);

x3 = (string)oldKeyConverter.ConvertTo(null, null, Key.Back, typeof(string));
x4 = (string)newKeyConverter.ConvertTo(null, null, Key.Back, typeof(string));
Assert.AreEqual(x3, x4);

x3 = (string)oldKeyConverter.ConvertTo(null, null, Key.DeadCharProcessed, typeof(string));
x4 = (string)newKeyConverter.ConvertTo(null, null, Key.DeadCharProcessed, typeof(string));
Assert.AreEqual(x3, x4);

x3 = (string)oldKeyConverter.ConvertTo(null, null, Key.None, typeof(string));
x4 = (string)newKeyConverter.ConvertTo(null, null, Key.None, typeof(string));
Assert.AreEqual(x3, x4);

//Null destinationType
Assert.ThrowsException<ArgumentNullException>(() => (string)oldKeyConverter.ConvertTo(null, null, (Key)(-1), null));
Assert.ThrowsException<ArgumentNullException>(() => (string)newKeyConverter.ConvertTo(null, null, (Key)(-1), null));
//Invalid destinationType
Assert.ThrowsException<NotSupportedException>(() => (string)oldKeyConverter.ConvertTo(null, null, (Key)(-1), typeof(Int32)));
Assert.ThrowsException<NotSupportedException>(() => (string)newKeyConverter.ConvertTo(null, null, (Key)(-1), typeof(Int32)));
//Value out of range
Assert.ThrowsException<NotSupportedException>(() => (string)oldKeyConverter.ConvertTo(null, null, (Key)(-1), typeof(string)));
Assert.ThrowsException<NotSupportedException>(() => (string)newKeyConverter.ConvertTo(null, null, (Key)(-1), typeof(string)));
//Value is null
Assert.ThrowsException<NotSupportedException>(() => (string)oldKeyConverter.ConvertTo(null, null, null, typeof(string)));
Assert.ThrowsException<NotSupportedException>(() => (string)newKeyConverter.ConvertTo(null, null, null, typeof(string)));
```

## Risk

Low, changes have been thoroughly tested.

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/wpf/pull/9697)

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/WindowsBase/System/Windows/Input/KeyConverter.cs
+++ b/src/Microsoft.DotNet.Wpf/src/WindowsBase/System/Windows/Input/KeyConverter.cs
@@ -45,10 +45,10 @@ namespace System.Windows.Input
                 return false;
 
             // When invoked by the serialization engine we can convert to string only for known type
-            if (context is null || context.Instance is not Key key)
+            if (context is null || context.Instance is null)
                 return false;
 
-            return IsDefinedKey(key);
+            return IsDefinedKey((Key)context.Instance);
         }
 
         /// <summary>
--- a/src/Microsoft.DotNet.Wpf/src/WindowsBase/System/Windows/Input/KeyConverter.cs
+++ b/src/Microsoft.DotNet.Wpf/src/WindowsBase/System/Windows/Input/KeyConverter.cs
@@ -2,24 +2,35 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
-using System.ComponentModel;
-using System.Globalization;
+//
+//
+// Description: KeyConverter - Converts a Key string
+// to the *Type* that the string represents and vice-versa
+//
+
+using System;
+using System.ComponentModel;    // for TypeConverter
+using System.Globalization;     // for CultureInfo
+using System.Reflection;
+using System.Windows;
+using System.Windows.Input;
+using System.Windows.Markup;
+using MS.Utility;
+using MS.Internal.WindowsBase;
 
 namespace System.Windows.Input
 {
     /// <summary>
-    /// Key Converter class for converting between a string and the Type of a Key
+    /// Converter class for converting between a <see langword="string"/> and <see cref="Key"/>.
     /// </summary>
-    /// <ExternalAPI/>
     public class KeyConverter : TypeConverter
     {
-        /// <summary>
-        /// CanConvertFrom()
-        /// </summary>
-        /// <param name="context"></param>
-        /// <param name="sourceType"></param>
-        /// <returns></returns>
-        /// <ExternalAPI/> 
+        ///<summary>
+        /// Used to check whether we can convert a <see langword="string"/> into a <see cref="Key"/>.
+        ///</summary>
+        ///<param name="context">ITypeDescriptorContext</param>
+        ///<param name="sourceType">type to convert from</param>
+        ///<returns><see langword="true"/> if the given <paramref name="sourceType"/> can be converted from, <see langword="false"/> otherwise.</returns>
         public override bool CanConvertFrom(ITypeDescriptorContext context, Type sourceType)
         {
             // We can only handle string
@@ -27,11 +38,11 @@ namespace System.Windows.Input
         }
 
         /// <summary>
-        /// TypeConverter method override. 
+        /// Used to check whether we can convert specified value to <see langword="string"/>.
         /// </summary>
         /// <param name="context">ITypeDescriptorContext</param>
         /// <param name="destinationType">Type to convert to</param>
-        /// <returns>true if conversion is possible</returns>
+        /// <returns><see langword="true"/> if conversion to <see langword="string"/> is possible, <see langword="false"/> otherwise.</returns>
         public override bool CanConvertTo(ITypeDescriptorContext context, Type destinationType)
         {
             // We can convert to a string
@@ -46,38 +57,36 @@ namespace System.Windows.Input
         }
 
         /// <summary>
-        /// ConvertFrom()
+        /// Converts <paramref name="source"/> of <see langword="string"/> type to its <see cref="Key"/> represensation.
         /// </summary>
-        /// <param name="context"></param>
-        /// <param name="culture"></param>
-        /// <param name="source"></param>
-        /// <returns></returns>
-        /// <ExternalAPI/> 
+        /// <param name="context">Parser Context</param>
+        /// <param name="culture">Culture Info</param>
+        /// <param name="source">Key String</param>
+        /// <returns>A <see cref="Key"/> representing the <see langword="string"/> specified by <paramref name="source"/>.</returns>
         public override object ConvertFrom(ITypeDescriptorContext context, CultureInfo culture, object source)
         {
             if (source is not string stringSource)
                 throw GetConvertFromException(source);
-            
+
             ReadOnlySpan<char> fullName = stringSource.AsSpan().Trim();
-            return GetKeyFromString(fullName);      
+            return GetKeyFromString(fullName);
         }
 
         /// <summary>
-        /// ConvertTo()
+        /// Converts a <paramref name="value"/> of <see cref="Key"/> to its <see langword="string"/> represensation.
         /// </summary>
-        /// <param name="context"></param>
-        /// <param name="culture"></param>
-        /// <param name="value"></param>
-        /// <param name="destinationType"></param>
-        /// <returns></returns>
-        /// <ExternalAPI/> 
+        /// <param name="context">Serialization Context</param>
+        /// <param name="culture">Culture Info</param>
+        /// <param name="value">Key value </param>
+        /// <param name="destinationType">Type to Convert</param>
+        /// <returns>A <see langword="string"/> representing the <see cref="Key"/> specified by <paramref name="value"/>.</returns>
         public override object ConvertTo(ITypeDescriptorContext context, CultureInfo culture, object value, Type destinationType)
         {
             ArgumentNullException.ThrowIfNull(destinationType);
 
             if (value is null || destinationType != typeof(string))
                 throw GetConvertToException(value, destinationType);
-       
+
             Key key = (Key)value;
             return key switch
             {
@@ -89,7 +98,7 @@ namespace System.Windows.Input
                 Key.Back => "Backspace",
                 Key.LineFeed => "Clear",
                 Key.Escape => "Esc",
-                // We will add some heavy used interned strings too (F10-F12)
+                // We will add some heavily used interned strings (F10-F12)
                 Key.F10 => "F10",
                 Key.F11 => "F11",
                 Key.F12 => "F12",
@@ -100,6 +109,11 @@ namespace System.Windows.Input
             };
         }
 
+        /// <summary>
+        /// Helper function that performs the conversion of <paramref name="keyToken"/> to the <see cref="Key"/> enum.
+        /// </summary>
+        /// <param name="keyToken">The string to convert from.</param>
+        /// <returns>A <see cref="Key"/> value corresponding to the specified string, <see cref="Key.None"/> if <paramref name="keyToken"/> was empty.</returns>
         private static Key GetKeyFromString(ReadOnlySpan<char> keyToken)
         {
             // If the token is empty, we presume "None" as our value but it is a success
@@ -307,6 +321,11 @@ namespace System.Windows.Input
             return Enum.Parse<Key>(keyToken, true);
         }
 
+        /// <summary>
+        /// Helper function similar to <see cref="Enum.IsDefined{Key}(Key)"/>, just lighter and faster.
+        /// </summary>
+        /// <param name="key">The value to test against.</param>
+        /// <returns><see langword="true"/> if <paramref name="key"/> falls in enumeration range, <see langword="false"/> otherwise.</returns>
         private static bool IsDefinedKey(Key key)
         {
             return key >= Key.None && key <= Key.DeadCharProcessed;
--- a/src/Microsoft.DotNet.Wpf/src/WindowsBase/System/Windows/Input/KeyConverter.cs
+++ b/src/Microsoft.DotNet.Wpf/src/WindowsBase/System/Windows/Input/KeyConverter.cs
@@ -22,14 +22,8 @@ namespace System.Windows.Input
         /// <ExternalAPI/> 
         public override bool CanConvertFrom(ITypeDescriptorContext context, Type sourceType)
         {
-            if (sourceType == typeof(string))
-            {
-                return true;
-            }
-            else
-            {
-                return false;
-            }
+            // We can only handle string
+            return sourceType == typeof(string);
         }
 
         /// <summary>
--- a/src/Microsoft.DotNet.Wpf/src/WindowsBase/System/Windows/Input/KeyConverter.cs
+++ b/src/Microsoft.DotNet.Wpf/src/WindowsBase/System/Windows/Input/KeyConverter.cs
@@ -8,14 +8,9 @@
 // to the *Type* that the string represents and vice-versa
 //
 
-using System;
 using System.ComponentModel;    // for TypeConverter
 using System.Globalization;     // for CultureInfo
-using System.Reflection;
-using System.Windows;
-using System.Windows.Input;
-using System.Windows.Markup;
-using MS.Utility;
+
 using MS.Internal.WindowsBase;
 
 namespace System.Windows.Input
--- a/src/Microsoft.DotNet.Wpf/src/WindowsBase/System/Windows/Input/KeyConverter.cs
+++ b/src/Microsoft.DotNet.Wpf/src/WindowsBase/System/Windows/Input/KeyConverter.cs
@@ -75,31 +75,29 @@ namespace System.Windows.Input
         {
             ArgumentNullException.ThrowIfNull(destinationType);
 
-            if (destinationType == typeof(string) && value != null)
+            if (value is null || destinationType != typeof(string))
+                throw GetConvertToException(value, destinationType);
+       
+            Key key = (Key)value;
+            return key switch
             {
-                Key key = (Key)value;
-                if (key == Key.None)
-                {
-                    return String.Empty;
-                }
-
-                if (key >= Key.D0 && key <= Key.D9)
-                {
-                    return Char.ToString((char)(int)(key - Key.D0 + '0'));
-                }
-
-                if (key >= Key.A && key <= Key.Z)
-                {
-                    return Char.ToString((char)(int)(key - Key.A + 'A'));
-                }
-
-                String strKey = MatchKey(key, culture);
-                if (strKey != null)
-                {
-                    return strKey;
-                }
-            }
-            throw GetConvertToException(value, destinationType);
+                Key.None => string.Empty,
+                // This is a fast path for common keys before resort to Enum<Key>.ToString()
+                >= Key.D0 and <= Key.D9 => char.ToString((char)(key - Key.D0 + '0')),
+                >= Key.A and <= Key.Z => char.ToString((char)(key - Key.A + 'A')),
+                // We format some keys differently than defined in the enum
+                Key.Back => "Backspace",
+                Key.LineFeed => "Clear",
+                Key.Escape => "Esc",
+                // We will add some heavy used interned strings too (F10-F12)
+                Key.F10 => "F10",
+                Key.F11 => "F11",
+                Key.F12 => "F12",
+                // Last resort, use Enum<Key>.ToString() if the range is defined
+                _ when IsDefinedKey(key) => key.ToString(),
+                // Everything else failed, we throw an exception
+                _ => throw GetConvertToException(value, destinationType)
+            };
         }
 
         private static Key GetKeyFromString(ReadOnlySpan<char> keyToken)
@@ -309,23 +307,9 @@ namespace System.Windows.Input
             return Enum.Parse<Key>(keyToken, true);
         }
 
-        private static string MatchKey(Key key, CultureInfo culture)
+        private static bool IsDefinedKey(Key key)
         {
-            if (key == Key.None)
-                return String.Empty;
-            else
-            {
-                switch (key)
-                {
-                    case Key.Back: return "Backspace";
-                    case Key.LineFeed: return "Clear";
-                    case Key.Escape: return "Esc";
-                }
-            }
-            if ((int)key >= (int)Key.None && (int)key <= (int)Key.DeadCharProcessed)
-                return key.ToString();
-            else
-                return null;
+            return key >= Key.None && key <= Key.DeadCharProcessed;
         }
     }
 }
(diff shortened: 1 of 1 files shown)


src/Microsoft.DotNet.Wpf/src/WindowsBase/System/Windows/Input/KeyConverter.cs
```diff
@@ -4,246 +4,326 @@
...
-                        case "ZOOM": keyFound = Key.Zoom; break;
-                        case "PA1": keyFound = Key.Pa1; break;
-                        default: keyFound = Enum.Parse<Key>(keyToken, true); break;
-                    }
-
-                    if ((int)keyFound != -1)
-                    {
-                        return keyFound;
-                    }
-                    return null;
```
h3xds1nz, line 225: This is a deadcode, as a result, the block `throw new NotSupportedException(SR.Format(SR.Unsupported_Key, fullName))` in `ConvertFrom` would never be thrown and instead we get `ArgumentException` from `Enum.Parse<Key>`, however I think we should adjust the behavior using `Enum<Key>.TryParse` to throw this on `false` as it seemed intended and correct behavior with other converters.

Opinions?

src/Microsoft.DotNet.Wpf/src/WindowsBase/System/Windows/Input/KeyConverter.cs
```diff
@@ -4,246 +4,326 @@
...
-                if (strKey != null)
-                {
-                    return strKey;
-                }
-            }
-            throw GetConvertToException(value, destinationType);
+                Key.None => string.Empty,
+                // This is a fast path for common keys before resort to Enum<Key>.ToString()
+                >= Key.D0 and <= Key.D9 => char.ToString((char)(key - Key.D0 + '0')),
+                >= Key.A and <= Key.Z => char.ToString((char)(key - Key.A + 'A')),
```
h3xds1nz, line 91: This could be sped up from 6ns to 2ns per invocation using static array (takes 10ns to create and 330B in memory)
```csharp
private static readonly string[] s_letters = ["A", "B", "C", "D", "E", "F", "G", "H", "I", "J", "K", "L", "M", "N", "O", "P", "Q", "R", "S", "T", "U", "V", "W", "X", "Y", "Z"];
private static readonly string[] s_digits = ["0", "1", "2", "3", "4", "5", "6", "7", "8", "9"];
```
and using array-lookup instead
```
>= Key.D0 and <= Key.D9 => s_digits[key - Key.D0],
>= Key.A and <= Key.Z => s_letters[key - Key.A],
```

Opinions?

| Method            | source | Mean [ns] | Error [ns] | StdDev [ns] | Gen0   | Code Size [B] | Allocated [B] |
|------------------ |------- |----------:|-----------:|------------:|-------:|--------------:|--------------:|
| Original| A      |  5.796 ns |  0.1326 ns |   0.1108 ns | 0.0014 |         381 B |          24 B |
| PR__EDIT| A      |  2.281 ns |  0.0134 ns |   0.0119 ns |      - |         575 B |             - |

src/Microsoft.DotNet.Wpf/src/WindowsBase/System/Windows/Input/KeyConverter.cs
miloush: I think it would be a compatibility concern to throw a different type of exception. `ArgumentException` seems appropriate anyway...

h3xds1nz: @miloush Other converters throw `NotSupportedException` with appropriate strings, e.g. `Unsupported_Modifier` in `ModifierKeysConverter` or `Unsupported_MouseAction` in `MouseActionConverter`.

miloush: An argument for aligning is that `NotSupportedException` is the documented one: https://learn.microsoft.com/en-us/dotnet/api/system.windows.input.keyconverter.convertfrom

h3xds1nz: Yeah, I'm inclined to align this one but maybe in a different PR since I wanna go over gestures init and conversion which is fugly, which requires me to touch the converter with an internal method to do it cleanly with spans and without duplicating functionality.

src/Microsoft.DotNet.Wpf/src/WindowsBase/System/Windows/Input/KeyConverter.cs
Symbai: > 5ns to 2ns per invocation [...] Opinions?

Because you asked for it: Smaller number always look more impressive when it comes to optimization. But if we speak about 3 nanoseconds more or less, I think we're at the level of micro optimization. Declaring additional arrays just to optimize two lines of code to get a -3 nanoseconds improvement is kind of an overkill. Especially as this is not such a hot path of code. For readability I'd prefer the original one. Otherwise: Good work!

miloush: I am not sure that `char.ToString((char)(key - Key.D0 + '0'))` is more readable than `s_digits[key - Key.D0]`, but that is likely a matter of personal opinion.

Performance-wise I agree this probably doesn't matter, especially since this is Key to string path, so not used during XAML loading, typically it would be showing shortcuts in menu entries and UI and for that the performance will get drowned in much bigger costs.

@h3xds1nz Curious, what is the allocation in the Original case for? Are the strings not interned?

h3xds1nz: @miloush `char.ToString` will forcefully init a new string instance everytime, there's no pool even for printable ASCII, that's where the 24B is coming from. Most characters are used solo as strings too (e.g. formats) throughout the assemblies, so the cost is really just mostly with the array itself. if you go that path.

I agree it is not a hot path and hence it was more like a point to consider; however I also think on the side of readibility the array solution makes more sense.

But in my own opinion about the implementation, I'd go with the `char` solution too, since the 3 ns doesn't matter here as you both pointed out. I like to discuss.

Symbai: > I am not sure that [...] is more readable than

You are right it's a matter of personal opinion. Thus I wouldn't have said something but h3xds1nz asked for it. But to explain my opinion: It's not the replacement of the lines, its the declaration of the two arrays. These are outside of the function. People who are reading the code have to figure out what they are for / where they are used etc. Do this a couple of times and the code becomes hard(er) to read. That's why I prefer to avoid it when it's not necessary.

h3xds1nz: Thanks guys, we will not move this direction.

src/Microsoft.DotNet.Wpf/src/WindowsBase/System/Windows/Input/KeyConverter.cs
h3xds1nz: I'll create a separate issue and PR for this particular case, can fix merge conflicts later.

src/Microsoft.DotNet.Wpf/src/WindowsBase/System/Windows/Input/KeyConverter.cs
```diff
@@ -4,246 +4,326 @@
...
-using System.Windows.Input;
-using System.Windows.Markup;
-using MS.Utility;
+
 using MS.Internal.WindowsBase;
 
 namespace System.Windows.Input
 {
     /// <summary>
-    /// Key Converter class for converting between a string and the Type of a Key
```
ThomasGoulet73, line 30: Since you're already updating the XML doc, could you use the docs from [learn.microsoft.com](https://learn.microsoft.com/en-us/dotnet/api/system.windows.input.keyconverter) instead ? learn.microsoft.com is the source of truth so we should use it. If you want to modify the documentation you should open a PR in https://github.com/dotnet/dotnet-api-docs.

src/Microsoft.DotNet.Wpf/src/WindowsBase/System/Windows/Input/KeyConverter.cs
```diff
@@ -4,246 +4,326 @@
...
         {
-            if (keyToken.Length == 0)
-            {
+            // If the token is empty, we presume "None" as our value but it is a success
+            if (keyToken.IsEmpty)
                 return Key.None;
+
+            // In case we're dealing with a lowercase character, we uppercase it
+            char firstChar = keyToken[0];
+            if (firstChar >= 'a' && firstChar <= 'z')
```
ThomasGoulet73, line 120: You can use:
```suggestion
            if (char.IsAsciiLetterLower(firstChar))
```

src/Microsoft.DotNet.Wpf/src/WindowsBase/System/Windows/Input/KeyConverter.cs
```diff
@@ -4,246 +4,326 @@
...
+                firstChar ^= (char)0x20;
```
ThomasGoulet73, line 121: What's the performance gain of doing this instead of:

```suggestion
                firstChar = firstChar.ToLowerInvariant();
```

(I didn't benchmark it)

src/Microsoft.DotNet.Wpf/src/WindowsBase/System/Windows/Input/KeyConverter.cs
```diff
@@ -4,246 +4,326 @@
...
+
+            // If this is a single-character we're dealing with, match digits/letters
+            if (keyToken.Length == 1 && char.IsLetterOrDigit(firstChar))
+            {
+                // Match a digit (0-9) or an ASCII letter (lower/uppercase)
+                if (char.IsDigit(firstChar) && (firstChar >= '0' && firstChar <= '9'))
```
ThomasGoulet73, line 127: You can use:
```suggestion
                if (char.IsAsciiDigit(firstChar))
```

src/Microsoft.DotNet.Wpf/src/WindowsBase/System/Windows/Input/KeyConverter.cs
```diff
@@ -4,246 +4,326 @@
...
+                    return Key.D0 + firstChar - '0';
+                else if (char.IsLetter(firstChar) && (firstChar >= 'A' && firstChar <= 'Z'))
```
ThomasGoulet73, line 129: You can use:
```suggestion
                else if (char.IsAsciiLetter(firstChar))
```

src/Microsoft.DotNet.Wpf/src/WindowsBase/System/Windows/Input/KeyConverter.cs
```diff
@@ -4,246 +4,326 @@
...
-                    return ((Key)key);
-                }
-                else
-                {
-                    throw new NotSupportedException(SR.Format(SR.Unsupported_Key, fullName));
-                }
-            }
-            throw GetConvertFromException(source);
+            if (source is not string stringSource)
+                throw GetConvertFromException(source);
```
ThomasGoulet73, line 64: Could you revert this usage of GetConvertFromException and use the previous code `throw new NotSupportedException(SR.Format(SR.Unsupported_Key, fullName));`. This is a change of behavior since it changes the exception message and AFAIK GetConvertFromException returns a non-localized message while the exceptions in WPF are localized. If you want to do the change you should probably do it in a separate PR just to keep this PR a 1:1 replacement.

src/Microsoft.DotNet.Wpf/src/WindowsBase/System/Windows/Input/KeyConverter.cs
h3xds1nz: `Converts a Key object to and from other types.`
vs.
`Converter class for converting between a string and Key`. (could be `conversion`, right)

While I agree with you about the source of truth, looking at the source, you clearly know this only works between `string` and `Key`, so the msdn description is less technical, and while it is correct in the spirit of what the class does, it doesn't specify it for the developer the implementation reality.

To be honest, I don't think I really wanna spend my time writing PRs for docs unless there's something horribly wrong, while noble and helpful, I'd do nothing else. But if I'm doing a PR on a small class, I want to have it right for others.

src/Microsoft.DotNet.Wpf/src/WindowsBase/System/Windows/Input/KeyConverter.cs
h3xds1nz: This makes more sense and will be faster, thanks, will fix.

src/Microsoft.DotNet.Wpf/src/WindowsBase/System/Windows/Input/KeyConverter.cs
h3xds1nz: That would xor the 5th bit again (0x20), so the correct one will be `char.IsAsciiLetterUpper`, which removes this redundancy.

src/Microsoft.DotNet.Wpf/src/WindowsBase/System/Windows/Input/KeyConverter.cs
h3xds1nz: If it is an ASCII codepoint, it's irrelevant: 40B (60/20) in code size, and instead of basically zero you "spend time" even on no-op (again, not important - 0.05ns vs. 0.004ns). **Irrelevant**.

However, give me a non-ASCII codepoint and it's 10ns slower path to getting an exception, which I think is important as this is a public converter that's kind of used widely and not always with sanitized input with people catching left and right.

src/Microsoft.DotNet.Wpf/src/WindowsBase/System/Windows/Input/KeyConverter.cs
h3xds1nz: @ThomasGoulet73 Can I resolve the conversation? I believe I've addressed your concern with it.

src/Microsoft.DotNet.Wpf/src/WindowsBase/System/Windows/Input/KeyConverter.cs
```diff
@@ -4,246 +4,326 @@
...
-        /// <param name="context"></param>
-        /// <param name="sourceType"></param>
-        /// <returns></returns>
-        /// <ExternalAPI/> 
+        ///<summary>
+        /// Used to check whether we can convert a <see langword="string"/> into a <see cref="Key"/>.
+        ///</summary>
+        ///<param name="context">ITypeDescriptorContext</param>
+        ///<param name="sourceType">type to convert from</param>
+        ///<returns><see langword="true"/> if the given <paramref name="sourceType"/> can be converted from, <see langword="false"/> otherwise.</returns>
```
harshit7962, line 28: nit: 
```suggestion
        ///<returns><see langword="true"/> if the given <paramref name="sourceType"/> can be converted, <see langword="false"/> otherwise.</returns>
```

src/Microsoft.DotNet.Wpf/src/WindowsBase/System/Windows/Input/KeyConverter.cs
```diff
@@ -4,246 +4,326 @@
...
+            if (source is not string stringSource)
+                throw GetConvertFromException(source);
+
+            ReadOnlySpan<char> fullName = stringSource.AsSpan().Trim();
+            return GetKeyFromString(fullName);
         }
 
         /// <summary>
-        /// ConvertTo()
+        /// Converts a <paramref name="value"/> of <see cref="Key"/> to its <see langword="string"/> represensation.
```
harshit7962, line 71: nit:
```suggestion
        /// Converts a <paramref name="value"/> of <see cref="Key"/> type to its <see langword="string"/> represensation.
```

harshit7962 (approved):
LGTM.
@h3xds1nz, could you please address the minor nits and the merge conflict? Let me know if you’d like me to resolve the conflicts.

src/Microsoft.DotNet.Wpf/src/WindowsBase/System/Windows/Input/KeyConverter.cs
h3xds1nz: Fixed

src/Microsoft.DotNet.Wpf/src/WindowsBase/System/Windows/Input/KeyConverter.cs
h3xds1nz: Fixed

h3xds1nz:
@harshit7962 Thanks, merge conflicts fixed. I've additionally fixed the word "representation" on two occurrences as I've misspelled it.

harshit7962 (approved):
Reapproving post resolution of conflicts and review comments.

harshit7962:
Thank you @h3xds1nz for the contribution and prompt resolution of conflicts.

h3xds1nz:
@harshit7962 Thank you for reviewing; if we could queue up #9683 (already went through test pass) for review I could get started on MouseGesture/KeyGesture converters finally.

harshit7962:
Sure, will prioritize #9683 along with other PRs that you have mentioned already.

PR #9710, opened 2024-09-04
Build changes to enable GenAPI in repo

dipeshmsft:
## Description

- Added MicrosoftDotNetGenAPIVersion in Version.props to allow GenAPI to find the right task at build time.

- Add GenAPIHeaderFile and GenAPIHeader properties to GenApi.props to use .NET C# file header rather than auto generated header. Added GenAPIHeaderText.txt file with the header text.

- Added a condition in GenAPI.props, where we set _GenerateReferenceAssemblySource false when we are building a _wpftmp project

- Removed 2 attributes from GlobalAttrExclusions.txt to minimize the changes in ref assembly on regeneration. ( We will add them back in future after investigation )

- Added a hack where we generate ref assemblies for _wpftmp assemblies by setting a target dir, and then later after generation is complete deleted ref assembly files for these _wpftmp assemblies. I added this hack because, when we set the flag for generating ref assemblies using GenAPI, that flag gets enabled for temp assemblies as well, and while building, MSBuild can't find the location to generate the ref as that directory is no more present. So, as a quick fix, I set the target dir, in GenApi.targets and then delete it once the generation is complete. PS : We can find a better way to deal with this situation.

## Customer Impact
Developers contributing to WPF, can easily regenerate the reference assemblies and pick out the required changes to include in the reference assemblies

## Regression
Maybe. GenAPI would have been working initially, but this is not a code/product regression

## Testing
Locally regenerated the ref assemblies.

## Risk
N/A . This is an out of build process that needs to be done manually.

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/wpf/pull/9710)

### DIFF
--- a/eng/Versions.props
+++ b/eng/Versions.props
@@ -50,7 +50,8 @@
   <PropertyGroup>
     <MicrosoftDotNetApiCompatVersion>9.0.0-beta.23621.2</MicrosoftDotNetApiCompatVersion>
     <MicrosoftDotNetCodeAnalysisPackageVersion>9.0.0-beta.23621.2</MicrosoftDotNetCodeAnalysisPackageVersion>
-  </PropertyGroup>
+    <MicrosoftDotNetGenAPIVersion>9.0.0-beta.23621.2</MicrosoftDotNetGenAPIVersion>
+ </PropertyGroup>
   <!-- Sourcelink -->
   <PropertyGroup>
     <MicrosoftSourceLinkGitHubVersion>8.0.0-beta.23409.2</MicrosoftSourceLinkGitHubVersion>
--- a/eng/WpfArcadeSdk/tools/GenApi.props
+++ b/eng/WpfArcadeSdk/tools/GenApi.props
@@ -25,10 +25,13 @@
 
     <GlobalApiExclusionsFile Condition="'$(GlobalApiExclusionsFile)'==''">$(WpfArcadeSdkRoot)tools\GenApi\GlobalApiExclusions.txt</GlobalApiExclusionsFile>
     <GlobalAttrExclusionsFile Condition="'$(GlobalAttrExclusionsFile)'==''">$(WpfArcadeSdkRoot)tools\GenApi\GlobalAttrExclusions.txt</GlobalAttrExclusionsFile>
+    <GenAPIHeaderFile Condition="'$(GenAPIHeaderFile)'==''">$(WpfArcadeSdkRoot)tools\GenApi\GenAPIHeaderText.txt</GenAPIHeaderFile>
 
     <GenAPIExcludeMembers>true</GenAPIExcludeMembers>
     <GenAPIExcludeApiList Condition="'$(GlobalApiExclusionsFile)'!=''">$(GlobalApiExclusionsFile)</GenAPIExcludeApiList>
     <GenAPIExcludeAttributesList Condition="'$(GlobalAttrExclusionsFile)'!=''">$(GlobalAttrExclusionsFile)</GenAPIExcludeAttributesList>
+    <GenAPIHeader>$(GenAPIHeaderFile)</GenAPIHeader>
+
 
     <!-- 
       GenAPI is run on demand.  Note that making a public API change without the associated reference assembly change will result in an ApiCompat error. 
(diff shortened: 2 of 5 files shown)


ViktorHofer:
cc @ericstj

eng/Versions.props
```diff
@@ -50,7 +50,8 @@
   <PropertyGroup>
     <MicrosoftDotNetApiCompatVersion>9.0.0-beta.23621.2</MicrosoftDotNetApiCompatVersion>
     <MicrosoftDotNetCodeAnalysisPackageVersion>9.0.0-beta.23621.2</MicrosoftDotNetCodeAnalysisPackageVersion>
-  </PropertyGroup>
+    <MicrosoftDotNetGenAPIVersion>9.0.0-beta.23621.2</MicrosoftDotNetGenAPIVersion>
```
ericstj, line 53: Please include a subscription for this version in versiondetails.xml so that it's not stuck at this version and gets updated automatically.

eng/WpfArcadeSdk/tools/GenApi.props
```diff
@@ -37,7 +40,8 @@
     <_GenerateReferenceAssemblySource>false</_GenerateReferenceAssemblySource>
     <_GenerateReferenceAssemblySource Condition="'$(GenerateReferenceAssemblySource)'=='true'
                                                         and '$(GenAPIEnabledProjects)'!=''
-                                                        and $(GenAPIEnabledProjects.Contains('$(MSBuildProjectName);'))">true</_GenerateReferenceAssemblySource>
+                                                        and $(GenAPIEnabledProjects.Contains('$(MSBuildProjectName);'))
+                                                        and !($(MSBuildProjectName.Contains('_wpftmp')))">true</_GenerateReferenceAssemblySource>
```
ericstj, line 44: How would this ever be true if you're checking `GenAPIEnabledProjects` contains the entire `$(MSBuildProjectName);`?

eng/WpfArcadeSdk/tools/GenApi.targets
```diff
@@ -4,13 +4,30 @@
                       Condition="'$(_GenerateReferenceAssemblySource)'=='true'"/>
   </ItemGroup>
 
-  <PropertyGroup Condition="'$(_GenerateReferenceAssemblySource)'=='true'">
+  <PropertyGroup Condition="'$(_GenerateReferenceAssemblySource)'=='true'
+                                or $(MSBuildProjectName.Contains('_wpftmp'))">
```
ericstj, line 8: I don't think you want to run GenAPI when `$(MSBuildProjectName.Contains('_wpftmp')`, right?

eng/WpfArcadeSdk/tools/GenApi.targets
```diff
@@ -19,4 +36,18 @@
     <MakeDir Directories="$(GenAPITargetDir)" />
   </Target>
 
+  <!-- 
+    Target to remove ref files generated while building _wpftmp projects.
+    These files are generated when we call the GenerateTemporaryTargetAssembly task on
+    the main assembly.
+    PS : This is the part of the hack above, to delete the unwanted ref assembly files.
+  -->
+  <Target Name="RemoveTempGenAPITargetDirectory"
```
ericstj, line 45: You shouldn't need this.

eng/Versions.props
ericstj: Also, we'd prefer if you use `Microsoft.DotNet.GenAPI.Task` instead.  The version from arcade is [no longer being maintained](https://github.com/dotnet/arcade/blob/main/src/Microsoft.DotNet.GenAPI/README.md) and is missing support for the latest language features.

eng/WpfArcadeSdk/tools/GenApi.targets
```diff
@@ -4,13 +4,30 @@
...
       Instead, WPF generates into the ref directory in the project's source tree.
     -->
     <GenAPITargetDir>$(MSBuildProjectDir)ref\</GenAPITargetDir>
+
+    <!-- 
+      Sets GenAPITargetDir for _wpftmp assemblies, these directories are not needed
+      and removed after generation.
+      This is a quick hack to avoid failures during GenerateReferenceAssemblySource for 
+      _wpftmp assemblies. Added this because once we set GenerateReferenceAssemblySource
+      parameter from CLI, it is enabled for temp assemblies as well, and we fail to filter
```
ericstj, line 20: Did you try just calling the `GenerateReferenceAssemblySource` target?  That target will just run GenAPI from your already built assembly.  That's how folks use this in runtime.

dipeshmsft: I will give it a try. I made these changes a year back but never merged it; I have forgotten some of the details by now. I will take a look and make necessary updates.

dipeshmsft:
@ericstj closing this PR for now, will try using the Task instead and raise another PR

ericstj:
Let us know if you have any questions / challenges making that work.  We'll help you get it working.

PR #9711, opened 2024-09-04
Improve start-up time by not figuring out NetFX version for AppContext switches

h3xds1nz:
## Description

All WPF apps will spend now spend considerable time on startup evaluating whether they're running on NetFX or not by initializing `AppContextDefaultValues` in each assembly (`PresentationCore`/`PresentationFramework`/`WindowsBase`), querying `TargetFrameworkName` property via reflection on AppDomain/AppContext (depends which runtime, one forwards the other).

The only place where the check makes sense is in `PresentationBuildTasks` on net472, which can be easily opted-in by using a preprocessor macro, rest of the assemblies do not need it and can confidently init as `.NETCoreApp`, version doesn't matter.

- We can keep the current code for `PBT` on `net472` via preprocessor condition and omit it from core builds
- We do not need to retrieve `TargetFrameworkName` via reflection on any target framework anymore

While I could have removed the cases in core libraries for NetFX, I presume we can do it in a different PR that cleans-up the whole logic for all of them, this is supposed to be neat and clean in-place solution.

## Customer Impact

Improved start-up time, decreased start-up allocations.

## Regression

No.

## Testing

Local build.

## Risk

Low, this merely removes basically dead code on core libraries.

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/wpf/pull/9711)

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/Common/src/System/AppContextDefaultValues.cs
+++ b/src/Microsoft.DotNet.Wpf/src/Common/src/System/AppContextDefaultValues.cs
@@ -19,15 +19,19 @@ namespace System
     {
         public static void PopulateDefaultValues()
         {
-            string platformIdentifier, profile;
-            int version;
-
-            ParseTargetFrameworkName(out platformIdentifier, out profile, out version);
+#if NETFX
+            // Get Target Framework information
+            ParseTargetFrameworkName(out string platformIdentifier, out string profile, out int version);
 
             // Call into each library to populate their default switches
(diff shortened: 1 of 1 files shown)


singhashish-wpf (approved):

harshit7962:
@h3xds1nz Thank you for actively contributing.

h3xds1nz:
@harshit7962 Thank you, this is a nice startup time improvement. I can now follow-up with a cleanup PR.

PR #9719, opened 2024-09-05
Fixed Font Scaling issue when FontSize is not set explicitly

### LINKED ISSUE #9192: .NET 9 Preview 4 Wpf Themed using PresentatinFramework.Fluent Fonts do not scale with System Font scaling
### Description

When using the the PresentationFramework.Fluent in .NET 9 Preview 4 the Fonts do not scale with Settings/Accesability/TextSize when usig WPF

### Reproduction Steps

Any control showing text will reveal the issue

### Expected behavior

Scalling as in .NET 8

### Actual behavior

Test remains the same size

### Regression?

_No response_

### Known Workarounds

_No response_

### Impact

I guess this is a work in progress

### Configuration

.NET 9 Preview 4
Windows 11
x64
Unkown

### Other information

_No response_

dipeshmsft:
Fixes #9192  

## Description
Fluent styles use 14 as the default size and so in order to handle this we check if the Fluent theme is enabled at the Application level and in that case we used to return the size 14 without scaling it. Hence, even though Text Scaling is set to something other than original size, we did not adjust the font size to match the scaling.

In this PR, I have replaced the DefaultFluentFontSize with DefaultFluentFontSizeFactor and use this to modify the system font size that we get from SystemParameters

## Customer Impact
Developers and users that use scaling other than 100% will see the text in scaled size ( whenever the size is not explicitly set on the control )

## Regression
Technically no, as Fluent style is a new feature.

## Testing
Local Ad-Hoc testing with sample applications

## Risk
Minimal. No one is using this currently ( until we count the .NET 9 Preview releases )

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/wpf/pull/9719)

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/SystemFonts.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/SystemFonts.cs
@@ -428,7 +428,7 @@ namespace System.Windows
                 // TODO : Find a better solution to this. Difference in default size of font in Fluent and other themes.
                 if(ThemeManager.IsFluentThemeEnabled)
                 {
-                    return ThemeManager.DefaultFluentThemeFontSize;
+                    return ConvertFontHeight(SystemParameters.NonClientMetrics.lfMessageFont.lfHeight * ThemeManager.DefaultFluentFontSizeFactor) ;
                 }
                 return ConvertFontHeight(SystemParameters.NonClientMetrics.lfMessageFont.lfHeight);
             }
@@ -1021,6 +1021,22 @@ namespace System.Windows
                 // Could not get the DPI to convert the size, using the hardcoded fallback value
                 return FallbackFontSize;
             }
+        }
+
+        private static double ConvertFontHeight(double height)
+        {
+            int dpi = SystemParameters.Dpi;
+
+            if (dpi != 0)
+            {
+                return (double)(Math.Abs(height) * 96 / dpi);
+            }
+            else
+            {
+                // Could not get the DPI to convert the size, using the hardcoded fallback value
+                return FallbackFontSize;
+            }
+
         }
 
         private const double FallbackFontSize = 11.0;   // To use if unable to get the system size
--- a/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/SystemFonts.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/SystemFonts.cs
@@ -425,12 +425,21 @@ namespace System.Windows
         {
             get
             {
+                return ConvertFontHeight(SystemParameters.NonClientMetrics.lfMessageFont.lfHeight);
+            }
+        }
+
+        internal static double ThemeMessageFontSize
+        {
+            get
+            {
                 // TODO : Find a better solution to this. Difference in default size of font in Fluent and other themes.
                 if(ThemeManager.IsFluentThemeEnabled)
                 {
-                    return ConvertFontHeight(SystemParameters.NonClientMetrics.lfMessageFont.lfHeight * ThemeManager.DefaultFluentFontSizeFactor) ;
+                    return MessageFontSize * ThemeManager.DefaultFluentFontSizeFactor ;
                 }
-                return ConvertFontHeight(SystemParameters.NonClientMetrics.lfMessageFont.lfHeight);
+
+                return MessageFontSize;
             }
         }
 
@@ -1021,22 +1030,6 @@ namespace System.Windows
                 // Could not get the DPI to convert the size, using the hardcoded fallback value
                 return FallbackFontSize;
             }
-        }
-
-        private static double ConvertFontHeight(double height)
-        {
-            int dpi = SystemParameters.Dpi;
-
-            if (dpi != 0)
-            {
-                return (double)(Math.Abs(height) * 96 / dpi);
-            }
-            else
-            {
-                // Could not get the DPI to convert the size, using the hardcoded fallback value
-                return FallbackFontSize;
-            }
-
         }
 
         private const double FallbackFontSize = 11.0;   // To use if unable to get the system size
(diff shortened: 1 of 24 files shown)


src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/SystemFonts.cs
```diff
@@ -1023,6 +1023,22 @@ private static double ConvertFontHeight(int height)
             }
         }
 
+        private static double ConvertFontHeight(double height)
+        {
+            int dpi = SystemParameters.Dpi;
+
+            if (dpi != 0)
+            {
+                return (double)(Math.Abs(height) * 96 / dpi);
```
h3xds1nz, line 1032: The cast here does nothing. Same as it does nothing in the original function you've copied.

The original function returns `double` but the calculations are done in `int`, so it is truncated and then returned as double. Yours does the computations in double.

miloush:
Wait a minute I missed the original code too, are you changing what `SPI_NONCLIENTMETRICS` returns based on the theme? I don't think that's very good. Why is this not dealt with in the theme?

dipeshmsft:
@miloush, if we handled it in the theme, then the inheritence of FontSize would not have worked as the ThemeStyles \ Styles have priority over inheritence. So, I wanted to handle it at the lowest level possible.

src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/SystemFonts.cs
```diff
@@ -428,7 +428,7 @@ public static double MessageFontSize
                 // TODO : Find a better solution to this. Difference in default size of font in Fluent and other themes.
```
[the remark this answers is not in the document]
dipeshmsft: I don't think this is the right way to handle this, so I think we should keep the comment here.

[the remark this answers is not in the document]
h3xds1nz: Yeah if we're not solving it here, then for sure haha.

miloush:
I am not quite sure I understand the scenario that doesn't work, but even if that was the case, can you not have this property in the theme assembly and read it from there? Where is this used?

miloush:
The answer to where it is used is

https://github.com/dotnet/wpf/blob/4b83839af3c63cda0890f628c83ffb3deef9d06b/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Documents/TextElement.cs#L643-L652

src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/SystemFonts.cs
dipeshmsft: Removed this

h3xds1nz: Yeah, looks better now.

Though not for this PR, but I wonder about the original intention, the truncation is obviously not ideal for scaling but then again, it's been there since the beginning (even for NetFX).

dipeshmsft:
I have modified the above PR to revert the changes in MessageFontSize and introduced an internal property which does the Fluent theme enabled check and returns font size accordingly. This way, the original API is unchanged now, and will enable font scaling for controls when using Fluent themes.

src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/SystemFonts.cs
```diff
@@ -422,15 +422,24 @@ public static TextDecorationCollection StatusFontTextDecorations
...
+
+        internal static double ThemeMessageFontSize
         {
             get
             {
                 // TODO : Find a better solution to this. Difference in default size of font in Fluent and other themes.
                 if(ThemeManager.IsFluentThemeEnabled)
                 {
-                    return ThemeManager.DefaultFluentThemeFontSize;
+                    return MessageFontSize * ThemeManager.DefaultFluentFontSizeFactor ;
```
harshit7962, line 439: Is there a reason we are not doing this in `MessageFontSize` itself? I can think of a case where someone is using the above public property and using Fluent. This will still posses the scaling issues there.

harshit7962 (changes requested):

src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/SystemFonts.cs
miloush: Yes because `MessageFontSize` is documented to return `SPI_GETNONCLIENTMETRICS` value.

h3xds1nz: As @miloush says, we've raised this in previous comments. Those properties are representing `SystemParametersInfo`, besides being adjusted for DPI, it is partially a raw abstraction that shouldn't change based on active theme.

A new API (e.g. property) would be suitable if we want to return it publicly imho.

miloush: On top of that, we should avoid mixing theme-specific code and logic into non-theme assemblies.

src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/ThemeManager.cs
```diff
@@ -204,6 +204,11 @@ internal static ResourceDictionary GetThemeDictionary(ThemeMode themeMode)
         return rd;
     }
 
+    internal static bool IsFluentThemeDictionayIncluded()
```
h3xds1nz, line 207: Dictionary*

harshit7962 (approved):

PR #9736, opened 2024-09-07
Improve startup and runtime performance by removing XamlTypeInvoker dead code

ThomasGoulet73:
## Description
Removes XamlTypeInvoker dead code. This code is hit when creating an instance of a non-known type from XAML/BAML. Removing this code improves both startup and runtime performance.

Known types are types in an hard-coded list that makes them more optimized by WPF. This includes types from WPF and some BCL common types (Like int, bool, string, etc). Everything else is a non-known type so things like custom controls are non-known types.

The reason this is dead code is because `DefaultCtorXamlActivator.EnsureConstructorDelegate` always returns false because of this condition here:
https://github.com/dotnet/wpf/blob/e500f8e58e439566ca68cdeff09303fb6f46110a/src/Microsoft.DotNet.Wpf/src/System.Xaml/System/Xaml/Schema/XamlTypeInvoker.cs#L330

Since the removal of CAS from .Net, `ConstructorInfo.IsSecurityCritical` always returns true and `ConstructorInfo.IsSecuritySafeCritical` always returns false so the the condition is always true which means that the whole `DefaultCtorXamlActivator` class doesn't do anything.

Here's the .Net code for `ConstructorInfo.IsSecurityCritical` and `ConstructorInfo.IsSecuritySafeCritical` which always return constants:
https://github.com/dotnet/runtime/blob/a349912e4f8610f39a6ac3cb73d8d4d0968ccec4/src/coreclr/System.Private.CoreLib/src/System/Reflection/RuntimeConstructorInfo.CoreCLR.cs#L228-L230

 #### Benchmarks
I benchmarked a new WPF from the Visual Studio template where I added an empty class like `public class CustomLabel : Label { }` and I added this control 20 times in the XAML of the main window. This PR shaves about 30-40 ms when starting the app.

There's a small cache per-type that skips some of the code so I also benchmarked a new WPF from the Visual Studio template where I added 20 empty classes like `public class CustomLabel1 : Label { }` and I added each control in the XAML of the main window. This PR shaves about 50-60 ms when starting the app.

I expect more complex apps with more non-known type in XAML to have even bigger performance gain. Projects that have a ton of non-known types are third-party control vendors (Like DevExpress, Telerik, etc) because they have a ton of custom controls.

## Customer Impact
Better startup and runtime performance.

## Regression
No.

## Testing
Local testing using sample app that hits this specific code.

## Risk
Low.

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/System.Xaml/Resources/ExceptionStringTable.txt
+++ b/src/Microsoft.DotNet.Wpf/src/System.Xaml/Resources/ExceptionStringTable.txt
@@ -249,7 +249,6 @@ DeferredLoad=Deferred load threw an exception.
 DeferredSave=Save of deferred-load content threw an exception.
 FactoryReturnedNull=The factory method '{0}' that matches the specified binding constraints returned null.
 ConstructorInvocation=The invocation of the constructor on type '{0}' that matches the specified binding constraints threw an exception.
-NoDefaultConstructor=No default constructor found for type '{0}'. You can use the Arguments or FactoryMethod directives to construct this type.
 NoConstructor=No matching constructor found on type '{0}'. You can use the Arguments or FactoryMethod directives to construct this type.
 DeferringLoaderInstanceNull=Cannot get a XamlDeferringLoader from XamlValueConverter '{0}' because its ConverterInstance property is null.
 TypeConverterFailed2=Failed to convert '{0}' to type '{1}'.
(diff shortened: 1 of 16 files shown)


ThomasGoulet73:
I rebased to fix the conflicts.

ThomasGoulet73:
I rebased to fix the conflicts.

ThomasGoulet73:

PR #9768, opened 2024-09-12
Optimize parsing and conversion in DurationConverter, reduce allocations

h3xds1nz:
## Description

Parsing optimization for `Duration` and its conversion from/to `string`. Conversion from is now alloc-free (besides the non-removable ones), conversion to is massively sped up by modification of `ToString` in `Duration`.
- Besides the obvious improvements, adds/completes code documentation for the whole class.
- The `TimeSpanConverter` from `ComponentModel` does nothing useful but `Parse`/`ToString` on `TimeSpan`/`string` instance; the exceptions are not localized. Using it means we have to use the string instance as ref structs cannot be boxed (`ReadOnlySpan<T>`), and that would result in additional trim for example.

### ConvertTo benchmarks
| Method   | source              | Mean [ns]  | Error [ns] | StdDev [ns] | Code Size | Gen0   | Allocated |
|--------- |-------------------- |-----------:|-----------:|------------:|----------:|-------:|----------:|
| Original | 17.22:10:15.4571230 | 449.078 ns |  1.3738 ns |   1.2178 ns |         107 B | 0.0134 |         224 B |
| PR__EDIT | 17.22:10:15.4571230 |  22.712 ns |  0.3210 ns |   0.3003 ns |         107 B | 0.0038 |          64 B |
| Original | Automatic           |   2.436 ns |  0.0249 ns |   0.0221 ns |         107 B |      - |             - |
| PR__EDIT | Automatic           |   2.183 ns |  0.0136 ns |   0.0114 ns |         107 B |      - |             - |

### ConvertFrom benchmarks
| Method   | source            | Mean [ns]  | Error [ns] | StdDev [ns] | Code Size | Gen0   | Allocated |
|--------- |------------------ |-----------:|-----------:|------------:|----------:|-------:|----------:|
| Original | \x207:12:14:45.3448  | 216.157 ns |  1.1200 ns |   0.9929 ns |       1,018 B | 0.0219 |         368 B |
| PR__EDIT | \x207:12:14:45.3448  | 207.656 ns |  0.8965 ns |   0.8386 ns |         570 B | 0.0138 |         232 B |
| Original | \x20Forever          |  10.444 ns |  0.2368 ns |   0.2326 ns |       1,289 B | 0.0043 |          72 B |
| PR__EDIT | \x20Forever          |   7.213 ns |  0.1097 ns |   0.1026 ns |         722 B | 0.0019 |          32 B |
| Original | Automatic\x20         |  19.617 ns |  0.1948 ns |   0.1822 ns |       1,920 B | 0.0043 |          72 B |
| PR__EDIT | Automatic\x20         |   7.794 ns |  0.1190 ns |   0.1055 ns |       1,132 B | 0.0019 |          32 B |
| Original | 6:12:14:45.3448   | 199.160 ns |  0.8665 ns |   0.7236 ns |         696 B | 0.0153 |         256 B |
| PR__EDIT | 6:12:14:45.3448   | 188.803 ns |  0.7456 ns |   0.6974 ns |         606 B | 0.0138 |         232 B |
| Original | Automatic         |   7.548 ns |  0.6944 ns |   2.0475 ns |         776 B | 0.0019 |          32 B |
| PR__EDIT | Automatic         |   5.561 ns |  0.1681 ns |   0.4955 ns |         900 B | 0.0019 |          32 B |
| Original | Forever           |   7.773 ns |  0.5075 ns |   1.4964 ns |         611 B | 0.0019 |          32 B |
| PR__EDIT | Forever           |   7.246 ns |  0.4819 ns |   1.4208 ns |         520 B | 0.0019 |          32 B |

## Customer Impact

## Regression

No.

## Testing

Local build, CI, extensive assert testing:
```csharp
 CultureInfo russianCulture = new CultureInfo("ru-RU");
 //ConvertFrom Tests
 Duration x1 = (Duration)oldDurationConverter.ConvertFrom(null, null, "Automatic");
 Duration x2 = (Duration)newDurationConverter.ConvertFrom(null, null, "Automatic");
 
 Assert.AreEqual(x1, x2);
 x1 = (Duration)oldDurationConverter.ConvertFrom(null, null, " Forever");
 x2 = (Duration)newDurationConverter.ConvertFrom(null, null, " Forever");
 Assert.AreEqual(x1, x2);
 x1 = (Duration)oldDurationConverter.ConvertFrom(null, null, "Automatic ");
 x2 = (Duration)newDurationConverter.ConvertFrom(null, null, "Automatic ");
 Assert.AreEqual(x1, x2);
 x1 = (Duration)oldDurationConverter.ConvertFrom(null, CultureInfo.InvariantCulture, "07:12:14:45.3448 ");
 x2 = (Duration)newDurationConverter.ConvertFrom(null, CultureInfo.InvariantCulture, "07:12:14:45.3448 ");
 Assert.AreEqual(x1, x2);
 x1 = (Duration)oldDurationConverter.ConvertFrom(null, russianCulture, " 6:12:14:45,3448 ");
 x2 = (Duration)newDurationConverter.ConvertFrom(null, russianCulture, " 6:12:14:45,3448 ");
 Assert.AreEqual(x1, x2);
 x1 = (Duration)oldDurationConverter.ConvertFrom(null, CultureInfo.InvariantCulture, " 6 ");
 x2 = (Duration)newDurationConverter.ConvertFrom(null, CultureInfo.InvariantCulture, " 6 ");
 Assert.AreEqual(x1, x2);
 x1 = (Duration)oldDurationConverter.ConvertFrom(null, CultureInfo.InvariantCulture, " 6:12 ");
 x2 = (Duration)newDurationConverter.ConvertFrom(null, CultureInfo.InvariantCulture, " 6:12 ");
 Assert.AreEqual(x1, x2);
 // Invalid TimeSpan string specified
 Assert.ThrowsException<FormatException>(() => oldDurationConverter.ConvertFrom(null, null, " á "));
 Assert.ThrowsException<FormatException>(() => newDurationConverter.ConvertFrom(null, null, " á "));
 // Valid TimeSpan but wrong culture
 Assert.ThrowsException<FormatException>(() => oldDurationConverter.ConvertFrom(null, CultureInfo.InvariantCulture, " 6:12:14:45,3448 "));
 Assert.ThrowsException<FormatException>(() => newDurationConverter.ConvertFrom(null, CultureInfo.InvariantCulture, " 6:12:14:45,3448 "));
 // Wrong input type
 Assert.ThrowsException<NotSupportedException>(() => oldDurationConverter.ConvertFrom(null, null, 12345));
 Assert.ThrowsException<NotSupportedException>(() => newDurationConverter.ConvertFrom(null, null, 12345));
 // Input was null
 Assert.ThrowsException<NotSupportedException>(() => oldDurationConverter.ConvertFrom(null, null, null));
 Assert.ThrowsException<NotSupportedException>(() => newDurationConverter.ConvertFrom(null, null, null));

 // CanConvertTo Tests
 // True cases
 bool x6 = oldDurationConverter.CanConvertTo(null, typeof(InstanceDescriptor));
 bool x7 = newDurationConverter.CanConvertTo(null, typeof(InstanceDescriptor));
 Assert.AreEqual(x6, x7);
 x6 = oldDurationConverter.CanConvertTo(null, typeof(string));
 x7 = newDurationConverter.CanConvertTo(null, typeof(string));
 Assert.AreEqual(x6, x7);
 // False case
 x6 = oldDurationConverter.CanConvertTo(null, typeof(Int32));
 x7 = newDurationConverter.CanConvertTo(null, typeof(Int32));
 Assert.AreEqual(x6, x7);

 // ConvertTo Tests
 string x3 = (string)oldDurationConverter.ConvertTo(null, null, Duration.Automatic, typeof(string));
 string x4 = (string)newDurationConverter.ConvertTo(null, null, Duration.Automatic, typeof(string));
 Assert.AreEqual(x3, x4);

 x3 = (string)oldDurationConverter.ConvertTo(null, null, Duration.Forever, typeof(string));
 x4 = (string)newDurationConverter.ConvertTo(null, null, Duration.Forever, typeof(string));
 Assert.AreEqual(x3, x4);

 //It converts according to InvariantCulture everytime
 x3 = (string)oldDurationConverter.ConvertTo(null, CultureInfo.InvariantCulture, new Duration(new TimeSpan(17, 22, 10, 15, 457, 123)), typeof(string));
 x4 = (string)newDurationConverter.ConvertTo(null, CultureInfo.InvariantCulture, new Duration(new TimeSpan(17, 22, 10, 15, 457, 123)), typeof(string));
 string x5 = (string)oldDurationConverter.ConvertTo(null, russianCulture, new Duration(new TimeSpan(17, 22, 10, 15, 457, 123)), typeof(string));
 string x9 = (string)newDurationConverter.ConvertTo(null, russianCulture, new Duration(new TimeSpan(17, 22, 10, 15, 457, 123)), typeof(string));
 //Sanity test
 Assert.AreEqual(x3, x4);
 Assert.AreEqual(x5, x9);
 //Cross test
 Assert.AreEqual(x3, x9);
 Assert.AreEqual(x4, x5);

 //This is a special case, because original call was a base call, it does not throw for NULL but returns string.Empty
 x3 = (string)oldDurationConverter.ConvertTo(null, null, null, typeof(string));
 x4 = (string)newDurationConverter.ConvertTo(null, null, null, typeof(string));
 Assert.AreEqual(x3, x4);

 //Null destinationType
 Assert.ThrowsException<ArgumentNullException>(() => (string)oldDurationConverter.ConvertTo(null, null, new Duration(new TimeSpan(17, 22, 10, 15, 457, 123)), null));
 Assert.ThrowsException<ArgumentNullException>(() => (string)newDurationConverter.ConvertTo(null, null, new Duration(new TimeSpan(17, 22, 10, 15, 457, 123)), null));
 //Invalid destinationType
 Assert.ThrowsException<NotSupportedException>(() => (string)oldDurationConverter.ConvertTo(null, null, new Duration(new TimeSpan(17, 22, 10, 15, 457, 123)), typeof(Int32)));
 Assert.ThrowsException<NotSupportedException>(() => (string)newDurationConverter.ConvertTo(null, null, new Duration(new TimeSpan(17, 22, 10, 15, 457, 123)), typeof(Int32)));
```

## Risk

Low, the behaviour has been matched as verified by the tests.

**There is a slight behavioral change in `ConvertTo` as in my opinion that is a bug**, since the original function delegates this to `Duration.ToString` override, which obtains a `TypeDescriptor` for `TimeSpan`, that in normal scenario is `TimeSpanConverter` but it is possible to change it and therefore change the output of the `DurationConverter` for `TimeSpan` values.
This doesn't apply onto `ConvertFrom` as a raw instance of `TimeSpanConverter` was used instead, originally.

The new behavior is to always use `TimeSpan.ToString` override, so the `ConvertTo`/`ConvertFrom` have consistent behavior. This is also consistent with other related `Timeline` types - e.g. `RepeatBehavior`, that just uses `TimeSpan.ToString` override. If anything, one shall directly change the `TypeDescriptor` for `Duration` to change the whole conversion logic.

Sample code would influence the output is `ConvertTo` (only, `ConvertFrom` would stay intact) [courtesy of @miloush]:

```csharp
AttributeCollection xx = TypeDescriptor.GetAttributes(typeof(TimeSpan));
TypeDescriptor.AddAttributes(typeof(TimeSpan), new TypeConverterAttribute(typeof(HahaConverter)));
string s = (string)oldDurationConverter.ConvertTo(null, null, 
                  new Duration(new TimeSpan(17, 22, 10, 15, 457, 123)), typeof(string));

class HahaConverter : TypeConverter
{
    public override object ConvertTo(ITypeDescriptorContext context,
    CultureInfo culture, object value, Type destinationType) => "haha";
 }
```

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/wpf/pull/9768)

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Duration.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Duration.cs
@@ -483,7 +483,7 @@ namespace System.Windows
         {
             if (HasTimeSpan)
             {
-                return TypeDescriptor.GetConverter(_timeSpan).ConvertToString(_timeSpan);
+                return _timeSpan.ToString();
             }
             else if (_durationType == DurationType.Forever)
             {
--- a/src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Duration.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Duration.cs
@@ -481,18 +481,22 @@ namespace System.Windows
         /// <returns>A string representation of this Duration.</returns>
         public override string ToString()
         {
-            if (HasTimeSpan)
-            {
-                return _timeSpan.ToString();
-            }
-            else if (_durationType == DurationType.Forever)
-            {
+            return HasTimeSpan ? TypeDescriptor.GetConverter(_timeSpan).ConvertToString(_timeSpan) : ToStringInvariant();
+        }
+
+        /// <summary>
+        /// This method does not use the current <see cref="System.TimeSpan"/>'s <see cref="TypeDescriptor"/> for retrieving the
+        /// current converter, but rather uses the standard <see cref="TimeSpan.ToString()"/> override for <see cref="DurationConverter"/> needs.
+        /// </summary>
+        /// <returns>A culture-invariant representation of the <see cref="Duration"/> instance.</returns>
+        internal string ToStringInvariant()
+        {
+            if (_durationType == DurationType.Forever)
                 return "Forever";
-            }
-            else // IsAutomatic
-            {
+            else if (_durationType == DurationType.Automatic)
                 return "Automatic";
-            }
+
+            return _timeSpan.ToString();
         }
 
         #endregion
--- a/src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/DurationConverter.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/DurationConverter.cs
@@ -41,34 +41,43 @@ namespace System.Windows
         /// ConvertFrom
         /// </summary>
         /// <ExternalAPI/>
-        public override object ConvertFrom(
-            ITypeDescriptorContext td, 
-            CultureInfo cultureInfo, 
-            object value)
+        public override object ConvertFrom(ITypeDescriptorContext td, CultureInfo cultureInfo, object value)
         {
-            string stringValue = value as string;
+            // In case value is not a string, we can try to check InstanceDescriptor or we just throw NotSupportedException
+            if (value is not string stringValue)
+                return base.ConvertFrom(td, cultureInfo, value);
 
-            // Override the converter for sentinel values
-            if (stringValue != null)
-            {
-                stringValue = stringValue.Trim();
-                if (stringValue == "Automatic")
-                {
-                    return Duration.Automatic;
-                }
-                else if (stringValue == "Forever")
-                {
-                    return Duration.Forever;
-                }
-            }
+            // Sanitize the input
+            ReadOnlySpan<char> valueSpan = stringValue.AsSpan().Trim();
 
-            TimeSpan duration = TimeSpan.Zero;
-            if(_timeSpanConverter == null)
+            // In case it is not a pre-defined value, we will try to parse the TimeSpan and if it throws,
+            // we will catch it as TimeSpanConverter does and rethrow with the inner exception information.
+            if (valueSpan.Equals("Automatic", StringComparison.Ordinal))
+                return Duration.Automatic;
+            else if (valueSpan.Equals("Forever", StringComparison.Ordinal))
+                return Duration.Forever;
+            else
+                return ParseTimeSpan(valueSpan, cultureInfo);
+        }
+
+        /// <summary>
+        /// Faciliaties parsing from <paramref name="valueSpan"/> to <see cref="TimeSpan"/> and initializes new <see cref="Duration"/> instance.
+        /// </summary>
+        /// <param name="valueSpan">The string to convert from.</param>
+        /// <param name="cultureInfo">The culture specifier to use.</param>
+        /// <returns>A newly initialized <see cref="Duration"/> instance from the <paramref name="valueSpan"/> string.</returns>
+        /// <remarks>This function is decoupled from the <see cref="ConvertFrom(ITypeDescriptorContext, CultureInfo, object)"/> for performance reasons.</remarks>
+        /// <exception cref="FormatException">Thrown when parsing of <paramref name="valueSpan"/> to <see cref="TimeSpan"/> instance fails.</exception>
+        private static Duration ParseTimeSpan(ReadOnlySpan<char> valueSpan, CultureInfo cultureInfo)
+        {
+            try
             {
-                 _timeSpanConverter = new TimeSpanConverter();
+                return new Duration(TimeSpan.Parse(valueSpan, cultureInfo));
             }
-            duration = (TimeSpan)_timeSpanConverter.ConvertFrom(td, cultureInfo, value);
-            return new Duration(duration);
+            catch (FormatException e)
+            {
+                throw new FormatException($"{valueSpan} is not a valid value for {nameof(TimeSpan)}.", e);
+            }
         }
 
         /// <summary>
@@ -124,6 +133,5 @@ namespace System.Windows
             // Pass unhandled cases to base class (which will throw exceptions for null value or destinationType.)
             return base.ConvertTo(context, cultureInfo, value, destinationType);
         }
-        private static TimeSpanConverter _timeSpanConverter;
     }
 }
--- a/src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/DurationConverter.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/DurationConverter.cs
@@ -4,43 +4,48 @@
 
 //
 
-using System.ComponentModel;
 using System.ComponentModel.Design.Serialization;
+using System.ComponentModel;
 using System.Globalization;
+using System.Diagnostics;
 using System.Reflection;
 
 namespace System.Windows
 {
     /// <summary>
-    /// Provides a type converter to convert Duration to and from other representations.
+    /// Provides a type converter to convert from <see cref="Duration"/> to <see langword="string"/> and vice versa.
     /// </summary>
     public class DurationConverter : TypeConverter
     {
         /// <summary>
-        /// CanConvertFrom - Returns whether or not this class can convert from a given type
+        /// Returns whether this class can convert specific <see cref="Type"/> into <see cref="Duration"/>.
         /// </summary>
-        /// <ExternalAPI/>
+        /// <param name="td">Context information used for conversion.</param>
+        /// <param name="t">Type being evaluated for conversion.</param>
+        /// <returns><see langword="true"/> if the given <paramref name="sourceType"/> can be converted from, <see langword="false"/> otherwise.</returns>
         public override bool CanConvertFrom(ITypeDescriptorContext td, Type t)
         {
             return t == typeof(string);
         }
 
         /// <summary>
-        /// TypeConverter method override.
+        /// Returns whether this class can convert specified value to <see langword="string"/>.
         /// </summary>
-        /// <param name="context">ITypeDescriptorContext</param>
-        /// <param name="destinationType">Type to convert to</param>
-        /// <returns>true if conversion is possible</returns>
-        /// <ExternalAPI/>
+        /// <param name="context">Context information used for conversion.</param>
+        /// <param name="destinationType">Type being evaluated for conversion.</param>
+        /// <returns><see langword="true"/> if conversion to <see langword="string"/> is possible, <see langword="false"/> otherwise.</returns>
         public override bool CanConvertTo(ITypeDescriptorContext context, Type destinationType)
         {
             return destinationType == typeof(InstanceDescriptor) || destinationType == typeof(string);
         }
 
         /// <summary>
-        /// ConvertFrom
+        /// Converts <paramref name="value"/> of <see langword="string"/> type to its <see cref="Duration"/> represensation.
         /// </summary>
-        /// <ExternalAPI/>
+        /// <param name="td">Context information used for conversion.</param>
+        /// <param name="cultureInfo">The culture specifier to use.</param>
+        /// <param name="value">The string to convert from.</param>
+        /// <returns>A <see cref="Duration"/> representing the <see langword="string"/> specified by <paramref name="value"/>.</returns>
         public override object ConvertFrom(ITypeDescriptorContext td, CultureInfo cultureInfo, object value)
         {
             // In case value is not a string, we can try to check InstanceDescriptor or we just throw NotSupportedException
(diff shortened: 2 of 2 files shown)


src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Duration.cs
```diff
@@ -512,7 +512,7 @@ public override string ToString()
         {
             if (HasTimeSpan)
             {
-                return TypeDescriptor.GetConverter(_timeSpan).ConvertToString(_timeSpan);
+                return _timeSpan.ToString();
```
miloush, line 515: This obviously breaks any custom descriptors for `TimeSpan`.

```C#
class HahaConverter : TypeConverter
{
    public override object ConvertTo(ITypeDescriptorContext context, CultureInfo culture, object value, Type destinationType) => "haha";
}

{
    TypeDescriptor.AddAttributes(typeof(TimeSpan), new TypeConverterAttribute(typeof(HahaConverter)));
    string s = new Duration(default).ToString(); // returns haha before PR but not after PR
}
```

and look:

https://github.com/thomhurst/ReadableTimeSpan/blob/7eb0c92ebf4132010dbe46b84bf4db8d7dad5b58/TomLonghurst.ReadableTimeSpan/ReadableTimeSpan.cs#L32

https://github.com/Splamy/Sybtitle/blob/ba306e0228c6215246a312ff14dd7e266bdeec49/Sybtitle/HumanTimeSpan.cs#L185-L192

h3xds1nz: It seemed like an oversight to me, I was checking other related `Timeline` types like `RepeatBehavior`, and there it calls regular `ToString` on `TimeSpan`. But it is a behavioral change I didn't consider, indeed.

In my own opinion, I wouldn't want the behavior of `Duration`'s `ToString` change based on a `TimeSpan`'s current converter set via attributes; since `ConvertFrom` won't change accordingly.

My bigger problem is that we're doing always conversion on current culture tbh, there's so much inconsistencies it's killing me with the converters.

Whether it's exceptions throwing, sometimes you throw in `CanConvert`* methods because of bugs (e.g. casts to types).

h3xds1nz: Regarding my culture problem: I think I remember why, `TimeSpan` got globalized later on afaik, maybe `net4`?

So the support was then never added to the converters most likely.

miloush: Yeah obviously the `ConvertFrom` and `ConvertTo` should be consistent. I would think that is a bug in the converter more than anything else. It's reasonable for people to just use `Duration` for non-negative timestamps in general, even in non-WPF applications relying on `ToString()`.

I am not that confident on it being an oversight, you need extra effort and knowledge to write this. Maybe it was needed for the designer?

Unlike numbers and dates, the `TimeSpan` is not culturally sensitive unless you explicitly ask for it, it uses the constant invariant format.

I agree that a better consistency would be desirable across converters.

h3xds1nz: Thank you by the way, I love your reviews, you catch some great things everytime.

In your opinion then, we would revert the change on `Duration`'s `ToString` and handle it independently in the Converter? I think one internal method can solve this without breaking anything. We can maybe follow-up the `ToString` in another issue.

I was going through the WPF tests and none of them regarding `Duration` counts on this.

Yeah, I just went through the docs to refresh my memory on `TimeSpan` and was correct indeed, the culture-sensitivity didn't arrive until `net4`.

Based on that, I went to check the old .NET 3.5 libraries, and yeah, the culture overload gets added in `net4` and subsequently used in `TimeSpanConverter`'s `ConvertFrom` but obviously all the WPF code is left untouched, indirectly swallowing this change because it was passing the `cultureInfo` there. So that's how `ConvertFrom` became sensitive to culture passed in but `ConvertTo` is always culture-insensitive. So yeah this feels like a bug.

Yeah I was hoping later on to summarize the differences in one issue between the converters but first I have to get through them, write tests for them to catch all edge cases and get some output out of this, then maybe some action.

miloush: It wasn't a breaking change, right? Without specifying a format, converting to string still uses the invariant format regardless of culture, and parsing still supports the invariant one but also became tolerant to a localized format.

Yeah I would think reverse the `Duration.ToString()` for now, especially since you are selling this as an optimization in `DurationConverter`, and we should probably use `TimeSpanConverter` in the converter both ways for consistency. That would also save you the type descriptor lookup on every call.

h3xds1nz: Well partially it is, because let's say you've been passing to `DurationConverter`/`TimeSpanConverter` a string with, let's say "07:12:14:45.3448" (that could have been generated via `ToString()`) and `CurrentCulture` (or `null`, that's more likely), the tokenizer in `TimeSpan` will call https://learn.microsoft.com/en-us/dotnet/api/system.globalization.datetimeformatinfo.getinstance?view=net-8.0 under the hood.

Pre-`net4` this would work, as it was ignored obviously since you weren't passing the `CultureInfo` beyond `TimeSpanConverter`, but when it swapped from `Parse(string s)` overload in `net4` to `Parse(string input, IFormatProvider? formatProvider)` and the culture is recognized, this blows on systems with different decimal separator than `.` for example.

Essentially `ConvertTo`/`ConvertFrom` gives you incompatible values (throws) since then if you need cultural sensitivity or you were counting on it pre-`net4`. Sure there was a compat switch but I'm not sure whether it covered the converter case.

h3xds1nz: Yes, because you're using `var`. Just kidding.

```csharp
//This is the original behaviour pre-net 4.0 of TimeSpan converter, since it was the only existent overload
//Therefore passing any CultureInfo had no effect
CultureInfo czechCulture = new CultureInfo("cs-CZ");
TimeSpan timeStamp =  TimeSpan.Parse("07:12:14:45.3448");
string xx =  timeStamp.ToString();

//This is post net4 behaviour, when the overload in TimeSpanConverter was changed to this one
//So any null passed in to the converter, or specific CultureInfo with different separators etc. 
//ends up with a beautiful exception
Thread.CurrentThread.CurrentCulture = czechCulture;
timeStamp = TimeSpan.Parse("07:12:14:45.3448", null);
xx = timeStamp.ToString();
```

miloush: You might be getting exception because your string is invalid, cs-CZ uses decimal comma... 

```C#
Thread.CurrentThread.CurrentCulture = CultureInfo.GetCultureInfo("cs-CZ");
string sInv = "07.12:14:45.3448";
string sLoc = "07:12:14:45,3448";
var t1 = TimeSpan.Parse(sInv, null);
var t2 = TimeSpan.Parse(sLoc, null);
var var = t1.ToString(); // 7.12:14:45.3448000
```

doesn't throw either...?

h3xds1nz: My string will parse as `InvariantCulture` passed in for example; but I'm not sure whether you still get my point.

`DurationConverter`/`TimeSpanConverter` pre-net4 uses (there's nothing else to use but you're still passing `CultureInfo` to the converter, e.g. `null`): `TimeSpan.Parse("07:12:14:45.3448");`

`DurationConverter`/`TimeSpanConverter` post-net4 uses the new overload (now your `CultureInfo` starts to matter, even if it was null): `TimeSpan.Parse("07:12:14:45.3448", cultureInfo);`

And that's a behavioral change, because now the string started to throw on different cultures instead of being converted.

miloush: Well either we are talking each about something else or one of us is wrong. I am claiming that if your string is a valid `TimeSpan` in invariant/pre-net4 format, then both `TimeSpan.Parse(s)` and `TimeSpan.Parse(s, cultureInfo)` behave the same for any culture provided, i.e. they always succeed, no new exceptions are thrown. The parser always tries the invariant/pre-net4 format, as I linked in the code. What changed is that now strings that were not valid before will get parsed if they match the format in the culture supplied.

I am also happy to cease the discussion because it doesn't matter very much here. :)

h3xds1nz: Yeah let's not waste anymore time on this haha; anyways here you go and I hope you will understand now the edge-case.

[Sharplab demo](https://sharplab.io/#v2:EYLgxg9gTgpgtADwGwBYA0AXEUCuA7AHwAEAmARgFgAoIgBgAIiyA6AcQBsJgBDdgSwBe3DHwh4A3NTqMWAFQAWsbgBM+eAOaSaAZkYl6AYXoBvavXMykjFPQCy3NQAoAlCbMWPAZQxQ16+vwYMFC89AC89ABEtADsIGQk8SggKACszNooKAAckVoeHu4F9AD0JQDqMADksPTcALbBfGDceADORQUKSsrMBjhQsHgY3TAqfQNDGP3sGAMw4fR4MADuhjiz8wCSeABmEI6RMHhwAKqekc75xfSdHmVMMY1Q9MoQMG30GPJ8HVQ3PgAnsZ6AAFWBBPDKLb1AxiABuwQwADEoBB6o5AsFeGglqt1ptYDt9ocwG04AYAFqXXG7XhtGBXegAXzuFhaGDA8nojmR0HqwgAoggwDAAA4iMSuYxMACch2WiJe8m4YrFxzal3ErP+BSBJjBEOO0NhCKRqPRmL4QRC7FxyzWMzmRL2B0iZIp1OcuJ8OEZ4hZbPMHK5PL5UAFGGFoolojw0rlhxWMHYkEaXwgeIwKC1Oo8eYsnSYVlkfEanjFrUNMEhJrheCVKLRGKYDBE5creG8vg0uKd21d9DAG2dMGJEFxwAgEHY9BwDIAcjWUM5OqZdcU+LsefOYEvs6uNwV1zcCkQYvRSx3WsxQdwoAzHO2YBXWt2/Ljh4Sx66mcUg4GR4eCmDIASep4eOel5li+na3vej7Pq+XY+H4VwAQW5g6phQA===)

[SharpLab demo 2](https://sharplab.io/#v2:EYLgxg9gTgpgtADwGwBYA0AXEUCuA7AHwAEAmARgFgAoIgBgAIiyA6AcQBsJgBDdgSwBe3DHwh4A3NTqMWAFQAWsbgBM+eAOaSaAZkYl6AYXoBvavXMykjFPQCy3NQAoAlCbMWPAZQxQ16+vwYMFC89AC89ABEtADsIGQk8SggKACsaNooKAAckVoeHu4F9AD0JQDqMADksPTcALbBfGDceADORQUKSsrMBjhQsHgY3TAqfQNDGP3sGAMw4fR4MADuhjiz8wCSeABmEI6RYG1wBgBakc75xfSdNwVlTDGNUPTKEDBt9BjyfB1UNx8AE9jPQAAqwIJ4ZRbeoGMQAN2CGAAYlAIPVHIFgrw0OtNrAdvtmDsEdxfK1phs5rA8bteG0YFd6ABfO4WFoYMDyeiOFHQerCACiCDAMAADiIxK5jEwAJyHZZI17ybji8UwdqXcRsgEFYEmcGQzUwuGI5FojFYvhBELsPEzGkwIkQEl4MkU4aO+Z4nw4JniVns8yc7m8/lQQUYEViyWiPAy+WHFYwdiQRrfCBLGAYFDa3UeAsWTpMKyyPiNTzi1pGnMm2Hw90W9GYpgMESV6t4by+DQO6nbPZZsADwlDvHACAQdj0HCMgByOZQzk6pj1xT4u15c5gi9zK/XBTX9w8RBi9HLndazDB5MZjg7MCrrR7fjxI4JzqHzOKwaDh48VNGT/Y8TwsM8LwrJ8uxvO8YAfKDn27Hw/CuP8i3MXUMKAA)

Kuldeep-MS (dismissed):

Kuldeep-MS:
@h3xds1nz - Could you please resolve the conflicts?

h3xds1nz:
@Kuldeep-MS Done :)

Kuldeep-MS (approved):

PR #9769, opened 2024-09-12
Redefine Duration as readonly struct to prevent defensive copies

h3xds1nz:
## Description

Since `Duration` is immutable by design, we should adjust it to `readonly struct`. This will also prevent defensive copies, even though the struct is small, so it doesn't really matter. It will also improve code quality. **This is not a breaking change**.

- A private constructor for the static property creation will be added.
- Some code cleanup included.

## Customer Impact

Possibly a bit of perf and a clear directive that `Duration` and all its methods are not adjusting the state of the `struct`.

## Regression

No.

## Testing

Local build, testing with #9768.

## Risk

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/wpf/pull/9769)


miloush:
LGTM

dipeshmsft (dismissed):

dipeshmsft (approved):

himgoyalmicro (approved):

dipeshmsft:
Thanks @h3xds1nz for this contribution. We discussed with the API review board about these API changes and the guidance we got is that we don't need to go for API review process for this, but we shall be careful when making structs readonly as once marked readonly we won't be able to revert back, as it is with other API changes.

h3xds1nz:
@dipeshmsft Thank you.

Yeah I understand, it should always be considered whether the struct represents an immutable state or it may be subject to change even if that's in the future. I'll pick a few more that are carrying constant information as I had them in mind but wanted to wait until you have time to get answers from the board.

P.S. If you outline the documentation changes needed I can submit it next time so you don't have to burden with it.

PR #9817, opened 2024-09-19
Convert static float variables in FloatUtil into compile-time constants

h3xds1nz:
## Description

Converts three static variables into compile-time constants as there's no reason they shall be static variables in the first place. This will allow us to save a few precious bytes of memory. `IsOne`/`IsZero` are unused right now but I'll keep them in.

## Customer Impact

Decreased static allocations by a few bytes and tiny bit of perf!

## Regression

No.

## Testing

Local build.

## Risk

None.
_Besides if someone was using reflection to change them — miloush doesn't like when I say "None", haha._

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/wpf/pull/9817)

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/Shared/MS/Internal/FloatUtil.cs
+++ b/src/Microsoft.DotNet.Wpf/src/Shared/MS/Internal/FloatUtil.cs
@@ -20,9 +20,9 @@ namespace MS.Internal
 {
     internal static class FloatUtil
     {
-        internal const float FLT_EPSILON = 1.192092896E-07f;
-        internal const float FLT_MAX_PRECISION = 0x00FFFFFF;
-        internal const float INVERSE_FLT_MAX_PRECISION = 1.0f / FLT_MAX_PRECISION;
+        private const float FLT_EPSILON = 1.192092896E-07f;
+        private const float FLT_MAX_PRECISION = 0x00FFFFFF;
+        private const float INVERSE_FLT_MAX_PRECISION = 1.0f / FLT_MAX_PRECISION;
 
         /// <summary>
         /// AreClose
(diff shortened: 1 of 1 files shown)


miloush:
We had a discussion in the past on static vs const, I can't quickly find it, does anybody remember? I believe there was an argument to keep static, possibly because it could and was being modified using reflection. We should write these decisions down somewhere in guidelines.

This obviously doesn't apply here since these are "factual" constants, but why do they need to be internal (as opposed to private)?  I can't see any other class using them.

Either way LGTM.

h3xds1nz:
They could be private, I'll check and adjust accordingly.

dipeshmsft:
> but why do they need to be internal (as opposed to private)?


@miloush , I think it was internal so as, if any other class would require those constants in future, be it for any operation / check / assignment, they can access it, without redining the const.

h3xds1nz:
@dipeshmsft yet when I checked it upon miloush's comment; https://github.com/dotnet/wpf/blob/188112f9f8c86f35b6805ef1b5154446e3ef9292/src/Microsoft.DotNet.Wpf/src/System.Windows.Input.Manipulations/System/Windows/Input/Manipulations/ManipulationSequence.cs#L772

And note this `FloatUtil` is in the shared folder, so you would expect the projects to just link it, haha.

dipeshmsft:
Given this, won't it be better to make the const's internal here and use it in the above file that you mentioned, and remove the const defined there? This is just a question at this moment. It's probably fine either way because it isn't a maintenance overhead as we won't be changing these values.

h3xds1nz:
@dipeshmsft I'll leave this one up to you.

While I'm generally in for sharing the utility classes, WPF codebase is known to be especially schizophrenic (oh this wouldn't pass Policheck review) to not bring 1 small utility class into a library unless it absolutely needs to, would even rather extract a piece of code from one of the main NetFX libraries even though that lib is 99% gonna be loaded anyways by one of core ones).

This code snippet is from the separate System.Windows.Input.Manipulations library. Which even was its own `Lazy<T>` impl again, oh that's gonna go away.

harshit7962 (approved):
LGTM

harshit7962:
Thank you @h3xds1nz for the contribution.

h3xds1nz:
@harshit7962 Thanks!

PR #9826, opened 2024-09-20
Fixing Fluent style for CheckBox, Calendar and text input controls

dipeshmsft:
### Description

This PR fixes the following issues in Fluent styles :

- Fixes ListViewItem keys duplication
- Fixes bindings in CheckBox
- Fixes inconsistent border brushes for TextBox, ComboBox and other controls
- Fixes Height and Width being ignored when HorizontalAlignment and VerticalAlignment are set on DatePicker, Calendar and TextBox controls

PS : Bugs were found as a result of local testing with applications. 

### Customer Impact
Developers who will want to use the new Fluent styles will be able to use the controls correctly, and they will need less customization in their already existing apps to use the new theme. Brings more parity with WinUI styles and decreases few inconsistencies in the styles.

### Regression
Technically no, but any developer who will try to port their application to use the new styles will face these issue and the properties won't work for them. For few of these issues only way for them would be to derive from Fluent styles and write new styles with ControlTemplates

### Testing
Tested with different configurations for controls in sample and WPF Gallery application

### Risk
Minimal. No one is using the styles currently.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/wpf/pull/9826)

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/Themes/PresentationFramework.Fluent/Resources/Theme/Dark.xaml
+++ b/src/Microsoft.DotNet.Wpf/src/Themes/PresentationFramework.Fluent/Resources/Theme/Dark.xaml
@@ -361,21 +361,49 @@
     <SolidColorBrush x:Key="CardFooterBackground" Color="{StaticResource CardBackgroundFillColorSecondary}" />
 
     <!--  CheckBox  -->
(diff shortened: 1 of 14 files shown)


dipeshmsft:
Servicing approved on the mail thread

singhashish-wpf (approved):

pchaurasia14 (approved):

PR #9838, opened 2024-09-24
Remove dead code in FontSourceCollection/FontSource (+ hollow CAS remnants)

h3xds1nz:
## Description

Removes `skipDemand` from `FontSource` as it is not used for anything anymore, it was used in `NetFx` for figuring out whether to demand permissions.

Also removes `isWindowsFonts` from `FontSourceCollection`, because this is already a dead code in latest `NetFx` as the internal constructor from `FamilyCollection` that allowed `true` has been removed at the same time it stopped deriving from `IFontCacheElement` and it has been replaced by `FamilyCollection.FromWindowsFonts` which uses DirectWrite factory, so it never reaches `FontSourceCollection` on this code path. It is also way faster if you would run it.

## Customer Impact

Smaller code size of `PresentationCore`, cleaner code size for developers, faster JITTing, decreased allocs.

## Regression

No.

## Testing

Local build, app run, work with custom font sources.

## Risk

Low, as it is a dead code. I do not believe anything relies on the original shape of the ctors either.

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/wpf/pull/9838)


himgoyalmicro:
Thank you @h3xds1nz for your contribution 😄

h3xds1nz:
@himgoyalmicro Thank you for including it :)

PR #9849, opened 2024-09-27
Live filtering race condition fix

### LINKED ISSUE #9454: CollectionViewSource live filtering can hide item(s) when adding items and quickly changing their filter status afterwards.
### Description

A user of our software noticed that sometimes when opening a page in our WPF application that an item was missing from a `DataGrid`.
Since this DataGrid needs to be sorted and show either all items, or items that require attention based on a toggle, we used a `CollectionViewSource` like so:
```
<CollectionViewSource
    x:Key="ItemsFilterAndSorter"
    Filter="FilteringEvent"
    IsLiveFilteringRequested="True"
    Source="{Binding Items}">
    <CollectionViewSource.SortDescriptions>
        <componentModel:SortDescription PropertyName="Name" />
    </CollectionViewSource.SortDescriptions>
    <CollectionViewSource.LiveFilteringProperties>
        <system:String>IsVisibleThroughFilter</system:String>
    </CollectionViewSource.LiveFilteringProperties>
</CollectionViewSource>
``` 

The Filter is an event that is defined like so:
```c#
private void FilteringEvent(object sender, FilterEventArgs e)
{
    e.Accepted = (e.Item as IFilterable)?.IsVisibleThroughFilter == true;
}
``` 

Then upon opening the page, the application waits for an network reply asynchronously and fills an `ObservableCollection<T>` from a secondary thread. For that we do use the `BindingOperations.EnableCollectionSynchronization(Items, sync)` system to let WPF know it's asynchronous.

Then our items add/update function can be boiled down to this:
```c#
for (int i = 0; i < 100; i++)
{
    lock(sync)
    {
        Items.Add(new Item(i));
    }
}

foreach (var item in Items)
{
    item.IsVisibleThroughFilter = true;
}
``` 

By default the `IsVisibleThroughFilter` is false, and is set to true after adding.
What I expect, is that the binding to `IsVisibleThroughFilter` happens after the initial check that is done in the `FilteringEvent`. And because of that there's a small gap in time that allows an Item to be added that is not visible at start, then marked as visible and then subscribed to changes of the `IsVisibleThroughFilter` property, making the `CollectionViewSource` not update the actual view.

So, I'm not sure I'm using `CollectionViewSource` as intended, but the documentation also doesn't seem to forbid what I'm doing.

### Reproduction Steps

Using the code from the description you can build a reproduction, but it takes time as the time gap at which this can happen seems to be quite small. Our product can run quite heavily so the `Dispatcher` is more occupied than with the code samples I've given.

### Expected behavior

I expect the `CollectionViewSource` to subscribe to the property before taking the initial value of the filter, ensuring all items are displayed as requested.

### Actual behavior

There's a small chance that item(s) are not displayed, until the `IsVisibleThroughFilter` property is changed.

### Regression?

I'm not sure, for now our product is stuck on .NET 4.7.1

### Known Workarounds

- Set the `IsVisibleThroughFilter` property correctly before adding to the `ObservableCollection<T>`.
In our case this works fine because other updates take a while, but milage may vary. If the user quickly toggles the filter manually theoretically it could also occur.

- Change the `IsVisibleThroughFilter` property

### Impact

The users of our software can't fully know if data in the list is displayed accurately.

### Configuration

- .NET 4.7.1
- Windows 11 Pro - Version 22H2 (OS Build 22621.3880)
- x64
- AFAIK not specific to my configuration

### Other information

_No response_

miloush:
Fixes #9454

## Description

When items are added to a filtering collection, changing a live filtering property too fast might result in the collection view becoming out of sync of the data.

This is because the subscription to changes in the list happens after items are added to a private copy of the list, providing a window between the copy and the subscription in which changes can occur unnoticed. This PR fixes it by moving the subscription before creating the copy.

## Customer Impact

By not taking this fix, customers cannot rely on data being correctly filtered during live shaping.

## Regression

No.

## Testing

Tested with provided repro project at https://github.com/VanderlandeWouterPost/CollectionViewSourceFilterAndSortIssue. Verified the issue occurs before the PR but not after.

## Risk

Can't think of any significant risk. When the subscription event occurs, it marks the collection as dirty and queues a dispatcher task to update the view. Both before and after PR, this task executes after the collection is copied.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/wpf/pull/9849)

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Data/ListCollectionView.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Data/ListCollectionView.cs
@@ -2584,6 +2584,11 @@ namespace System.Windows.Data
                                                 : (IList)(new ArrayList(size));
                 lsList = localList as LiveShapingList;
 
+                if (lsList != null)
+                {
+                    lsList.LiveShapingDirty += new EventHandler(OnLiveShapingDirty);
+                }
+
                 // filter the collection's array into the local array
                 for (int k = 0; k < size; ++k)
                 {
@@ -2605,11 +2610,6 @@ namespace System.Windows.Data
                 if (ActiveComparer != null)
                 {
                     localList.Sort(ActiveComparer);
-                }
-
-                if (lsList != null)
-                {
-                    lsList.LiveShapingDirty += new EventHandler(OnLiveShapingDirty);
                 }
 
                 _internalList = localList;


h3xds1nz:
LGTM, I was looking into this earlier when you've posted the workaround.

himgoyalmicro (approved):

PR #9855, opened 2024-09-27
Remove XBAP-related dead code from BrowserInteropHelper/SafeSecurityHelper

h3xds1nz:
## Description

Removes some XBAP-related dead code that hasn't been removed in #1009, mainly from `BrowserInteropHelper` and my favourite `SafeSecurityHelper`. Also removes the registry paths that are no longer looked up by WPF on .NET Core.

## Customer Impact

Decreased size of assemblies, better code quality of codebase for developers.

## Regression

No.

## Testing

Local build.

## Risk

I believe there is none, XBAP is not supported on .NET Core, this is just leftover code that is uncalled.


src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Interop/BrowserInteropHelper.cs
```diff
@@ -154,93 +115,10 @@ internal static UnsafeNativeMethods.IServiceProvider HostHtmlDocumentServiceProv
...
-        }
-
-        private static SecurityCriticalDataForSet<HostingFlags> _hostingFlags;
         private static SecurityCriticalDataForSet<bool> _isInitialViewerNavigation;
-        private static SecurityCriticalDataForSet<bool?> _isScriptInteropDisabled;
         
         private static SecurityCriticalDataForSet<UnsafeNativeMethods.IServiceProvider> _hostHtmlDocumentServiceProvider;
         private static SecurityCriticalDataForSet<bool> _initializedHostScript;
-
-        [DllImport(ExternDll.PresentationHostDll, EntryPoint="ForwardTranslateAccelerator")]
```
ThomasGoulet73, line 242: I believe you can go even further and delete PresentationHostDll and all its references and also delete anything that would require the file PresentationHost_cor3.dll to be used.

AFAIK PresentationHost_cor3.dll never shipped with any version of .Net Core (I checked in every version sine 3.0 and I didn't see it in any runtime packs).

h3xds1nz: Yeah, I was about to. It never shipped indeed. Thanks for checking as well, Thomas.

h3xds1nz: Fun fact, while digging into the PresentationHost stuff as whole, I've found out most of the `SecureEnvironment` (and AD RMS) cannot actually work and will throw (I believe `Msdrm` is not really supported anymore anyways on Windows newer than 8), either for missing dll (`PresentationHost_cor3`) or you get `STATUS_STACK_BUFFER_OVERRUN` somewhere in `secproc.dll` while calling the methods.

The DRM stuff is something I've never done (and most I'd say) but yeah, even on `NetFx` I had to manually load `PresentationHost_v0400` via `LoadLibrary` to at least retrieve `E_DRM_SERVICE_NOT_FOUND`, it won't load it on its own. At least it doesn't crash though after manually loading the library though!

There's literally zero documentation to it, kind of a black hole. Peak documentation is "TBD".

h3xds1nz:
Resolved merge conflicts after #9609

src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Interop/BrowserInteropHelper.cs
h3xds1nz: Anyways, resolving the conversation, I'll address this in a separate PR, almost done with it.

harshit7962 (approved):

harshit7962:
@h3xds1nz Thank you for the contribution.

h3xds1nz:
@harshit7962 Thanks for including it.

PR #9860, opened 2024-09-30
Redefine PixelFormatChannelMask as readonly struct, optimize methods

h3xds1nz:
## Description

Since `PixelFormatChannelMask` is immutable by design, we should adjust it to `readonly struct`. It will improve code quality. **This is not a breaking change**. I have also went ahead and optimized the methods and added documentation.

- Official documentation was the same as in code, hence I've added some fixes (e.g. 'then' -> 'than') too.
- `Equals` is now twice as fast for 8 bit colors; and a bit more than twice as fast for 32 bit colors.
- `Mask` property is now 3 times faster and requires half the memory allocations.
- Do note that the underlying type of `Mask` changes from `PartialList<byte>` to `byte[]`.
     - The difference being that the `IList<byte>` wrapper now will allow setting via indexer, while `PartialList<byte>` would throw an exception previously, so you would have to explicitly go and create new array if needed.
     - I do not think it matters, the array is owned by the caller as we make a private copy upon retrieval.
- WPF base libraries reference `System.Linq` by the way.

### Mask property retrieval (4 retrievals)
| Method    | Mean [ns] | Error [ns] | StdDev [ns] | Gen0   | Code Size [B] | Allocated [B] |
|---------- |----------:|-----------:|------------:|-------:|--------------:|--------------:|
| Original  | 129.70 ns |   1.494 ns |    1.398 ns | 0.0153 |       1,035 B |         256 B |
| PR__EDIT3 |  45.35 ns |   0.590 ns |    0.523 ns | 0.0076 |         476 B |         128 B |

### Equals (32bit color) (2 comparisons)
| Method    | Mean [ns] | Error [ns] | StdDev [ns] | Code Size [B] | Allocated [B] |
|---------- |----------:|-----------:|------------:|--------------:|--------------:|
| Original  | 10.683 ns |  0.1041 ns |   0.0974 ns |         416 B |             - |
| PR__EDIT3 |  4.650 ns |  0.0182 ns |   0.0152 ns |         765 B |             - |

### Equals (8 bit) (2 comparisons)
| Method    |  Mean [ns] | Error [ns] | StdDev [ns] | Code Size [B] | Allocated [B] |
|---------- |----------:|-----------:|------------:|--------------:|--------------:|
| Original  |  9.485 ns |  0.0806 ns |   0.0754 ns |         416 B |             - |
| PR__EDIT3 |  4.222 ns |  0.0231 ns |   0.0205 ns |         756 B |             - |

## Customer Impact

Increased performance, decreased allocations, cleaner code for developers.

## Regression

No.

## Testing

Local build, playing with the struct.

## Risk

As for the underlying type swap for `Mask` property, I do not think it negatively impacts anything, behavioral change has been outlined in the description above (imho a positive one).


src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Media/PixelFormat.cs
```diff
@@ -71,68 +74,63 @@ internal PixelFormatChannelMask(byte[] mask)
...
         /// The bitmask for a color channel
-        /// It will never be greater then 0xffffffff
+        /// It will never be greater than 0xFFFFFFFF.
         /// </summary>
         public IList<byte> Mask
         {
             get
             {
-                return _mask != null ? new PartialList<byte>((byte[])_mask.Clone()) : null;
+                return _mask?.ToArray();
```
ThomasGoulet73, line 83: I don't think returning a mutable IList here would break anything but I don't see the value in changing the existing behavior. Changing it to returning a mutable IList also means the API will now need to always return a mutable IList because in the future code could depend on the fact that they can modify the IList. Going non-mutable -> mutable is not a breaking change but the inverse is a breaking change.

This is what I would do (I also changed it to use ReadOnlyCollection here since it's built-in and we don't need the "partial list" feature of PartialList):
```suggestion
                return _mask != null ? new ReadOnlyCollection<byte>(_mask.AsSpan().ToArray()) : null;
```

h3xds1nz: It's a good point that in the future you would have this problem, indeed.

However, in that case, I do not understand why would you make a copy of the array if you're a returning readonly wrapper? Guess that's an oversight.

PR #9669, opened 2024-08-29
Optimize parsing of color strings into KnownColor enum, remove allocations

h3xds1nz:
## Description

Removes conversion of the color string into uppercase and its subsequent compare with `StringComparison.OrdinalIgnoreCase`.
The replacement was done by a simple regex.

| Method   | colorString | Mean [ns] | Error [ns] | StdDev [ns] | Gen0   | Code Size [B] | Allocated [B] |
|--------- |----------- |----------:|-----------:|------------:|-------:|--------------:|--------------:|
| Original | SlateGray  | 18.853 ns |  0.3373 ns |   0.3155 ns | 0.0024 |       6,066 B |          40 B |
| PR__EDIT | SlateGray  |  2.155 ns |  0.0072 ns |   0.0060 ns |      - |       6,695 B |             - |

## Customer Impact

Increased performance on serialization, zero allocations.

## Regression

No.

## Testing

Local build/CI.

## Risk

Low.

PR #9481, opened 2024-07-28
Replace occurrences of params T[] with params ReadOnlySpan<T> where possible

h3xds1nz:
## Description

Continues the journey that started in #9468 and converts some internal call-chains from `params T[]` to `params ReadOnlySpan<T>` where I felt it is without any side-effects, largely decreasing allocations in other scenarios besides just tracing and improving both performance and allocations since we do not need to create a new heap array of `T[]` everytime.

### Simple benchmark with 0, 1, and 3 params

| Method   | Mean [ns] | Error [ns] | StdDev [ns] | Gen0   | Code Size [B] | Allocated [B] |
|--------- |----------:|-----------:|------------:|-------:|--------------:|--------------:|
| Original |  7.547 ns |  0.1129 ns |   0.1056 ns | 0.0029 |         143 B |          48 B |
| PR__EDIT |  1.532 ns |  0.0096 ns |   0.0085 ns |      - |         120 B |             - |

<details><summary>Benchmark code</summary>

```c#
private static object target1;
private static object target2;
private static object target3;

[GlobalSetup]
public void Setup()
{
    target1 = new object();
    target2 = new object();
    target3 = new object();
}

[Benchmark]
public int Original()
{
    int test1_1 = Test1(1);
    int test1_2 = Test1(1, null);
    int test1_3 = Test1(1, target1, target2, target3);

    return  test1_1 + test1_2 + test1_3;
}


[Benchmark]
public int PR__EDIT()
{
    int test2_1 = Test2(1);
    int test2_2 = Test2(1, null);
    int test2_3 = Test2(1, target1, target2, target3);

    return test2_1 + test2_2 + test2_3;
}

[MethodImpl(MethodImplOptions.AggressiveInlining)]
private int Test1(int stuff, params object[] args)
{
    if (args is null || args.Length == 0)
        return stuff;

    return args[0].GetHashCode();
}

[MethodImpl(MethodImplOptions.AggressiveInlining)]
private int Test2(int stuff, params ReadOnlySpan<object> args)
{
    if (args.IsEmpty)
        return stuff;

    return args[0].GetHashCode();
}

```
</details>

## Customer Impact

Improved performance, decreased allocations.

## Regression

No.

## Testing

Local build, test app run.

## Risk

Low.

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/wpf/pull/9481)

PR #9865, opened 2024-10-01
Remove XBAP-related dead code from Application/BrowserInteropHelper, finish cleanup

h3xds1nz:
## Description

This continues the work started in #9855, removing rest of the internal properties in `BrowserInteropHelper` and thanks to this we can remove dead code in other places of the codebase (`Application`, `AppSecurityManager`, etc.) I've separated the two as those deletions might be a bit harder to grasp, while the first one is easier to understand imho.

- `SiteOfOriginContainer.BrowserSource` is never set, thus it is always `NULL`.
     - This means the public property `Uri` in `BrowserInteropHelper` in always `NULL`.
     - `GetReferer` method in `BindUriHelper` depends on this value, if `BrowserSource` is NULL, it returns `NULL`.
     - So the value in `httpRequest.Referer` of `WpfWebRequestHelper` would end up being `NULL` anyways, no-op / dead code.
- `Application.MimeType` is never set, thus it is always `MimeType.Unknown`.
     - This means `IsViewer` of `BrowserInteropHelper` is always `false`.
     - That also means that `IsInitialViewerNavigation` is always `false` and its underlying value does not matter at all for any operations. So we can remove any setters; and then we can remove it from both conditions; which makes it unused.
- `Application.ServiceProvider` is never set (classes doing it don't exist anymore), so it is always `NULL`.
- Classes calling the `GetService` also don't exist anymore, so the whole method does not matter.
- Any methods removed in `AppSecurityManager` just call themselves, besides that we were still calling `ClearSecurityManager` from Application but it is never initialized anymore so there's nothing to clear.
- As a result, `SecurityMgrSite` is never initialized anymore, so we can remove it as well.

## Customer Impact

Decreased size of assemblies, tiny bit of perf, cleaner codebase for developers.

## Regression

No.

## Testing

Local build, app run.

## Risk

Low, the code is a dead code or it doesn't do anything.

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/wpf/pull/9865)


h3xds1nz:
Looks like one of the changes in #9220 was here. Rebased and resolved merge conflicts.

h3xds1nz:
Resolved merge conflicts #01

h3xds1nz:
Fix merge conflicts after #9521

h3xds1nz:
Resolved merge conflicts after #9855 

@harshit7962 Can we go with this one and #9943 next?

harshit7962:
@h3xds1nz, We are done with the test pass of #9943 and will prioritize it. I am yet to review this PR and will include it in the next test run.

h3xds1nz:
@harshit7962 Awesome, let me know at which part you'd like me to resolve the conflicts.

I feel its counter-productive to do it now should a style-PR get merged meanwhile as that will mess up the resolutions.

harshit7962:
Sure, will let you know when the PR is ready to be merged. It would be better to resolve the conflicts at that time.

src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Interop/BrowserInteropHelper.cs
```diff
@@ -41,53 +32,13 @@ static BrowserInteropHelper()
         /// <summary>
         /// Returns true if the app is a browser hosted app.
         /// </summary>
-        /// <remarks>
```
harshit7962, line 44: Was this intentionally removed?

h3xds1nz: Yeah, `HostingFlags` was an internal static property that is no longer present since #9855, just forgot to clean up before.

harshit7962 (approved):

src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Interop/BrowserInteropHelper.cs
harshit7962: Yeah makes sense. Other changes seems fine. Will speed up the test pass and try to get this in.

harshit7962:
@h3xds1nz, we are done with the test pass and its all green. We are ready to take this change in.

h3xds1nz:
> @h3xds1nz, we are done with the test pass and its all green. We are ready to take this change in.

@harshit7962 Great, should be merged/rebased properly I hope. I can't compile right now so I have to verify with CI I did pick the correct usings.

harshit7962 (approved):

harshit7962:
@h3xds1nz, thanks for the contribution.

We are working on getting in the PRs that have passed the tests. Please let us know if any of these you would like us to prioritize.

Also, feel free to mention any PR that you would like us to include in the next test run.

h3xds1nz:
@harshit7962 Great, glad to see this one in.

The untested ones in order of importance (perf gains):

1) #9364
2) #10003
3) #9949
4) #9199
5) #9395
6) #9844 (this is a big gain but the customer impact is lower as its more of an edge case)

Do note there are two unit test PRs that could be merged imho:

1) #9895
2) #9891 (this one includes a fix that slipped in through the WPF tests)

Regarding the ones that have passed tests so I can do further work/would be good to merge:

1) #9822
1.1) #9739
2) #9669
3) #9481 (@dipeshmsft was reviewing it)
4) #9468 (this provides a regression fix)

PR #9866, opened 2024-10-01
Remove AppContextDefaultValues/AccessibilitySwitches values from NetFX (dead code)

h3xds1nz:
## Description

This is a follow-up to #9711, cleaning up the dead code from assemblies that are compiled for .NET Core only.
The PBT one has been kept for obvious reasons.

The effective process is removing the switch statement, values for `NetFX` and copying over contents of `InitializeNetFxSwitchDefaultsForNetCoreRuntime` directly into the partial method.

I have also removed the warning supressions where they were not necessary.

## Customer Impact

Smaller size of assemblies, 0.0001 ns faster start up time, cleaner codebase for developers.

## Regression

No.

## Testing

Local build, app startup.

## Risk

Low, this is just a cleanup after previous PR.

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/wpf/pull/9866)

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/AppContextDefaultValues.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/AppContextDefaultValues.cs
@@ -2,68 +2,30 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
+// Since multiple assemblies (WindowsBase, PresentationFramework, PresentationCore, PresentationBuildTasks)
+// share this file/class and have InternalsVisibleTo set between them, there will be a compiler warning
+// that the type is found both in the source and in a referenced assembly. The compiler will prefer 
+// the version of the type defined in the source once the warning is supressed and won't compile the rest.
+
+// In order to disable the warning for this type we are disabling the warning for this entire file.
+#pragma warning disable CS0436 // Type conflicts with imported type
+
 using MS.Internal;
 
 namespace System
 {
-    // There are cases where we have multiple assemblies that are going to import this file and
-    // if they are going to also have InternalsVisibleTo between them, there will be a compiler warning
-    // that the type is found both in the source and in a referenced assembly. The compiler will prefer
-    // the version of the type defined in the source
-    //
-    // In order to disable the warning for this type we are disabling this warning for this entire file.
-    #pragma warning disable 436
-
     internal static partial class AppContextDefaultValues
     {
+        /// <summary>
+        /// This is a partial method. This method is responsible for populating the default values based on a TFM.
+        /// It is partial because each library should define this method in their code to contain their defaults.
+        /// </summary> 
         static partial void PopulateDefaultValuesPartial(string platformIdentifier, string profile, int targetFrameworkVersion)
         {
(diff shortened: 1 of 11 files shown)


src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/AppContextDefaultValues.cs
```diff
@@ -2,68 +2,30 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
+// Since multiple assemblies (WindowsBase, PresentationFramework, PresentationCore, PresentationBuildTasks)
+// share this file/class and have InternalsVisibleTo set between them, there will be a compiler warning
+// that the type is found both in the source and in a referenced assembly. The compiler will prefer 
+// the version of the type defined in the source once the warning is supressed and won't compile the rest.
+
+// In order to disable the warning for this type we are disabling the warning for this entire file.
+#pragma warning disable CS0436 // Type conflicts with imported type
```
harshit7962, line 11: @h3xds1n, Imo the more granular we could go with respect to warning suppressions the better it is. Can we not explore disabling the warning at class level or `PopulateDefaultValuesPartial` method itself?

harshit7962 (changes requested):

src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/AppContextDefaultValues.cs
h3xds1nz: @harshit7962 I'll check later today whether we could constrain it better but at least class-level will be still required iirc.

h3xds1nz: @harshit7962 I see why it didn't work previously for me; I didn't include the cleanup of the warnings in the switch-specific files. I've fixed it in this file and also constrained it in `LocalAppContext` to suppress just in the static constructor; should do the job better.

I've unified the comment where the suppression is actually needed.

Thanks as always for the great feedback!

h3xds1nz: Also, just thinking, `LocalAppContext` should not be `partial` but `static`.

There seems to be no reason for this to be an instance class as all the members are static and it is not using any partial stuff either, each library gets a full compiled copy of it. I'd push a change for this small thing if you agree with me.

harshit7962: @h3xds1nz, seems fair. We can go ahead and have the class as static and non-partial. Resolving this conversation as of now.

harshit7962 (dismissed):

src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/AppContextDefaultValues.cs
h3xds1nz: @harshit7962 Pushed the change, you can re-approve.

harshit7962 (approved):

harshit7962:
@h3xds1nz Thank you for the contribution.

h3xds1nz:
@harshit7962 Thank you!

PR #9432, opened 2024-07-20
Replace ArrayList in ItemContainerGenerator with List<EmptyGroupItem>

h3xds1nz:
## Description

Replaces `ArrayList` with `List<EmptyGroupItem>` to hold empty group items when needed. Improves code quality by swapping to generic collection and decreases memory footprint.

Sample benchmark showing difference between `ArrayList` and `List<EmptyGroupItem>`.

### 10 additions comparison

| Method   | Mean [ns] | Error [ns] | StdDev [ns] | Gen0   | Code Size [B] | Gen1   | Allocated [B] |
|--------- |----------:|-----------:|------------:|-------:|--------------:|-------:|--------------:|
| Original | 103.59 ns |   2.013 ns |    3.578 ns | 0.0483 |         740 B | 0.0001 |         808 B |
| PR__EDIT |  95.02 ns |   1.907 ns |    2.480 ns | 0.0483 |         342 B | 0.0001 |         808 B |

<details><summary>Benchmark code</summary>

```c#
[Benchmark]
public void Original()
{
    ArrayList? _alEmptyGroupItems = null;

    for (int i = 0; i < 10; i++)
        BenchV1(ref _alEmptyGroupItems);
}

[Benchmark]
public void PR__EDIT()
{
    List<EmptyGroupItem>? _emptyGroupItems = null;

    for (int i = 0; i < 10; i++)
        BenchV2(ref _emptyGroupItems);
}

[MethodImpl(MethodImplOptions.AggressiveInlining)]
private void BenchV1(ref ArrayList? _alEmptyGroupItems)
{
    EmptyGroupItem emptyGroupItem = new EmptyGroupItem();

    if (_alEmptyGroupItems == null)
        _alEmptyGroupItems = new ArrayList();

    _alEmptyGroupItems.Add(emptyGroupItem);
}

[MethodImpl(MethodImplOptions.AggressiveInlining)]
private void BenchV2(ref List<EmptyGroupItem>? _emptyGroupItems)
{
    EmptyGroupItem emptyGroupItem = new EmptyGroupItem();

    if (_emptyGroupItems == null)
        _emptyGroupItems = new();

    _emptyGroupItems.Add(emptyGroupItem);
}

public class EmptyGroupItem
{
    public string? Item1 { get; set; }

    public Int32? Item2 { get; set; }

    public Int32? Item3 { get; set; }

    public List<int>? Item4 { get; set; }
}
```

</details>

## Customer Impact

Improved performance, getting rid of `ArrayList`.

[...]

PR #9871, opened 2024-10-01
Fix RemoveWriterByIndex bug, replace ArrayList in BitmapMetadataBlockWriter

h3xds1nz:
## Description

Replaces backing `ArrayList` in `BitmapMetadataBlockWriter` and `BitmapMetadataBlockWriterEnumerator`.

Also fixes a bug in `RemoveWriterByIndex` function, which would previously not remove the writer.

Sample benchmark showing difference between `ArrayList` and `List<T>` with reference type can be found f.e. in #9432.

## Customer Impact

Improved performance, decreased allocations, a bug fix.

## Regression

No.

## Testing

Local build, work with `BitmapMetadata`.

## Risk

Low but note that things might actually work properly and a bit faster now.

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/wpf/pull/9871)

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Media/Imaging/BitmapMetadata.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Media/Imaging/BitmapMetadata.cs
@@ -146,7 +146,7 @@ namespace System.Windows.Media.Imaging
             {
                 _fixedSize = fixedSize;
                 _containerFormat = containerFormat;
-                _metadataBlocks = new ArrayList();
+                _metadataBlocks = new List<SafeMILHandle>();
             }
 
             internal BitmapMetadataBlockWriter(BitmapMetadataBlockWriter blockWriter, object syncObject)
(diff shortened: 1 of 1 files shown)


src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Media/Imaging/BitmapMetadata.cs
```diff
@@ -437,22 +437,22 @@ UInt32 index
                     return (int) WinCodecErrors.WINCODEC_ERR_UNSUPPORTEDOPERATION;
                 }
 
-                _metadataBlocks.Remove(index);
```
h3xds1nz, line 440: Basically previously it would look for the index of the writer, as if it was an item (which would never succeed), the correct implementation is to remove the writer at the index.

harshit7962 (approved):
LGTM

harshit7962:
Thank you @h3xds1nz for the contribution.

h3xds1nz:
@harshit7962 Thank you :)

PR #9887, opened 2024-10-03
Redefine FigureLength as readonly struct to prevent defensive copies

h3xds1nz:
## Description

Since `FigureLength` is immutable by design, we should adjust it to `readonly struct`. This will also prevent defensive copies. It will also improve code quality. **This is not a breaking change**.

- I've also simplified the ctor a bit, which will also offer a very minor performance benefit (it does not matter).
- Minor cleanup of the usual `IEquatable` functions to save precious code lines and simplify it.
- The value was being set to `0.0` in case of `Auto` but when queried, it is set to `1.0`.
     - I couldn't see a reason for it, the only thing that made sense was the hash code creation but the hash code is not unique even for basic values. It is too easy to get duplicates. This should be fixed later, I'll create a tracking issue.

## Customer Impact

Possibly a bit of perf and a clear directive that `FigureLength` and all its methods are not adjusting the state of the `struct`.

## Regression

No.

## Testing

Local build, base testing, type creation, etc.

## Risk

### DIFF
--- a/src/Microsoft.DotNet.Wpf/cycle-breakers/PresentationFramework/PresentationFramework.cs
+++ b/src/Microsoft.DotNet.Wpf/cycle-breakers/PresentationFramework/PresentationFramework.cs
@@ -405,9 +405,8 @@ namespace System.Windows
         PageRight = 2,
     }
     [System.ComponentModel.TypeConverterAttribute(typeof(System.Windows.FigureLengthConverter))]
-    public partial struct FigureLength : System.IEquatable<System.Windows.FigureLength>
+    public readonly partial struct FigureLength : System.IEquatable<System.Windows.FigureLength>
     {
-        private int _dummyPrimitive;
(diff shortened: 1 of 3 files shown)


miloush:
Funny hash code. Okay so if someone binary-serialized an instance of this structure to the disk, would upgrade .NET and try to deserialize it back, then the deserialized instances of Auto one would not equal new instances of Auto, correct?

h3xds1nz:
@miloush I've found one similar implementation, must have been done by the same engineer as the comments are same.

Equality is done over the property, not the field, so those two would still match. Obviously the payload itself would change but when deserialized and loaded, those two would still be equal.

miloush:
Ah missed the property, thanks!

h3xds1nz:
No worries, thanks for reviewing the changes and asking the right questions :)

PR #9890, opened 2024-10-03
Replace PartialList with ReadOnlyCollection

ThomasGoulet73:
## Description
ReadOnlyCollection does the same thing as PartialList but it's more optimized due to the fact that it does not support wrapping a range of a list and instead just forwards the read-only call to the wrapped list. It's also better to use a collection from the BCL than our own where possible.

I also removed the constructor that allowed passing a list without a range to discourage the use of PartialList when we don't need the range feature and I removed an unused static method in PartialList.

Here's benchmark where I do a foreach on PartialList vs ReadOnlyCollection:
```
| Method             | Mean     | Error    | StdDev   | Ratio | Gen0   | Allocated | Alloc Ratio |
|------------------- |---------:|---------:|---------:|------:|-------:|----------:|------------:|
| PartialList        | 34.57 ns | 0.334 ns | 0.313 ns |  1.00 | 0.0043 |      72 B |        1.00 |
| ReadOnlyCollection | 20.38 ns | 0.181 ns | 0.161 ns |  0.59 | 0.0024 |      40 B |        0.56 |
```

## Customer Impact
Better perf.

## Regression
No.

## Testing
Local testing.

## Risk
Low.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/wpf/pull/9890)


src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Media/Imaging/BitmapPalette.cs
```diff
@@ -382,7 +383,7 @@ private struct ImagePaletteColor
         // the behavior that we want.
         private SafeMILHandle _palette = null; // IWICPalette*
 
-        private IList<Color> _colors = new PartialList<Color>(new List<Color>());
+        private IList<Color> _colors = new ReadOnlyCollection<Color>(new List<Color>());
```
kronic, line 386: use ReadOnlyCollection<Color>.Empty

ThomasGoulet73: Done, thanks.

ThomasGoulet73:
I rebased on main to fix the conflicts.

ThomasGoulet73:

ThomasGoulet73:

himgoyalmicro (approved):
LGTM

himgoyalmicro:
Thank you @ThomasGoulet73 for your continuous contributions 😄

ThomasGoulet73:
Thanks

PR #9676, opened 2024-08-30
Optimize conversion of MouseAction enum from/to string, remove allocations

h3xds1nz:
## Description

Parsing optimization for `MouseAction` and its conversion from/to `string`. Allocation-free.
- Removes trimming on string, now trims on the span, resulting in no allocations.
- Removes conversion `ToUpperInvariant` and uses `StringComparison.OrdinalIgnoreCase` instead.
- This will allow for string interning between `Convert`/`ConvertFrom` and other literals, saving static memory.
- The JITTed codegen takes up less memory as we removed few things.
- Adds/completes code documentation for the whole class.

### Last value in the enumeration

| Method   | source            | Mean [ns] | Error [ns] | StdDev [ns] | Gen0   | Code Size | Allocated |
|--------- |------------------ |----------:|-----------:|------------:|-------:|----------:|----------:|
| Original | MiddleDoubleClick | 19.632 ns |  0.2160 ns |   0.2020 ns | 0.0033 |       1,270 B |          56 B |
| PR__EDIT | MiddleDoubleClick |  2.776 ns |  0.0166 ns |   0.0155 ns |      - |         988 B |             - |

### Last value in the enumeration with space (trimmed)

| Method   | source             | Mean [ns] | Error [ns] | StdDev [ns] | Gen0   | Code Size | Allocated |
|--------- |------------------- |----------:|-----------:|------------:|-------:|----------:|----------:|
| Original | MiddleDoubleClick\x20  | 28.230 ns |  0.5535 ns |   0.5177 ns | 0.0067 |       2,176 B |         112 B |
| PR__EDIT | MiddleDoubleClick\x20  |  5.659 ns |  0.0315 ns |   0.0280 ns |      - |       1,448 B |             - |

## Customer Impact

Increased performance, zero allocations, less static memory.

## Regression

No.

## Testing

Local build, CI, assert test:
```csharp
//Special case
string emptyToNone = string.Empty;
MouseAction constantName = MouseAction.None;

MouseAction lowerCase = (MouseAction)ConvertFrom(null, null, emptyToNone.ToLowerInvariant());
MouseAction upperCase = (MouseAction)ConvertFrom(null, null, emptyToNone.ToUpperInvariant());
MouseAction normalColor = (MouseAction)ConvertFrom(null, null, emptyToNone);

Assert.AreEqual<MouseAction>(constantName, lowerCase);
Assert.AreEqual<MouseAction>(constantName, upperCase);
[...]

PR #9891, opened 2024-10-03
Add test coverage for public API surface of MouseActionConverter

h3xds1nz:
## Description

Adds test coverage for `MouseActionConverter`. I've also fixed an unintentional behavioral change I've introduced with #9676, which would mean it wouldn't throw `InvalidEnumArgumentException` due to passing wrong value type to the exception.

- As I've outlined previously, there are basically two (in my opinion) bugs in the class
    - `CanConvertTo` throws `InvalidCastException` due to an unboxing cast when you pass an `int`
    - The same thing happens in `ConvertTo` when you pass an `int` within range

## Customer Impact

Improved test coverage on public surface.

## Regression

Fixing my own oversight at the same time.

## Testing

Local build.

## Risk

Low.

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/wpf/pull/9891)

### DIFF
--- a/src/Microsoft.DotNet.Wpf/tests/UnitTests/PresentationCore.Tests/System/Windows/Input/MouseActionConverter.Tests.cs
+++ b/src/Microsoft.DotNet.Wpf/tests/UnitTests/PresentationCore.Tests/System/Windows/Input/MouseActionConverter.Tests.cs
@@ -1,9 +1,10 @@
-﻿using System.Collections;
-using System.Windows.Input;
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
 using System.ComponentModel;
 using System.ComponentModel.Design.Serialization;
 
-namespace PresentationCore.Tests.System.Windows.Input
+namespace System.Windows.Input
 {
     public class MouseActionConverterTests
     {
--- a/src/Microsoft.DotNet.Wpf/tests/UnitTests/PresentationCore.Tests/System/Windows/Input/MouseActionConverter.Tests.cs
+++ b/src/Microsoft.DotNet.Wpf/tests/UnitTests/PresentationCore.Tests/System/Windows/Input/MouseActionConverter.Tests.cs
@@ -1,97 +1,14 @@
 ﻿// Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the MIT license.
 
-using System.ComponentModel;
 using System.ComponentModel.Design.Serialization;
+using System.ComponentModel;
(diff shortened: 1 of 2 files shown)


dipeshmsft:
> As I've outlined previously, there are basically two (in my opinion) bugs in the class
> - CanConvertTo throws InvalidCastException due to a cast when you pass an int
> - The same thing happens in ConvertTo when you pass an int within range

Good point, this looks like an issue in the class implementation.

src/Microsoft.DotNet.Wpf/tests/UnitTests/PresentationCore.Tests/System/Windows/Input/MouseActionConverter.Tests.cs
```diff
@@ -0,0 +1,166 @@
...
+
+using System.ComponentModel;
+using System.ComponentModel.Design.Serialization;
+
+namespace System.Windows.Input
+{
+    public class MouseActionConverterTests
+    {
+        [Fact]
+        public void ConvertTo_ConvertFrom_ReturnsExpected()
```
dipeshmsft, line 12: @h3xds1nz, is there any reason why we are not using `[Theory]` along with data here ? This seems a little shift from the general behavior.

dipeshmsft (commented):
Rest of the changes look good to me.

src/Microsoft.DotNet.Wpf/tests/UnitTests/PresentationCore.Tests/System/Windows/Input/MouseActionConverter.Tests.cs
```diff
@@ -0,0 +1,166 @@
...
+
+            Assert.Throws<NotSupportedException>(() => converter.ConvertTo(value, destinationType));
+        }
+
+        [Fact]
+        public void ConvertTo_ThrowsInvalidCastException()
+        {
+            MouseActionConverter converter = new();
+
+            Assert.Throws<InvalidCastException>(() => converter.ConvertTo(null, null, (int)(MouseAction.MiddleDoubleClick), typeof(string)));
```
dipeshmsft, line 71: It will be better that we add a comment here to tell that usually this is not an expected behavior.

h3xds1nz: I'm practically waiting for this to be approved so I can raise other PR fixing the behavior haha but I'll add it as I solve the other point.

dipeshmsft (approved):

PR #9902, opened 2024-10-05
[Managed DWrite] Migrate part of DWriteTypeConverter to managed

ThomasGoulet73:
Contributes to dotnet/wpf#5305

## Description
Migrate part of DWriteTypeConverter to managed. I also removed some dead code from my previous Managed DWrite PR. I changed the name of `DWriteTypeConverter` in PresentationCore to `DWriteTypeConverterEx` because it conflicted with the one in DirectWriteForwarder.

## Customer Impact

## Regression
No.

## Testing
Local testing.

## Risk
Low.


src/Microsoft.DotNet.Wpf/src/PresentationCore/MS/internal/Text/TextInterface/DWriteTypeConverterEx.cs
```diff
@@ -0,0 +1,60 @@
...
+            switch (dwriteMeasuringMode)
+            {
+                case DWRITE_MEASURING_MODE.DWRITE_MEASURING_MODE_NATURAL:
+                    return TextFormattingMode.Ideal;
+                case DWRITE_MEASURING_MODE.DWRITE_MEASURING_MODE_GDI_CLASSIC:
+                    return TextFormattingMode.Display;
+                // We do not support Natural Metrics mode in WPF
+                // However, the build system complained about not having an explicit case 
+                // for DWRITE_TEXT_MEASURING_METHOD_USE_DISPLAY_NATURAL_METRICS
+                case DWRITE_MEASURING_MODE.DWRITE_MEASURING_MODE_GDI_NATURAL:
```
JeremyKuhne, line 53: nit: slightly better to fall through here and only have one throw.

JeremyKuhne (approved):
Looks good to me.

Hopefully in a few weeks we'll have CsWin32 and xunit support in place which will make changes like this possible without pre-emptive test passes.

src/Microsoft.DotNet.Wpf/src/PresentationCore/MS/internal/Text/TextInterface/DWriteTypeConverterEx.cs
ThomasGoulet73: I wanted to keep the case for the comment but I didn't think about just merging `case DWRITE_MEASURING_MODE.DWRITE_MEASURING_MODE_GDI_NATURAL` with `default`. I pushed a commit the just merges the 2 and removes the duplicate throw.

ThomasGoulet73:
@JeremyKuhne Do you mind reviewing dotnet/wpf#6260 as well ? It's in the same vein.

ThomasGoulet73:

PR #9907, opened 2024-10-07
Avoid calling soon-to-be-removed string.Trim(ReadOnlySpan<char>) methods

bartonjs:
## Description

.NET 9 previews added new string.Trim/TrimStart/TrimEnd methods accepting (params) `ReadOnlySpan<char>`.  Those methods are being removed from .NET 9 GA (and .NET 10/main), so they should no longer be called by WPF.

## Customer Impact

WPF won't build.

## Regression

No

## Testing

It compiles.

## Risk

Extremely low, mechanical change.

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/wpf/pull/9907)

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/System.Xaml/System/Xaml/MS/Impl/KnownStrings.cs
+++ b/src/Microsoft.DotNet.Wpf/src/System.Xaml/System/Xaml/MS/Impl/KnownStrings.cs
@@ -35,7 +35,7 @@ namespace System.Xaml.MS.Impl
         public const string DefaultPrefix = "p";
 
         public const string ReferenceName = "__ReferenceID";
-        public static ReadOnlySpan<char> WhitespaceChars => [' ', '\t', '\n', '\r', '\f'];
+        public static readonly char[] WhitespaceChars = [' ', '\t', '\n', '\r', '\f'];
         public const char SpaceChar = ' ';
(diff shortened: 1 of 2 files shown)


ThomasGoulet73:
I believe these also need to be fixed (Unless I misunderstood what's being removed):
https://github.com/dotnet/wpf/blob/278c355919ad7496717b370ae9185857d26cf179/src/Microsoft.DotNet.Wpf/src/System.Xaml/System/Xaml/Parser/MeScanner.cs#L509-L510

They use KnownStrings.WhitespaceChars: https://github.com/dotnet/wpf/blob/278c355919ad7496717b370ae9185857d26cf179/src/Microsoft.DotNet.Wpf/src/System.Xaml/System/Xaml/MS/Impl/KnownStrings.cs#L38

bartonjs:
Ah, yep.  Good catch; thanks.

dipeshmsft:
@bartonjs, is there some places, from where I can check what all methods are being removed?

bartonjs:
https://github.com/dotnet/runtime/pull/108537/files#diff-cec8e6f471b4193246bdc0107b0dd7cbe131fb7fd189b288b37269c333d1171d

dipeshmsft (approved):

PR #9926, opened 2024-10-10
Optimize assembly/namespace pair creation and updating in XamlSchema

h3xds1nz:
## Description

Adjusts inefficient `List<AssemblyNamespacePair>` copies in `AddAssemblyNamespacePair` that result in up to double gen0 allocations needed for the job, offering perf benefit and decreased allocations.

- I have changed it to be a `readonly struct`, there's like zero benefit but I think it serves better purpose.
- The `WeakReference` is now generic to prevent unnecessary casting.
- Since `_assemblyNamespaces` is under a write-lock and copies are created each time to allow reading concurrently by multiple readers, we can just swap to an array for overall efficiency since only 1 item is added each time.

### Initial creation
| Original | 108.37 ns |   2.179 ns |    2.594 ns | 0.0086 |         317 B |         144 B |
| PR_EDIT |  89.78 ns |   1.833 ns |    2.687 ns | 0.0038 |         223 B |          64 B |

### Initial creation plus 9x AddAssemblyNamespacePair

| Original |  1,419.8 ns |   28.21 ns |    37.66 ns | 0.1678 |       2,668 B |        2824 B |
| PR_EDIT |  1,100.8 ns |   18.28 ns |    16.20 ns | 0.0935 |       2,329 B |        1584 B |

## Customer Impact

## Regression

No.

## Testing

Local build, basic method testing.

## Risk

Low, there's only a swap from class to readonly struct being stored in a private array (that was originally a list).

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/wpf/pull/9926)

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/System.Xaml/System/Xaml/MS/Impl/AssemblyNamespacePair.cs
+++ b/src/Microsoft.DotNet.Wpf/src/System.Xaml/System/Xaml/MS/Impl/AssemblyNamespacePair.cs
@@ -1,17 +1,16 @@
 // Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the MIT license.
 
-#nullable disable
-
+using System.Diagnostics;
 using System.Reflection;
 
 namespace System.Xaml.MS.Impl
 {
     [DebuggerDisplay("{ClrNamespace} {Assembly.FullName}")]
-    internal class AssemblyNamespacePair
+    internal readonly struct AssemblyNamespacePair
     {
-        private WeakReference _assembly;
-        private string _clrNamespace;
+        private readonly WeakReference _assembly;
+        private readonly string _clrNamespace;
 
         public AssemblyNamespacePair(Assembly asm, string clrNamespace)
         {
@@ -19,9 +18,9 @@ namespace System.Xaml.MS.Impl
             _clrNamespace = clrNamespace;
         }
 
-        public Assembly Assembly
+        public Assembly? Assembly
         {
-            get { return (Assembly)_assembly.Target; }
+            get { return (Assembly?)_assembly.Target; }
         }
 
         public string ClrNamespace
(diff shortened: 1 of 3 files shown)


h3xds1nz:
@harshit7962 @himgoyalmicro I've fixed the dummy NRE mistake in the last commit, so this could finally go through tests.

dipeshmsft:
@h3xds1nz, this PR looks good to me. Can you resolve the merge conflicts here ?

**PS:** From what I gather, copying a List is heavier as compared to array and that's why the shift ? If yes, then maybe we can add this as a comment where AssemblyNamespacePair[] array is defined.

h3xds1nz:
@dipeshmsft `List<T> is merely a wrapper` over `T[]` which in normal scenarios (besides ref returns etc.) doesn't really add much overhead over array and will often be more beneficial overall if you need to add items and resize them as you already pre-allocate some scratch space but access it as a limited-number array.

This is a special case, or rather, lazy misuse ot `List<T>` capabilities as you:
1) only build the list at the beginning
2) only add 1 item per iteration
3) you're attempting atomic updates, so you're forced to create a new instance each time

This means you're here initializing the `List<T>` from another `List<T>`, which creates a precisely-sized array onto you which then add a new item which incurs a new, practically redundant, copy of the just freshly created `List<T>`.

Therefore exploiting the fact that we only add 1 item, we can just allocate an array with `n + 1` elements, copy then `n` and add the `1`, without having to copy twice.

Plus we get better control of the sizing (empty list on first addition allocates array or 4 vs our array of 1) and then we don't grow by * 2 number of items but by 1.

dipeshmsft:
> This means you're here initializing the List<T> from another List<T>, which creates a precisely-sized array onto you which then add a new item which incurs a new, practically redundant, copy of the just freshly created List<T>.

Okay, so now using array, we pre-allocate the extra space for the 1 extra element, preventing the extra copy. Thanks a lot for the explaination.

h3xds1nz:
@dipeshmsft Anytime :) (though usually it is better to answer questions regarding PRs in the next few weeks than 8 months later, one might miss context after that time if its something complicated, which isn't the case here but couldn't help myself to nitpick about the timeline 😀)

Conflicts resolved.

dipeshmsft (approved):

PR #9947, opened 2024-10-12
Adjust MouseWheelEventArgs.Delta from static to an instance readonly field

h3xds1nz:
## Description

Similarly as in #7910, the `MouseWheelEventArgs.Delta` is a static field, meaning a creation of a new event changes the `Delta` value for all previously created events. This is clearly a bug, the field shall not be `static` but an instanced `readonly` as the event should describe the state at the exact point of time it was generated.

## Customer Impact

Creation of a new `MouseWheelEventArgs` will not influence the value of `Delta` for previous instances.

## Regression

No.

## Testing

Local build.

## Risk

Low.

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Input/MouseWheelEventArgs.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Input/MouseWheelEventArgs.cs
@@ -2,9 +2,6 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
-
-using System;
-
 namespace System.Windows.Input
 {
     /// <summary>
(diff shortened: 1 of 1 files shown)


miloush:
> the event should describe the state at the exact point of time it was generated.

That is not such a clear cut, since as discussed in a few other issues here, mouse events/args tend to return the current state of the device rather than the state at the time the args were created. See the buttons state properties, coming from `MouseDevice.GetButtonStateFromSystem` or even the `GetPosition`, which reports the last received mouse position for any instance of the args. I can see how a similar behavior could have been the intent here (and in that light, the #7910 fix is questionable).

That said, wheel delta is not a state, it's an event, and it seems reasonable that you should be able to accumulate them, and accumulating with the current implementation gives you potentially incorrect results.

h3xds1nz:
Totally forgot; just wanted to say I fully agree with the statement @miloush, was just trying too hard to sell it.

PR #9948, opened 2024-10-12
Fix FontSource/FontFileStream embedded resource memory leak

### LINKED ISSUE #7289: Using embedded font resources leaks memory
* .NET Core Version: 7.0
* Windows version: Windows 10 version 21H2 (OS Build 19044.2251)
* Is this bug related specifically to tooling in Visual Studio (e.g. XAML Designer, Code editing, etc...)? No
 
 **Problem description:**
When a custom font is loaded as a Resource and used in the application, the application slowly leaks memory via an `UnmanagedMemoryStream` object that grows apparently without bounds, regardless of GC collection.

The issue does not appear to affect fonts with Build Action set to Content (although publishing fonts this way may be against their respective EULAs).

 **Actual behavior:**
When a text block is rapidly changed between a series of embedded fonts, Visual Studio's Diagnostic Tools shows a `UnmanagedMemoryStream` object growing by approximately 35–40 KB per second.
 
 **Expected behavior:**
Once all the custom fonts are loaded, changing the `FontFamily` property of a `TextBlock` shouldn't increase memory usage at all.
 
 **Minimal repro:**
1. Create a new WPF project (the issue can be reproduced in .NET 7.0 and .NET Framework v4.8 at least, but [I can find references to it online from earlier][1] referencing now-unavailable MS Connect posts).
2. Add some typefaces to the project (I've got a combination of OTF and TTF fonts in my test app), and mark their Build Action as `Resource`.
3. Add those fonts to your application resources in `App.xaml` (exact code will differ based on the fonts you choose):
    ```xaml
    <Application.Resources>
      <FontFamily x:Key="Flama">./Fonts/#Flama</FontFamily>
      <FontFamily x:Key="FlamaMedium">./Fonts/#Flama Medium</FontFamily>
      <FontFamily x:Key="IntroCaps">./Fonts/#Intro Caps</FontFamily>
      <FontFamily x:Key="IntroBlackInline">./Fonts/#Intro Inline Caps</FontFamily>
      <FontFamily x:Key="Pacifico">./Fonts/#Pacifico</FontFamily>
    </Application.Resources>
    ```
4. In `MainWindow.xaml`, create a `TextBlock`:
    ```xaml
    <TextBlock x:Name="TestBlock" FontSize="48">Test block</TextBlock>
    ```
5. Add a new `Loaded` event handler for `MainWindow` that cycles through your embedded fonts to use as the text block's font family as fast as possible:
    ```csharp
    private void MainWindow_Loaded(object sender, RoutedEventArgs e)
    {
        var resourceEnumerator = Application.Current.Resources.GetEnumerator();
        var fontChanger = new DispatcherTimer();
        fontChanger.Tick += (s, e) =>
        {
            if (!resourceEnumerator.MoveNext())
            {
                resourceEnumerator.Reset();
                resourceEnumerator.MoveNext();
            }

            if (resourceEnumerator.Value is FontFamily fontFamily)
            {
                TestBlock.FontFamily = fontFamily;
            }
        };
        fontChanger.Start();
    }
    ```
6. Run the app and observe the memory usage steadily increasing over time. Diagnostic tools will show the largest culprit for the growth being the `UnmanagedMemoryStream`:
    wpf-font-memory-leak
7. Change the font resources to have a Build Action of `Content`, and set the Copy to Output Directory property to `Copy if newer`, and re-run the test.
8. Observe that now the memory usage does not appear to grow over time.

  [1]: https://stackoverflow.com/a/31452979/181495

h3xds1nz:
Fixes #7289

## Description

Fixes memory leak when using fonts from embedded resources that are materialized onto `FontSource` -> `FontFamily`, etc.

The issue is basically in `FontSource.GetUnmanagedStream()`:
- We're passing around pure `UnmanagedMemoryStream` instance, but life without destructor is kind of hard.
    - This can be removed once `FontFileStream` from `DirectWriteForwarder` would be migrated to C#.

More context: The fix could be done by disposing the `UnmanagedMemoryStream` instantly and wrapping it in a new instance so it's cleared from the `List<Stream>` of `PackagePart` but I think it's nice to tie the lifetime together.

## Customer Impact

Customers will finally get their fix after 10 years.

https://stackoverflow.com/questions/31452443/wpf-textblock-memory-leak-when-using-font
https://stackoverflow.com/questions/49550759/wpf-load-font-from-resources
https://stackoverflow.com/questions/45297837/what-causes-unmanagedmemorystream-in-wpf-memory-leak
https://stackoverflow.com/questions/39157846/wpf-glyphs-control-font-uriits-slow-and-remains-in-memory

## Regression

Has been around since at least .NET 4.5.

## Testing

Using the repro in #7289; sample apps use.

## Risk

Low though we're passing around stuff onto C++/CLI, that's always awful.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/wpf/pull/9948)

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/PresentationCore/MS/internal/FontCache/FontSource.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationCore/MS/internal/FontCache/FontSource.cs
@@ -184,7 +184,7 @@ namespace MS.Internal.FontCache
                 {
                     WebResponse response = WpfWebRequestHelper.CreateRequestAndGetResponse(_fontUri);
                     fontStream = response.GetResponseStream();
-                    if (String.Equals(response.ContentType, ObfuscatedContentType, StringComparison.Ordinal))
+                    if (string.Equals(response.ContentType, ObfuscatedContentType, StringComparison.Ordinal))
                     {
                         // The third parameter makes sure the original stream is closed
                         // when the deobfuscating stream is disposed.
@@ -192,19 +192,19 @@ namespace MS.Internal.FontCache
                     }
                 }
 
-                UnmanagedMemoryStream unmanagedStream = fontStream as UnmanagedMemoryStream;
+                // We don't want any memory leaks
+                // TODO: Remove FinalizableUnmanagedStream once FontFileStream is migrated from C++/CLI.
+                if (fontStream is UnmanagedMemoryStream unmanagedMemoryStream)
+                    return new FinalizableUnmanagedStream(unmanagedMemoryStream);
 
-                if (unmanagedStream != null)
-                    return unmanagedStream;
-
+                // Convert the DeobfuscatingStream to byte[]; add it to our cache, dispose it
                 bits = StreamToByteArray(fontStream);
+                lock (s_resourceCache)
+                {
+                    s_resourceCache.Add(_fontUri, bits, false);
+                }
 
-                fontStream?.Close();
-            }
-
-            lock (s_resourceCache)
-            {
-                s_resourceCache.Add(_fontUri, bits, false);
+                fontStream.Close();
             }
 
             return ByteArrayToUnmanagedStream(bits);
@@ -358,7 +358,7 @@ namespace MS.Internal.FontCache
             Assembly fontResourceAssembly = Assembly.GetExecutingAssembly();
             ResourceManager rm = new($"{ReflectionUtils.GetAssemblyPartialName(fontResourceAssembly)}.g", fontResourceAssembly);
 
-            return rm?.GetStream($"fonts/{fontFilename}");
+            return rm.GetStream($"fonts/{fontFilename}");
         }
 
         #endregion Private Methods
@@ -370,6 +370,26 @@ namespace MS.Internal.FontCache
         //------------------------------------------------------
 
         #region Private Classes
+
+        /// <summary>
+        /// Calls <see cref="UnmanagedMemoryStream.Dispose"/> from the destructor.
+        /// </summary>
+        // TODO: Remove this once FontFileStream is migrated from C++/CLI.
+        private sealed class FinalizableUnmanagedStream : UnmanagedMemoryStream
+        {
+            private readonly UnmanagedMemoryStream _unmanaged;
+
+            internal unsafe FinalizableUnmanagedStream(UnmanagedMemoryStream unmanagedStream)
+            {
+                _unmanaged = unmanagedStream;
+                Initialize(_unmanaged.PositionPointer, _unmanaged.Length, _unmanaged.Length, FileAccess.Read);
+            }
+
+            ~FinalizableUnmanagedStream()
+            {
+                _unmanaged.Dispose();
+            }
+        }
 
         private class PinnedByteArrayStream : UnmanagedMemoryStream
         {


Symbai:
> Customers will finally get their fix after 10 years.

❤️❤️❤️

src/Microsoft.DotNet.Wpf/src/PresentationCore/MS/internal/FontCache/FontSource.cs
```diff
@@ -194,27 +194,27 @@ public UnmanagedMemoryStream GetUnmanagedStream()
...
+                    _resourceCache.Add(_fontUri, bits, false);
+                }
 
-                fontStream?.Close();
-            }
-
-            lock (_resourceCache)
-            {
-                _resourceCache.Add(_fontUri, bits, false);
+                fontStream.Close();
```
miloush, line 217: GetCompositeFontResourceStream() could return null from ResourceManager.GetStream() if I see correctly. Although when that happens, StreamToByteArray crashes already, so I guess removing the null check here is fine.

src/Microsoft.DotNet.Wpf/src/PresentationCore/MS/internal/FontCache/FontSource.cs
```diff
@@ -381,6 +381,26 @@ private Stream GetCompositeFontResourceStream()
 
         #region Private Classes
 
+        /// <summary>
+        /// Calls <see cref="UnmanagedMemoryStream.Dispose"/> from the destructor.
+        /// </summary>
+        /// <remarks>TODO: Remove this once FontFileStream is migrated from C++/CLI.</remarks>
```
miloush, line 387: Don't the XML comments get compiled into public docs? If so, perhaps simple comment would be better.

miloush:
LGTM

src/Microsoft.DotNet.Wpf/src/PresentationCore/MS/internal/FontCache/FontSource.cs
h3xds1nz: Ugh, no idea to be honest as the docs that are shown are always different than the ones in the project, so I do not think so.

If that's the case however, changing it to a plain comment is in order indeed.

src/Microsoft.DotNet.Wpf/src/PresentationCore/MS/internal/FontCache/FontSource.cs
h3xds1nz: Yeah, the null-check on that and in `GetCompositeFontResourceStream` on freshly created RM were weird to say at the least, though you shouldn't be able to init a `FontSource` that's a composite font without it actually existing iirc.

Fixing a possible NRE shall be handled in a different PR if we want to be on the safe side imho.

poutine70:
What is the status of this issue? Anyfix soon?

himgoyalmicro (approved):
LGTM

himgoyalmicro:
@h3xds1nz this PR is good to be merged. Can you please resolve the merge conflicts and the conversations.

src/Microsoft.DotNet.Wpf/src/PresentationCore/MS/internal/FontCache/FontSource.cs
h3xds1nz: Resolving, as I've noted, this "issue" was always there, shall be solved in a different PR; I'll probably cook up something once #9844 is merged.

src/Microsoft.DotNet.Wpf/src/PresentationCore/MS/internal/FontCache/FontSource.cs
h3xds1nz: Changed it to a plain comment, maybe if we get DWrite Forwarder rewritten before 2030, I'll get to remove it myself.

h3xds1nz:
@himgoyalmicro Resolved :)

P.S. I'd consider backporting this to NetFX given that you've backported the XPS one as well. For some apps this is rather a big issue.

himgoyalmicro (approved):

himgoyalmicro:
Thank you @h3xds1nz for your continuous contributions! 😄
Regarding the backporting, we will discuss it with the internal review team.

h3xds1nz:
@himgoyalmicro Thank you, maybe I'll fix some more stuff then 😀

Issue #9939, opened 2024-10-11
Fluent Theme: `ListView` inside `TabControl` has erratic resize behavior when scrolling vertically


bent-rasmussen:
### Description

When nesting a `ListView` inside a `TabControl`, it can exhibit strange resize behavior when scrolling vertically. Since RC2 the `ListView` seemingly wants to use as little horizontal space as possible but this leads to usability issues as can be seen below.

### Reproduction Steps

Create a .NET 9 WPF application and use the following markup in `MainWindow.xaml`:

```xaml
<Window
    x:Class="Fluent_ListViewScrollBarJumps.MainWindow"
    xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
    xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
    xmlns:d="http://schemas.microsoft.com/expression/blend/2008"
    xmlns:local="clr-namespace:Fluent_ListViewScrollBarJumps"
    xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006"
    Title="MainWindow"
    Width="800"
    Height="450"
    ThemeMode="Light"
    mc:Ignorable="d">
    <Grid Margin="10">

        <TabControl>
            <TabItem Header="Test">
                <ListView ItemsSource="{Binding Path=Items, Mode=OneWay}" />
            </TabItem>
        </TabControl>

    </Grid>
</Window>
 ````

In `MainWindow.xaml.cs` code-behind, use this code:

```csharp
using System.Collections.ObjectModel;
using System.Windows;

namespace Fluent_ListViewScrollBarJumps;

/// <summary>
/// Interaction logic for MainWindow.xaml
/// </summary>
public partial class MainWindow : Window
{
    public MainWindow()
    {
        InitializeComponent();

        void Add(int max)
        {
            var n = Random.Shared.Next(1, max);
            Items.Add(LoremIpsum[..n]);
        }

        for (var i = 0; i < 1_000; i++)
        {
            if (i <= 100)
            {
                Add(10);
            }
            else if (i <= 500)
            {
                Add(50);
            }
            else
            {
                Add(120);
            }
        }

        DataContext = this;
    }

[...]

PR #9950, opened 2024-10-14
Fixed ListView content alignment regression

dipeshmsft:
Fixes #  #9939 

## Description
In RC2, we modified some TemplateBinding's in Fluent styles, to allow developers to modify properties for default Fluent styles. While modifying ListView styles, the changes were made to make it similar to Aero2 style, but that caused a regression when using the ListView inside another container. Reverting that change to mitigate the issue.

## Customer Impact
Developers will get the correct ListView styles

## Regression
Technically no, but it is a regression in .NET 9 RC2

## Testing
Tested with sample repro application

## Risk
Minimal. The new styles are not in use. Reverts a change made in RC2

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/wpf/pull/9950)


dipeshmsft:
\backport to release\9.0

Issue #9802, opened 2024-09-18
[API Proposal]: Simplify Grid Column and Row Definitions


himgoyalmicro:
### Background and motivation

Following up on discussion in #5612 and #1739 this proposal is to simplify the definition syntax of RowDefinitions and ColumnDefinitions by:

- Allowing rows and columns within a Grid to be defined by a collection that is delimited by commas or space.
- Creating a Typeconvertor for ColumnDefinitionCollection and RowDefinitionCollection so they can process String as its input.

## Goal
The goal of this feature is to make Grid markup less verbose, allowing developers to create grids with simpler syntax.

### Current Syntax
Defining columns and rows is overly tedious, repetitive and not very productive, as is shown below:
```xaml
<Grid>
    <Grid.ColumnDefinitions>
          <ColumnDefinition Width="1*" />
          <ColumnDefinition Width="2*" />
          <ColumnDefinition Width="Auto" />
          <ColumnDefinition Width="*" />
          <ColumnDefinition Width="300" />
    </Grid.ColumnDefinitions>
    <Grid.RowDefinitions>
          <RowDefinition Height="1*" />
          <RowDefinition Height="Auto" />
          <RowDefinition Height="25" />
          <RowDefinition Height= "14" />
          <RowDefinition Height="20" />
    </Grid.RowDefinitions>
</Grid>
```
### Proposed Syntax
The same functionality as above with the proposed succinct syntax is shown below:
```xaml
<Grid ColumnDefinitions="1*, 2*, Auto, *, 300" RowDefinitions="1*, Auto, 25, 14, 20"> </Grid>
<!-- or -->
<Grid ColumnDefinitions="1* 2* Auto * 300" RowDefinitions="1* Auto 25 14 20"> </Grid>
```

## Scope
| Feature     | Priority      |
| -------------- | -------------- |
| Ability for rows and columns to be defined as a collection or list of values | Must |
| Content properties of RowDefinition and ColumnDefinition are set to Height and Width, respectively | Must |
| Original syntax will still be fully supported and functional | Must |
| Hot Reload will still be fully supported and functional | Must | 
| Include support for min/max height/width | Won't |
| Include support for data binding within definitions | Won't |

### API Proposal

Provide new typeconverters for `ColumnDefinitionCollection` and `RowDefinitionCollection`. 
This will allow us to convert string input into the corresponding collection.  

```diff
namespace System.Windows.Controls
{

[...]

PR #9956, opened 2024-10-15
Implemented shorthand syntax for ColumnDefinitions and RowDefinitions

himgoyalmicro:
Implements #9802 

## Description

It will simplify the definition syntax of RowDefinitions and ColumnDefinitions by:

- Allowing rows and columns within a Grid to be defined by a collection that is delimited
- Creating a Typeconvertor for ColumnDefinitionCollection and RowDefinitionCollection so they can process String as its input.

## Customer Impact

This feature will make Grid markup less verbose, allowing developers to create grids with simpler syntax.

## Regression

N/A

###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/wpf/pull/9956)


h3xds1nz:
Great to see this going forward!

Before checking the actual code, I'd have two things:

1) Can we not use `var`? Feels like JavaScript before 2015.
2) Formatting issues (50% spaces missing before parentheses), just let VS auto format the new files imho.

src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Controls/RowDefinitionCollectionConverter.cs
```diff
@@ -0,0 +1,117 @@
...
+        /// <param name="value"> The Thickness to convert. </param>
+        public override object ConvertFrom(ITypeDescriptorContext context, CultureInfo culture, object value)
+        {
+            if(value != null && value is string input){
+                IProvideValueTarget ipvt = context?.GetService(typeof(IProvideValueTarget)) as IProvideValueTarget;
+                Grid grid = ipvt?.TargetObject as Grid;
+                if(grid != null)
+                {
+                    var collection = new RowDefinitionCollection(grid); // Pass Grid instance
+                    var converter = new GridLengthConverter();
```
h3xds1nz, line 69: `GridLengthConverter`, as most converters, is state-less, if anything, it should have one `static readonly` instance as it was done in some of the other converters. With that being said, most converters expose `internal static `methods (and more will have with my additions), that are called from instance methods, which bypasses unnecessary boxing, will allow ref structs to be passed in. Same in `ColumnDefinitions`.

src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Controls/RowDefinitionCollectionConverter.cs
```diff
@@ -0,0 +1,117 @@
...
+
+                    TokenizerHelper th = new TokenizerHelper(input, culture);
```
h3xds1nz, line 71: Not sure what's the timeframe on this getting merged, but #9364 but introduce a non-allocating tokenizer, which could be taken advantage of here already if this is merged after.

src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Controls/RowDefinitionCollectionConverter.cs
```diff
@@ -0,0 +1,117 @@
...
+        /// <param name="culture"> The CultureInfo which is respected when converting. </param>
+        /// <param name="value"> The RowDefintionCollection to convert. </param>
+        /// <param name="destinationType">The type to which to convert the RowDefintionCollection instance. </param>
+        public override object ConvertTo(ITypeDescriptorContext context, CultureInfo culture, object value, Type destinationType)
+        {
+            ArgumentNullException.ThrowIfNull(value);
+            ArgumentNullException.ThrowIfNull(destinationType);
+            if (destinationType == typeof(string) && value is RowDefinitionCollection RowDefinitions)
+            {
+                var parts = new string[RowDefinitions.Count];
```
h3xds1nz, line 102: If anything, a `StringBuilder` would do a better job here, though it should be handled on stack / using `ArrayPool<char>` as fallback for longer ones (`DefaultInterpolatedStringHandler` impl for example). Same in `ColumnDefinitions`.

src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Controls/Grid.cs
```diff
@@ -297,6 +308,15 @@ public RowDefinitionCollection RowDefinitions
 
                 return (_data.RowDefinitions);
             }
+            set
+            {
+                if (value == null){
+                    _data.RowDefinitions = new RowDefinitionCollection(this);
+                    return;
+                }
+                _data ??= new ExtendedData();
```
h3xds1nz, line 317: I didn't check how _data gets initialized, but if this is supposed to be lazy-init, you get NRE if `value` was `null`.

src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Controls/RowDefinitionCollectionConverter.cs
```diff
@@ -0,0 +1,117 @@
...
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System.ComponentModel;
+using System.Windows.Controls;
+using System.Windows.Markup;
+using System.Globalization;
+using MS.Internal;
+
+#pragma warning disable 1634, 1691  // suppressing PreSharp warnings
```
h3xds1nz, line 11: Unless you guys are still using PreSharp (don't find that likely), I don't see why we need to suppress a warning about suppressing warnings.

src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Controls/Grid.cs
```diff
@@ -282,11 +283,21 @@ public ColumnDefinitionCollection ColumnDefinitions
 
                 return (_data.ColumnDefinitions);
             }
+            set
+            {
+                if (value == null){
+                    _data.ColumnDefinitions = new ColumnDefinitionCollection(this);
+                    return;
+                }
+                _data ??= new ExtendedData();
```
h3xds1nz, line 292: Same problem as on `RowDefinitions`.

src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Controls/ColumnDefinitionCollectionConverter.cs
```diff
@@ -0,0 +1,114 @@
...
+using System.Windows.Controls;
+using System.Windows.Markup;
+using System.Globalization;
+using MS.Internal;
+
+#pragma warning disable 1634, 1691  // suppressing PreSharp warnings
+
+namespace System.Windows.Controls
+{
+    internal class ColumnDefinitionCollectionConverter : TypeConverter
```
h3xds1nz, line 15: This should be `sealed` unless there's an explicit reason. The .NET policy is to seal types unless there's demand for creating derives (or it makes sense), as you can always unseal them later but not the other way around.

Given this is internal right now, it should definitely be `sealed`.

src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Controls/RowDefinitionCollectionConverter.cs
```diff
@@ -0,0 +1,117 @@
...
+        /// CanConvertFrom - Returns whether or not this class can convert from a given type.
+        /// </summary>
+        /// <returns>
+        /// bool - True if thie converter can convert from the provided type, false if not.
+        /// </returns>
+        /// <param name="context"> The ITypeDescriptorContext for this call. </param>
+        /// <param name="sourceType"> The Type being queried for support. </param>
+        public override bool CanConvertFrom(ITypeDescriptorContext context, Type sourceType)
+        {
+            return sourceType == typeof(string) || base.CanConvertFrom(context, sourceType);
```
h3xds1nz, line 29: Why the `base` call here? Same in `ColumnDefinitions`.

src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Controls/RowDefinitionCollectionConverter.cs
```diff
@@ -0,0 +1,117 @@
...
+        /// CanConvertTo - Returns whether or not this class can convert to a given type.
+        /// </summary>
+        /// <returns>
+        /// bool - True if this converter can convert to the provided type, false if not.
+        /// </returns>
+        /// <param name="context"> The ITypeDescriptorContext for this call. </param>
+        /// <param name="destinationType"> The Type being queried for support. </param>
+        public override bool CanConvertTo(ITypeDescriptorContext context, Type destinationType)
+        {
+            return destinationType == typeof(string) || base.CanConvertTo(context, destinationType);
```
h3xds1nz, line 42: Why the `base` call here? Same in `ColumnDefinitions`.

src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Controls/RowDefinitionCollectionConverter.cs
```diff
@@ -0,0 +1,117 @@
...
+        {
+            ArgumentNullException.ThrowIfNull(value);
+            ArgumentNullException.ThrowIfNull(destinationType);
+            if (destinationType == typeof(string) && value is RowDefinitionCollection RowDefinitions)
+            {
+                var parts = new string[RowDefinitions.Count];
+
+                for (int i = 0; i < RowDefinitions.Count; i++)
+                {
+                    parts[i] = RowDefinitions[i].Height.ToString();
```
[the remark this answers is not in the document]
himgoyalmicro: I fixed that using StringBuilder and GridLengthConverter.ToString function

src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Controls/RowDefinitionCollectionConverter.cs
himgoyalmicro: Fixed that

src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Controls/RowDefinitionCollectionConverter.cs
himgoyalmicro: Used the `StringBuilder` there

src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Controls/ColumnDefinitionCollectionConverter.cs
himgoyalmicro: Done

src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Controls/RowDefinitionCollectionConverter.cs
himgoyalmicro: Removed it

src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Controls/RowDefinitionCollectionConverter.cs
himgoyalmicro: Removed it

src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Controls/RowDefinitionCollectionConverter.cs
```diff
@@ -0,0 +1,121 @@
...
+        /// <exception cref="ArgumentException">
+        /// An ArgumentException is thrown if the object is not null and is not a valid type,
+        /// or if the destinationType isn't one of the valid destination types.
+        /// </exception>
+        /// <param name="typeDescriptorContext"> The ITypeDescriptorContext for this call. </param>
+        /// <param name="cultureInfo"> The CultureInfo which is respected when converting. </param>
+        /// <param name="value"> The Thickness to convert. </param>
+        public override object ConvertFrom(ITypeDescriptorContext typeDescriptorContext, CultureInfo cultureInfo, object value)
+        {
+            if (value != null && value is string input)
```
h3xds1nz, line 62: This can be just `if (value is string input)`, since pattern matching discards anything that doesn't represent an actual value for the type (and `null` isn't considered a value), even in compile-time defined types. It makes the code clearer once you're used to it. Same in the other converter.

That is also why in case of nullable value types, you can just use pattern matching to avoid the `.Value` and `.HasValue` syntax in the underlying block.

src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Controls/RowDefinitionCollectionConverter.cs
```diff
@@ -0,0 +1,121 @@
...
+        /// <param name="value"> The RowDefinitionCollection to convert. </param>
+        /// <param name="destinationType">The type to which to convert the RowDefinitionCollection instance. </param>
+        public override object ConvertTo(ITypeDescriptorContext typeDescriptorContext, CultureInfo cultureInfo, object value, Type destinationType)
+        {
+            ArgumentNullException.ThrowIfNull(value);
+            ArgumentNullException.ThrowIfNull(destinationType);
+            if (destinationType == typeof(string) && value is RowDefinitionCollection rowDefinitions)
+            {
+                char listSeparator = TokenizerHelper.GetNumericListSeparator(cultureInfo);
+                StringBuilder sb = new StringBuilder();
```
h3xds1nz, line 102: The `StringBuilder` can have a defaulted capacity to the most common scenario, which imho is gonna be about `5 * rowDefinitions.Count`. This will prevent several re-allocations of the buffer. Same in the other converter.

h3xds1nz:
@himgoyalmicro The changes look good to me now, I've left two more comments, along with the unresolved ones that's I think all from my side. Thank you.

src/Microsoft.DotNet.Wpf/src/PresentationFramework/ref/PresentationFramework.cs
```diff
@@ -5106,6 +5106,22 @@ public void Refresh() { }
         public override bool ShouldSerializeContent() { throw null; }
         public void StopLoading() { }
     }
+    internal partial class ColumnDefinitionCollectionConverter : System.ComponentModel.TypeConverter
```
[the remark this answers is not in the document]
himgoyalmicro: Removed it

src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Controls/Grid.cs
himgoyalmicro: Fixed it

src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Controls/Grid.cs
himgoyalmicro: Fixed it

src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Controls/RowDefinitionCollectionConverter.cs
himgoyalmicro: Fixed it

src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Controls/RowDefinitionCollectionConverter.cs
h3xds1nz: @himgoyalmicro I believe you can resolve this, it will require to spanify the GridLengthConverter function as well, so that's imho out of scope, I'll just do it later when that PR is merged.

himgoyalmicro: Thank you @h3xds1nz !!

h3xds1nz:
@himgoyalmicro

One question; did the internal setter not work for the `RowDefinitions`/`ColumnDefinitions` properties? As discussed in #9802 by miloush.

Speaking of the setter, this is just a massive nitpick, I believe the properties to unify the style could be just:
```csharp
_data ??= new ExtendedData();
_data.ColumnDefinitions = value ?? new ColumnDefinitionCollection(this);
```
And the getters then could be the same:
```csharp
_data ??= new ExtendedData();
_data.ColumnDefinitions ??= new ColumnDefinitionCollection(this);

return _data.ColumnDefinitions;
```

h3xds1nz:
It would be also good to add this, so you don't have to solo-close the issues after this would be eventually merged.

```
Fixes #9802
Fixes #1739
```

himgoyalmicro:
Closing this PR in favor of #10317 
Thank you @h3xds1nz for all the insightful contribution and comments. 😄
Can you please take a look at the other PR?

PR #9959, opened 2024-10-15
Use ThreadStatic fields over LocalDataStoreSlot for improved performance

h3xds1nz:
## Description

Replaces `LocalDataStoreSlot` with `[ThreadStatic]` fields for improved performance and increased type safety.
- Static cost of initializing a `LocalDataStoreSlot` is about `500 ns` and `184 B`, that will forever live.
- The retrieval process using `Thread.GetData` is about 40% more expensive.
- There is no reason to use `LocalDataStoreSlot` these days (though it uses `ThreadLocal<object?>` under the hood since .NET Core 3.0 afaik so the performance is not THAT bad, while `NetFX` still goes with the original impl).

In the other commits, I've just cleaned up some CAS remnants (`Critical`*/`Unsecure`* methods/properties) from `ComponentDispatcher`, that just unnecessarily complicate the code these days. Since the formatting was different style on each, I've unified that.

### 10 property retrievals before/after (already initialized)
| Method   |  Mean [ns] | Error [ns] | StdDev [ns] | Code Size [B] | Allocated [B] |
|--------- |----------:|-----------:|------------:|--------------:|--------------:|
| Original |  25.15 ns |   0.061 ns |    0.054 ns |         334 B |             - |
| PR_EDIT  |  14.07 ns |   0.157 ns |    0.139 ns |         130 B |             - |

## Customer Impact

Improved performance, increased code base quality and type safety for developers.

## Regression

No.

## Testing

Local build, sample app run and testing with all the modified types.

## Risk

Low.

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/wpf/pull/9959)

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/InterOp/HwndKeyboardInputProvider.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/InterOp/HwndKeyboardInputProvider.cs
@@ -510,8 +510,8 @@ namespace System.Windows.Interop
         internal void ProcessKeyAction(ref MSG msg, ref bool handled)
         {
             // Remember the last message
-            MSG previousMSG = ComponentDispatcher.UnsecureCurrentKeyboardMessage;
-            ComponentDispatcher.UnsecureCurrentKeyboardMessage = msg;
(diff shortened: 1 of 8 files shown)


h3xds1nz:
Rebased and resolver merge conflicts.

PR #9981, opened 2024-10-22
Optimize working with EllipseGeometry/RectangleGeometry, reduce allocs

h3xds1nz:
## Description

Removes some unnecessary `gen0` allocations when working with `EllipseGeometry`/`RectangleGeometry` (such as adding them to a `PathGeometry` or combining geometries which serializes them (`Point` array alloc) and when rendering `TickBar`. Also introduces `internal` constructors for `PathFigure`/`PathGeometry` that can be used with inline arrays.

Also changes some `private static` fields to constants, decreasing static allocations.

The savings are not that massive right now, but it will allow for further optimizations (and there's a lot of space for it).

### Define geometry, add to group, get bounds

| Method       | Mean [ns]  | Error [ns] | StdDev [ns] | Gen0   | Code Size [B] | Allocated [B] |
|---------------- |-----------:|-----------:|------------:|-------:|--------------:|--------------:|
| Original | 9,371.1 ns |   67.37 ns |    56.25 ns | 0.4578 |       9,483 B |        7848 B |
| PR_EDIT | 8,986.0 ns |   55.25 ns |    48.98 ns |  0.4120    |   9,120 B  |        7120 B |

<details>
<summary>Benchmark code</summary>

```csharp
public Rect Original()
{
  PathGeometry abcd =  new PathGeometry();
  RectangleGeometry rectangle =   new RectangleGeometry(s_myRect, 10, 10);
  EllipseGeometry ellipse = new EllipseGeometry(s_myPoint, 10, 10);

  // uses GetPathFigureCollection
  abcd.AddGeometry(rectangle);
  abcd.AddGeometry(ellipse);

  // Retrieve bounds, another alloc
  Rect myBounds = rectangle.Bounds;
  Rect myEllipse = ellipse.Bounds;

  myBounds.Union(myEllipse);

  return myBounds;
}
```

</details>

## Customer Impact

## Regression

No.

## Testing

Local build, sample apps, usage of types.

## Risk

Should be low, all the changes were done with using stack-allocated memory

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Media/EllipseGeometry.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Media/EllipseGeometry.cs
@@ -2,25 +2,10 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
-//                                             
-
-using System;
 using MS.Internal;
-using System.ComponentModel;
-using System.ComponentModel.Design.Serialization;
-using System.Reflection;
-using System.Collections;
-using System.Text;
-using System.Globalization;
-using System.Windows.Media;
 using System.Windows.Media.Composition;
-using System.Windows;
-using System.Text.RegularExpressions;
-using System.Windows.Media.Animation;
-using System.Diagnostics;
-using System.Runtime.InteropServices;
-using System.Security;
-using SR=MS.Internal.PresentationCore.SR;
+
+using SR = MS.Internal.PresentationCore.SR;
 
 namespace System.Windows.Media 
 {
--- a/src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Media/EllipseGeometry.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Media/EllipseGeometry.cs
@@ -19,9 +19,7 @@ namespace System.Windows.Media
         /// <summary>
         ///
         /// </summary>
-        public EllipseGeometry()
-        {
-        }
+        public EllipseGeometry() { }
 
         /// <summary>
         /// Constructor - sets the ellipse to the paramters with the given transformation
@@ -30,7 +28,7 @@ namespace System.Windows.Media
         {
             if (rect.IsEmpty) 
             {
-                throw new System.ArgumentException(SR.Format(SR.Rect_Empty, "rect"));
+                throw new ArgumentException(SR.Format(SR.Rect_Empty, nameof(rect)));
             }
 
             RadiusX = (rect.Right - rect.X) * (1.0 / 2.0);
@@ -41,10 +39,7 @@ namespace System.Windows.Media
         /// <summary>
         /// Constructor - sets the ellipse to the parameters
         /// </summary>
-        public EllipseGeometry(
-            Point center, 
-            double radiusX, 
-            double radiusY)
+        public EllipseGeometry(Point center, double radiusX, double radiusY)
         {
             Center = center;
             RadiusX = radiusX;
@@ -54,11 +49,7 @@ namespace System.Windows.Media
         /// <summary>
         /// Constructor - sets the ellipse to the parameters
         /// </summary>
-        public EllipseGeometry(
-            Point center,
-            double radiusX,
-            double radiusY,
-            Transform transform) : this(center, radiusX, radiusY)
+        public EllipseGeometry(Point center, double radiusX, double radiusY, Transform transform) : this(center, radiusX, radiusY)
         {
             Transform = transform;
         }
@@ -81,14 +72,13 @@ namespace System.Windows.Media
                 if (transform == null || transform.IsIdentity) 
                 {
                     Point currentCenter = Center;
-                    Double currentRadiusX = RadiusX;
-                    Double currentRadiusY = RadiusY;
+                    double currentRadiusX = RadiusX;
+                    double currentRadiusY = RadiusY;
 
-                    boundsRect = new Rect(
-                        currentCenter.X - Math.Abs(currentRadiusX),
-                        currentCenter.Y - Math.Abs(currentRadiusY),
-                        2.0 * Math.Abs(currentRadiusX),
-                        2.0 * Math.Abs(currentRadiusY));
+                    boundsRect = new Rect(currentCenter.X - Math.Abs(currentRadiusX),
+                                          currentCenter.Y - Math.Abs(currentRadiusY),
+                                          2.0 * Math.Abs(currentRadiusX),
+                                          2.0 * Math.Abs(currentRadiusY));
                 }
                 else
                 {
@@ -99,9 +89,7 @@ namespace System.Windows.Media
                     // it's easier to let unmanaged code do the work for us.
                     //
 
-                    Matrix geometryMatrix;
-
-                    Transform.GetTransformValue(transform, out geometryMatrix);
+                    Transform.GetTransformValue(transform, out Matrix geometryMatrix);
 
                     boundsRect = EllipseGeometry.GetBoundsHelper(
                         null /* no pen */,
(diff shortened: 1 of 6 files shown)


h3xds1nz:
@robert-abeo I wasn't originally even gonna respond, but here you go.

1) The team is free to close my PRs if they deem them unfit, I'm sure they do not need your permission.
2) There's a saying you probably are not familiar with (this time, not your fault); it goes roughly as "he opened a restaurant but people started coming in". It's a nice but a very true oxymoron, much of like what you've said here.

3) You must be the reason x86_64 architecture still starts up in real mode. Every bug fix is a risk, every upstream level change (JIT, BCL) risks breaking all downstream consumers (e.g production apps) for, as you say, no benefit. Why even introduce new types, why even introduce ref structs, ability to `stackalloc` memory, and actually using those new features under the hood? I'd start at the core, that's dotnet/runtime. For some reason, they wanted `NET` unlike `NetFX` to be performance driven. You should go ahead and create an issue to tell everyone how you feel about the fact that about 70% of the BCL has already been rewritten using new language and runtime features for often low change-to-change benefit but overall great results, since most changes are often a building block for the next one.
Most optimizations are incremental, sometimes you need to shave off 1 kB out of 8 kB so then you can get additional 5 kB down and 60% perf improvement in the next; not everytime you get x7 perf improvement and 10% of previously allocated memory the first time around. But why do I explain it to you? Besides everyone else's, I'm just breaking the platform my company builds apps on and about 40% of the infrastructure depends on.

4) I've resolved your conversation because it has made zero sense. The WPF codebase has obviously had changing style over the years, some parts are written with pure C/C++ conventions (member names, function names, etc.) even though they do not incorporate or relate to any unmanaged calls simply because developers writing them were used to ATL/MFC code style, some parts are generated with that same conventions; some follow early `NetFX` style (namings etc.), then there are parts written after the release of C# 3.0 that incorporate `var` (there are few but there are some) which is generally discouraged (nobody wants to think about JavaScript), and so much more variations that I could spend days on.

There are sometimes `static` functions that start with `_` (hello C-style, I'd like you see you write this in a .NET project); members that have `m_` prefix; there are constants that have `c_` prefix, there are constants that have `_` prefix, and then there are finally constants in `PascalCase` as it is recommended by the guidelines these days.

But guess what, shocker, most of the time, the conventions vary not just in a namespace, but in a single file, and in the particular file (EllipseGeometry.cs) where you've highligted the line 72; I'd recommend checking line 360; 205; 157 (original lines, those were not edited since that part of WPF was opensourced), and tell me, please, what is the coding style? Why are those function definitions as long as the particular `ctor` I've auto-formatted (yes, an analyzer, imagine that). Guess opposite of so called coding style in one file. If anything, I've just unified it.

If we wanna have a look at so called "Microsoft's standard code format rules", guess those would be mostly incorporated in standard tools such as Visual Studio and Roslyn analyzers, correct? You obviously weren't thinking about incorporating random Rider rules and line breaks for 1024x768 CRT monitors, calling them current "Microsoft's standard code format rules"? Plus there's of course runtime coding style which I can't really seem to have violated. https://github.com/dotnet/runtime/blob/main/docs/coding-guidelines/coding-style.md

Not even in C code bases you'd see auto formatters mostly to wrap 1 per line. Plus, as of like 4 years ago, there's `--ignore-revs-file with` with `git blame` where you put those 2-3 big auto format commits and nobody gets hurt. So, for example:

- I've renamed constants (where some were defined fields and some `const`), both prefixed with `c_` as PascalCase
    - That would fall under `IDE1006`
- I've used pattern matching, that would be `IDE0019`
- I've inlined out variables, that would be C# 7? and `IDE0018`
- I've simplified some object inits into `new()` on right-hand side, that would be `IDE0090`
- I've simplified `PathGeometryData` init, that was `IDE0017`
- I've renamed some data types (e.g. `UInt32`) to language keywords, that's IDE0049
- I've simplified some collection initializations so that would be `IDE0300`
    - Aas that's the only way to get Roslyn to emit `InlineArray` parameters, not because I'm collection expressions fan
- I've changed compile-time constants that are defined as "static private byte" (not even `readonly` as it wasn't available at the time of writing) into actual constant definitions (`const`).
    - There is no compile-time checking without `readonly` so anyone could overwrite it in the future
    - It consumes unnecessary memory (perf opt)
    - JIT cannot treat them as constants
    - `static` first is mostly C-style, not really the newest C#
- I've changed `Point * name` (don't know what this is) and `Point *name` (C-Style) to `Point* name` - VS auto format
- I've added whitespaces between numbers on multiplication and between parentheses and code blocks
   - e.g. `fixed(byte` - VS auto format
- At line 384 and 390 (original lines) in `RectangleGeometry`, what is the coding style? Confused. Same issue as you highlighted above with `EllipseGeometry`. One is that way, second time it's the other way. This time it's an `if` condition.

And yes, it is absolutely fine to put parameters onto the next line when a single one gets too long!

Lastly, given the time and effort you must have put into your two comments, you could have put up PRs fixing half of your Fluent-styling issues instead and attempt to help everyone using WPF's new Fluent theme for a much better experience incorporating it into their applications. I'd even give you a thumbs up as it shows effort! But I understand, it is always easier to randomly criticise from distance. Please, instead of responding to my reaction here, I'd recommend creating a discussion, name it "Performance optimizations PRs should be banned as the team has not been 100% focused on the issues I have created" if you want and have a go at it. But I shall not take this from a drive-by shooter. Fixing community-reported bugs and alongside doing some optimizations does not make me the ultimate authority but it surely gives me more ground to have an educated opinion on this, and given your inability to even go through one file before saying how I've violated the codestyle that's violated 15 times in that file to begin with, it definitely doesn't hold a high value to me as a constructive feedback.

I couldn't be more constructive, simply because, well, I "focus on bigger gains" but without preparing the ground at the down level, with PRs like these, you can never get there. One last example, let's see if you can comprehend it: `Span<T>`.

With that being said, I'll continue on "spamming" when I feel like it unless the team's position is to close my PRs, thank you. There's 1.2k open issues and for about 80% of what I'm submitting it is contributing to fixing those issues but since it is only partially as the fixes require ground preparation often times, it does not justify slapping "Fixes #" on it.

ThomasGoulet73:
I agree with some of what @robert-abeo said, though I really disagree with the tone used.

I'll open an issue to discuss revamping contributor guidelines but I'll leave some thoughts here:

For now, I'll say that I'm in favor of the performance enhancement in this PR but I'm against the unrelated changes around it. In my experience, this PR would have been refused in a lot of other repos (And probably in dotnet/runtime and WPF follows most of the guidelines of dotnet/runtime). Here's some things I noticed:
1. There's "+233 -488" of diff for a PR that could have ~40-50 lines of diff to achieve the same performance result.
2. Just in the first file, there is 125 lines of diff of unrelated changes: https://github.com/dotnet/wpf/pull/9981/files#diff-89ae88ad0a25ce6473bd237409a759def3e5a3c5386f30152befaec9fd0b5d6eL5-R130. IMO that's way beyond "reasonable" refactoring for a performance PR.
3. [This code](https://github.com/dotnet/wpf/pull/9981/files#diff-c4563a8e1c51aa43bcb283d7e3d336676cd261722806d4a452c083dfa1b6543bL143) is not against the guidelines so why rename the variable to your prefered name ? I personally use "pXYZ", am I supposed to change it back from "ptrXYZ" to "pXYZ" if I want to modify this code in 6 months ?

The last bullet point [here](https://github.com/dotnet/wpf/blob/e16222f888d89a3c06efd8fb252f67ed30f39050/Documentation/acceptance_criteria.md#general-guidelines) says "DO NOT submit pure formatting/typo changes to code that has not been modified otherwise." so technically some of the code in this PR is against the guideline.

This is a very old and very big code base, you'll find code that you would've written differently than what was written 10 years ago and you'll find code in the future that you would've written differently than what was written today.

I'm not trying to discourage you from contributing since your PRs are valuable, I'm just trying to make sure they are done in the best way possible.

miloush:
I criticized other people for mixing actual changes and formatting changes in the past, in much smaller PRs, because it makes it difficult to review (I particularly remember one that could have been single change but was a few extra lines with formatting changes, with which the team agreed and asked the author to only do the functional change, though I cannot seem to find that PR now), so it sounds fair to uphold that view: Yes, I would certainly prefer if we could keep styling changes to minimum, preferably matching the code around. The repercussions from my side, however, is probably just unwillingness to spend time reviewing PRs (which otherwise tend to be straightforward in case of @h3xds1nz).

But I also agree that the team is free to do what they see fit. So far these PRs seem to be getting merged. Moreover, it was indicated that performance contributions are now a priority: https://github.com/dotnet/wpf/discussions/9920.

I wouldn't want to lose contributors like @h3xds1nz by making what they do not enjoyable, so I hope we can find some balance between what the team is willing to accept and authors to produce while allowing reviewers to help out.

rampaa:
> That means do not change things that don't need to be changed because every change is a risk.

This is such an unhelpful comment. If you have concerns about specific code changes and their ramifications, by all means, please name them, that's actually very helpful for everyone involved. But if you don't actually have anything to contribute, maybe refrain from discouraging people who actually contribute from contributing. Thankfully, you don't seem to be in a position to decide (micro) optimizations are not needed, and people who actually get to decide those things have been merging those changes, that probably should have been your cue about the fact that your view about micro optimizations is not shared by others, especially by those who actually get to decide those things.

> Some might disagree with that and I understand why but it's Microsoft's policy for the most part.

"Every change involves some risks, so we will avoid micro optimizations" is certainly not a Microsoft policy. 200+ pages long `Performance Improvements in .NET X` posts are a good testament to that. AFAICS, squeezing every bit of performance where it's possible has been the general direction of .NET.

miloush:
It would certainly help some contributors and avoid all discussions if we introduced an .editorconfig and formatted the whole repo at once. @pchaurasia14 any thoughts on that?

I am less keen on renaming things without good reason, it makes it difficult to track things across versions and does have the potential to break applications without good reason. Obviously using reflection is explicitly not supported, but we should have a reason to break it more than a personal preference or code style.

h3xds1nz:
Your observation is correct, I did not reaxt to you this time. And I shall not, @rampaa has said it for me.

ThomasGoulet73:
I opened dotnet/wpf#9994 to continue this discussion without poluting this PR.

(@h3xds1nz, I'll respond to your comment in this issue)

PR #10008, opened 2024-10-28
Use ReadOnlySpan<HwndWrapperHook> instead of a temporary heap array

h3xds1nz:
## Description

Avoids creation of temporary heap array when initiating `HwndWrapper`, contributing to startup performance.

- I've also replaced `""` with `string.Empty` to use singletons in the API calls.
- I've also modified the hwnd class-name creation to avoid some allocations, logic of course stays the same. (to be honest, the best would be to cache the assembly name in case you're creating multiple windows, this needs to be retrieved everytime though #8567 has optimized this to at least be just 1 call per invocation but could be zero after 1st retrieval, really).
- Tons of unnecessary usings from touched files were removed.
- Also removes `HWND` struct as a bonus (one of many, this one is unsed and also not particularly perfect).

## Customer Impact

## Regression

No.

## Testing

Local build, app run.

## Risk

Low.

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/Common/Graphics/wgx_render.cs
+++ b/src/Microsoft.DotNet.Wpf/src/Common/Graphics/wgx_render.cs
@@ -721,13 +721,7 @@ namespace MS.Internal
         // TRUE if the effect supports multi-resolution
 
         [MarshalAs(UnmanagedType.Bool)] bool fSupportMultiResolution;
-};
-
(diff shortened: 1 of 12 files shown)


zlatanov:
> * I've also replaced `""` with `string.Empty` to use singletons in the API calls.

I would avoid doing such changes. `string.Empty` and `""` have been treated exactly the same by the runtime for a while now.

PR #9914, opened 2024-10-08
Remove NetCoreReference

JeremyKuhne:
This removes the `NetCoreReference` infrastructure and replaces it with `DefaultReferenceExclusion` to remove the one problematic implicit reference to WindowsBase.

Manually picking references was blocking using System.Private.Windows.Core from the WinForms repo. Not having this also greatly simplifies the projects.

This also tweaks the solution to add folders docs and eng items.

Fixes #9168
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/wpf/pull/9914)

PR #10023, opened 2024-10-31
Enable xunit testing in VS with the dispatcher

JeremyKuhne:
Enable xunit testing in VS with the dispatcher
This change enables writing and debugging unit tests in the Test Explorer in Visual Studio. I added a new test project for WindowsBase that has a unit test that launches the SplashScreen as an example.

In order to make this happen I had to reference the xunit.stafact package which introduced a number of challenges to overcome. The package has a reference to the desktop sdk and needs to be used in a project that targets `net10.0-windows`, not `net10.0`. To make everything work seamlessly I had to:

- Strip the platform reference
- Apply actual versions from the sdk
- Get rid of AnyCPU and explictly use x86 (everything has a native dependency)
- Enable transitive copying of references for the unit test projects
- Target the 10.0 SDK (would get version confusion otherwise)
- Add an override setting for WinForms which still targets the 9.0 SDK
- Remove other targets that were trying to fixup WindowsBase references from the .NET SDK

This builds on https://github.com/dotnet/wpf/pull/9914, which is included.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/wpf/pull/10023)


eng/Versions.props
```diff
@@ -121,5 +124,6 @@
   </PropertyGroup>
   <PropertyGroup>
     <MicrosoftDotNetWpfDncEngVersion>10.0.0-alpha.1.24474.3</MicrosoftDotNetWpfDncEngVersion>
+    <MicrosoftWindowsCsWin32PackageVersion>0.3.106</MicrosoftWindowsCsWin32PackageVersion>
```
JeremyKuhne, line 127: Fyi: Not actually using this in this PR. Now that I have the tests figured out, I'll be creating another updated PR converting the SplashScreen to using CsWin32 and the shared assembly coming from WinForms. Noticed I had left it in after I created the PR.

eng/Versions.props
```diff
@@ -43,6 +45,7 @@
     <SystemSecurityCryptographyXmlPackageVersion>10.0.0-alpha.1.24528.7</SystemSecurityCryptographyXmlPackageVersion>
     <SystemSecurityPermissionsPackageVersion>10.0.0-alpha.1.24528.7</SystemSecurityPermissionsPackageVersion>
     <SystemWindowsExtensionsPackageVersion>10.0.0-alpha.1.24528.7</SystemWindowsExtensionsPackageVersion>
+    <SystemRuntimeCompilerServicesUnsafePackageVersion>6.0.0</SystemRuntimeCompilerServicesUnsafePackageVersion>
```
JeremyKuhne, line 48: This is also related to using the shared assembly from WinForms, not actually used yet.

src/Microsoft.DotNet.Wpf/tests/UnitTests/WindowsBase.Tests/Resources/Needle.png
JeremyKuhne, line 1: Used AI to generate this so there would be no need to list the source.

src/Microsoft.DotNet.Wpf/tests/UnitTests/WindowsBase.Tests/System/Windows/SplashScreenTests.cs
```diff
@@ -0,0 +1,20 @@
...
+
+public class SplashScreenTests
+{
+    [Fact]
+    public void Constructor_NullImageSource_ThrowsArgumentNullException()
+    {
+        Assert.Throws<ArgumentNullException>(() => new SplashScreen(null));
+    }
+
+    [WpfFact]
```
JeremyKuhne, line 14: Here is the sample `[WpfFact]` test. This will pop and close the splash screen. With a normal `[Fact]`  it would fail when attempting to dispatch.

JeremyKuhne:
Clearly there are some CI issues, but this does work locally. 🤷🏼 I'll be digging through that tomorrow. :)

src/Microsoft.DotNet.Wpf/src/WindowsBase/LibraryAssemblyInfo.cs
```diff
@@ -0,0 +1,7 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System;
+using System.Runtime.CompilerServices;
+
+[assembly: InternalsVisibleTo("WindowsBase.Tests, PublicKey=00000000000000000400000000000000")]
```
h3xds1nz, line 7: Is there a reason for creating additional file instead of using established `OtherAssemblyAttrs`?

JeremyKuhne: It is included in the `-ref` project.

JeremyKuhne:
The build machine is failing to load `System.Text.Json` when restoring. I can't reproduce this locally unfortunately.


Some general notes about why I updated to build against the .NET 10 SDK:

Despite installing and running the 9.0 SDK in the `.dotnet` folder, projects actually reference 10.0 via:

https://github.com/dotnet/wpf/blob/5c87afff67404fa36431173683bc498bd8998022/eng/Versions.props#L29

   and

https://github.com/dotnet/wpf/blob/5c87afff67404fa36431173683bc498bd8998022/eng/WpfArcadeSdk/tools/RuntimeFrameworkReference.targets#L15-L22

That means the output assemblies have a target framework of `[assembly: TargetFramework(".NETCoreApp,Version=v9.0", FrameworkDisplayName = ".NET 9.0")]` but all of the assembly references to the platform are `10.0`.

Anything that brings in a reference with be keyed off of `net9.0`, but we have all `10.0` assemblies and assembly references. We can fix these up via a bit of whack-a-mole, but I'm not even sure what the runtime implications are. I'm experimenting a bit more to see if I can get all of the references implicitly pulled into the tests to be `10.0`.

src/Microsoft.DotNet.Wpf/src/WindowsBase/LibraryAssemblyInfo.cs
h3xds1nz: Ah, I've overlooked we're referencing the same file here for ref. Makes sense, thanks.

Makes me wonder whether it's a correct approach to reference it though but it doesn't really matter.

JeremyKuhne:
@lonitra could not repro the CI failure locally either.

JeremyKuhne:
The VS version on the CI machines needs to be updated to unblock this as the SDK we're trying to use has a newer version of MSBuild with an updated System.Text.Json dependency. Following up internally.

eng/WpfArcadeSdk/tools/ApiCompat.targets
```diff
@@ -84,7 +84,7 @@
     <!-- If DotNetTool is undefined, we default to assuming 'dotnet' is on the path -->
     <DotNetTool Condition="'$(DotNetTool)' == ''">dotnet</DotNetTool>
 
-    <_ApiCompatPath Condition="'$(MSBuildRuntimeType)' == 'core'">$(NuGetPackageRoot)\microsoft.dotnet.apicompat\$(MicrosoftDotNetApiCompatVersion)\tools\netcoreapp2.1\Microsoft.DotNet.ApiCompat.dll</_ApiCompatPath>
+    <_ApiCompatPath Condition="'$(MSBuildRuntimeType)' == 'core'">$(NuGetPackageRoot)\microsoft.dotnet.apicompat\$(MicrosoftDotNetApiCompatVersion)\tools\net10.0\Microsoft.DotNet.ApiCompat.dll</_ApiCompatPath>
```
ViktorHofer, line 87: This seems to be the old apicompat which we delete from Arcade more than a year ago. There's definitely not a package out there with a net10.0 TFM in it.

eng/WpfArcadeSdk/tools/RuntimeFrameworkReference.targets
```diff
@@ -3,6 +3,15 @@
...
     usually lags behind the version that WPF gets from CoreFx packages
   -->
+
+  <PropertyGroup>
+    <UpdatedTargetingPackVersion Condition="'$(MicrosoftNETCoreAppRefVersion)'!=''
+                                            And '$(NoTargets)'!='true'
+                                            And '$(TargetFrameworkIdentifier)' == '.NETCoreApp'
+                                            And $([MSBuild]::VersionGreaterThanOrEquals('$(TargetFrameworkVersion)', '3.0'))
+                                            And '$(MSBuildProjectExtension)'!='.vcxproj'">$(MicrosoftNETCoreAppRefVersion)</UpdatedTargetingPackVersion>
+  </PropertyGroup>
```
ViktorHofer, line 13: You should be able to revert all the changes in this file and merge latest main into your branch. My PR that fixed this got merged yesterday.

eng/WpfArcadeSdk/tools/ApiCompat.targets
JeremyKuhne: Odd. With this change it seems to work.

JeremyKuhne:
Finally figured out how to get everything happy here. @ViktorHofer's fixes did enable me to backpedal on changing to 10.0. I struggled a bit with ARM until I figured out there was a race condition created by having the platform with multiple casings (`ARM64` and `arm64`). 👀

eng/WpfArcadeSdk/tools/RuntimeFrameworkReference.targets
```diff
@@ -3,6 +3,15 @@
     Explicitly import Microsoft.NETCore.Platforms to prevent using the one from the SDK, which
     usually lags behind the version that WPF gets from CoreFx packages
   -->
+
+  <PropertyGroup>
+    <UpdatedTargetingPackVersion Condition="'$(MicrosoftNETCoreAppRefVersion)'!=''
```
ViktorHofer, line 8: As mentioned, all the changes in this file should be reverted.

JeremyKuhne: Sorry, conflated this with another change. Reverted.

JeremyKuhne:
The key changes in this PR:

- Makes the ref projects only reference the ref projects and makes them AnyCPU
- Makes all implementation projects explicitly x86, x64, and arm64 (outside of PresentationBuildTasks, which remains AnyCPU)
- Strips out references to the desktop platform in the unit tests to ensure we test what we build and avoid version conflicts
- Establishes the pattern for using StaFact, which enables tests that require the dispatcher (UI)

(This took a while to finalize due to other infra issues.)

eng/pipeline-pr.yml
```diff
@@ -167,13 +167,14 @@ jobs:
...
       - script: eng\scripts\cibuild.cmd
-          -configuration $(_BuildConfig) 
+          -configuration $(_BuildConfig)
           -prepareMachine
           $(_PublishArgs)
           $(_SignArgs)
           $(_OfficialBuildIdArgs)
           $(_PlatformArgs)
           $(_InternalRuntimeDownloadArgs)
+          /bl:$(Build.SourcesDirectory)\artifacts\log\$(_BuildConfig)\BuildSrc-$(_Platform).binlog
```
ViktorHofer, line 177: Why is this needed? The Arcade default that most/all(?) repos follow is to just produce a Build.binlog file.

eng/pipeline-pr.yml
```diff
@@ -183,26 +184,26 @@ jobs:
...
-          -configuration $(_BuildConfig) 
+          -configuration $(_BuildConfig)
           -prepareMachine
           $(_PublishArgs)
           $(_SignArgs)
           $(_OfficialBuildIdArgs)
           $(_PlatformArgs)
           $(_InternalRuntimeDownloadArgs)
-          /bl:$(Build.SourcesDirectory)\artifacts\log\$(_BuildConfig)\Test.binlog
+          /bl:$(Build.SourcesDirectory)\artifacts\log\$(_BuildConfig)\Test-$(_Platform).binlog
```
ViktorHofer, line 194: Same question as above but for the Test.binlog.

src/Microsoft.DotNet.Wpf/cycle-breakers/PresentationFramework/PresentationFramework.cs
```diff
@@ -512,7 +512,7 @@ public FrameworkContentElement() { }
         [System.Windows.LocalizabilityAttribute(System.Windows.LocalizationCategory.NeverLocalize)]
         public string Name { get { throw null; } set { } }
         public bool OverridesDefaultStyle { get { throw null; } set { } }
-        public new System.Windows.DependencyObject Parent { get { throw null; } }
+        public System.Windows.DependencyObject Parent { get { throw null; } }
```
ViktorHofer, line 515: What made this change necessary?

src/Microsoft.DotNet.Wpf/src/DirectWriteForwarder/DirectWriteForwarder.vcxproj
```diff
@@ -116,6 +116,11 @@
     </ItemGroup>
   </Target>
   <Import Project="$(VCTargetsPath)\Microsoft.Cpp.targets" />
+  <!--<ItemGroup>
+    <KnownAppHostPack Update="@(KnownAppHostPack)">
+      <AppHostPackVersion>10.0.100-alpha.1.24509.5</AppHostPackVersion>
+    </KnownAppHostPack>
+  </ItemGroup>-->
```
ViktorHofer, line 123: Remove?
```suggestion
```

src/Microsoft.DotNet.Wpf/tests/UnitTests/Directory.Build.targets
```diff
@@ -24,15 +23,19 @@
...
-          AfterTargets="ResolveTargetingPackAssets"
-          Returns="@(Reference)">
+  <Target Name="RemoveDesktopFramework" AfterTargets="ProcessFrameworkReferences" BeforeTargets="FixupRuntimeFrameworkVersions">
     <ItemGroup>
-      <Reference Remove="@(Reference)" Condition="'%(Reference.AssemblyName)' == 'WindowsBase' And '%(Reference.FrameworkReferenceName)' == 'Microsoft.NETCore.App'" />
+      <RuntimeFramework Remove="@(RuntimeFramework)" Condition="'%(RuntimeFramework.Identity)' == 'Microsoft.WindowsDesktop.App'"/>
     </ItemGroup>
   </Target>
 
+  <Target Name="RemoveTransitiveDesktopFramework" AfterTargets="ResolvePackageAssets" BeforeTargets="AddTransitiveFrameworkReferences">
```
ViktorHofer, line 36: Please add a comment describing why this is necessary.

src/Microsoft.DotNet.Wpf/tests/UnitTests/WindowsBase.Tests/WindowsBase.Tests.csproj
```diff
@@ -0,0 +1,45 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+  <PropertyGroup>
+    <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
+    <Nullable>enable</Nullable>
+    <Platforms>x64;x86;arm64</Platforms>
+    <RootNamespace />
+    <TargetFramework Condition="!$(TargetFramework.Contains('windows'))">$(TargetFramework)-windows</TargetFramework>
```
ViktorHofer, line 7: This looks weird. Where is TargetFramework set and when does it not get set to the `-windows` suffix?

src/Microsoft.DotNet.Wpf/tests/UnitTests/WindowsBase.Tests/WindowsBase.Tests.csproj
```diff
@@ -0,0 +1,45 @@
...
+      <DesignTime>True</DesignTime>
+      <AutoGen>True</AutoGen>
+      <DependentUpon>Resources.resx</DependentUpon>
+    </Compile>
+    <Compile Update="Resources.Designer.cs">
+      <DesignTime>True</DesignTime>
+      <AutoGen>True</AutoGen>
+      <DependentUpon>Resources.resx</DependentUpon>
+    </Compile>
+  </ItemGroup>
```
ViktorHofer, line 32: The recommendation is to not check Resources.Designer.cs files in and let them get auto-generated.

This can be achieved by adding the `GenerateSource=true` metadata to the EmbbededResource items below.

src/Microsoft.DotNet.Wpf/cycle-breakers/PresentationFramework/PresentationFramework.cs
JeremyKuhne: The refs are no longer referring to the implementation assemblies (as they were in some cases).

eng/pipeline-pr.yml
ViktorHofer: Yes, please undo

src/Microsoft.DotNet.Wpf/tests/UnitTests/Directory.Build.targets
JeremyKuhne: It is above, but I'll call it out.

src/Microsoft.DotNet.Wpf/tests/UnitTests/WindowsBase.Tests/WindowsBase.Tests.csproj
JeremyKuhne: It never gets set with `-windows` in WPF. I would think that it should, but I left that as a follow-up. I _have_ to use `-windows` here or `StaFact` won't give me the right assemblies.

JeremyKuhne: https://github.com/dotnet/wpf/issues/10053

ViktorHofer: Understood. So why the Contains condition?

src/Microsoft.DotNet.Wpf/tests/Directory.Build.props
```diff
@@ -3,5 +3,7 @@
   <PropertyGroup>
     <IsShipping>false</IsShipping>
     <CLSCompliant>false</CLSCompliant>
+    <CopyTransitiveReferences>true</CopyTransitiveReferences>
```
Kuldeep-MS, line 6: why it is required?

eng/WpfArcadeSdk/tools/SdkReferences.targets
```diff
@@ -112,18 +112,18 @@
...
     <ItemGroup>
-      <Reference Include="$(PkgMicrosoft_Private_Winforms)\ref\$(TargetFramework)\%(MicrosoftPrivateWinFormsReference.Identity).dll"
-                 Condition="Exists('$(PkgMicrosoft_Private_Winforms)\ref\$(TargetFramework)\%(MicrosoftPrivateWinFormsReference.Identity).dll')">
-        <NuGetPackageId>Microsoft.Private.Winforms</NuGetPackageId>
-      </Reference>
-      <Reference Include="$(PkgMicrosoft_Private_Winforms)\ref\net8.0\%(MicrosoftPrivateWinFormsReference.Identity).dll"
-                 Condition="!Exists('$(PkgMicrosoft_Private_Winforms)\ref\$(TargetFramework)\%(MicrosoftPrivateWinFormsReference.Identity).dll')
-                 And $(TargetFramework) == 'net9.0'
-                 And Exists('$(PkgMicrosoft_Private_Winforms)\ref\net8.0\%(MicrosoftPrivateWinFormsReference.Identity).dll')">
+      <Reference Include="$(PkgMicrosoft_Private_Winforms)\lib\$(_WinFormsTargetFrameworkVersion)\%(MicrosoftPrivateWinFormsReference.Identity).dll"
```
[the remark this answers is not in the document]
JeremyKuhne: Because we're going to actually use the implementation for System.Private.Windows.Core

src/Microsoft.DotNet.Wpf/tests/Directory.Build.props
JeremyKuhne: Simpler, not required.

eng/WpfArcadeSdk/tools/SdkReferences.targets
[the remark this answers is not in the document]
ViktorHofer: Just double checking, does this mean that all references from that package are now implementation ones? That doesn't sound right. That said, what are those assemblies used for?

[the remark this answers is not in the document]
JeremyKuhne: No, we're just using the implementation assemblies for internal build purposes as we're actually going to run code from System.Private.Windows.Core.dll.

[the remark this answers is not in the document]
JeremyKuhne: The ref assemblies will still be deployed as the ref for the sdk.

Kuldeep-MS (approved):

PR #10078, opened 2024-11-16
Change primitives to build platform specific

JeremyKuhne:
If just building AnyCPU only the x86 transport package gets the runtime assembly. Rather than work around that, build explicitly for each platform so WindowsDesktop can get what it needs.


lonitra:
Helps unblock https://github.com/dotnet/windowsdesktop/pull/4786

lonitra (approved):
👍

Kuldeep-MS (approved):
LGTM

Issue #10017, opened 2024-10-30
Align WPF and WinForms code style guidelines


JeremyKuhne:
WPF and WinForms aren't far off. We both point to the runtime repo and Framework Design Guidelines, but WinForms has elaborated a bit more:

https://github.com/dotnet/winforms/blob/main/docs/coding-style.md

If we share the same guidelines, it will facilitate moving between the two repos for both internal and external contributors.

@Kuldeep-MS, @pchaurasia14

miloush:
I think I personally don't see any issues with those guidelines, assuming they are not enforced retrospectively in a breaking manner (e.g. turning public fields into properties).

One point of discussion we had recently was the naming convention for pointer variables, I don't see those covered.

Btw. why does it use backslashes for comments rather than forward slashes?

h3xds1nz:
Besides what miloush has pointed out, imho using collection expressions in cases like:
`return [];`
over e.g. 
`return Array.Empty<T>();`

or assigning e.g. private fields like:

`_hooks = [];`
over
`_hooks = new List<T>();`

should be discouraged.

I'm also not a fan of primary constructors but I guess it's whatever.

Otherwise I feel the guidelines are pretty well defined.

JeremyKuhne:
> I think I personally don't see any issues with those guidelines, assuming they are not enforced retrospectively in a breaking manner (e.g. turning public fields into properties).

@miloush No, we'll **never** make binary breaking changes. I'll call that out.

> One point of discussion we had recently was the naming convention for pointer variables, I don't see those covered.

I'm going to call that out explicitly- I was just thinking about that again last night. I've tried a variety of options at scale, and I currently believe pointers should be named like any other variable, _except_ in the case of pinning an existing variable. As pinning is usually constrained to a small block, I now use initials where the existing name is the best name:

``` C#
fixed (char* n = name)
{
}

// Not
fixed (char* namePtr = name)

// Avoid (can lead to much longer P/Invoke statements)
fixed (char* namePointer = name)
```

> Btw. why does it use backslashes for comments rather than forward slashes?

Whoops :) I'll fix that

> using collection expressions 

[...]

PR #10011, opened 2024-10-29
General using cleanup in C# files

JeremyKuhne:
This removes unnecessary using statements and moves them outside of the namespaces when they were inside them.

I manually cleaned up the headers a bit for several hundred files, but the cleanup there isn't exhaustive by any means. Changes are only to the tops of files.

To make this work a bit more cleanly I've added GlobalUsings.cs files to some of the assemblies.

NOTE: This was done using the code tools in Visual Studio.

Having usings inside of namespaces makes it difficult to use CsWin32, which uses the published Widows metadata names. Unsurprisingly, namespaces all start with `Windows.` which doesn't resolve when trying to do a using inside of a namespace definition where `Windows` is part of the name.

Finally, there are three HRESULT implementations in WPF, which are all visible in some assemblies. I had to put a couple of static usings inside of namespaces to resolve this. Moving to CsWin32 we can use the generated definition and avoid the conflicts.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/wpf/pull/10011)

PR #10080, opened 2024-11-16
Align .editorconfig with WinForms

JeremyKuhne:
This effectively copies the `.editorconfig` from WinForms to align the baseline.

As there are many deviations from the rules in the sources, added a temporary `.editorconfig` in the product src directory. This will need to have rules pulled bit by bit as the projects are fixed. Most of the rules are pretty easy to fix but would require some pretty massive code fix changes. They are reasonably safe to do, but they will create disruption for existing PRs as they'll need to rebase.

Created another `.editorconfig` specifically for the cycle breaker projects. Fixed the headers in the cycle breakers, but otherwise mostly left them alone.

Test projects were updated to pass without needing any overrides. As a general rule, we try to not disable rules for test projects as we want to keep good habits over the entire codebase.

The sooner the automated fixes are done, the better. It is better to have this done before Preview 1 and it is better to aggressively push to do them. Don't wait for full test passes in-between changes, do a full pass once they are committed. The recommendation:

1. For each rule where there is an automated fix, run the fix and create a PR that says that you ran the code fix
2. Sanity review and merge each as quickly as possible (check off, merge as soon as green)
3. Run a full test pass when finished

Clearing the automated fixes should be able to be done in a couple weeks at most. Once the rules are turned on it will significantly improve the contribution process. Less cleanup to review, less iteration over rules, etc.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/wpf/pull/10080)


JeremyKuhne:
This is a step towards #10017.

JeremyKuhne:
#10011 is one of the larger semi-automated fixes that would apply towards enabling the disabled rules in this PR.

harshit7962:
Thank you @JeremyKuhne for the help here, we will iteratively resolve the pending failures.

PR #10104, opened 2024-11-21
Start using System.Private.Windows.Core.TestUtilities

lonitra:
`System.Private.Windows.Core.TestUtilities` is a test utilities package that comes from WinForms that is meant to be shared with WPF to take advantage of test utilities that we have in WinForms such as `TestAccessors`. This change takes the test utilities package in `WindowsBase.Tests` (bumping xunit version is needed) and shows how TestAccessors can be used to validate private fields within a class.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/wpf/pull/10104)

### DIFF
--- a/eng/Versions.props
+++ b/eng/Versions.props
@@ -84,7 +84,7 @@
   </PropertyGroup>
   <!-- XUnit-related (not extensions) -->
   <PropertyGroup>
-    <XUnitVersion>2.9.0</XUnitVersion>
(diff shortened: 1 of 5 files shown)


JeremyKuhne:
@pchaurasia14

dipeshmsft (approved):

PR #10112, opened 2024-11-25
Opt-In: Dyanmic Resource Optimization

### LINKED ISSUE #10020: Break Issue in Net9.0
### Description
We've updated all our projects to support .NET 9.0. Our projects use a custom theme, which has been causing issues specifically in .NET 9.0, even though the same setup works seamlessly in .NET 8.0 and other frameworks. We've attached sample projects to illustrate this issue.
### Replication Procedure

1. Run the sample.
2. Click the ComboBox to open the dropdown.

### Screenshot
Image

[WpfAppNet90.zip](https://github.com/user-attachments/files/17572867/WpfAppNet90.zip)
[WpfAppNet80.zip](https://github.com/user-attachments/files/17572875/WpfAppNet80.zip)

harshit7962:
Fixes #10020 

## Description
The usage of dynamic resource was optimized as part of our initiative in the changes [here](https://github.com/dotnet/wpf/pull/5610). As in the issue above, we have observed issues related to freezing of resources which are not meant to frozen at a particular stage and this is causing some breaking changes. Following up on the discussions in the issue mentioned above, we have made this feature as an opt-in to garner more such feedbacks if and when available and work on fixing it for the next release.

## Customer Impact
Fixes crashing applications without the need to disable the provided opt-out switch

## Regression
Yes

## Testing
Local Build Pass
Sample Application Testing
Failure Testing
Community Test Pass

## Risk
None, changing things back to original case by default. Using Dynamic Resource optimizations via the switch can cause occasional crashing of application, in which case, the switch can be defaulted to true.

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/AppContextDefaultValues.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/AppContextDefaultValues.cs
@@ -11,7 +11,7 @@ namespace System
         /// <summary>
         /// This is a partial method. This method is responsible for populating the default values based on a TFM.
         /// It is partial because each library should define this method in their code to contain their defaults.
-        /// </summary> 
+        /// </summary>
         static partial void PopulateDefaultValuesPartial(string platformIdentifier, string profile, int targetFrameworkVersion)
         {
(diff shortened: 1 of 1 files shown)


hiteshwpfmsft (dismissed):

dipeshmsft:
Closing this PR as we will keep this switch as opt-out in main branch

PR #10119, opened 2024-11-26
[StyleCleanUp] Addressing SA warnings Part 1

harshit7962:
## Description
This work is a part of our initiative to set code-style guidelines, align WPF and WinForms, and ensure PR standards with respect to styles. This in turn will help us in better maintainability and readability of the repo overall. The changes follow up from the PR #10080 and references to the issue #10017.

The current changes addresses the following Errors/Warnings in the src folder of WPF:
- **SA1408:** Conditional expressions should declare precedence
- **SA1404:** Code analysis suppression should have justification
- **SA1131:** Constant values should appear on the right-hand side of comparisons

A good way to go about reviewing this is to go commit by commit which pans over individual errors/warnings, and hence easing out the overall understanding.

## Customer Impact
_None_

## Regression
_None_

## Testing
Local Build Pass
The current set of changes looks fairly mechanical and probably don't need a Test Pass, but these set of PRs can be clubbed to do so at a later stage.

## Risk
Low.

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/System.Xaml/System/Xaml/Context/ObjectWriterContext.cs
+++ b/src/Microsoft.DotNet.Wpf/src/System.Xaml/System/Xaml/Context/ObjectWriterContext.cs
@@ -267,7 +267,7 @@ namespace MS.Internal.Xaml.Context
         {
             get
             {
-                if (null == _serviceProviderContext)
+                if (_serviceProviderContext != null)
                 {
                     _serviceProviderContext = new ServiceProviderContext(this);
                 }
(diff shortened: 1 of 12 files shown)


src/Microsoft.DotNet.Wpf/src/System.Xaml/System/Xaml/Context/ObjectWriterContext.cs
```diff
@@ -267,7 +267,7 @@ internal ServiceProviderContext ServiceProviderContext
         {
             get
             {
-                if (null == _serviceProviderContext)
+                if (_serviceProviderContext != null)
```
h3xds1nz, line 270:
```suggestion
                if (_serviceProviderContext is null)
```

himgoyalmicro (approved):

PR #10120, opened 2024-11-26
Fixing window theme changes when system's theme is changed

siagupta0202:
## Description
Setting the Application ThemeMode to Light/Dark and changing the system's theme would not reflect on the window when the window is set to System ThemeMode. 

## Regression

Yes
## Testing
Local build pass. Tested with Sample applications

###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/wpf/pull/10120)


harshit7962 (approved):
Verified that the failing case is fixed. Also, skimmed through the performance aspect of the change. Seems like the practical use cases shouldn't have too much of a hit.

Issue #8548, opened 2023-12-14
Accessibility Review and Testing


singhashish-wpf:
- [ ] Check for A11y sanity using the A11y insights tools.
- [ ] Get the A11Y Test Pass Feature test run on the updated Styles.
- [ ] Fix A11y issues reported by the team

miloush:
Ensure accessibility settings apply, at least the ones in Settings > Accessibility > Visual effects
- Always show scrollbars
- Transparency effects
- Animation effects

PR #10130, opened 2024-11-28
Modified Expander animation mechanism in Fluent to avoid Argument Null Exception

dipeshmsft:
Fixes https://github.com/microsoft/WPF-Samples/issues/672

## Description
While changing the ResourceDictionaries, during the invalidation, Tag property was being set to null ( default value ) which was causing an error with the animation. Replaced it with a dummy border and instead of using Tag, now I use Width property of the border which is a double and it's default value is not null. This avoids the ArgumentNullException when the expander is expanded and the Theme is changed.

## Customer Impact
Not taking this fix would cause a ArgumentNullException when the Expander is expanded, and the Theme is changed

## Regression
No

## Testing
Sample App Testing

## Risk
Low

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/wpf/pull/10130)

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/Themes/PresentationFramework.Fluent/Controls/AnimationFactorToValueConverter.cs
+++ b/src/Microsoft.DotNet.Wpf/src/Themes/PresentationFramework.Fluent/Controls/AnimationFactorToValueConverter.cs
@@ -5,6 +5,7 @@ using System.Text;
 using System.Threading.Tasks;
 using System.Windows.Data;
 using System.Globalization;
+using System.Diagnostics;
 
 namespace Fluent.Controls
 {
@@ -17,7 +18,7 @@ namespace Fluent.Controls
                 return 0.0;
             }
 
-            if (values[1] is not double factor)
+            if (values[1] is not double factor || factor == double.NaN)
             {
                 return 0.0;
             }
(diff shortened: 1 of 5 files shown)


harshit7962 (dismissed):

miloush:
Just a reminder that there should be no animations when animations are disabled (SystemParameters.ClientAreaAnimation)

dipeshmsft:
> Just a reminder that there should be no animations when animations are disabled (SystemParameters.ClientAreaAnimation)

Thanks @miloush for bringing this up. I will take a look and update this PR accordingly. However, we are planning this fix for .NET 9 servicing as it is, because we need to merge it today.

dipeshmsft:
@miloush this seems like an issue with other controls as well. Will take this up as a part of the above issue : https://github.com/dotnet/wpf/issues/8548. For now, I am merging this PR.

himgoyalmicro (approved):

thurrott:
I am experiencing this problem with expanders being expanded as well. I've read through this thread and it's not clear how/where this is "fixed." What is the fix and/or workaround for this issue?

dipeshmsft:
@thurrott the fix is out now. It was release with .NET SDK 9.0.1. You can try the newer version of SDK for this.

Issue #10096, opened 2024-11-20
Fluent theme not working on Windows 10


LoriBru:
### Description

Hello, I am currently working on a small WPF app and I decided to give .NET9 and the new Fluent theme a go.
I am currently working in parallel on two machines, one is Windows 10 and one is Windows 11.
While everything looks fine on Windows 11, on Windows 10 the Fluent theme does not work.
Is this expected? I cannot find any reference in the documentation regarding the expected behavior of this new Theme on Windows 10. I understood that this theme would be applied also to Windows 10.
Both machines are working with Visual Studio 17.12 and .NET 9 installed.

On Windows 11:
![Image](https://github.com/user-attachments/assets/fc01090b-6f2e-4371-babc-62ad1e389e31)

On Windows 10:
![Image](https://github.com/user-attachments/assets/21a1d259-dc3d-427e-96fc-725f85acf853)
![Image](https://github.com/user-attachments/assets/43439477-b18a-429e-ba8d-e495d6b05a93)

Winodws 10 info:
![Image](https://github.com/user-attachments/assets/4dfdc065-2056-4486-82ac-61123a230a4c)

### Reproduction Steps

Create a minimal app with Fluent theme using .NET9 and start it on both Winodws 11 and Windows 10.

### Expected behavior

Fluent theme working on Windows 10.

### Actual behavior

Fluent theme not working on Windows 10.

miloush:
I suspect https://github.com/dotnet/wpf/issues/9635#issuecomment-2472295895 will work as a workaround for you.

harshit7962:
@LoriBru, thank you for giving the new themes a try. Yes, certain aspects of Fluent Theme on Windows 10 needs an additional switch to be added, as pointed by @miloush. We are working on getting this fixed in the framework itself. We will try to get this fix in the subsequent service release of .NET 9.

LoriBru:
Thank you both @miloush and @harshit7962 !
I will give the suggestion a try, indeed it is at least disappointing that the main new feature introduced with .NET9 on WPF requires a workaround right out of the box to work as expected on Windows 10.

miloush:
@LoriBru Please note that the Win11 theming is experimental.

LoriBru:
[...]

PR #10136, opened 2024-11-29
[release/9.0] Fixes window backdrop and background in Windows 10 and window theme interactions with system changes

dipeshmsft:
Fixes #  #10096 

## Description
In Windows 10, when using Fluent theme the background of the window appears black. This PR fixes the issue by including a DataTrigger that checks the OS version and in case of Windows 10, it sets the background accordingly as Windows 10 doesn't support the backdrop operations through DWM.

Secondly, when Application is Light / Dark ( not System ) the Window's with ThemeMode set to system didn't react to system theme changes. This was because, when application is in light or dark, the state of fluent theme state for application didn't change and we returned it from the function without checking the window's thememode. We enabled that behavior in this PR.

## Customer Impact
Developers can use Fluent theme in Windows 10 without using the backdrop switch. Enables the correct behavior for Window System theme mode when Application's theme mode is set to Light and Dark. 

## Regression
Yes. This worked in .NET 9 Preview 4

## Testing
Sample app testing + Unit tests

## Risk
Minimal. Fixes the behavior of Fluent theme in Windows 10 and when application theme is set to Light / Dark

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/wpf/pull/10136)

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/ThemeManager.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/ThemeManager.cs
@@ -28,12 +28,10 @@ internal static class ThemeManager
 
                 FluentThemeState newFluentThemeState = new FluentThemeState(Application.Current.ThemeMode.Value, useLightColors);
 
-                if (s_currentFluentThemeState == newFluentThemeState)
+                if (s_currentFluentThemeState != newFluentThemeState)
                 {
-                    return;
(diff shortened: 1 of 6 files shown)


harshit7962 (approved):
Combination of previously reviewed PRs fixing Fluent in Windows 10 and Window's ThemeMode bug with system mode. 
Had tested these fixes individually in respective PRs.
LGTM.

PR #10155, opened 2024-12-10
Convert to using CsWin32 HRESULT definition

JeremyKuhne:
This converts to using the CsWin32 HRESULT definition. It creates two HRESULT helpers to provide compatible throw behavior for existing code (PresentationCore threw things one way, other assemblies threw a different way).

Also:

- Remove dead interop defines (a lot)
- Add Primitives test project
- Force test projects to use open source key (which System.Private.Windows.Core expects)
- Tweak build files a bit to try and ensure test dependencies are in place
- Add strings to System.Windows.Primitives to facilitate unifying hresult handling in the future (need to redirect InvalidWmpVersionException)

It would be nice to always throw the same way, but that would change exceptions that currently get thrown. Combining the behavior from both would give whatever Marshal gives with special handling for WIC/MIL errors and unwrapping of Win32 COMExceptions into Win32Exceptions.

Converting the existing interop to use CsWin32 types will make transitioning far easier. The next change I'm working on is using HWND.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/wpf/pull/10155)


JeremyKuhne:
FYI @harshit7962

src/Microsoft.DotNet.Wpf/src/PresentationCore/MS/Win32/UnsafeNativeMethodsTablet.cs
```diff
@@ -120,33 +119,33 @@ public override bool IsInvalid
...
+            // context.
+            //
+            // Debug.Assert(_recognizerHandle != null && !_recognizerHandle.IsInvalid);
+            Debug.Assert(handle != 0);
+            HRESULT result = UnsafeNativeMethods.DestroyContext(handle);
 
             // Now, dereference the attached recognizer.
             _recognizerHandle = null;
 
-            return MS.Internal.HRESULT.Succeeded(hr);
```
lonitra, line 137: Double checking - how did this define success? Was S_FALSE included?

src/Microsoft.DotNet.Wpf/src/PresentationCore/MS/internal/Ink/GestureRecognizer/NativeRecognizer.cs
```diff
@@ -409,14 +408,21 @@ private int AddStrokes(MS.Win32.Recognizer.ContextSafeHandle recContext, StrokeC
                     GetPacketData(stroke, out packetDescription, out countOfBytes, out packets, out xForm);
                     if (packets == IntPtr.Zero)
                     {
-                        return -2147483640; //E_FAIL - 0x80000008.  We never raise this in an exception
+                        // This isn't actually E_FAIL, it's the Macintosh version of it from the headers. Keeping for compat.
+                        return (HRESULT)(-2147483640); // E_FAIL - 0x80000008.  We never raise this in an exception
```
lonitra, line 412: Interesting 👀 If this is used in other places we should consider creating some sort of const

lonitra (commented):
Got through some, looks good so far. Will circle back to review the rest.

src/Microsoft.DotNet.Wpf/src/PresentationCore/MS/Win32/UnsafeNativeMethodsTablet.cs
JeremyKuhne: >= 0

src/Microsoft.DotNet.Wpf/src/PresentationCore/MS/internal/Ink/GestureRecognizer/NativeRecognizer.cs
JeremyKuhne: It isn't. I can see how they made the mistake though. (Another argument for source generated interop.) I thought about changing it, but for now I commented and left it as is.

src/Microsoft.DotNet.Wpf/src/PresentationFramework/MS/Internal/Printing/UnsafeNativeMethods.cs
```diff
@@ -1,8 +1,8 @@
-// Licensed to the .NET Foundation under one or more agreements.
+﻿// Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the MIT license.
-// See the LICENSE file in the project root for more information.
```
lonitra, line 3: Looks like accidental deletion?

lonitra (commented):
LGTM mostly just some small style comments

src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Media/StreamAsIStream.cs
```diff
@@ -170,15 +168,11 @@ public int Clone(out IntPtr stream)
                 return SecurityHelper.GetHRForException(e);
             }
 
-            #pragma warning restore 6500
-
-            return NativeMethods.E_NOTIMPL;
+            return HRESULT.E_NOTIMPL;
         }
 
         public int Commit(uint grfCommitFlags)
```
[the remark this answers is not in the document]
JeremyKuhne: There are quite a few places that are technically HRESULT that I didn't get. I only did things where they mattered for consumption. Large swaths of code will change when transitioning to CsWin32 which always has the right definition- letting that process pick up the rest.

src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/dataobject.cs
```diff
@@ -661,15 +662,15 @@ void IComDataObject.GetData(ref FORMATETC formatetc, out STGMEDIUM medium)
                     medium.tymed = TYMED.TYMED_ISTREAM;
 
                     IStream istream = null;
-                    hr = Win32CreateStreamOnHGlobal(IntPtr.Zero, true /*deleteOnRelease*/, ref istream);
-                    if ( NativeMethods.Succeeded(hr) )
+                    result = Win32CreateStreamOnHGlobal(IntPtr.Zero, true /*deleteOnRelease*/, ref istream);
```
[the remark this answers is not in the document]
JeremyKuhne: You're going to redo this file shortly, so I wasn't super aggressive.

lonitra (approved):
👍

JeremyKuhne:
Note that I haven't changed every instance of `int` to `HRESULT`. If we weren't looking at the result, I didn't necessarily change it as most of this will go away when switching to generated interop.

PR #10186, opened 2024-12-17
[StyleCleanUp] Addressing CA warnings Part 2

### LINKED ISSUE #10294: CA1507: Use nameof in place of string
Link to the issue description: https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1507

## Rule Description 

Rule CA1507 flags the use of a string literal as an argument to a method or constructor where a [nameof](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/operators/nameof) (NameOf in Visual Basic) expression would add maintainability. The rule fires if all of the following conditions are met:

- The argument is a string literal or constant.
- The argument corresponds to a string-typed parameter of the method or the constructor that's being invoked (that is, there is no conversion involved at the call site).
- Either:
  - The declared name of the parameter is paramName and the constant value of the string literal matches the name of a parameter of the method, lambda, or local function within which the method or constructor is being invoked.
  - The declared name of the parameter is propertyName and the constant value of the string literal matches the name of a property of the type within which the method or constructor is being invoked.

himgoyalmicro:
Fixes #10294 

## Description

This work is a part of our initiative to set code-style guidelines, align WPF and WinForms, and ensure PR standards with respect to styles. This in turn will help us in better maintainability and readability of the repo overall. The changes follow up from the PR https://github.com/dotnet/wpf/pull/10080 and references to the issue https://github.com/dotnet/wpf/issues/10017.

The current changes address the following Errors/Warnings in the src folder of WPF:

- **CA1507**: Use nameof in place of string

A good way to go about reviewing this is to go commit by commit which pans over individual errors/warnings and hence easing out the overall understanding.

## Customer Impact

None

## Regression

None

## Testing

Local Build Pass
The current set of changes looks fairly mechanical and probably don't need a Test Pass, but these set of PRs can be clubbed to do so at a later stage.

## Risk

Low

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/wpf/pull/10186)


harshit7962 (approved):
LGTM

DanFTRX:
Is there a reason all of the Format arguments that were string versions of type names or other parameters/variables were not converted to nameof as well?   Tons of examples on just the edited lines in this PR.

h3xds1nz:
@DanFTRX Read the implementation notes on https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1507 to understand what is the reason. Especially the note about parameter names required for the analyzer to fire.

DanFTRX:
@h3xds1nz It seems a lot like following the word of the rules rather than the spirit of the rules only correct the strings that caused this warning.  The point of the diagnostic is that using nameof improves code maintainability in cases where the code element may be renamed in the future, but the string literal is mistakenly not renamed.  That motivation applies equally to these other string constants.   I would posit that a pass on these other string constants should also be done otherwise this particular rule is just theatre.

h3xds1nz:
@DanFTRX Analyzers don't catch everything, likely won't for a while as its a static analysis too, or they catch too much.

Of course I agree with you that a pass on the occurences that the analyzer does not catch should be done as well, but does not necessarily need to be done in this PR, where the clear scope was to compile with `CA1507` enabled ("addressing CA warnings", not "making sure we're using nameof on every symbol where it makes sense", which will require a bit of manual work.)

Sorry if you've misunderstood my reply but given you've asked for a reason behind "why those weren't converted as well", I merely answered that part.

DanFTRX:
> @DanFTRX Analyzers don't catch everything, likely won't for a while as its a static analysis too, or they catch too much.
> 
> Of course I agree with you that a pass on the occurences that the analyzer does not catch should be done as well, but does not necessarily need to be done in this PR, where the clear scope was to compile with `CA1507` enabled ("addressing CA warnings", not "making sure we're using nameof on every symbol where it makes sense", which will require a bit of manual work.)
> 
> Sorry if you've misunderstood my reply but given you've asked for a reason behind "why those weren't converted as well", I merely answered that part.

Ah understood, the answer wasn't "we don't want to" but rather "the scope of this PR was limited".  That makes sense.

PR #10213, opened 2024-12-30
Messagebox object support

### LINKED ISSUE #10212: Messagebox object support
Would be great if the MessageBox class had a method, that enables the developer to show not just strings, but any object.

BartaBlzs:
Fixes #10212

## Description

Adding a method to the MessageBox class, that enables the developer to show not just strings, but any object.

## Customer Impact

Very little impact, it saves ".ToString()" if you want to show an object in a MessageBox.

## Testing

None 💀

## Risk

Don't think there's any risk.

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/wpf/pull/10213)

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/MessageBox.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/MessageBox.cs
@@ -1,4 +1,4 @@
-// Licensed to the .NET Foundation under one or more agreements.
+﻿// Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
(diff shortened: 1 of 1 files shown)


BartaBlzs:
@dotnet-policy-service agree

miloush:
This requires API review. Make a commit that contains only the suggested change, do some testing. It crashes when the object is null. Why only this override?

src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/MessageBox.cs
```diff
@@ -1,695 +1,705 @@
...
+        {
+            return ShowCore(IntPtr.Zero, messageBoxText, String.Empty, MessageBoxButton.OK, MessageBoxImage.None, 0, 0);
+        }
+
+        /// <devdoc>
+        ///    <para>
+        ///       Displays a message box with the string representation of the specified object.
+        ///    </para>
+        /// </devdoc>
+        public static MessageBoxResult Show(object messageBoxObject)
```
lindexi, line 166: Sorry, I do not think it is the good design. Because this override do as `public static MessageBoxResult Show(string messageBoxText)` can do, and this override hide the implementation details.

pchaurasia14:
@BartaBlzs - Since this is not planned, we're closing this until we hear more compelling arguments in favor of this API.

PR #10229, opened 2025-01-06
Fix Potential Crash in Fluent High Contrast

harshit7962:
## Description
The theme generator for `Fluent` themes partially relies on common brushes, colors, and keys written in theme files, i.e. `Light.xaml`, `Dark.xaml` and `HC.xaml`. The current implementation of `HC.xaml` contains a duplication of `SystemColorWindowColor` key.
The resource dictionary loaded at runtime for High Contrast theme is Fluent.HC.xaml, which right now is correct but if/when we generate the same, it will result in duplication of that key and hence runtime crashing of applications on high contrast theme.

## Customer Impact
Fixes any future regression and crash on high contrast theme (Fluent)

## Testing
Local Build Pass

## Risk
Low, removal of a xaml key.
Does not violate API contract as the key is already present.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/wpf/pull/10229)


himgoyalmicro (approved):
LGTM

Issue #10087, opened 2024-11-18
Break Issue in Net9.0 for RadioButton used in a PopUp


umanagarajan12:
### Description

We have updated all our projects to support .NET 9.0. As previously noted, our projects use a custom theme that functions well in .NET 8.0 and other frameworks but has been causing issues specifically in .NET 9.0.

We have logged this issue [#10020](https://github.com/dotnet/wpf/issues/10020) and received the recommendation: *"The issue is due to a recent change to optimize DynamicResource usage (#5610). To prevent the crash, you can opt-out of this behavior by adding the following to your `.csproj` file:*

```xml
<ItemGroup>
    <RuntimeHostConfigurationOption Include="Switch.System.Windows.Controls.DisableDynamicResourceOptimization" Value="true" />
</ItemGroup>
```

This solution has resolved the issue in some sample projects. However, we still encounter problems in other cases when using themes:

In our DataGrid control, we use RadioButtons in the filter functionality, but we are encountering issues when clearing filters.  

**Steps to Reproduce:**  
1. Open the Filter Popup and apply a filter using the Advanced Filter options (e.g., Text Filter or Numeric Filter).  
2. Use the first filter option to apply a filter.  
3. Select the "And" option, which deselects the default "Or" option.  
4. Click the **OK** button to apply the filter.  
5. Reopen the Filter Popup and click **Clear Filter**.  

**Expected Behavior:**  
- The "And" option should be deselected, and only the "Or" option should remain selected by default.  

**Actual Behavior:**  
- Both "And" and "Or" options remain selected simultaneously, which is incorrect.  

**Expected Behavior Screenshot(Net 80):**
![Image](https://github.com/user-attachments/assets/94082c3a-707e-4fd0-aaa8-12d9993b8788)

**Actual Behavior Screenshot(Net 90):**  
![Image](https://github.com/user-attachments/assets/82396cf7-6ee0-41e3-86aa-f58faaf8f82a)

**Note:** We have also included `<RuntimeHostConfigurationOption Include="Switch.System.Windows.Controls.DisableDynamicResourceOptimization" Value="true" />` in the `.NET 9.0` project `ItemGroup`.

We’ve attached sample projects demonstrating this issue and video references for Issue replication. 

Could you please investigate further?
[Samples.zip](https://github.com/user-attachments/files/17800582/Samples.zip)
[Video References.zip](https://github.com/user-attachments/files/17800581/Video.References.zip)

harshit7962:
[...]

PR #10252, opened 2025-01-10
Fix interaction of RadioButton, shared bindings and hanging GC refs

### LINKED ISSUE #2995: RadioButtons with the same `GroupName` will affect each other over windows if window is closed
* .NET Core Version: 3.1.202
* Windows version: 10.0.18363 / 10.0.19624

 **Problem description:**

If some `RadioButton`s are in the same group, they should check and uncheck each other. If we create two windows both contain the same XAML RadioButton groups, we'll find that the `RadioButton`s are separated into two windows. However, if one window is closed, the `RadioButton`s in the other window will affect those in the closed one.

This code below shows how to set the `GroupName`, but it's recommended to read the minimal reproduction at the end of this issue to know all the details.

```xml
<RadioButton GroupName="A" Content="Option 1" />
<RadioButton GroupName="A" Content="Option 2" />
<RadioButton GroupName="A" Content="Option 3" />
```

By running the minimal reproduction source code, you'll see the expected behavior and the broken behavior and I've posted the gifs below.

 **Expected behavior:**
 
2020-05-14-radio-buttons-in-opened-windows

 **Actual behavior:** 
 
2020-05-14-radio-buttons-in-closed-windows

 **Minimal repro:**

The GitHub repo:

- https://github.com/walterlv/Walterlv.Issues.RadioButton.GroupName

The keys:

1. Write some `RadioButton`s in a window XAML;
2. We want to know whether the `RadioButton`s in different windows affect each other, so I bind the first `RadioButton.IsChecked` to a singleton. This means:
    - If the radio buttons affect only in a single window, the A window checks a radio button then the binding source viewmodel propertychanged, and then the B window checks a radio button due to the viewmodel new property value.
    - But if the radio buttons affect over windows, the A window checks a radio button then it change the binding source viewmodel and then checks the B window radio button. But the radio button in B window unchecks the radio button in A window, so the binding source viewmodel property is changed to `false`. At last, all values are `false`. As a result, no radio button can be checked.

If all the windows are open, all behave as expected:

image

If the previous windows are closed, then the radio buttons start to affect each other:

image

The XAML code:

```xml
<Window x:Class="Walterlv.Issues.MainWindow"
        xmlns:local="clr-namespace:Walterlv.Issues"
        mc:Ignorable="d"
    <StackPanel>
        <Border>
            <RadioButton GroupName="A" IsChecked="{Binding Bar, Source={x:Static local:Foo.Instance}}" Content="Option 1" />
        </Border>
        <Border>
            <RadioButton GroupName="A" Content="Option 2" />
        </Border>
        <Border>
            <RadioButton GroupName="A" Content="Option 3" />
        </Border>
    </StackPanel>
</Window>
```

The C# code:

```csharp
using System.ComponentModel;
using System.Windows;

namespace Walterlv.Issues
{
    {
        public MainWindow()
        {
            InitializeComponent();
        }
    }

    public class Foo : INotifyPropertyChanged
    {
        public static Foo Instance { get; } = new Foo();

        public bool Bar
        {
            get => _bar;
            set
            {
                if (!Equals(_bar, value))
                {
                    _bar = value;
                    PropertyChanged?.Invoke(this, new PropertyChangedEventArgs(nameof(Bar)));
                }
            }
        }

        private bool _bar;
    }
}
```

### LINKED ISSUE #9035: Radiobutton with GroupName sets value false second time window is opened
### Description

A wpf Window that has a set of RadioButton controls with GroupName set and IsChecked bound to an underlying property will set that property to `false` when a second instance of that window is opened after the first instance is closed.

If the GroupName attribute of the RadioButtons is not set, the property is not changed when opening a second window.

This is very similar to issue #2995 that was reported as fixed in October 2023 and possibly related to (same as?) issues #3751 and #7462 that were closed due to lack of repro.

### Reproduction Steps

A minimal repro project is available here: https://github.com/OpterJohan/RadiobuttonIssueRepro
Steps to reproduce:
* run the wpf app
* press the "Open details window" button
(A window named Details with 2 radio buttons opens. The radio button "Option 1" is set)
* close the "Details" window
* press the "Open details window" button again
(The Details-window opens again, but this time both the radio buttons "Option 1" and "Option 2" are not set)

If DetailsWindow.xaml is changed to that the radio buttons do not have GroupName, "Option 1" is set the second time the window is opened as well.

### Expected behavior

Bound property should be unaffected by old closed windows.
The behaviour should be the same regardless if a GroupName is given.

### Actual behavior

Bound property is set false.

### Regression?

Possible regression of #2995.

### Known Workarounds

Run GC.Collect before opening the window again seems to work in the minial repro project. Unclear if it would always work.

### Impact

Makes is more difficult to create a UI with multiple sets of radio buttons.

### Configuration

.net 8.0.2
Windows 11 10.0.22631.0, x64.

### Other information

_No response_

h3xds1nz:
Fixes #9035
Fixes #2995

Marking it as draft right now due to the changes from #10251 being included. Only the latest commit is relevant.

## Description

This fixes actually two things, not just one and I'm fine with only fixing the #2995 / #9035 if desirable.

### First fix

First fix resolves #2995 / #9035 report that was also attempted in #8202 but was reverted in #10105 due to obvious issues.
Original method compares the visual root that's found using the following:
```csharp
DependencyObject rootScope = KeyboardNavigation.GetVisualRoot(this);
DependencyObject otherRoot = KeyboardNavigation.GetVisualRoot(radioButton);
```
The problem being that this search will result in `null` on Popups (hence why the original fix caused #10087), it will return `null` on radio buttons with closed Windows and depending on the focus situation it may return `null` on a fresh window.

Using `InputElement.GetRootVisual` instead we will retrieve the proper visual roots for Popups (PopupRoot), for open windows right after creation and for non-GCed `RadioButton`'s but detached from their visual root it will retrieve the container.

### Second fix/improvement
The original issue attempts to **share** a binding between two distincts visual roots using a singleton instance, which works until you have a hanging not-yet GCed `RadioButton` due to the reasons described in the first fix.

```csharp
BindingExpression rootBindingSource = GetBindingExpression(IsCheckedProperty);
BindingExpression otherBindingSource = radioButton.GetBindingExpression(IsCheckedProperty);
if (rootBindingSource is not null && otherBindingSource is not null &&
    rootBindingSource.SourceItem == otherBindingSource.SourceItem &&
    rootBindingSource.SourcePropertyName == otherBindingSource.SourcePropertyName)
    continue;
```

Excuse the checking verbosity, however,, this will allow for **shared** bindings (e.g. here - https://github.com/dotnet/wpf/issues/2995#issuecomment-628559516) to continue working properly even under the same visual root instead of just bound property being always false while retaining the usual **RadioButton** behavior and functionality. Updated with more stuff here [RadioButton.GroupName.Tests](https://github.com/h3xds1nz/RadioButton.GroupName.Tests), you'll have to update the `References`.

## Customer Impact

Resolving long standing issues concerning `RadioButton` and bindings.

## Regression

No.

## Testing

Local build, extensive testing with multiple reproduction samples.

## Risk

More testing is needed of course, but it would be a nice improvement to have this working. The second fix changes behavior, but changes it from non-working and appearing buggy to actually working nicely.

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Controls/RadioButton.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Controls/RadioButton.cs
@@ -67,14 +67,11 @@ namespace System.Windows.Controls
 
         private static void Register(string groupName, RadioButton radioButton)
         {
-            if (_groupNameToElements == null)
-                _groupNameToElements = new Hashtable(1);
+            _groupNameToElements ??= new Dictionary<string, ArrayList>(1);
 
             lock (_groupNameToElements)
             {
-                ArrayList elements = (ArrayList)_groupNameToElements[groupName];
-
-                if (elements == null)
+                if (!_groupNameToElements.TryGetValue(groupName, out ArrayList elements))
                 {
                     elements = new ArrayList(1);
                     _groupNameToElements[groupName] = elements;
@@ -92,15 +89,13 @@ namespace System.Windows.Controls
 
         private static void Unregister(string groupName, RadioButton radioButton)
         {
-            if (_groupNameToElements == null)
+            if (_groupNameToElements is null)
                 return;
 
             lock (_groupNameToElements)
             {
                 // Get all elements bound to this key and remove this element
-                ArrayList elements = (ArrayList)_groupNameToElements[groupName];
-
-                if (elements != null)
+                if (_groupNameToElements.TryGetValue(groupName, out ArrayList elements))
                 {
                     PurgeDead(elements, radioButton);
                     if (elements.Count == 0)
@@ -135,12 +130,14 @@ namespace System.Windows.Controls
             if (!string.IsNullOrEmpty(groupName))
             {
                 Visual rootScope = KeyboardNavigation.GetVisualRoot(this);
-                if (_groupNameToElements == null)
-                    _groupNameToElements = new Hashtable(1);
+
+                _groupNameToElements ??= new Dictionary<string, ArrayList>(1);
+
                 lock (_groupNameToElements)
                 {
                     // Get all elements bound to this key and remove this element
-                    ArrayList elements = (ArrayList)_groupNameToElements[groupName];
+                    ArrayList elements = _groupNameToElements[groupName];
+                    // Haha, notice the unlikely but possible null-deref here
                     for (int i = 0; i < elements.Count; )
                     {
                         WeakReference weakReference = (WeakReference)elements[i];
@@ -283,7 +280,9 @@ namespace System.Windows.Controls
 
         #region private data
 
-        [ThreadStatic] private static Hashtable _groupNameToElements;
+        [ThreadStatic]
+        private static Dictionary<string, ArrayList> _groupNameToElements;
+
         private static readonly UncommonField<string> _currentlyRegisteredGroupName = new UncommonField<string>();
 
         #endregion private data
--- a/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Controls/RadioButton.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Controls/RadioButton.cs
@@ -67,12 +67,12 @@ namespace System.Windows.Controls
 
         private static void Register(string groupName, RadioButton radioButton)
         {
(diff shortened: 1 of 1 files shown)


miloush:
The equivalence of the expressions seems a little fragile. Shouldn't we do reference equality for the source item? How are we doing on XPath bindings, don't they have the property name null?

h3xds1nz:
That should work imho but I'll definitely check it just to make sure, not 100% on that one. Great point though, appreciated.

PR #10266, opened 2025-01-11
[Cleanup] Convert typeof(Foo).Name to nameof(Foo) (IDE0082)

h3xds1nz:
## Description

Analyzer fix, converts occurrences of `typeof(Foo).Name` to the compile-time generated `nameof(Foo)` alternative.

## Customer Impact

Cleaner codebase, impact is negligible perf-wise as its exception cases.

## Regression

No.

## Testing

Local build.

## Risk

Minimal.

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/TemplateContentLoader.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/TemplateContentLoader.cs
@@ -22,14 +22,14 @@ namespace System.Windows
             T result = provider.GetService(typeof(T)) as T;
             if (result == null)
             {
-                throw new InvalidOperationException(SR.Format(SR.DeferringLoaderNoContext,typeof(TemplateContentLoader).Name, typeof(T).Name));
+                throw new InvalidOperationException(SR.Format(SR.DeferringLoaderNoContext,nameof(TemplateContentLoader), typeof(T).Name));
             }
             return result;
         }
 
         public override XamlReader Save(object value, IServiceProvider serviceProvider)
         {
-            throw new NotSupportedException(SR.Format(SR.DeferringLoaderNoSave, typeof(TemplateContentLoader).Name));
+            throw new NotSupportedException(SR.Format(SR.DeferringLoaderNoSave, nameof(TemplateContentLoader)));
         }
     }
 }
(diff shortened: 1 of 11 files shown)


src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/TemplateContentLoader.cs
```diff
@@ -22,14 +22,14 @@ private static T RequireService<T>(IServiceProvider provider) where T : class
             T result = provider.GetService(typeof(T)) as T;
             if (result == null)
             {
-                throw new InvalidOperationException(SR.Format(SR.DeferringLoaderNoContext,typeof(TemplateContentLoader).Name, typeof(T).Name));
+                throw new InvalidOperationException(SR.Format(SR.DeferringLoaderNoContext,nameof(TemplateContentLoader), typeof(T).Name));
```
ThomasGoulet73, line 25: Nit, since you're changing that line:
```suggestion
                throw new InvalidOperationException(SR.Format(SR.DeferringLoaderNoContext, nameof(TemplateContentLoader), typeof(T).Name));
```

h3xds1nz: Thanks!

h3xds1nz:
@himgoyalmicro Fixed merge conflict :)

h3xds1nz:
That's unfortunate, this PR will have to wait until this flows in due to the implementation of `nameof` for unbound generics: https://github.com/dotnet/roslyn/pull/76920

tl;dr the failure is not real, those two generate different results.

h3xds1nz:
@himgoyalmicro Once https://github.com/dotnet/sdk/pull/46195 is merged and the SDK flows in, it will work. For now we gotta wait.

ThomasGoulet73:
> @himgoyalmicro It builds now, the new SDK flowed in but of course #10514 broke the test runs now haha.

I have a fix ready that I'll push soon. This was caused by the switch from xUnit test runner to VSTest test runner introduced in #10514 which changed the order in which test were ran and exposed an existing bug in the tests.

ThomasGoulet73:
@h3xds1nz: Just a ping to let you know that the test failures in this PR were fixed in main, pulling main should fix the failures.

himgoyalmicro (approved):
LGTM

himgoyalmicro:
Many thanks to @h3xds1nz for your relentless contributions. Your support is invaluable 😄

PR #10267, opened 2025-01-11
Avoid zero-length array allocations and storing them in static fields (CA1825)

### LINKED ISSUE #10286: CA1825: Avoid zero-length array allocations
More info: https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1825

h3xds1nz:
Fixes #10286

## Description

This PR mainly resolves `CA1825` but in some cases goes a step further over standard analyzer fix.

- In addition to using singleton over allocating zero-length array everytime, we want to avoid storing those in `readonly` fields (whether they're marked or not), as the pointer to such array is already accessible via `Array.Empty<Type>()`.
- `XamlTypeMapper` was going step further and cloning this zero-length array even. Happy days.
- `RowSpanVector` was performing unnecessary assigment if the branch was not taken.

## Customer Impact

## Regression

No.

## Testing

Local build.

## Risk

Should be low.

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/wpf/pull/10267)

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/.editorconfig
+++ b/src/Microsoft.DotNet.Wpf/src/.editorconfig
@@ -95,9 +95,6 @@ dotnet_diagnostic.CA1822.severity = suggestion
 # CA1823: Avoid unused private fields
 dotnet_diagnostic.CA1823.severity = suggestion
 
-# CA1825: Avoid zero-length array allocations
-dotnet_diagnostic.CA1825.severity = suggestion
-
 # CA1834: Use StringBuilder.Append(char) for single character strings
 dotnet_diagnostic.CA1834.severity = suggestion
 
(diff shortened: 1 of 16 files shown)


omariom:
Could `[]` be used insted of `Array.Empty`?

h3xds1nz:
> Could `[]` be used insted of `Array.Empty`?

I'm not a fan. In cases like assigning to already defined field or parameter you lose the visible type information.

harshit7962 (approved):
Changes LGTM

harshit7962:
Thank you @h3xds1nz for these contributions. We will be taking the next batch of changes for the test pass soon.

h3xds1nz:
@harshit7962 Thanks you, happy to hear :)

PR #10281, opened 2025-01-15
Correcting ITextPointer when using cursor indicator

harshit7962:
## Description
Using accessibility option of _text cursor indicator_ in `RichTextBox` sometimes lead to crashing applications due to Invariant assert [here](https://github.com/dotnet/wpf/blob/main/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Controls/TextRangeAdaptor.cs/#L462). Adding `Normalize()` call in changed location moves the `end` to the `start` if this invariant assert fails and hence essentially corrects the cursor position.

## Customer Impact
Accessibility fix. Allowing the usage of text cursor indicator with RichTextBox.

## Regression
_None_

## Testing
Local Build Pass
Sample Application Testing

## Risk
Low
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/wpf/pull/10281)

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Controls/TextRangeAdaptor.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Controls/TextRangeAdaptor.cs
@@ -353,7 +353,7 @@ namespace MS.Internal.Automation
                     },
                     delegate(object val1, object val2) { return (double)val1 == (double)val2; })
                 );
-            // OutlineStyles 
+            // OutlineStyles
             _textPatternAttributes.Add(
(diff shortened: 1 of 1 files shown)


singhashish-wpf (approved):
Backporting NETFX changes.

miloush:
So the only code change I can see is to call `Normalize()` on clone. Isn't the invariant supposed to hold in general? So perhaps there is a bug that if fixed, the object would be good to clone as is...

harshit7962:
@miloush, we have raised another change for fixing the same. PR: https://github.com/dotnet/wpf/pull/10333.
Closing this PR.

cc: @h3xds1nz

PR #10303, opened 2025-01-16
Use source-generated regex instead of runtime compilation (SYSLIB1045)

### LINKED ISSUE #10302: SYSLIB1045: Use [GeneratedRegex] for compile time source generation
More info: https://learn.microsoft.com/en-us/dotnet/fundamentals/syslib-diagnostics/syslib1040-1049

h3xds1nz:
Fixes #10302

## Description

Use `GeneratedRegex` attribute to allow for source-generation of regexes at compile time for improved performance.

We only use regexes in two places which could effectively be removed but for now we should at least take advantage of the highest-performing form of regular expressions on .NET.

In the PBT case, we have to use a pre-processor statement to comfort NetFX scenario, which will later need naming-styles suppression.

Note: `RegexOptions.Compiled` is ignored in source-generated regexes, so I've removed it.

## Customer Impact

Improved performance, cleaner codebase.

## Regression

No.

## Testing

Local build.

## Risk

While the second occurence was fixed manually because analyzer won't flag it (guess nobody thought someone is declaring regex inline with a const), the change is straightforward.

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/wpf/pull/10303)

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/PresentationBuildTasks/Microsoft/Build/Tasks/Windows/UidManager.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationBuildTasks/Microsoft/Build/Tasks/Windows/UidManager.cs
@@ -925,7 +925,7 @@ namespace Microsoft.Build.Tasks.Windows
 
 
     // writing to a file, removing or updating uid
-    internal sealed class UidWriter
+    internal sealed partial class UidWriter
     {
         internal UidWriter(UidCollector collector, Stream source, Stream target)
         {
@@ -1101,7 +1101,7 @@ namespace Microsoft.Build.Tasks.Windows
             // escape all the Xml entities in the value
             string attributeValue = EscapedXmlEntities.Replace(
                 uid.Value,
-                EscapeMatchEvaluator
+                s_escapeMatchEvaluator
                 );
 
             string clause = string.Format(
@@ -1128,11 +1128,6 @@ namespace Microsoft.Build.Tasks.Windows
 
         private void WriteNewAttributeValue(string value)
         {
-            string attributeValue = EscapedXmlEntities.Replace(
-                value,
-                EscapeMatchEvaluator
-                );
-
             _targetWriter.Write(
                 string.Format(
                     TypeConverterHelper.InvariantEnglishUS,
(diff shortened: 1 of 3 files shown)


src/Microsoft.DotNet.Wpf/src/PresentationBuildTasks/Microsoft/Build/Tasks/Windows/UidManager.cs
```diff
@@ -1305,8 +1300,13 @@ private enum WriterAction
...
         }
 
-        private static Regex          EscapedXmlEntities   = new Regex("(<|>|\"|'|&)", RegexOptions.CultureInvariant | RegexOptions.Compiled);
-        private static MatchEvaluator EscapeMatchEvaluator = new MatchEvaluator(EscapeMatch);
+#if !NETFX
+        [GeneratedRegex("(<|>|\"|'|&)", RegexOptions.CultureInvariant)]
+        private static partial Regex EscapedXmlEntities { get; }
+#else
+        private static readonly Regex EscapedXmlEntities = new("(<|>|\"|'|&)", RegexOptions.CultureInvariant | RegexOptions.Compiled);
+#endif
```
harshit7962, line 1308: I am guessing this fix was directly from the analyzer? Did not quite understand the need of flag here.

h3xds1nz: @harshit7962 Which flag?

harshit7962: I meant the NETFX thing

h3xds1nz: Right, you cannot use `GeneratedRegex` for NetFX, the source generator is only available for .NET 7+.

harshit7962: Yeah in that case, `WinRTSpellerInterop.cs` should have that as well.

harshit7962: Ohh, that was an oversight, I thought both the changes were in PBT. It's fine, we can take this forward. Feel free to resolve the conflict.

h3xds1nz: Now it makes sense 😀

h3xds1nz:
@harshit7962 Resolved

harshit7962 (approved):

h3xds1nz:
@harshit7962 Thanks :)

PR #10170, opened 2024-12-12
[StyleCleanUp] Address IDE warnings Part 2

harshit7962:
## Description
This work is a part of our initiative to set code-style guidelines, align WPF and WinForms, and ensure PR standards with respect to styles. This in turn will help us in better maintainability and readability of the repo overall. The changes follow up from the PR #10080 and references to the issue #10017.

The current changes addresses the following Errors/Warnings in the src folder of WPF:
- **IDE0019:** Use pattern matching to avoid as followed by a null check
- **IDE0020:** Use pattern matching to avoid is check followed by a cast (with variable)
- **IDE0029:** Use coalesce expression

A good way to go about reviewing this is to go commit by commit which pans over individual errors/warnings, and hence easing out the overall understanding. 

## Customer Impact
_None_

## Regression
_None_

## Testing
Local Build Pass
Will be taken in the next test pass

PR #10168, opened 2024-12-12
[StyleCleanUp] Address IDE Warnings Part 1

harshit7962:
## Description
This work is a part of our initiative to set code-style guidelines, align WPF and WinForms, and ensure PR standards with respect to styles. This in turn will help us in better maintainability and readability of the repo overall. The changes follow up from the PR #10080 and references to the issue #10017.

This PR addresses **IDE0017: Use object initializers.**
Link to the issue: https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/style-rules/ide0017

## Customer Impact
_None_

## Regression
_None_

## Testing
Local Build Pass
Will be taken in test pass
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/wpf/pull/10168)

PR #10310, opened 2025-01-17
Remove some allocations related to GudelineSet and DoubleCollection

h3xds1nz:
## Description

- Accept stack-allocated arrays internally for `GudelineSet`.
- Accept params for `DoubleCollection` with pre-allocated capacity in `FrugalStructList<T>`.
- Use them in the framework code.

#### Sample benchmark for `GuidelineSet` creation

| ORIGINAL |  298.1 ns |    4.57 ns |     4.28 ns | 0.0329 |         866 B |         552 B |
| PR__EDIT |  257.5 ns |    3.55 ns |     3.15 ns | 0.0281 |       3,842 B |         472 B |

```csharp
[Benchmark]
public GuidelineSet Original() => new GuidelineSet([4.5, 6.5], [4.5, 6.5]);

[Benchmark]
public GuidelineSet PR__EDIT() => new GuidelineSet(new double[] { 4.5, 6.5 }, new double[] { 4.5, 6.5 });
```

## Customer Impact

## Regression

No.

## Testing

Local build.

## Risk

Low, just swaps from heap arrays to stack arrays. Part of the speed up comes from the fact the new `DoubleCollection` doesn't check the thread in `params` ctor as it makes no sense (further backed with all the code comments).

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Media/Generated/DoubleCollection.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Media/Generated/DoubleCollection.cs
@@ -948,12 +948,6 @@ namespace System.Windows.Media
                 }
             }
 
-
-
-
-
-
-
             WritePostscript();
         }
 
--- a/src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Media/Generated/DoubleCollection.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Media/Generated/DoubleCollection.cs
@@ -897,6 +897,20 @@ namespace System.Windows.Media
         }
 
         /// <summary>
+        /// Initializes a new instance using a <paramref name="values"/>. Elements are copied.
+        /// </summary>
+        /// <param name="values"></param>
+        internal DoubleCollection(params ReadOnlySpan<double> values)
+        {
+            _collection = new FrugalStructList<double>(values.Length);
+
+            foreach (double item in values)
+            {
+                _collection.Add(item);
+            }
+        }
+
+        /// <summary>
         /// Creates a DoubleCollection with all of the same elements as collection
         /// </summary>
         public DoubleCollection(IEnumerable<double> collection)
--- a/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Controls/Primitives/TickBar.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Controls/Primitives/TickBar.cs
@@ -430,11 +430,11 @@ namespace System.Windows.Controls.Primitives
             Pen pen = new Pen(Fill, 1.0d);
 
             bool snapsToDevicePixels = SnapsToDevicePixels;
-            DoubleCollection xLines = snapsToDevicePixels ? new DoubleCollection() : null;
-            DoubleCollection yLines = snapsToDevicePixels ? new DoubleCollection() : null;
+            DoubleCollection xLines = snapsToDevicePixels ? new DoubleCollection(4) : null;
+            DoubleCollection yLines = snapsToDevicePixels ? new DoubleCollection(6) : null;
 
             // Is it Vertical?
-            if ((Placement == TickBarPlacement.Left) || (Placement == TickBarPlacement.Right))
+            if (Placement is TickBarPlacement.Left or TickBarPlacement.Right)
             {
                 // Reduce tick interval if it is more than would be visible on the screen
                 double interval = TickFrequency;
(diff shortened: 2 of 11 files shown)


src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Media/Generated/DoubleCollection.cs
ThomasGoulet73, line 1: This file is auto-generated, could you add your custom constructor in a partial in a separate file (Like it was done for GuidelineCollection.cs) ?

src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Controls/Primitives/TickBar.cs
```diff
@@ -430,11 +430,11 @@ protected override void OnRender(DrawingContext dc)
             Pen pen = new Pen(Fill, 1.0d);
 
             bool snapsToDevicePixels = SnapsToDevicePixels;
-            DoubleCollection xLines = snapsToDevicePixels ? new DoubleCollection() : null;
-            DoubleCollection yLines = snapsToDevicePixels ? new DoubleCollection() : null;
+            DoubleCollection xLines = snapsToDevicePixels ? new DoubleCollection(4) : null;
+            DoubleCollection yLines = snapsToDevicePixels ? new DoubleCollection(6) : null;
```
ThomasGoulet73, line 434: Why did you change the initial capacity and what made you choose these numbers ?

src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Media/Generated/DoubleCollection.cs
h3xds1nz: @ThomasGoulet73 I really wonder whether all the effort I'm putting into updating `MilCodeGen` to correspond with my changes in my PRs is gonna have any use given how stagnant the PR fixing it is.

Raise your concern in #10186 please as well. #10168 and #10170 were already merged without those changes as nobody was there to remind it. I did my part in #10221 for example to adjust the generation.

Side-note, I think it would be best if we could first fix the generation process in a PR (e.g. the original), so it doesn't impact the generated code itself (should hopefully help the team to approve it quicker if there are no compiled-code changes) and then follow up with other PR that runs it and does necessary adjustment so that we match as closely with the current formatting/code as is and we can start running it on build then.

And yeah I'll put up a separate file for it, I just didn't pay attention this time.

src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Controls/Primitives/TickBar.cs
h3xds1nz: The underlying code; skips the 1->3 storage reallocations. I could change it to be even more efficient but didn't wanna bother with it here.

src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Media/Generated/DoubleCollection.cs
ThomasGoulet73: > I really wonder whether all the effort I'm putting into updating `MilCodeGen` to correspond with my changes in my PRs is gonna have any use given how stagnant the PR fixing it is.

Yes I need to update my PR to fix MilCodeGen because there's currently a bunch of conflicts.

> Raise your concern in #10186 please as well. #10168 and #10170 were already merged without those changes as nobody was there to remind it. I did my part in #10221 for example to adjust the generation.

I didn't notice it in the issues you linked but I think those are fine since the changes are for all generated files and I can fix them in my PR (Btw thanks for fixing the code generation code in your PRs since it makes my life easier). Your change in this PR was a bit different IMO because you modified a single generated file to add a constructor that other classes generated from the same template do not have.

> Side-note, I think it would be best if we could first fix the generation process in a PR (e.g. the original), so it doesn't impact the generated code itself (should hopefully help the team to approve it quicker if there are no compiled-code changes) and then follow up with other PR that runs it and does necessary adjustment so that we match as closely with the current formatting/code as is and we can start running it on build then.

Yes that might be a good idea. I'll think about it, though I'm not sure how useful it would be to only merge the fix to MilCodeGen without the changes to the generated files.

h3xds1nz: > I didn't notice it in the issues you linked but I think those are fine since the changes are for all generated files and I can fix them in my PR (Btw thanks for fixing the code generation code in your PRs since it makes my life easier). Your change in this PR was a bit different IMO because you modified a single generated file to add a constructor that other classes generated from the same template do not have.

I'd be happy when we would get it working as well, trust me, so no need to say thanks haha, I think it's the right way to go. Yeah, agreed my change was a bit different and I'm happy you've raised the concern.

> Yes that might be a good idea. I'll think about it, though I'm not sure how useful it would be to only merge the fix to MilCodeGen without the changes to the generated files.

Anything that can move us closer to the goal, haha.

Issue #10313, opened 2025-01-18
Stack overflow exception when creating and using multiple instances of XamlDirective in parallel


ThomasGoulet73:
### Description

This bug isn't really specific to XamlDirective but it's the class where I found the issue and it's the easiest to cause an issue. You can have the bug when using instances of XamlMember or of any of its descendants (In this case XamlDirective).

The bug is that XamlMember._reflector can be wrongly initialized when hitting this point from 3 threads and with very specific timing:
https://github.com/dotnet/wpf/blob/b08d70d354d9f0f34ecf4bcd5f214e57dd5dab1e/src/Microsoft.DotNet.Wpf/src/System.Xaml/System/Xaml/XamlMember.cs#L154

Here's the code for MemberReflector.UnknownReflector:
https://github.com/dotnet/wpf/blob/b08d70d354d9f0f34ecf4bcd5f214e57dd5dab1e/src/Microsoft.DotNet.Wpf/src/System.Xaml/System/Xaml/Schema/MemberReflector.cs#L77-L104

Here's the required timing of those 3 threads to have an invalid MemberReflector returned from MemberReflector.UnknownReflector:
Thread 1 and 2 invoke MemberReflector.UnknownReflector and find that s_UnknownReflector is null. Thread 1 assigns s_UnknownReflector to a new instance. Thread 3 invokes MemberReflector.UnknownReflector, finds that s_UnknownReflector is not null and returns s_UnknownReflector. Thread 2 assigns s_UnknownReflector to a new instance.

In this scenario thread 1 and 2 both return the same instance from MemberReflector.UnknownReflector while thread 3 returned an old instance from MemberReflector.UnknownReflector that wasn't fully initialized.

<br/>
<br/>

Now that I've explained how to have a partially initialized MemberReflector, here's how to have a stack overflow exception using XamlDirective:
We enter XamlDirective.Type when _reflector.Type is null which means that we enter this if block:
https://github.com/dotnet/wpf/blob/b08d70d354d9f0f34ecf4bcd5f214e57dd5dab1e/src/Microsoft.DotNet.Wpf/src/System.Xaml/System/Xaml/XamlMember.cs#L251-L254

In the if block it calls the virtual method XamlDirective.LookupType() which is implemented in XamlDirective to return base.Type: 
[...]

PR #10314, opened 2025-01-18
Fix XamlDirective stack overflow exception

ThomasGoulet73:
Fixes dotnet/wpf#10313

## Description
I changed MemberReflector.UnknownReflector to be thread safe by initializing the property from the static constructor instead of doing the lazy initialization manually. The static constructor of a type is guaranteed to be called only once.

See the issue for an explanation of the bug and how it's hit.

## Customer Impact
Could fix a crash or weird behavior at runtime but it's very hard to hit (See issue).

## Regression
No.

## Testing
Used the repro in the issue which crashed before this PR and passes with this PR.

## Risk
Low.


miloush:
Wouldn't it be enough if you did
```C#
private static MemberReflector s_UnknownReflector;
internal static MemberReflector UnknownReflector => s_UnknownReflector ??= CreateUnknownReflector();
```

h3xds1nz:
Is there a reason why we're not keeping lazy-init, just making it TS instead?

ThomasGoulet73:
It was lazy because static fields only get assigned when necessary and in this case CreateUnknownReflector was only called on the first call to UnknownReflector and it was thread-safe because static constructors are thread safe.

I pushed a commit to change it to make the lazy init explicit and more clear.

ThomasGoulet73:

PR #10328, opened 2025-01-23
Remove allocations in BaseUriHelper, improve resource resolution performance

h3xds1nz:
Currently blocked by #9822, hence draft.

## Description

Optimizes methods in `BaseUriHelper`, removes most of the allocations directly in this class and opens up more optimization possibilities; directly builds upon #9739 and #9822.

- Removes `IsSameKeyToken` from SecurityHelper in favour of `ReflectionUtils.IsSamePublicKeyToken`.
- `BaseUriHelper.GetAssemblyNameAndPart` is now fully allocation free.
    - This allows for more optimizations in the future and already makes some paths without any allocations.
    - Instead of returning 3 strings, we return slices in `AssemblyPackageInfo` `ref struct`.
- In `FontResourceCache`, we can now use `AlternateLookup` to benefit from this change as well.
- `BaseUriHelper.BaseUri` was never set outside static constructor, where it was initialized to the `s_packAppBaseUri` value,
hence I've removed the property altogether, which also led to removal of `BindUriHelper.BaseUri` which was just proxy.
- Code under `CF_Envelope_Activation_Enabled` macro was removed, most of the code is already missing in .NET Core than what was in reference files for .NET 3.5 and it is never compiled on NetFX nor on .NET Core.
- Changes in `GetLoadedAssembly` bring a bit of an improvement but there's more benchmarking needed on how to handle this chain most efficiently, so keeping as it is right now.

Cherry-picking some cases:

### GetAssemblyNameAndPart

| Method   | someUri             | Mean [ns] | Error [ns] | StdDev [ns] |  Gen0   | Allocated [B] |
|--------- |-------------------- |----------:|-----------:|------------:|---------:|--------------:|
| Original | /Mic(...)xaml [103] | 104.95 ns |   2.274 ns |    6.706 ns |     0.0448 |         752 B |
| PR__EDIT | /Mic(...)xaml [103] |  45.42 ns |   0.475 ns |    0.445 ns |         - |             - |

### AssemblyMatchesKeyString vs ComparePublicKeyTokens

| Method   | someAssembly         | Mean [ns] | Error [ns] | StdDev [ns] | Gen0   | Allocated [B] |
|--------- |--------------------- |----------:|-----------:|------------:|-------:|---------------:|
| Original | Syste(...)7798e [89] | 40.755 ns |  0.3990 ns |   0.3537 ns | 0.0019 |          32 B |
| PR__EDIT | Syste(...)7798e [89] |  8.135 ns |  0.0835 ns |   0.0740 ns |      - |                 - |

## Customer Impact

Improved performance, decrased allocations.

## Regression

No.

## Testing

Local build, some assert testing; I'll provide tests before this goes out of draft.

## Risk

Low to medium, there are quite a few changes but most of them are rather straightforward.

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Navigation/BaseUriHelper.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Navigation/BaseUriHelper.cs
@@ -1,6 +1,7 @@
 ﻿// Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the MIT license.
-// See the LICENSE file in the project root for more information.
+
+using PackUriHelper = System.IO.Packaging.PackUriHelper;
 
 using System.Globalization;
 using System.Text;
@@ -13,8 +14,6 @@ using MS.Internal.AppModel;
 using MS.Internal.IO.Packaging;
 using MS.Internal.PresentationCore;
 
-using PackUriHelper = System.IO.Packaging.PackUriHelper;
-
 namespace System.Windows.Navigation
 {
     /// <summary>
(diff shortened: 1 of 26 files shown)


src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Navigation/BaseUriHelper.cs
```diff
@@ -13,62 +14,47 @@
...
         ///     Default Value: null.
         /// </summary>
-        public static readonly DependencyProperty BaseUriProperty =
-                    DependencyProperty.RegisterAttached(
-                                "BaseUri",
-                                typeof(Uri),
-                                typeof(BaseUriHelper),
-                                new PropertyMetadata((object)null));
-
+        public static readonly DependencyProperty BaseUriProperty = DependencyProperty.RegisterAttached("BaseUri", typeof(Uri), typeof(BaseUriHelper), new PropertyMetadata(null));
```
ThomasGoulet73, line 51: Removing the cast to object means that this code will use [this constructor](https://learn.microsoft.com/en-us/dotnet/api/system.windows.propertymetadata.-ctor?view=windowsdesktop-9.0#system-windows-propertymetadata-ctor(system-windows-propertychangedcallback)).

I'm not sure if it changes the behavior but might as well keep the same behavior and you can juste use a named parameter (Which IMO is also cleaner than just having null):
```suggestion
        public static readonly DependencyProperty BaseUriProperty = DependencyProperty.RegisterAttached("BaseUri", typeof(Uri), typeof(BaseUriHelper), new PropertyMetadata(defaultValue: null));
```

ThomasGoulet73: Also, what's wrong with the original indentation ? Most dependency/attached property registration in this repo use the same indentation style, I don't see why we should change it. Personally I prefer the previous style and other repos like WinForms have a recommended max line length ([WinForms uses recommended < 120 and max < 150](https://github.com/dotnet/winforms/blob/main/docs/coding-style.md#line-breaks))

h3xds1nz: It won't change behaviour but I like the named argument there, let's use that. Thanks.

As for indentation; since the class is 90% changes, it was auto format. Reverted it as the line is indeed too long but I wouldn't go as far as saying "use the same indentation style", I can select a random file e.g. `DataGrid` files and I'll see 6 different ways to indent a dependency property. 😀

PR #10333, opened 2025-01-24
Normalize start Text pointer to support Text Cursor Indicator

harshit7962:
## Description
Using accessibility option of _text cursor indicator_ in `RichTextBox` sometimes lead to crashing applications due to Invariant assert [here](https://github.com/dotnet/wpf/blob/main/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Controls/TextRangeAdaptor.cs/#L462). Adding `Normalize()` call in changed location moves the `end` to the `start` if this invariant assert fails and hence essentially corrects the cursor position.

## Customer Impact
Accessibility fix. Allowing the usage of text cursor indicator with RichTextBox.

## Regression
_None_

## Testing
Local Build Pass
Sample Application Testing
Test Pass

## Risk
Low

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Controls/TextRangeAdaptor.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Controls/TextRangeAdaptor.cs
@@ -353,7 +353,7 @@ namespace MS.Internal.Automation
                     },
                     delegate(object val1, object val2) { return (double)val1 == (double)val2; })
                 );
-            // OutlineStyles 
+            // OutlineStyles
             _textPatternAttributes.Add(
                 TextPatternIdentifiers.OutlineStylesAttribute,
                 new TextAttributeHelper(
@@ -466,12 +466,12 @@ namespace MS.Internal.Automation
             _end = end.CreatePointer();
             _textPeer = textPeer;
         }
-        #endregion Constructors        
+        #endregion Constructors
 
         #region Internal methods
-        
+
         /// <summary>
-        /// This wrapper is to cover up a shortcoming in MoveToInsertionPosition code. 
+        /// This wrapper is to cover up a shortcoming in MoveToInsertionPosition code.
         /// If the position is inside a Hyperlink, it is being moved out of the hyperlink and to the previous
         /// insertion position which can be on the previous line (or unit) and this is creating problems
         /// This is a temporary solution until we find a better solution to handle the case
@@ -487,7 +487,7 @@ namespace MS.Internal.Automation
             return false;
         }
         #endregion Internal methods
-        
+
 
 
 
@@ -545,7 +545,7 @@ namespace MS.Internal.Automation
                     break;
 
                 case TextUnit.Format:
-                    // Formatting changes can be introduced by elements. Hence it is fair to 
+                    // Formatting changes can be introduced by elements. Hence it is fair to
                     // assume that formatting boundaries are defined by non-text context.
 
                     if (expandStart)
@@ -749,7 +749,7 @@ namespace MS.Internal.Automation
                     break;
 
                 case TextUnit.Format:
-                    // Formatting changes can be introduced by elements. Hence it is fair to 
+                    // Formatting changes can be introduced by elements. Hence it is fair to
                     // assume that formatting boundaries are defined by non-text context.
                     while (position.GetPointerContext(direction) == TextPointerContext.Text)
                     {
@@ -915,9 +915,9 @@ namespace MS.Internal.Automation
 
                         if (textSegments != null && textSegments.Count > 0)
                         {
-                            //When comparing, we need to take into account if the pointer is not right at 
+                            //When comparing, we need to take into account if the pointer is not right at
                             //the end of the page (or beginning) because of normalization
-                            
+
                             if (direction == LogicalDirection.Forward)
                             {
                                 while (position.CompareTo(textSegments[textSegments.Count - 1].End) != 0)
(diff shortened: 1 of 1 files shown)


miloush:
I see change on lines 1813-15. So you get the bounding rectangle, then potentially change it and then return the previous rectangle values? Doesn't sound like the intended API experience.

I also don't see how changing the order of Normalize and GetBoundingRectangle in this method changes the invariant after exiting the method. Even if the adapter called back for a clone, it would be after Normalize before the change, so putting Normalize at the beginning Clone should have not fixed it.

Can you share some of the investigation on what breaks the invariant or when?

h3xds1nz:
I agree with miloush. I'm a bit puzzled. Sharing the investigation would be great to understand this properly.

> So you get the bounding rectangle, then potentially change it and then return the previous rectangle values? Doesn't sound like the intended API experience.

And this exactly concerns me.

harshit7962:
So the `GetBoudingRectangles` eventually leads the call to `NormalizePosition` method of `TextPointerBase`, where based on the `LogicalDirection`, the start pointer deviates from its expected value.

**Flow:**
- TextRangeAdaptor -> ITextRangeProvider.GetBoundingRectangles
- This calls GetBoundingRectangles of TextAdaptor
- which calls MoveToInsertionPosition of TextRangeAdaptor from [here](https://github.com/dotnet/wpf/blob/main/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Controls/TextAdaptor.cs#L115)
- This calls the MoveToInsertionPosition of TextPointer which again calls MoveToInsertionPosition of TextPointerBase
- This is where [NormalizePosition](https://github.com/dotnet/wpf/blob/main/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Documents/TextPointerBase.cs#L1136) is called and our invariance breaks.

harshit7962:
Building upon the above comment, basically the idea of the change is that `GetBoundingRectangles` in `TextRangeAdaptor` uses a start textpointer and an end textpointer, during the calculation of which it changes the start textpointer causing the invariance. After the calculation is done with, we Normalize the start pointer back to what the original value was.

How do we certain that the start is going back to its original value post Normalize:
The `MoveToInsertionPosition` which causes the change in GetBoundingRectangles, is also called in `Normalize` with LogicalDirection as reverse of what it was earlier.

cc:/ @miloush, @h3xds1nz

h3xds1nz:
@harshit7962 sorry, somehow managed to miss this.

From a brief overview it makes sense with your explanation but I'm still not sure whether that's the correct solution; imho it shouldn't happen in the first place. Do we have a repro? I'd just like to step through.

harshit7962:
Thanks for willing to take a look into it, for the repro, use the following xaml
```xml
<RichTextBox> 
    <FlowDocument> 
        <Paragraph> 
            <Bold>a</Bold>
            <Bold>b</Bold> 
        </Paragraph> 
   </FlowDocument> 
</RichTextBox> 
```

Turn on the text cursor indicator from settings. Settings -> Accessibility -> Text Cursor
Launch the application and move the indicator between a and b, it should crash the application.

harshit7962:
@h3xds1nz, @miloush - If there are no more concerns, we are planning on moving forward with this.

h3xds1nz:
@harshit7962 Apologies, didn't have the bandwidth yet, I'll take a look today/tomorrow max. to no longer delay this.

miloush:
@harshit7962 I can repro on .NET 9 but not on .NET Framework, is that expected?

h3xds1nz:
@miloush https://github.com/dotnet/wpf/pull/10281#pullrequestreview-2552327198

harshit7962:
@miloush so in the framework code, the `Clone` method does the normalization as we tried in #10281 (As pointed out by @h3xds1nz). But as was discussed earlier, this probably is not the best solution and hence we tried the current change.

h3xds1nz:
@harshit7962 Alright, I can see that calling `Normalize` (which I'd just move into the `TextRangeAdaptor` ctor by the way, therefore the assert can just be in `DEBUG`, so we know something is wrong there - if we're looking for a workaround, so its essentially same as the original "fix" but we just always perform normalization on construction) workarounds the issue but doesn't solve it.

However I'm not keen on spending more time on stuff I don't particularly understand right now (never worked with the internal document nodes - I have no idea why the `end` contains a `Run` (within `TextTreeTextElementNode`) and the difference between `start`'s `TextTreeTextNode` which doesn't contain `TextElement` when this assert crumbles). My knowledge around `FlowDocument` is limited in general.

Hopefully miloush knows more.

miloush:
OK I managed to have a look.

This is the Normalize() method:

https://github.com/dotnet/wpf/blob/d41123486e07b2fca6b76e435cda292dc3487f1a/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Controls/TextRangeAdaptor.cs#L1489-L1499

MoveToInsertionPosition() is the bit that manipulates the start/end separately and can potentially break the invariant. Therefore, after that is done, you need to make sure that end >= start.

Now in the context of this scenario, the only point where MoveToInsertionPosition() is called outside of Normalize is in GetBoundingRectangles(), before the actual rectangles are calculated:

https://github.com/dotnet/wpf/blob/d41123486e07b2fca6b76e435cda292dc3487f1a/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Controls/TextAdaptor.cs#L115-L116 

So I suggest that the invariant ensuring code is added after these two lines:

```c#
if (start.CompareTo(end) > 0)
{
    end.MoveToPosition(start);
}
```

harshit7962:
@miloush thanks for the suggestion. A quick couple of things though
1. shouldn't it be `start.MoveToPosition(end)` given that (in this case) its the start pointer that has deviated from what was provided to the parent call. Also, this is what we were kind of achieving via the `Normalize` call.
2. I am a bit concerned about the `Rect[]` being returned by `GetBoudingRectangles` given that we change the `TextPointer` kind of before the actual calculations/logic. So to put it more clearly, earlier in `TextRangeAdaptor` we used to get the `rects` based on whateved the _start and _end was computed withing the function logic and now it will alter the start/end and then return the `rects`. This looks like a possible behavioral change.

Tbh I am still trying to reason the 2nd point out completely, so had just mentioned it here to get your inputs.

harshit7962:
@miloush Thanks again for the follow-up. This sounds like a candidate fix, and a better alternative. Will go ahead and get your suggestion tested.

miloush:
As long as the formatting changes are expected, LGTM!

dipeshmsft (approved):

harshit7962:
Thanks @miloush and @h3xds1nz for your help.

PR #10337, opened 2025-01-26
Disable code analysis for generated code

ThomasGoulet73:
## Description
Disables code analysis for generated code. Code analysis for generated code does not make sense since generated code will never be perfectly formatted and it prevents doing auto formatting to generated code without modifying the code generator.

We'll probably want to format the generated code someday but it should probably be done only when dotnet/wpf#6135 is merged and we can modify the code generator instead of modifying the generated code manually.

## Customer Impact
None.

## Regression
No.

## Testing
Local build + CI.

## Risk
None.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/wpf/pull/10337)

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -2284,3 +2284,6 @@ dotnet_diagnostic.IDE0251.severity = suggestion
 
 # IDE1006: Naming Styles
 dotnet_diagnostic.IDE1006.severity = suggestion
+
+[**/Generated/*.*]
+generated_code = true


dipeshmsft:
Even when we have enabled MilCodeGen, we would still need to keep some rules disabled for the generated code right ? Like the rule for unnecessary imports, is something that needs to be disabled in that case.

/cc @harshit7962

ThomasGoulet73:
The typical behavior of code analyzers is to ignore generated files, code analysis uses a [specifc pattern](https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/configuration-options#exclude-generated-code) to detect generated files and the files generated by MilCodeGen does not match that pattern. This PR makes it so that code analysis knows that MilCodeGen files are generated.

Personally I don't see much value in enabling code analysis for generated files. Generated code will probably never be perfect and IMO it should be done manually (By removing the line I added locally and manually checking the warnings).

harshit7962 (approved):
LGTM, as suggested we can take care of styling in generated code later.

ThomasGoulet73:
Thanks @harshit7962

PR #10338, opened 2025-01-26
Fix unit tests native dependencies

ThomasGoulet73:
This is a workaround for dotnet/runtime#111825

## Description
Fixes errors in unit tests when trying to load wpfgfx_cor3.dll which depends on D3DCompiler_47_cor3.dll which can't be found at runtime which seems to be caused by a bug in the runtime when using `DllImportSearchPath.AssemblyDirectory | DllImportSearchPath.System32` as the search path. 

This search path is the default search path used by all projects in this repo and it's configured here:
https://github.com/dotnet/wpf/blob/81019fc2492bfc15620ef14053f4f3f22eeeac6c/eng/WpfArcadeSdk/tools/ExtendedAssemblyInfo.props#L16-L20

We can't just set `IncludeDllSafeSearchPathAttribute` to false in the test projects since the attribute is used to configure the search path used by default for each assembly. We would need to set it to false for all assemblies trying to load wpfgfx_cor3.dll.

**This problem does not occur for apps using WPF, see the issue for more info**

## Customer Impact
None, test only.

## Regression
No.

## Testing
Local build + CI + local test.

## Risk
None, test only.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/wpf/pull/10338)

### DIFF
--- a/src/Microsoft.DotNet.Wpf/tests/UnitTests/Directory.Build.props
+++ b/src/Microsoft.DotNet.Wpf/tests/UnitTests/Directory.Build.props
@@ -6,4 +6,9 @@
   <PropertyGroup>
     <WpfTestsDir>$(MsBuildThisFileDirectory)</WpfTestsDir>
   </PropertyGroup>
+
+  <ItemGroup>
+    <Compile Include="$(MSBuildThisFileDirectory)Shared\ModuleInitializer.cs" />  
+  </ItemGroup>
+
 </Project>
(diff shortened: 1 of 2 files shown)


ThomasGoulet73:
@dipeshmsft: This is the workaround I was talking about here: https://github.com/dotnet/wpf/pull/10304#issuecomment-2613764193

dipeshmsft:
Thanks @ThomasGoulet73, I took a look at this, and it fixes the issue I was talking about. I am going ahead with the merge. Once again, thanks for investigating this issue.

dipeshmsft (approved):

Issue #8552, opened 2023-12-14
Testing


singhashish-wpf:
- [ ] Each change should also have the necessary unit tests which will be run on the CI.
- [ ] Tests should cover all states of the controls. (Active, Hover, Selected etc)
- [ ] Test that the theme changes are happening correctly over different scenarios of system theme change and user theme change.
- [ ] Test that the themes are being loaded properly.
- [ ] Test that switching theme is not leading to inconsistencies.
- [ ] Test that the performance isn't getting degraded due to theme changes.
- [ ] Test 3rd party controls for proper working after theme changes
- [ ] Test all local wpf tests including DRTs, Microsuites and Feature Tests

Issue #9906, opened 2024-10-07
WPF NET 9.0 (latest preview)-based app that uses the built-in Fluent theme crashes when Theme is changed


danielkornev:
### Description

This problem appears on two kinds of OS:
* Windows 10 (AMD64, 24H2)
* Windows 11 (ARM64, 24H2)

Here's the error I've got in the beginning on both machines:
```
Monday, October 7, 2024, 10/7/2024 12:48:28 PM
Exception [System.InvalidOperationException]: The calling thread cannot access this object because a different thread owns it.
System.InvalidOperationException: The calling thread cannot access this object because a different thread owns it.
   at System.Windows.Threading.Dispatcher.<VerifyAccess>g__ThrowVerifyAccess|7_0()
   at System.Windows.ThemeManager.OnSystemThemeChanged()
   at System.Windows.SystemResources.SystemThemeFilterMessage(IntPtr hwnd, Int32 msg, IntPtr wParam, IntPtr lParam, Boolean& handled)
   at MS.Win32.HwndWrapper.WndProc(IntPtr hwnd, Int32 msg, IntPtr wParam, IntPtr lParam, Boolean& handled)
   at System.Windows.Threading.ExceptionWrapper.InternalRealCall(Delegate callback, Object args, Int32 numArgs)
   at System.Windows.Threading.ExceptionWrapper.TryCatchWhen(Object source, Delegate callback, Object args, Int32 numArgs, Delegate catchHandler)
   at System.Windows.Threading.Dispatcher.LegacyInvokeImpl(DispatcherPriority priority, TimeSpan timeout, Delegate method, Object args, Int32 numArgs)
   at MS.Win32.HwndSubclass.SubclassWndProc(IntPtr hwnd, Int32 msg, IntPtr wParam, IntPtr lParam)
```

Sadly there was no way to figure out why this error happened. My app showed this message and then crashed. App uses Microsoft.VisualBasic library to get a Single Instance behavior, so catching the exception would get us to app.Run method.

After several experiments, the app now crashes w/o even this error after the theme of the apps in System Settings is changed to Light/Dark.

### Reproduction Steps

1. Create a WPF 9.0 app that uses Visual Basic assembly's Application to get a Single Instance Application
2. Use Fluent theme as a ResourceDictionary across your app
3. Switch theme of all apps in System Settings from Light to Dark or vice versa

### Expected behavior

Theme changes where expected

### Actual behavior

App crashes (initially with an error message and now silently)

### Regression?

[...]

PR #10385, opened 2025-02-03
[ Draft ] Adding tests for Fluent theme

dipeshmsft:
Issue Link : #8552

### Description

Hello community members, we are working on writing tests for Fluent UI and ThemeMode property, and here is a basic structure of how I am planning to do this and would like to have your suggestions on how we can make this better. It is okay, if there is a completely different approach that should be taken up.

The tests that are present in this PR, can be divided into three parts : 
1. **ThemeMode Property Tests** : In this set of tests, the main aim is to test the initialization of ThemeMode properties on Window and Application classes. It will include initialization of Application and Window, switching ThemeMode, combination Application and Window ThemeMode properties.
2. **Fluent ResourceDictionary Tests** : In this set of tests, the main aim is to check the correctness of Fluent colors and brushes ( not control brushes ), loading of different Fluent ResourceDictionary, etc.
3. **Control Styles UI Tests** : These set of tests are meant to test the behavior and default property values of controls when Fluent this is applied. Here are more details on this : 
    - First, I have created a copy of all the Fluent colors and brushes, which will be used as the source of truth.
    - In each test, we launch a window with a control ( can have custom values set for the different properties )
    - For each test there is a dictionary, which corresponds to what values all the properties of controls and and different elements in the template of the control should have, and we verify these values by matching them with the values specified in the dictionary.
    - To avoid redundancy, there is a base ( default ) dictionary for a control, where the property values correspond to Fluent default values for that control. 
    - For subsequent tests ( like disabled button, mouse over, Background = Red ) we have separate dictionaries which have a subset of the values defined in the base dictionary. We combine the base dictionary with the dictionary for the current test and use that to verify the different properties of the control in that state.

From the approach that we have followed here, it may be better to put these tests under Integration Tests, but before that, I want your opinions on what limitations we can face with the above approach, or if there is some other approach that we can take to write these tests in a better fashion. 

/cc @h3xds1nz @hughbe @JeremyKuhne @ThomasGoulet73 @miloush @lindexi @batzen 
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/wpf/pull/10385)

### DIFF
--- a/src/Microsoft.DotNet.Wpf/tests/UnitTests/PresentationFramework.Fluent.Tests/ControlTests/BaseControlTests.cs
+++ b/src/Microsoft.DotNet.Wpf/tests/UnitTests/PresentationFramework.Fluent.Tests/ControlTests/BaseControlTests.cs
@@ -1,0 +1,230 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using PresentationFramework.Fluent.Tests.TestUtilities;
+using System.Windows.Controls;
+
+
+namespace PresentationFramework.Fluent.Tests.ControlTests;
+
+public class BaseControlTests : IDisposable
+{
+    public BaseControlTests()
+    {
+        TestWindow = SetupTestWindow();
+    }
+
+    public virtual List<FrameworkElement> GetStyleParts(Control element)
+    {
+        return new List<FrameworkElement>();
+    }
+
+    public virtual void VerifyControlProperties(FrameworkElement element, ResourceDictionary expectedProperties)
+    {
+        return;
+    }
+
+    #region Protected Methods
+
+    protected void AddControlToView(Window window, FrameworkElement fe)
+    {
+        ArgumentNullException.ThrowIfNull(window);
+
+        if (window.Content is not StackPanel panel)
+            throw new ArgumentException($"The window does not have StackPanel as it's root element.");
+        
+        panel.Children.Add(fe);
+    }
+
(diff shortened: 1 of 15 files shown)


h3xds1nz:
A bit of off-topic: The lucky test failure in debug is pretty much what https://github.com/dotnet/wpf/issues/9906 was about by the looks of it, and I stand by what I said, that this should be queued on the app thread.

src/Microsoft.DotNet.Wpf/tests/UnitTests/PresentationFramework.Fluent.Tests/ControlTests/BaseControlTests.cs
```diff
@@ -0,0 +1,230 @@
...
+
+        return window;
+    }
+
+    protected Window SetupTestWindow(ColorMode mode)
+    {
+        Window window = new Window() { Width = 800, Height = 600 };
+        StackPanel sp = new StackPanel() { Name = "RootPanel" };
+        SetColorMode(window, mode);
+        window.Content = sp;
```
Kuldeep-MS, line 81: I noticed that you are setting `window.ThemeMode` to `ThemeMode.None` in `ResetTestWindow`, but you are not setting it in `SetupTestWindow` to some valid value. Additionally, you are accepting the mode as an argument and adding the current window to the `TestWindows` map corresponding to the `mode`, but the `mode` is not being set in the window object.

dipeshmsft:
Closing this due to lack of discussion, not sure if this is the right way to go forward. Will be reopened once we are done with the internal reviews and sanity checks.

PR #10391, opened 2025-02-04
Use shared DataFormats code

JeremyKuhne:
This uses the shared DataFormats code from System.Private.Windows.Core.

Note that locking is handled by the shared code so there is no need to do so when adding the WPF specific types. The only reason they need added is that casing will stick with whatever the first request is. That is, if you ask for "XAML" that will be the string Windows gives back, not "Xaml".
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/wpf/pull/10391)

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/PresentationCore/GlobalSuppressions.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationCore/GlobalSuppressions.cs
@@ -1,0 +1,28 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System.Diagnostics.CodeAnalysis;
+
+[assembly: SuppressMessage("Style", "IDE1006:Naming Styles", Justification = "Public API", Scope = "member", Target = "~F:System.Windows.DataFormats.Text")]
+[assembly: SuppressMessage("Style", "IDE1006:Naming Styles", Justification = "Public API", Scope = "member", Target = "~F:System.Windows.DataFormats.UnicodeText")]
+[assembly: SuppressMessage("Style", "IDE1006:Naming Styles", Justification = "Public API", Scope = "member", Target = "~F:System.Windows.DataFormats.Dib")]
+[assembly: SuppressMessage("Style", "IDE1006:Naming Styles", Justification = "Public API", Scope = "member", Target = "~F:System.Windows.DataFormats.Bitmap")]
+[assembly: SuppressMessage("Style", "IDE1006:Naming Styles", Justification = "Public API", Scope = "member", Target = "~F:System.Windows.DataFormats.EnhancedMetafile")]
+[assembly: SuppressMessage("Style", "IDE1006:Naming Styles", Justification = "Public API", Scope = "member", Target = "~F:System.Windows.DataFormats.CommaSeparatedValue")]
+[assembly: SuppressMessage("Style", "IDE1006:Naming Styles", Justification = "Public API", Scope = "member", Target = "~F:System.Windows.DataFormats.Dif")]
+[assembly: SuppressMessage("Style", "IDE1006:Naming Styles", Justification = "Public API", Scope = "member", Target = "~F:System.Windows.DataFormats.FileDrop")]
+[assembly: SuppressMessage("Style", "IDE1006:Naming Styles", Justification = "Public API", Scope = "member", Target = "~F:System.Windows.DataFormats.Html")]
+[assembly: SuppressMessage("Style", "IDE1006:Naming Styles", Justification = "Public API", Scope = "member", Target = "~F:System.Windows.DataFormats.Locale")]
+[assembly: SuppressMessage("Style", "IDE1006:Naming Styles", Justification = "Public API", Scope = "member", Target = "~F:System.Windows.DataFormats.MetafilePicture")]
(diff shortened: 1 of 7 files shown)


src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/DataFormats.cs
```diff
@@ -1,472 +1,190 @@
...
-        /// Forms uses to store string objects. This
-        /// field is read-only.
-        /// </summary>
-        public static readonly string StringFormat = typeof(string).FullName;
-
-        /// <summary>
-        /// Specifies a format that encapsulates any type of Windows Design
-        /// object. This field is read-only.
-        /// </summary>
-        public static readonly string Serializable = "PersistentObject";
```
miloush, line 243: `DataFormatNames.Serializable` seems to be `WindowsForms10PersistentObject` rather than `PersistentObject`. Isn't that a breaking change?

miloush:
Where can I see the value of `PInvokeCore.INK_SERIALIZED_FORMAT`?

src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/dataobject.cs
```diff
@@ -2732,20 +2732,19 @@ private object GetDataFromHGLOBAL(string format, IntPtr hglobal)
                     {
                         data = ReadStringFromHandle(hglobal, false);
                     }
-                    else if (IsFormatEqual(format, DataFormats.UnicodeText) ||
-                             IsFormatEqual(format, DataFormats.ApplicationTrust))
+                    else if (IsFormatEqual(format, DataFormats.UnicodeText))
```
Kuldeep-MS, line 2735: why we are removing `IsFormatEqual(format, DataFormats.ApplicationTrust))` check?

JeremyKuhne: Because it was specific to CAS scenarios on .NET Framework. It has also been removed in WinForms.

miloush: Although not directly related to the code change, is the idea that objects from low-trust .NET Framework applications can be transferred to full-trust .NET Core applications?

JeremyKuhne: @miloush I'd have to dig deeper to give you a full answer, but you can see one of the types involved in the reference source for WPF.

https://referencesource.microsoft.com/#PresentationCore/Core/CSharp/MS/Internal/[ConstrainedDataObject.cs](https://referencesource.microsoft.com/#PresentationCore/Core/CSharp/MS/Internal/ConstrainedDataObject.cs)

miloush: that's what caused my prompt, since this will no longer be used for low-trust FW objects...

https://referencesource.microsoft.com/#PresentationCore/Core/CSharp/System/Windows/Clipboard.cs,810

src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/dataobject.cs
```diff
@@ -1469,23 +1470,22 @@ private int GetDataIntoOleStructsByTypeMedimHGlobal(string format, object data,
             {
                 hr = SaveStringToHandle(medium.unionmember, data.ToString(), false /* unicode */, doNotReallocate);
             }
-            else if (IsFormatEqual(format, DataFormats.UnicodeText)||
-                     IsFormatEqual(format, DataFormats.ApplicationTrust))
+            else if (IsFormatEqual(format, DataFormats.UnicodeText))
```
Kuldeep-MS, line 1473: why we are skippin `IsFormatEqual(format, DataFormats.ApplicationTrust)` check?

src/Microsoft.DotNet.Wpf/src/PresentationCore/GlobalSuppressions.cs
```diff
@@ -0,0 +1,28 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System.Diagnostics.CodeAnalysis;
+
+[assembly: SuppressMessage("Style", "IDE1006:Naming Styles", Justification = "Public API", Scope = "member", Target = "~F:System.Windows.DataFormats.Text")]
```
Kuldeep-MS, line 6: Why we are suppressing all these CodeAnalysis messages?

Kuldeep-MS (approved):
Changes looks good to me. Can you please add the reference to the PR which is handling the new common code for dataformats?

src/Microsoft.DotNet.Wpf/src/PresentationCore/GlobalSuppressions.cs
JeremyKuhne: Because they are existing public API that violate guidelines. Putting the suppressions here instead of inline makes the code easier to read.

JeremyKuhne:
> Changes looks good to me. Can you please add the reference to the PR which is handling the new common code for dataformats?

The primary refactoring was https://github.com/dotnet/winforms/pull/12802. The shared code had a number of other edits past that.

PR #10394, opened 2025-02-04
Avoid UCEERR_RENDERTHREADFAILURE

### LINKED ISSUE #9042: One way to fix UCEERR_RENDERTHREADFAILURE
I know that are many reasons for this error but I found that for my case it was related to using D3DImage.

Here is link to the project that demonstrates problem : https://github.com/vladimir-cheverdyuk-altium/dx9host
Basically it is Microsoft code copied from here:
https://learn.microsoft.com/en-us/dotnet/desktop/wpf/advanced/walkthrough-creating-direct3d9-content-for-hosting-in-wpf?view=netframeworkdesktop-4.8
https://learn.microsoft.com/en-us/dotnet/desktop/wpf/advanced/walkthrough-hosting-direct3d9-content-in-wpf?view=netframeworkdesktop-4.8

It has project for .NET Framework and .NET6.

**Steps**:
1. You need 2 monitors with different screen resolution.
2. Start app on main monitor and maximize it
3. Power off this monitor or change display settings to display only on second monitor
4. Windows will move the application to the second monitor and it will crash with UCEERR_RENDERTHREADFAILURE

**Source of the problem**
It looks like IDirect3D9.GetAdapterMonitor will return 0 because monitor is powered off (or unavailable) and function `pDisplaySet->GetDisplayIndexFromMonitor(hMon, uDisplayIndex)` in `CInteropDeviceBitmap::GetDisplayFromUserDevice` will return E_INVALIDARG and this eventually will lead to crash.

So to me it looks like issue is in D3DImage that connected to adapter 0 while adapter 0 is disconnected from monitor.

**Fix***
To fix it I declared `private EventHandler renderingEventHandler;` in `MainWindow.xaml.cs` then set rendering like this:
```
            renderingEventHandler = new EventHandler(CompositionTarget_Rendering);
            CompositionTarget.Rendering += renderingEventHandler;
```
and add these 2 lines to the end of constructor:
```
            Microsoft.Win32.SystemEvents.DisplaySettingsChanging += DisplaySettingsChanging;
            Microsoft.Win32.SystemEvents.DisplaySettingsChanged += DisplaySettingsChanged;
```

and lastly I added implementation:
```
        private void DisplaySettingsChanging(object sender, EventArgs e)
        {
            CompositionTarget.Rendering -= renderingEventHandler;
            d3dimg.Lock();
            d3dimg.SetBackBuffer(D3DResourceType.IDirect3DSurface9, IntPtr.Zero, true);
            d3dimg.Unlock();
        }

        private void DisplaySettingsChanged(object sender, EventArgs e)
        {
            POINT p = new POINT(imgelt.PointToScreen(new Point(0, 0)));

            HRESULT.Check(SetAdapter(p));

            CompositionTarget.Rendering += renderingEventHandler;
        }
```
and lastly I added 3rd parameter `true` to every call to `SetBackBuffer` and removed `d3dimg.IsFrontBufferAvailable` check in `CompositionTarget_Rendering`.

Does anybody know if this approach to detect monitor changes (using Microsoft.Win32.SystemEvents.DisplaySettingsChanging) is correct?

Just in case if somebody thing it is rare thing, we have 37933 crash reports from users for this issue. I'm not sure how many related to this particular reason but at least some are.
Here is one of reports from our customers: "I also get this issue on my Lenovo laptop when connecting/disconnecting external monitors." Some customers state that it happened when computer is going to sleep or perhaps when monitors going to sleep but I couldn't able to reproduce it.

petsuter:
Fixes #9042

Main PR

## Description

WPF applications with D3DImage can crash when a monitor is off or unavailable and IDirect3D9.GetAdapterMonitor returns NULL.
For example on laptops with connected HDMI external display and display setting “Extend”
and control panel “when I close the lid” = “Do nothing”, when closing the lid or on sleep / awake.

## Customer Impact

WPF applications crash.

## Regression

This bug exists in existing .NET releases like 4.8.

## Testing

Take a laptop with connected HDMI external display and set display setting “Extend”
and control panel “when I close the lid” = “Do nothing”, then close the lid and check if a WPF application with D3DImage crashes.

## Risk

Unclear.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/wpf/pull/10394)

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/WpfGfx/core/hw/InteropDeviceBitmap.cpp
+++ b/src/Microsoft.DotNet.Wpf/src/WpfGfx/core/hw/InteropDeviceBitmap.cpp
@@ -845,6 +845,10 @@ CInteropDeviceBitmap::GetDisplayFromUserDevice(
     IFC(pID3DUserDevice->GetDirect3D(&pID3DUserObject));
 
     HMONITOR hMon = pID3DUserObject->GetAdapterMonitor(m_uAdapter);
+    if(hMon == NULL)
+    {
+        hMon = ::MonitorFromWindow(NULL, MONITOR_DEFAULTTOPRIMARY);
+    }
 
     g_DisplayManager.GetCurrentDisplaySet(&pDisplaySet);
     UINT uDisplayIndex;


miloush:
The fix looks reasonable to me and should probably be backported to .NET Framework, did you try out your solution? The "Testing" part is referring to how the PR changes were tested.

petsuter:
No, I have no time to test it sorry. I just run into the crash all the time.
Maybe @Const-me tested it. See https://github.com/dotnet/wpf/issues/9042#issuecomment-2112139487

miloush:
or @vladimir-cheverdyuk-altium, especially since this one requires particular hardware setup

petsuter:
> particular hardware setup

A notebook and a monitor?

Wimmeh:
This also happens with a two monitor setup. Start the application on the primary monitor, disconnect the cable from the machine (the application gets moved to the second screen). Now you have the same exception. 

It alsof happens when you remote desktop to a two window machine, disconnect the session and reconnect.

Hope these extra cases help to fix the problem faster, it's really annoying.

harshit7962 (approved):

harshit7962:
Thank you @petsuter for the contribution!

PR #10404, opened 2025-02-04
Remove unnecessary S.R.CompilerServices.Unsafe packageref

ViktorHofer:
All the projects that reference the S.R.CompilerServices.Unsafe package target a version of .NETCoreApp (net10.0) which already have that assembly available inbox. There's no need for the package references.


ViktorHofer:
@JeremyKuhne PTAL

JeremyKuhne (approved):

PR #10417, opened 2025-02-06
Fix MilCodeGen and Csp build

ThomasGoulet73:
## Description
This PR aims to fix MilCodeGen which seems to be very outdated and it doesn't seem like it ever worked in this repo (It probably worked in the internal repo for Microsoft).

The goal of this PR is to fix MilCodeGen and Csp to be able to build and run MilCodeGen. I'll open other PRs to fix the templates to match the generated code and then other PRs which includes the modification to the generated code.

I cherry-picked the changes from dotnet/wpf#6135, reworked it to make it easier to use and updated the documentation.

## Customer Impact
None, this is non-shipping code.

## Regression
No.

## Testing
Local build + I generated the code locally using dotnet/wpf#6135 rebased over this PR.

## Risk
Low to none, this is non-shipping code and MilCodeGen already doesn't run so it can't get any worse.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/wpf/pull/10417)


src/Microsoft.DotNet.Wpf/src/WpfGfx/codegen/mcg/Runtime/FileCodeSink.cs
```diff
@@ -155,12 +120,6 @@ protected virtual void Dispose(bool disposing)
                         _streamWriter = null;
                     }
 
-                    if (_fileCreated)
-                    {
-                        // If we created the file, we now need to Sd Add it
-                        TfFile(_filePath, "add");
-                    }
-
                     LogCreation();
```
dipeshmsft, line 123: Probably, we can remove tf related code from LogCreation method as well.

src/Microsoft.DotNet.Wpf/src/WpfGfx/codegen/mcg/tools/InvokeCSP.cmd
```diff
@@ -11,7 +11,7 @@ if "%HostArch%"=="x86" set HostArch=i386
 set HostArch=i386
 
 set CspSrcPath=%_NtBinDir%\wpf\src\Graphics\tools\csp
-set CspExePath=%_NTTREE%\wpf
+if "%CspExePath%" == "" set CspExePath=%_NTTREE%\wpf
```
dipeshmsft, line 14: I think we should throw here, _NTTREE used to be the path of the location where the binaries are produced.

src/Microsoft.DotNet.Wpf/src/WpfGfx/codegen/mcg/xml/Elements.xsd
```diff
@@ -23,7 +23,7 @@
                   <!-- The namespace in which the element appears (e.g., "System.Windows") -->
                   <xs:attribute name="Namespace" type="xs:string" />
                   <!-- The path where the generator should output the code
-                       relative to %sdxroot%\wpf (e.g., "src\Core\CSharp\System\Windows\Generated")
```
dipeshmsft, line 26: We should replace references of `%sdxroot%\wpf` as well.

src/Microsoft.DotNet.Wpf/src/WpfGfx/codegen/mcg/generators/ProtocolFingerprint.cs
```diff
@@ -71,7 +71,7 @@ public override void Go()
             string generatedPath = 
                 Path.Combine(
                     _resourceModel.OutputDirectory,
-                    "src\\Graphics\\Include\\Generated"
+                    "src\\WpfGfx\\Include"
```
dipeshmsft, line 74: Is it intentional ?

dipeshmsft:
@ThomasGoulet73 I have reviewed the PR and left the comments. I am still to look into GenerateFiles.cmd, Resources.xml, Resources.xsd and related files, but I will be done by tomorrow.

src/Microsoft.DotNet.Wpf/src/WpfGfx/tools/csp/Project.cs
```diff
@@ -96,7 +96,7 @@ Parameters parameters
 
             // Add some implicit referenced assemblies. 
 
-            referencedAssemblies.Add(Path.Combine(parameters.ClrDir, "System.dll"));
+            referencedAssemblies.Add(string.IsNullOrEmpty(parameters.ClrDir) ? "System.dll" : Path.Combine(parameters.ClrDir, "System.dll"));
```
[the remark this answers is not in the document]
ThomasGoulet73: When it's not a full path, it's loaded from the .Net Framework runtime. Like in this example: https://learn.microsoft.com/en-us/dotnet/api/system.codedom.compiler.compilerparameters.referencedassemblies?view=windowsdesktop-9.0#examples

src/Microsoft.DotNet.Wpf/src/WpfGfx/codegen/mcg/mcg.proj
```diff
@@ -1,13 +1,15 @@
-<Project DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003" ToolsVersion="dogfood">
-  <Import Project="$(_NTDRIVE)$(_NTROOT)\wpf\src\Microsoft.Wpf.Product.Settings.targets" />
+<Project Sdk="Microsoft.Build.NoTargets">
 
-  <Target Name="Build">
-    <MSBuild Projects="$(WpfGraphicsRoot)\tools\csp\csp.csproj" />
-    <Exec Command="tools\GenerateFiles.cmd" />
-    <Exec Command="tools\GenerateElements.cmd" />
+  <Target Name="RunTool" BeforeTargets="AfterBuild">
```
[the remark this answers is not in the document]
ThomasGoulet73: I took this code from https://github.com/microsoft/MSBuildSdks/tree/main/src/NoTargets#example (The second code block). Microsoft.Build.NoTargets is an SDK that doesn't generate an assembly and in this case, I used it to run MilCodeGen when mcg.proj is built.

src/Microsoft.DotNet.Wpf/src/WpfGfx/codegen/mcg/generators/ProtocolFingerprint.cs
ThomasGoulet73: Yes, it generates the files wgx_sdk_version.cs and wgx_sdk_version.h which are currently in `src\Microsoft.DotNet.Wpf\src\WpfGfx\include`. I included it in this PR because I wanted it to generate the files in the right path even if the content of the files needed to be fixed.

src/Microsoft.DotNet.Wpf/src/WpfGfx/codegen/mcg/tools/InvokeCSP.cmd
ThomasGoulet73: I removed the if instead, it was redundant since later in the script we check that csp.exe exists in '%CspExePath%'.

src/Microsoft.DotNet.Wpf/src/WpfGfx/codegen/mcg/Runtime/FileCodeSink.cs
ThomasGoulet73: Yes, I'll remove the dead code.

src/Microsoft.DotNet.Wpf/src/WpfGfx/codegen/mcg/tools/GeneratedResources.txt
```diff
@@ -1,27 +0,0 @@
-src\Graphics\Include\Generated\wgx_resource_types.*
```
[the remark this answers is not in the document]
ThomasGoulet73: Good idea, I added it in the README. The files were removed because they were used for SD or TF checkout which isn't necessary since moving to Git.

src/Microsoft.DotNet.Wpf/src/WpfGfx/codegen/mcg/xml/Elements.xsd
ThomasGoulet73: Yes, SdxRoot is only found in comments but it doesn't hurt to fix them as well.

ThomasGoulet73:
Thanks @dipeshmsft, I believe I've applied all your change requests and answered your questions. I also pulled main to fix the conflicts.

Let me know if there's anything else.

src/Microsoft.DotNet.Wpf/src/WpfGfx/codegen/mcg/tools/GeneratedResources.txt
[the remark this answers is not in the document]
dipeshmsft: > The files were removed because they were used for SD or TF checkout which isn't necessary since moving to Git.

Yes, I got that after looking at the scripts.

src/Microsoft.DotNet.Wpf/src/WpfGfx/codegen/mcg/generators/ProtocolFingerprint.cs
dipeshmsft: The convention is that any generated file should be in a path with the parent directory as Generated, and as per that, the path mentioned here is correct. 

However, ProtocolFingerprint generator is not included in any .rsp file. So, actually it is not being used in any generation.
And as far as I can see, this file has not been generated since 2010.

I am trying to find if there was a modification in one of the .rsp files which removed this entry. ( These files have not been changed since 2010 as well )

dipeshmsft: What your thought on this ? Shall we create an issue to track this ?

dipeshmsft:
Thanks @ThomasGoulet73 for the quick turnaround on the comments and feedback. I have reviewed the PR again, and I have only one concern that I have left in the comments. Apart from that this PR is good to go.

src/Microsoft.DotNet.Wpf/src/WpfGfx/tools/csp/csp.csproj
```diff
@@ -1,18 +1,12 @@
...
-    <TargetFrameworkVersion>v3.5</TargetFrameworkVersion>
-    <CrossTargetWinFXRefsVersion>v3.5</CrossTargetWinFXRefsVersion>
-    <CrossTargetWinFXToolsVersion>v3.5</CrossTargetWinFXToolsVersion>
-  </PropertyGroup>
-
-  <Import Project="$(_NTDRIVE)$(_NTROOT)\wpf\src\Microsoft.wpf.Product.Settings.targets" />
+﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
+    <TargetFramework>net48</TargetFramework>
```
h3xds1nz, line 4: Is there any particular reason for specifically `net4.8`?

ThomasGoulet73: It requires some more work to target .Net Core and it's out of scope for this PR, but I'll probably do it eventually.

src/Microsoft.DotNet.Wpf/src/WpfGfx/codegen/mcg/generators/ProtocolFingerprint.cs
ThomasGoulet73: Ok I just checked and the reason I changed this file is because originally, when I reversed engineered the .rsp files, I added ProtocolFingerprint.cs to the .rsp file but since I updated my PR with the original .rsp files that you sent it no longer hits this line.

For now, I think I'll just revert the removal of "\\Generated" since, as you said, every generated files should be in a "Generated" folder. I'll keep the changes of "Graphics" to "WpfGfx" since it's a change I did globally because "Graphics" was renamed to "WpfGfx" so it doesn't make sense to keep "Graphics" in the path, even if it's not used right now.

I'll also open an issue to track the current work and future work (Which will include checking if we want to keep ProtocolFingerprint.cs).

ThomasGoulet73:
> Thanks @ThomasGoulet73 for the quick turnaround on the comments and feedback. I have reviewed the PR again, and I have only one concern that I have left in the comments. Apart from that this PR is good to go.

Great, I pushed a commit and answered your comment.

dipeshmsft (approved):

dipeshmsft:
Thanks @ThomasGoulet73 for support on this PR. I will wait for subsequent PRs on MilCodeGen.

ThomasGoulet73:
Thanks @dipeshmsft.

P.S.: I opened dotnet/wpf#10429 to track the rest of the work.

src/Microsoft.DotNet.Wpf/src/WpfGfx/tools/csp/csp.csproj
h3xds1nz: Thanks.

PR #10423, opened 2025-02-07
Remove deprecated .Net runtime

ThomasGoulet73:
## Description
Removes .Net Core 2.1.7 which is EOL since 2021. I cannot find a use for it anywhere in this repo and if it does break something we should fix it instead of relying on an EOL runtime.

It's downloaded on each build (Though locally cached) so it should improve the build speed a little bit, especially in CI.

## Customer Impact
None, build only.

## Regression
No.

## Testing
Local build + CI.

## Risk
Low, it would only break the WPF build and can always be reverted (Though we really shouldn't keep using .Net Core 2.1.7 since it's EOL).
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/wpf/pull/10423)

### DIFF
--- a/global.json
+++ b/global.json
@@ -3,7 +3,6 @@
     "dotnet": "10.0.100-alpha.1.25077.2",
     "runtimes": {
       "dotnet": [
-        "2.1.7",
         "$(VSRedistCommonNetCoreSharedFrameworkx64100PackageVersion)"
       ]
     },


h3xds1nz:
Is there any reason we keep this node around?
WF/runtime doesn't have it for example. Plus its 3 years old version if it meant anything.

```json
"vs": {
  "version": "17.2"
}
```

dipeshmsft:
> Is there any reason we keep this node around?
WF/runtime doesn't have it for example. Plus its 3 years old version if it meant anything.

Not sure about this.

h3xds1nz:
> It is used to detect the type of build engine to use, "Visual Studio" or "dotnet". The Arcade infrastructure checks if global.json contains the tool "vs" and if it does it uses the build engine "Visual Studio". For WPF, we need to use the "Visual Studio" build engine because there's .vcxproj projects and the "dotnet" build engine does not support .vcxproj projects (You can't do "dotnet build Project.vcxproj").
> 
> As for the version, we could probably raise the minimum version but I would prefer to do it in a separate PR since it could have an impact on the build and this PR as-is should have 0 impact on the build.

I see, thanks for the explanation, Thomas.

Yeah, definitely this doesn't have to be in this PR. But I guess it would be nice to test the waters. Locally I didn't see an issue.

dipeshmsft (approved):

dipeshmsft:
Thanks @ThomasGoulet73, if we face any issues in build, we will revert this.

ThomasGoulet73:
Thanks @dipeshmsft

Issue #10429, opened 2025-02-08
Fix MilCodeGen


ThomasGoulet73:
The goal of this issue is to track the work of fixing MilCodeGen and be able to run MilCodeGen when we need to modify generated code.

The necessary steps for this are:
1. Fix the build of MilCodeGen and Csp to be able to run it: dotnet/wpf#10417
2. Sync the code generators to the generated code since the generated code was modified manually instead of running MilCodegen: Not done yet
3. Update the generated code by running MilCodeGen: Not done yet

Followup work once the previous steps are done include:
- Update MilCodeGen to .Net Core. See https://github.com/dotnet/wpf/pull/10417#discussion_r1947354252
- Validate unused generators/dead code and check whether we should remove it or enable it. See https://github.com/dotnet/wpf/pull/10417#discussion_r1947365852
- Transition MilCodeGen to a Roslyn-based code generator and remove Csp. (Maybe in the future)

PR #10431, opened 2025-02-08
Run MilCodeGen for WpfGfx

ThomasGoulet73:

## Description
Generates the code in WpfGfx using the changes from dotnet/wpf#10430.

I recommend reviewing only the last commit and with whitespace changes hidden, use [this link](https://github.com/dotnet/wpf/pull/10431/commits/91ff20d245ee8f29633c29b5c8241a6eddaa7aba?diff=unified&w=1).

## Customer Impact

## Regression
No.

## Testing

## Risk
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/wpf/pull/10431)


src/Microsoft.DotNet.Wpf/src/WpfGfx/core/resources/marshal_generated.cpp
```diff
@@ -3189,7 +3189,7 @@ HRESULT CMilPixelShaderDuce::ProcessUpdate(
     RRETURN(hr);
 }
 
-HRESULT CMilPixelShaderDuce::RegisterNotifiers(CMilSlaveHandleTable *pHandleTable)
+HRESULT CMilPixelShaderDuce::RegisterNotifiers(__in_ecount(1) CMilSlaveHandleTable *pHandleTable)
```
ThomasGoulet73, line 3192: I needed to add `__in_ecount(1)` because some definitions of `RegisterNotifiers` had it and others didn't. I decided to add it to all instead of removing it to all. `__in_ecount(1)` is annotations used for build-time validations and shouldn't change behavior at runtime.

src/Microsoft.DotNet.Wpf/src/WpfGfx/include/Generated/wgx_misc.cs
```diff
@@ -204,15 +201,6 @@ internal struct MilColorF
     internal float g;
     internal float b;
     internal float a;
-
-    public override int GetHashCode()
```
ThomasGoulet73, line 208: Technically removing these methods is a perf regression but it doesn't functionally changes the behavior of the struct. I included the changes because wgx_misc.cs is unused right now.

In a followup in dotnet/wpf#10429, I'll come back to this and check if we should remove wgx_misc.cs.

ThomasGoulet73:
This is blocked by dotnet/wpf#10430.

src/Microsoft.DotNet.Wpf/src/WpfGfx/codegen/mcg/ResourceModel/Generated/Elements.cs
```diff
@@ -1,27 +1,22 @@
...
-// The .NET Foundation licenses this file to you under the MIT license.
-// See the LICENSE file in the project root for more information.
-
-
-//------------------------------------------------------------------------------
+﻿//------------------------------------------------------------------------------
 // <auto-generated>
 //     This code was generated by a tool.
-//     Runtime Version:4.0.30319.18408
+//     Runtime Version:4.0.30319.42000
```
dipeshmsft, line 4: Where are these version numbers coming from ?

dipeshmsft: Although these are comments so it doesn't matter much.

h3xds1nz: That's the netFX CLR version.

src/Microsoft.DotNet.Wpf/src/WpfGfx/include/Generated/wgx_misc.cs
dipeshmsft: > I included the changes because wgx_misc.cs is unused right now.

@ThomasGoulet73  what do you mean by this ? On a quick look I saw that there are many places where MilColorF is being used. Moreover there is a copy of this struct in here https://github.com/dotnet/wpf/blob/5f713e22cf3f6800277ec0acb5a4a03cbc8b096e/src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Media/Color.cs#L1129 which I am guessing is a copy of this struct.

It doesn't seem right to have two different definitions for this.

h3xds1nz: Yeah but not this file, the whole `wgx_misc.cs` is not included in the build of the managed libraries.

The generated files are currently not used here, we source them from Common/Graphics;

dipeshmsft: Thanks for the information, I didn't see that.

dipeshmsft: Resolving this conversation now.

dipeshmsft: > In a followup in https://github.com/dotnet/wpf/issues/10429, I'll come back to this and check if we should remove wgx_misc.cs.

Even if we do this, we would need a way to generate the files in Common/Graphics right ?

ThomasGoulet73: > > In a followup in #10429, I'll come back to this and check if we should remove wgx_misc.cs.
> 
> Even if we do this, we would need a way to generate the files in Common/Graphics right ?

It's unclear for now what should be done for the files in Common/Graphic, they're not in the Common/Graphics/Generated folder so it seems like they're not supposed to be generated and I didn't find any reference to these files anywhere in this repo, except for `wgx_misc.cs` but `wgx_misc.cs` is generated in path different than the one currently used. Maybe they were generated before and now they're not, maybe you could find more information about this in your internal repo.

dipeshmsft: I will take a look in the internal repo regarding this. 
Thanks @ThomasGoulet73, @h3xds1nz for your help.

dipeshmsft (approved):

PR #10433, opened 2025-02-08
Run MilCodeGen for WindowsBase

ThomasGoulet73:

## Description
Generates the code in WindowsBase using the changes from dotnet/wpf#10430.

I recommend reviewing only the last commit and with whitespace changes hidden, use [this link](https://github.com/dotnet/wpf/pull/10433/commits/be8df2a030b4570510fbe3113acec4a3967c4204?diff=unified&w=1).

## Customer Impact

## Regression
No.

## Testing

## Risk
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/wpf/pull/10433)

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/WindowsBase/System/Windows/Media/Converters/Generated/MatrixValueSerializer.cs
+++ b/src/Microsoft.DotNet.Wpf/src/WindowsBase/System/Windows/Media/Converters/Generated/MatrixValueSerializer.cs
@@ -1,4 +1,4 @@
-﻿// Licensed to the .NET Foundation under one or more agreements.
+// Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
@@ -9,7 +9,11 @@
 // Please see MilCodeGen.html for more information.
 //
 
+using MS.Internal;
+using System.ComponentModel;
+using System.Globalization;
 using System.Windows.Markup;
+using System.Windows.Media.Converters;
 
 namespace System.Windows.Media.Converters
 {
@@ -39,7 +43,7 @@ namespace System.Windows.Media.Converters
             }
 
             return true;
-}
+        }
 
         /// <summary>
         /// Converts a string into a Matrix.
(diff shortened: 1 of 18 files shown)


ThomasGoulet73:
This PR is blocked by dotnet/wpf#10430.

src/Microsoft.DotNet.Wpf/src/WindowsBase/System/Windows/Media/Converters/Generated/MatrixValueSerializer.cs
```diff
@@ -9,7 +9,11 @@
 // Please see MilCodeGen.html for more information.
 //
 
+using MS.Internal;
```
dipeshmsft, line 12: As you mentioned earlier, these extra imports are needed because one of the files being generated by the generator needs these imports ?

Also, I think there can be multiple generators that are involved while generating a file ?

ThomasGoulet73: One generator can generate multiple files and each files can require different usings. The generator needs to include the usings by every files, otherwise it would need to dynamically add usings according to what code gets generated. Technically it could be done, I haven't done it because I didn't see the value in doing it for my inital PRs and I'm not sure making the generators more complicated just to trim the list of usings is a positive.

src/Microsoft.DotNet.Wpf/src/WindowsBase/System/Windows/Media/Generated/Matrix.cs
```diff
@@ -298,6 +301,7 @@ public string ToString(IFormatProvider provider)
         /// </returns>
         string IFormattable.ToString(string format, IFormatProvider provider)
         {
+
```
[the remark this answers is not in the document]
dipeshmsft: Oh, I see thanks for explaining this.

dipeshmsft (approved):

dipeshmsft:
Thanks @ThomasGoulet73, I will go ahead to the next PR.

PR #10434, opened 2025-02-08
Run MilCodeGen for PresentationCore

ThomasGoulet73:

## Description
Generates the code in PresentationCore using the changes from dotnet/wpf#10430.

I recommend reviewing only the last commit and with whitespace changes hidden, use [this link](https://github.com/dotnet/wpf/pull/10434/commits/8f44ef015a0d68c8fc4d351972fc0ad3cade0125?diff=unified&w=1).

## Customer Impact

## Regression
No.

## Testing

## Risk
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/wpf/pull/10434)

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Generated/ContentElement.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Generated/ContentElement.cs
@@ -1,4 +1,4 @@
-﻿// Licensed to the .NET Foundation under one or more agreements.
+// Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
@@ -11,7 +11,6 @@
 
 using MS.Internal;
 using MS.Internal.KnownBoxes;
-using MS.Internal.PresentationCore;
 using MS.Utility;
 using System.ComponentModel;
 using System.Windows.Input;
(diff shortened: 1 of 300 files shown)


ThomasGoulet73:
This PR is blocked by dotnet/wpf#10430.

h3xds1nz:
That is to reduce the diff. `TextDecorationCollection` is like one of the few that got re-formatted because most of them are actually as `[ index ]`.

dipeshmsft:
> TextDecorationCollection is like one of the few that got re-formatted because most of them are actually as [ index ].

I thought this was reverted back to this state because of the limitation of the generator. This code gets generated from this place : 
https://github.com/dotnet/wpf/blob/5f713e22cf3f6800277ec0acb5a4a03cbc8b096e/src/Microsoft.DotNet.Wpf/src/WpfGfx/codegen/mcg/helpers/CollectionHelper.cs#L192
and if we remove the spaces, the generator crashes throwing a syntax error.

h3xds1nz:
Eitherway is possible tbh, I know where its coming from but I didn't try to modify it.

ThomasGoulet73:
I did hit the bracket limitation but the main reason I reverted `TextDecorationCollection` instead of fixing the code generator was because, as @h3xds1nz said, it was the only 1 of many files formatted manually so either I reverted this file or changed a bunch of other files and I chose the option with the smaller diff.

> and if we remove the spaces, the generator crashes throwing a syntax error.

I hit this problem also but there's way to workaround this limitation, you just need to concat the string with the `[` and `]` in a variable instead of inline and it will fix the parser error.

I do plan on working on formatting the generated code in the future.

PR #10477, opened 2025-02-19
Try fix DataGrid HTML clipboard data faulty

### LINKED ISSUE #10476: WPF DataGrid HTML clipboard data faulty
### Description

The HTML fragment DataGrid puts into the clipboard is not accepted by other applications. 

It appears that the fragment's offsets (StartHTML/EndHTML/StartFragment/EndFragment) do not match the content.

### Reproduction Steps

Select all data in a DataGrid, copy and then paste to Excel.

### Expected behavior

Excel (or other application) correctly accepts HTML data from DatGrid in clipboard.

### Actual behavior

Application cannot interpret and paste HTML data copied from DataGrid but display error.

### Regression?

Error was introduced with #8519.

### Known Workarounds

1. Derive DataGrid subclass and override OnExecutedCopy:
```
        protected override void OnExecutedCopy(ExecutedRoutedEventArgs args)
        {
            base.OnExecutedCopy(args);

            String htmlData = Clipboard.GetData(DataFormats.Html) as String;
            htmlData = htmlData.Replace("\r\n\r\n", "\r\n");
            Clipboard.SetData(DataFormats.Html, htmlData);
        }

2. Manually select csv or other format when pasting data copied from DataGrid in other applications.

### Impact

High impact at my company's backoffice which uses copy/paste to Excel intensely. On a more general scale, probably low.

### Configuration

.Net 9 Win11

### Other information

The problem with the wrong offsets is a result of double CR/LFs between the header items in Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Controls/DataGridClipboardHelper.cs (lines 86-102) - remove those and everything works as before.

### PR DESCRIPTION
**Pull request by lindexi**
Fixes #10476

Main PR 

## Description

The https://github.com/dotnet/wpf/pull/8519  introduces behavior changed. And this pr restore the origin behavior.

## Customer Impact

Fixes #10476

## Regression

See https://github.com/dotnet/wpf/pull/8519

## Testing

Only CI.

## Risk

Normal. This PR will changed the DataGrid clipboard behavior.

Cc @halgab

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Controls/DataGridClipboardHelper.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Controls/DataGridClipboardHelper.cs
@@ -85,30 +85,20 @@ namespace System.Windows.Controls
             string prefix = string.Create(CultureInfo.InvariantCulture,
                 $"""
                  Version:1.0
-
                  StartHTML:00000097
-
                  EndHTML:{bytecountEndOfHtml:00000000}
-
                  StartFragment:00000133
-
                  EndFragment:{bytecountEndOfFragment:00000000}
-
                  <HTML>
-
                  <BODY>
-
                  <!--StartFragment-->
                  """);
             content.Insert(0, prefix);
             content.Append(
                 """
 
-
                 <!--EndFragment-->
-
                 </BODY>
-
                 </HTML>
                 """);
         }

### CODE REVIEW

**Comment by h3xds1nz** [comment]
LGTM, this should be backported imho.

A unit test wouldn't hurt.

**Comment by miloush** [comment]
I don't really like the code this has been changed to, it's too fragile and might not survive git handling of new lines.

**Comment by kniessen** [comment]
Thx so much - for the ultra-quick reaction to this as well as just in general for maintaining the framework.

**Comment by lindexi** [comment]
@miloush What do you suggest? Thank you.

**Comment by halgab** [comment]
LGTM, even though I did like the added readability allowed by raw interpolated strings

**Comment by lindexi** [comment]
I'd like it too.

**Comment by h3xds1nz** [comment]
@lindexi As I've said previously, why don't you just use raw-interpolated string then as previously done and write 2-3 unit tests around to verify the correct line-breaks are emitted? That way it is safe and readable.

**Comment by lindexi** [comment]
@h3xds1nz Thank you. I just fall back to the old code. And I do not plan to add the unit test.

And we can not control the CRLF in the raw-interpolated string, so that I do not want to use the raw-interpolated string yet.

---

And, why not I use the syntax like:

```csharp
$"abcd \r\n" + 
$"xxx \r\n" + 
$"xxx \r\n";
```

Because we should pass the `bytecountEndOfHtml` and `bytecountEndOfFragment` to `string.Create(CultureInfo.InvariantCulture, xx)`

halgab (approved):

PR #10495, opened 2025-02-24
Refactor existing Clipboard code

JeremyKuhne:
This brings the existing Clipboard code up to date with the latest standards. It includes my other pending PR commits.

Cleaning up the existing code makes it easier to put together a PR that uses the shared code from System.Private.Windows.Core (this one is currently close to completion).

This shouldn't be squashed so it will retain the changes I made after breaking out the nested classes.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/wpf/pull/10495)


src/Microsoft.DotNet.Wpf/src/PresentationCore/MS/internal/SystemDrawingHelper.cs
```diff
@@ -1,108 +1,72 @@
...
     {
-        // return true if the data is a bitmap
-        internal static bool IsBitmap(object data)
-        {
-            SystemDrawingExtensionMethods extensions = AssemblyHelper.ExtensionsForSystemDrawing();
-            return (extensions != null) ? extensions.IsBitmap(data) : false;
-        }
+        /// <inheritdoc cref="SystemDrawingExtensionMethods.IsBitmap(object?)"/>
+        internal static bool IsBitmap(object? data) =>
+            AssemblyHelper.ExtensionsForSystemDrawing()?.IsBitmap(data) ?? false;
```
miloush, line 17: What's the benefit of these changes? Only makes it more difficult to debug to not have the `extensions` as explicit step and it gets optimized away anyway.

src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/DataObject.DataStore.cs
```diff
@@ -0,0 +1,170 @@
...
+using MS.Internal;
+
+namespace System.Windows;
+
+public sealed partial class DataObject
+{
+    private partial class DataStore : IDataObject
+    {
+        // Data hash table.
+        private readonly Dictionary<string, DataStoreEntry> _data = [];
```
miloush, line 18: Since we are looking into making the index public parameter, why would you remove it from DataStoreEntry and make the GetData ignore it? We will have to add it back to support it.

src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/DataObject.OleConverter.cs
```diff
@@ -0,0 +1,724 @@
...
+        /// <remarks>
+        ///  <para>
+        ///   We need a separate method to avoid loading the System.Drawing assembly when unnecessary.
+        ///  </para>
+        /// </remarks>
+        [MethodImpl(MethodImplOptions.NoInlining)]
+        private static BitmapSource GetBitmapSourceFromHbitmap(nint hbitmap) =>
+            Imaging.CreateBitmapSourceFromHBitmap(
+                hbitmap,
+                0,
```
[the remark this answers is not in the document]
h3xds1nz: I agree, `nint.Zero` would work better in this case for clarity as I understand that everyone moved (including me tbh) from `[U]IntPtr` to `n[u]int`.

src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/dataobject.cs
```diff
@@ -19,3741 +18,1783 @@
...
             }
-
-            return win32Pointer;
         }
 
-        /// <summary>
-        /// Call Win32 UnsafeNativeMethods.CreateStreamOnHGlobal() with Win32 error checking.
-        /// </summary>
-        private static int Win32CreateStreamOnHGlobal(IntPtr hGlobal, bool fDeleteOnRelease, ref IStream istream)
+        return [.. distinct];
```
h3xds1nz, line 1011: Is this change necessary? I mean, I will never understand in what world is this more readable than `.ToArray()` tbh.

src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/dataobject.cs
```diff
@@ -19,3741 +18,1783 @@
...
+        {
+            return (NativeMethods.E_INVALIDARG);
+        }
 
-            private Object GetData(string format, bool autoConvert, DVASPECT aspect, int index)
-            {
-                Object baseVar;
-                Object original;
-                DataStoreEntry dataStoreEntry;
+        if (Marshal.SystemDefaultCharSize == 1)
```
h3xds1nz, line 1484: This check should be removed, `SystemDefaultCharSize` is always `2` anyways.

src/Microsoft.DotNet.Wpf/src/Extensions/PresentationFramework-SystemDrawing/SystemDrawingExtension.cs
```diff
@@ -84,54 +54,45 @@ internal override IntPtr GetHBitmap(object data, out int width, out int height)
...
-            // for reference the tag is PropertyTagExifUserComment [0x9286] or 37510 (int)
-            PropertyItem piComment = img.GetPropertyItem(37510);
-            return new MemoryStream(piComment.Value);
+            // Read the GIF header
+            Bitmap bitmap = new(stream);
+
+            // Read the comment as that is where the ISF is stored.
+            // For reference the tag is PropertyTagExifUserComment [0x9286] or 37510 (int)
+            PropertyItem? piComment = bitmap.GetPropertyItem(37510);
+            return new MemoryStream(piComment!.Value!);
```
h3xds1nz, line 83: I think I'd expect a comment explaining the suppressions given that `GetPropertyItem` can actually return `null`.

h3xds1nz (commented):
I'm not sure to what extent is this supposed to be refactoring but most of the Marshal.* things could just be changed for Spans so that we can actually live in 2025.

src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/DataObject.OleConverter.cs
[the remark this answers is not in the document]
JeremyKuhne: Because it is an intrinsic. Ultimately it is better to not use `nint`/`nuint` (or worse `IntPtr`) for any of this. We should be using `Thing*` and handle wrappers (`HBITMAP`).

In any case, this entire class is being deleted in the next change. The shared code handles this and does it with real types.

src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/DataObject.DataStore.cs
JeremyKuhne: Because it has to be redone in the shared code anyway. Right now, it does nothing and the shared code has no such concept- this facilitates the move to the shared code.

src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/dataobject.cs
JeremyKuhne: It clears the warnings/messages for the file. I agree that this particular usage is debatable, but it is opinionated and the rule in general does make a big difference in other cases as it allows picking up params span improvements as they show up.

src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/dataobject.cs
```diff
@@ -19,3741 +18,1783 @@
...
+
+            // Select the new bitmap into a compatible DC and render the blt the original bitmap.
+            nint destinationObject = Win32SelectObject(new HandleRef(this, destinationDC), hBitmapNew);
+
+            try
             {
-                throw new System.ComponentModel.Win32Exception(win32Error);
+                // Bit-block transfer of the data onto the rectangle of pixels
+                // from the specified source device context into a destination device context.
+                Win32BitBlt(new HandleRef(this, destinationDC), width, height, new HandleRef(null, sourceDC), /* SRCCOPY */0x00CC0020);
```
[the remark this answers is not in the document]
JeremyKuhne: It is vanishing in the follow up PR.

src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/dataobject.cs
```diff
@@ -19,3741 +18,1783 @@
...
+            // Copy UTF8 encoding bytes to the memory.
+            UnsafeNativeMethods.CopyMemory(pointerUtf8, utf8Bytes, utf8ByteCount);
 
-                //------------------------------------------------------
-                //
-                //  Public Properties
-                //
-                //------------------------------------------------------
+            // Copy the null into the last of memory.
+            Marshal.Copy(new byte[] { 0 }, 0, (nint)((long)pointerUtf8 + utf8ByteCount), 1);
```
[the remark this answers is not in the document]
JeremyKuhne: This code is being removed.

src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/dataobject.cs
```diff
@@ -19,3741 +18,1783 @@
...
-            //
-            //  Private Fields
-            //
-            //------------------------------------------------------
+            try
+            {
+                UnsafeNativeMethods.CopyMemory(ptr, strBytes, pinvokeSize);
 
-            #region Private Fields
+                Marshal.Copy(new byte[] { 0 }, 0, (nint)((long)ptr + pinvokeSize), 1);
```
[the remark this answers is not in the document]
JeremyKuhne: This code is being removed.

src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/dataobject.cs
```diff
@@ -19,3741 +18,1783 @@
...
-                    if (dse == null && naturalDse != null)
-                    {
-                        dse = naturalDse;
-                    }
+            // Write out the strings.
+            for (int i = 0; i < files.Length; i++)
+            {
+                // Write out the each of file as the unicode and increase the pointer.
+                UnsafeNativeMethods.CopyMemoryW(currentPtr, files[i], files[i].Length * 2);
+                currentPtr = (nint)((long)currentPtr + (files[i].Length * 2));
```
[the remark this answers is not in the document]
JeremyKuhne: They are vanishing in the next PR.

src/Microsoft.DotNet.Wpf/src/Extensions/PresentationFramework-SystemDrawing/SystemDrawingExtension.cs
JeremyKuhne: This can totally throw. The answers here aren't great as modifying the throw is more likely to create grief than to be useful. We can put a comment in that says letting it throw as it would have for compat.

miloush: This is not public, we can make it return null. The caller already catches `NullReferenceException` which it will cause.

JeremyKuhne:
@h3xds1nz, @miloush a number of the comments you made are moot in https://github.com/dotnet/wpf/pull/10544. I did this part separately to try and make the actual change to use the shared code easier to accomplish and also not include updating style.

src/Microsoft.DotNet.Wpf/src/PresentationCore/MS/internal/SystemDrawingHelper.cs
JeremyKuhne: More class context on the screen at the same time. The impact to debugging doesn't seem to outweigh that benefit for something this simple.

JeremyKuhne:
> I'm not sure to what extent is this supposed to be refactoring but most of the Marshal.* things could just be changed for Spans so that we can actually live in 2025.

Yep, and I would change them if the code wasn't dropping off in #10544. :)

src/Microsoft.DotNet.Wpf/src/Extensions/PresentationFramework-SystemDrawing/SystemDrawingExtension.cs
JeremyKuhne: I'll catch it in the other PR @miloush. I'm closing this one out in favor of the other after chatting with @harshit7962

JeremyKuhne:
These commits are merged in #10544, closing this PR.

PR #10506, opened 2025-02-25
Removes MappingMode from LinearGradientBrush in HasRealizationContextChanged

### LINKED ISSUE #10502: Unexpected behavior when using LinearGradientBrush + RelativeTransform
### Description

While working on Fluent we ran into an issue when we were trying to use LinearGradientBrush alongwith RelativeTransform and MappingMode set to Absolute. 

In the issue, when we use a LinearGradientBrush defined with above properties and apply it on controls with different heights, the position of linear gradient remains the same from top, rather than from bottom.

Moreover, if a control with a different height gets processed / styled first, the position of linear gradient is now different. 

The first control using the brush gets the correct styling, however subsequent controls where this is used don't see the correct style. This seems like a caching issue in the sense that once the brush is evaluated for the first control, the  values of gradient stops are cached by the brush and they are not being reevaluated for other controls where the brush is used.

### Reproduction Steps

**Here are the steps to repro the issue :**

1. Define a LinearGradientBrush like following in your resources : 
```
<LinearGradientBrush x:Key="FaultyBorderBrush" MappingMode="Absolute"  StartPoint="0,0" EndPoint="0,3">
    <LinearGradientBrush.RelativeTransform>
        <ScaleTransform ScaleY="-1" CenterY="0.5" />
    </LinearGradientBrush.RelativeTransform>
    <LinearGradientBrush.GradientStops>
        <GradientStop Offset="0.33" Color="Red" />
        <GradientStop Offset="1.0" Color="Orange" />
    </LinearGradientBrush.GradientStops>
</LinearGradientBrush>
```

2. Now create multiple elements in your window which use the Brush ( make sure they have varying sizes ) like this : 
```
    <StackPanel Orientation="Horizontal">
        <Border Margin="20" BorderBrush="{DynamicResource FaultyBorderBrush}" 
                BorderThickness="10" Height="50" Width="50"/>
        <Border Margin="20" BorderBrush="{DynamicResource FaultyBorderBrush}" 
                BorderThickness="10" Height="100" Width="50"/>
        <Border Margin="20" BorderBrush="{DynamicResource FaultyBorderBrush}" 
                BorderThickness="10" Height="30" Width="50"/>
    </StackPanel>
```
3. Run the application. You will get something like this on the screen : 

Image

4. Now just reorder the elements in your application, Here is what I did ( just moved border with height 30 to first position )
```
    <StackPanel Orientation="Horizontal">
        <Border Margin="20" BorderBrush="{DynamicResource FaultyBorderBrush}" 
                BorderThickness="10" Height="30" Width="50"/>
        <Border Margin="20" BorderBrush="{DynamicResource FaultyBorderBrush}" 
                BorderThickness="10" Height="50" Width="50"/>
        <Border Margin="20" BorderBrush="{DynamicResource FaultyBorderBrush}" 
                BorderThickness="10" Height="100" Width="50"/>
    </StackPanel>
```

5. Run the application and observe : 

Image

### Expected behavior

In respect to the above example, every control no matter the height should have a red color at the bottom of the border.

### Actual behavior

The brush is getting cached according to the first brush and gradient stops are not being calculated again.

### Regression?

Don't know. Looks like a long standing issue.

### Known Workarounds

Can avoid using Absolute MappingMode for the brush.

### Impact

_No response_

### Configuration

- Which version of .NET is the code running on?  :   .NET 9 and 10
- What OS and version, and what distro if applicable?  :   Windows 11
- What is the architecture (x64, x86, ARM, ARM64)?  : Tried with x64
- Do you know whether it is specific to that configuration? : Don't think so.

### Other information

I discussed this issue with @miloush around half an year ago, but raising the issue got delayed from my end.

dipeshmsft:
Fixes #10502 

## Description
As described in the issue description, LinearGradientBrushes ( LGB ) caches it bounds when we use Absolute MappingMode. However, when we apply transformations to the brush, the brush does not work work as expected. This PR removes the caching of the brush, thus resolving the issue.

## Customer Impact
Developers using the LGB along with MappingMode=Absolute and RelativeTransform will get the correct behavior for the brush.

## Regression
No

## Testing
Not yet.

## Risk
- It will have a performance impact on the application using LGB with Absolute mapping mode.
- Applications using LGB will get a different behavior now.

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/wpf/pull/10506)

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/WpfGfx/core/resources/lineargradient.h
+++ b/src/Microsoft.DotNet.Wpf/src/WpfGfx/core/resources/lineargradient.h
@@ -80,20 +80,20 @@ public:
         __in_ecount(1) const BrushContext *pBrushContext
         ) const
     {
-                // If a mapping mode is relative to the brush
-                // *or* or when a relative transform is used 
-                // *and* bounds have changed, then the realization context has changed            
+                    // If a mapping mode is relative to the brush
+                    // *or* or when a relative transform is used 
+                    // *and* bounds have changed, then the realization context has changed            
         return  (   ((m_data.m_MappingMode == MilBrushMappingMode::RelativeToBoundingBox) 
                     || (m_data.m_pRelativeTransform != NULL)) &&
 
-                // Return true if the brush sizing bounds have changed
-                //
-                // We use exact equality here because fuzzy checks are expensive, coming up 
-                // with a fuzzy threshold that defines the point at which visible changes
-                // occur isn't straightforward (i.e., the brush sizing bounds aren't
-                // in device space), and exact equality handles the case we need to optimize
-                // for where a brush fills the exact same geometry more than once.                
-                !(IsExactlyEqualRectD(pBrushContext->rcWorldBrushSizingBounds, m_cachedBrushSizingBounds) ));
+                    // Return true if the brush sizing bounds have changed
+                    //
+                    // We use exact equality here because fuzzy checks are expensive, coming up 
+                    // with a fuzzy threshold that defines the point at which visible changes
+                    // occur isn't straightforward (i.e., the brush sizing bounds aren't
+                    // in device space), and exact equality handles the case we need to optimize
+                    // for where a brush fills the exact same geometry more than once.                
+                    !(IsExactlyEqualRectD(pBrushContext->rcWorldBrushSizingBounds, m_cachedBrushSizingBounds) ));
 
     }
 
--- a/src/Microsoft.DotNet.Wpf/src/WpfGfx/core/resources/lineargradient.h
+++ b/src/Microsoft.DotNet.Wpf/src/WpfGfx/core/resources/lineargradient.h
@@ -80,20 +80,20 @@ public:
         __in_ecount(1) const BrushContext *pBrushContext
         ) const
     {
-        // If a mapping mode is relative to the brush
-        // *or* or when a relative transform is used 
-        // *and* bounds have changed, then the realization context has changed            
+                // If a mapping mode is relative to the brush
+                // *or* or when a relative transform is used 
+                // *and* bounds have changed, then the realization context has changed            
         return  (   ((m_data.m_MappingMode == MilBrushMappingMode::RelativeToBoundingBox) 
                     || (m_data.m_pRelativeTransform != NULL)) &&
 
-        // Return true if the brush sizing bounds have changed
-        //
-        // We use exact equality here because fuzzy checks are expensive, coming up 
-        // with a fuzzy threshold that defines the point at which visible changes
-        // occur isn't straightforward (i.e., the brush sizing bounds aren't
-        // in device space), and exact equality handles the case we need to optimize
-        // for where a brush fills the exact same geometry more than once.                
-        !(IsExactlyEqualRectD(pBrushContext->rcWorldBrushSizingBounds, m_cachedBrushSizingBounds) ));
+                // Return true if the brush sizing bounds have changed
+                //
+                // We use exact equality here because fuzzy checks are expensive, coming up 
+                // with a fuzzy threshold that defines the point at which visible changes
+                // occur isn't straightforward (i.e., the brush sizing bounds aren't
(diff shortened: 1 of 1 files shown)


src/Microsoft.DotNet.Wpf/src/WpfGfx/core/resources/lineargradient.h
```diff
@@ -80,18 +80,19 @@ class CMilLinearGradientBrushDuce : public CMilGradientBrushDuce
...
-                    //
-                    // We use exact equality here because fuzzy checks are expensive, coming up 
-                    // with a fuzzy threshold that defines the point at which visible changes
-                    // occur isn't straightforward (i.e., the brush sizing bounds aren't
-                    // in device space), and exact equality handles the case we need to optimize
-                    // for where a brush fills the exact same geometry more than once.                
-                    !(IsExactlyEqualRectD(pBrushContext->rcWorldBrushSizingBounds, m_cachedBrushSizingBounds) ));
+        // If a mapping mode is relative to the brush sizing bounds *and* those
+        // bounds have changed, then the realization context has changed.            
+        // If a mapping mode is absolute and we perform transformation operations
```
pchaurasia14, line 85: Comments need to be updated to reflect the current implementation.

dipeshmsft:
@h3xds1nz I came across this while fixing the borders in the new Fluent styles. WinUI uses the brush in the above syntax, and gets the expected behavior. I am not sure of there mechanism though.

I will try to look for other ways for fixing the usage.

miloush:
I don't remember much honestly and will have to look once I am back travelling, but I don't think `Absolute` should depend on the size of the element, which is what I think @h3xds1nz is saying but what is happening. Certainly caching shouldn't affect the output, it should only save calculations.

src/Microsoft.DotNet.Wpf/src/WpfGfx/core/resources/lineargradient.h
```diff
@@ -80,18 +80,19 @@ class CMilLinearGradientBrushDuce : public CMilGradientBrushDuce
         __in_ecount(1) const BrushContext *pBrushContext
         ) const
     {
-                    // If a mapping mode is relative to the brush sizing bounds *and* those
-                    // bounds have changed, then the realization context has changed            
-        return  (   (m_data.m_MappingMode == MilBrushMappingMode::RelativeToBoundingBox) &&
```
ThomasGoulet73, line 85: How about we do this instead here:
```
(m_data.m_MappingMode == MilBrushMappingMode::RelativeToBoundingBox || m_data.m_pRelativeTransform != NULL) &&
```

A similar check is done in the same file:
https://github.com/dotnet/wpf/blob/f944708ac4d98b7665664f8dd698e9e920043b66/src/Microsoft.DotNet.Wpf/src/WpfGfx/core/resources/lineargradient.h#L55-L56

Doing so would also reduce the change of breaking something since it would only affect brushes with a relative transform and it won't affect performance of absolute brushes without a relative transform.

dipeshmsft: Oh thanks @ThomasGoulet73, I will give this a try and it will be less performance impacting change

h3xds1nz: That looks reasonable.

src/Microsoft.DotNet.Wpf/src/WpfGfx/core/resources/lineargradient.h
dipeshmsft: Reverted this change.

src/Microsoft.DotNet.Wpf/src/WpfGfx/core/resources/lineargradient.h
dipeshmsft: Addressed.

singhashish-wpf (approved):

PR #10507, opened 2025-02-26
Add a few Clipboard tests

JeremyKuhne:
Adds a few Clipboard tests. Based somewhat on WinForms unit tests.

WPF doesn't consistently swallow bad data on the clipboard. This is called out with comments and skips.

Change PresentationCore.Tests so it can use `[WpfFact]` and shared test code. Add reference to drawing extensions so Clipboard image tests can be run.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/wpf/pull/10507)


src/Microsoft.DotNet.Wpf/tests/UnitTests/PresentationCore.Tests/System/Windows/ClipboardTests.cs
```diff
@@ -0,0 +1,246 @@
...
+        object? result1 = Clipboard.GetData("testData1");
+        result1.Should().Be(testData1);
+        object? result2 = Clipboard.GetData("testData2");
+        result2.Should().Be(testData2);
+    }
+
+    [WpfFact]
+    public void SetData_Text_Format_AllUpper()
+    {
+        // The fact that casing on input matters is likely incorrect, but behavior has been this way.
```
miloush, line 219: how do the tests below show the case sensitivity?

JeremyKuhne: "Text" is the "right" name for text, you'll see that in the formats list below. The not-quite right case sensitivity is with our DataStore implementations.

ThomasGoulet73:
@JeremyKuhne: Just a ping to let you know that the test failures in this PR were fixed in main, pulling main should fix the failures.

PR #10517, opened 2025-02-27
Adding simsun-extg font to support more character

Kuldeep-MS:
## Issue Description

Some Chinese characters are being displayed as boxes in WPF due to unavailability of the font. 

## Background Infromation
When text is rendered, each character's codepoint is searched in either the default font or the specified font. If the codepoint is not found, it falls back to composite fonts to locate the codepoint. These composite fonts are collections of different fonts that cover various codepoint ranges and are associated with different languages, enabling the rendering of characters.

## Change Description

Adding Simsun-ExtG font and modifying the character range for the Simsun-ExtB font to GlobalUserInterface composite font. This will enable us to support more character from GB18030.

## Customer Impact

all the characters will be displayed correctly, without this many of the Chinese characters will be displayed as boxes.

## Regression

NA

## Testing

In progress

## Risk

Low

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/wpf/pull/10517)


src/Microsoft.DotNet.Wpf/src/PresentationCore/Fonts/GlobalUserInterface.CompositeFont
```diff
@@ -1716,8 +1716,11 @@
             -->
             <!-- CHS -->
             <FontFamilyMap
-                Unicode     = "20000-2A6DF"
-                Language    = "zh-Hans"
```
dipeshmsft, line 1720: Do we need to remove the above unicodes for it to work ? Can't we add to it ?

Kuldeep-MS: Its not removed. We have added a new font Simsun-ExtG and modified the unicode range for Simsun-ExtB.

lindexi: `Simsun-ExtG` added in Win11, and should we consider system versions compatibility issues here?

Kuldeep-MS: > `Simsun-ExtG` added in Win11, and should we consider system versions compatibility issues here?
> 
> See https://blogs.windows.com/windows-insider/2024/10/09/announcing-windows-11-insider-preview-build-27723-canary-channel/
> 
> > There is a new simplified Chinese font (Simsun-ExtG) with 9,753 ideographs supporting Unicode Extensions [G](https://www.unicode.org/charts/PDF/U30000.pdf), [H](https://www.unicode.org/charts/PDF/U31350.pdf) and [I](https://www.unicode.org/charts/PDF/U2EBF0.pdf). Included is the highly requested character for Biangbiang noodles. This character U+30EDD is said to be the most complicated Chinese character, as you can see in the image below. We are aware that some applications may not be able to display these new extension characters yet.

to my knowledge is being ported to windows10 as well.

miloush: > to my knowledge is being ported to windows10 as well.

That does not align with my information, but it does not matter. Users can uninstall fonts too. The DWrite has a font fallback mechanism to deal with missing fonts. Even if it didn't, you would see notdef glyphs which would be the same outcome as without this change, this file does not make the fonts in question required (in fact many are only available on demand now).

There has been other font and Unicode updates though that the composite fonts could and should take advantage of.

miloush:
SimSun-ExtG ranges look OK to me now, they are block aligned.

src/Microsoft.DotNet.Wpf/src/PresentationCore/Fonts/GlobalUserInterface.CompositeFont
```diff
@@ -1745,9 +1745,13 @@
...
                 Scale       = "1.0" />
             <!-- Other -->
+            <!-- CHT/KO/All
+                MingLiU-ExtB is not in UapFonts so Yu Gothic is specified as a fallback.
+                YaHei, Simsun-ExtB, and MingLiU_HKSCS-ExtB are included at the end of this catch-all category because each includes some characters
+                not present in other fonts. -->
             <FontFamilyMap
                 Unicode     = "20000-2A6DF"
-                Target      = "MingLiU-ExtB"
+                Target      = "MingLiU-ExtB, Yu Gothic UI, Microsoft YaHei UI, Simsun-ExtB, MingLiU_HKSCS-ExtB"
```
neerajmsgithub, line 1754: @Kuldeep-MS: We are changing fallback for MingLiU-ExtB also. Are they also not rendering correctly?

src/Microsoft.DotNet.Wpf/src/PresentationCore/Fonts/GlobalUserInterface.CompositeFont
```diff
@@ -1790,9 +1794,13 @@
...
                 Scale       = "1.0" />
             <!-- Other -->
+            <!-- All
+                MingLiU-ExtB is not in UapFonts so Yu Gothic is specified as a fallback.
+                YaHei, Simsun-ExtB, and MingLiU_HKSCS-ExtB are included at the end of this catch-all category because each includes some characters
+                not present in other fonts. -->
             <FontFamilyMap
                 Unicode     = "2A700-2B81F"
-                Target      = "MingLiU-ExtB"
+                Target      = "MingLiU-ExtB, Yu Gothic UI, Microsoft YaHei UI, Simsun-ExtB, MingLiU_HKSCS-ExtB"
```
neerajmsgithub, line 1803: Is fallback sequence updated in this PR specific to WPF?

singhashish-wpf (approved):

src/Microsoft.DotNet.Wpf/src/PresentationCore/Fonts/GlobalUserInterface.CompositeFont
Kuldeep-MS: we are changing this to be in sync with composite font used by other technologies such as dwrite.

src/Microsoft.DotNet.Wpf/src/PresentationCore/Fonts/GlobalUserInterface.CompositeFont
Kuldeep-MS: No it is not specific to WPF instead the changes are made to be in sync with dwrite.

PR #10514, opened 2025-02-27
Establish Code Coverage for wpf

singhashish-wpf:
###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/wpf/pull/10514)

Issue #10323, opened 2025-01-22
DependencyProperty tests are ignoring multiple variance of ArgumentException


h3xds1nz:
### The issue

In the new tests from #8215 for `DependencyProperty`, there's a lot of re-use of data that are being passed to `Register` methods of `DependencyProperty`.

This would be fine on its own when testing for other things, but in tests specifically for `Register` methods, an `ArgumentException` can be thrown either because the property was already registered (`SR.PropertyAlreadyRegistered`) or when auto generation of default property metadata fails (`SR.DefaultValueAutoAssignFailed`).

This can cause flaky unit test results that do not clearly express where the problem is and require further debugging.

### Other information

I will fix these myself by proper comparison of resource messages but I guess that will require first fixing the correct internals visibility between WindowsBase and WindowsBase.Tests or my VS is flaky at the moment. Assign to me please as I cannot without triage rights.

PR #10522, opened 2025-02-28
Fix DependencyProperty test

ThomasGoulet73:
## Description
Fixes the test failure seen in main currently. This was caused by the switch from xUnit test runner to VSTest test runner introduced in #10514 which changed the order in which test were ran and exposed an existing bug in the tests.

The problem is because the xUnit test runner ran the test `Register_InvokeStringTypeTypeValidateFail_Throws` before `Register_InvokeStringTypeType_Success` and VSTest test runner runs them in inverse order. This causes problem because both test registers a dependency property for the same type with the same property name, the thing is that `Register_InvokeStringTypeTypeValidateFail_Throws` expects the registration to fail and not register the dependency property in the cache while `Register_InvokeStringTypeType_Success` expects the registration to succeed and register the dependency property in the cache. 

When the order is `Register_InvokeStringTypeTypeValidateFail_Throws` -> `Register_InvokeStringTypeType_Success` it's fine because the dependency property is not in the cache for the second test.
When the order is `Register_InvokeStringTypeType_Success` -> `Register_InvokeStringTypeTypeValidateFail_Throws` it fails because  `Register_InvokeStringTypeTypeValidateFail_Throws` fails early because the property is already registered for the second test.

Here's the 2 conflicting dependency property name + type:
https://github.com/dotnet/wpf/blob/009ae5a190fcbd8f730d2f0959a96ffaebe244d5/src/Microsoft.DotNet.Wpf/tests/UnitTests/WindowsBase.Tests/System/Windows/DependencyPropertyTests.cs#L19

https://github.com/dotnet/wpf/blob/009ae5a190fcbd8f730d2f0959a96ffaebe244d5/src/Microsoft.DotNet.Wpf/tests/UnitTests/WindowsBase.Tests/System/Windows/DependencyPropertyTests.cs#L299-L300

Side note:
We should probably refactor how tests are done for DependencyProperty, this PR is just a patch to fix the current failure. IMO we should have 1 type per tests that register dependency properties, if we don't want to have a mess of types we could possibly create the types at runtime using IL emit.

## Customer Impact
None, tests only.

## Regression
WPF tests regression only.

## Testing
Running the tests locally.

## Risk
Low to none, tests only.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/wpf/pull/10522)

### DIFF
--- a/src/Microsoft.DotNet.Wpf/tests/UnitTests/WindowsBase.Tests/System/Windows/DependencyPropertyTests.cs
+++ b/src/Microsoft.DotNet.Wpf/tests/UnitTests/WindowsBase.Tests/System/Windows/DependencyPropertyTests.cs
@@ -296,7 +296,7 @@ public class DependencyPropertyTests
 
     public static IEnumerable<object?[]> Register_String_Type_Type_ValidateFail_TestData()
     {
-        yield return new object?[] { " ", typeof(int), typeof(DependencyObjectTests), null, 0 };
+        yield return new object?[] { " ", typeof(int), typeof(DependencyObjectTest1), null, 0 };
         yield return new object?[] { " ", typeof(int), typeof(DependencyObject), new PropertyMetadata(), 0 };
         yield return new object?[] { "Register_String_Type_Type_ValidateFail_TestData1", typeof(string), typeof(DependencyObjectTests), null, null };
         yield return new object?[] { "Register_String_Type_Type_ValidateFail_TestData2", typeof(string), typeof(DependencyObject), new PropertyMetadata(), null };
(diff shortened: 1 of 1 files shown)


ThomasGoulet73:
@dotnet/wpf-developers: It would be a good idea to prioritize this to fix the test failures in main and unblock other PRs.

dipeshmsft:
LGTM

singhashish-wpf (approved):

h3xds1nz:
@ThomasGoulet73 #10323 looks like you got to it sooner haha

PR #10523, opened 2025-02-28
[Managed DWrite] Migrate FontFileEnumerator to managed

ThomasGoulet73:
Contributes to #5305

## Description
Migrate DWrite FontFileEnumerator to managed.

## Customer Impact

## Regression
No.

## Testing
Local build + CI + I tested a few apps that uses custom fonts to test this code path.

## Risk
Low. For the most part, it is a copy of the C++ code manually rewritten to C# with as little changes as possible.


ThomasGoulet73:
@JeremyKuhne: Would you be interested in reviewing this ?

JeremyKuhne (approved):
The transliteration looks good.

src/Microsoft.DotNet.Wpf/src/PresentationCore/MS/internal/Text/TextInterface/FontFileEnumerator.cs
```diff
@@ -0,0 +1,78 @@
...
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System.Runtime.InteropServices;
+using MS.Internal.Interop.DWrite;
+
+namespace MS.Internal.Text.TextInterface
+{
+    [ClassInterface(ClassInterfaceType.None)]
+    [ComVisible(true)]
+    internal unsafe class FontFileEnumerator : IDWriteFontFileEnumeratorMirror
```
h3xds1nz, line 11: I believe we could seal this one.

ThomasGoulet73: I'd rather not change it because I try to stay as close as possible to the original code. I also haven't checked the implications of adding `sealed` to a COM object.

h3xds1nz: Sure, not really keen on it right now; would be much more pleasant if the migration was done already haha.

bytefish:
@JeremyKuhne Any news on this? Would be great to see it upstream. 😊

lindexi:
Waiting for long time.

PR #10543, opened 2025-03-06
Replace static ImagingCache with instanced WeakReferenceCache<K, V>

h3xds1nz:
## Description

Replaces static `ImagingCache` used within `BitmapDecoder` and `BitmapImage` with a more OOP-friendly solution of `WeakReferenceCache<K, V>`, using strongly-typed generic version for WeakRefs and using `Lock` instead of locking on instance.

This behaves the same way as the previous one. And another `ArrayList` definition gone.

## Customer Impact

Cleaner codebase for developers mostly, improved performance/allocs during cleanups.

## Regression

No.

## Testing

Local build.

## Risk

Should be low but review carefully.


src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Media/Imaging/WeakReferenceCache.cs
```diff
@@ -0,0 +1,84 @@
...
+    /// </summary>
+    private const int MaxCacheSize = 300;
+
+    private readonly Dictionary<K, WeakReference<V>> _cacheStore = new();
+    private readonly Lock _cacheLock = new();
+
+    /// <summary>
+    /// Adds an object to cache
+    /// </summary>
+    public void AddToCache(K key, WeakReference<V> value)
```
ThomasGoulet73, line 28: I feel like `ToCache` is redundant here:
```suggestion
    public void Add(K key, WeakReference<V> value)
```

src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Media/Imaging/WeakReferenceCache.cs
```diff
@@ -0,0 +1,84 @@
...
+
+            // add it
+            _cacheStore.Add(key, value);
+        }
+    }
+
+    /// <summary>
+    /// Removes an object from cache
+    /// </summary>
+    public void RemoveFromCache(K key)
```
ThomasGoulet73, line 65: I feel like `FromCache` is redundant here:
```suggestion
    public void Remove(K key)
```

src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Media/Imaging/WeakReferenceCache.cs
```diff
@@ -0,0 +1,84 @@
...
+        {
+            // if entry is there, remove it
+            _cacheStore.Remove(key);
+        }
+    }
+
+    /// <summary>
+    /// Attempts to retrieve an object from cache
+    /// </summary>
+    public bool TryGetValue(K key, out WeakReference<V>? value)
```
ThomasGoulet73, line 77: I feel like it would be cleaner if `value` was of type `V` instead of `WeakReference<V>`. This method would try to get the entry in the dictionary, call `TryGetTarget` on the weak reference, set `value` from the result and return true if we could return a value. If we couldn't return a value, remove the entry from the dictionary. The caller would then not have to worry about calling `RemoveFromCache` if the reference was collected and it would avoid acquiring the lock twice to get the value and remove.

src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Media/Imaging/WeakReferenceCache.cs
ThomasGoulet73, line 28: I feel like it would be cleaner if `value` was of type `V` instead of `WeakReference<V>`. It doesn't make sense to add a weak reference that could potentially be already collected.

src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Media/Imaging/WeakReferenceCache.cs
ThomasGoulet73, line 1: Since it's now more generic and unrelated to imaging, maybe move it to  https://github.com/dotnet/wpf/tree/main/src/Microsoft.DotNet.Wpf/src/PresentationCore/MS/internal ?

src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Media/Imaging/WeakReferenceCache.cs
```diff
@@ -0,0 +1,84 @@
...
+using System.Threading;
+
+namespace System.Windows.Media.Imaging;
+
+/// <summary>
+/// Provides keyed-caching for different objects stored as weak references.
+/// </summary>
+/// <remarks>The cache operations are thread-safe.</remarks>
+internal sealed class WeakReferenceCache<K, V> where K : notnull
+                                               where V : class
```
ThomasGoulet73, line 15: Nit, personally I put `where` on separate lines. It seems like it's the standard in this repo and in the WinForms repo, especially when there's more than 1 type parameter.

Something like this:
```suggestion
internal sealed class WeakReferenceCache<K, V>
    where K : notnull
    where V : class
```

src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Media/Imaging/WeakReferenceCache.cs
h3xds1nz: Yeah, good point

src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Media/Imaging/WeakReferenceCache.cs
h3xds1nz: Makes sense

src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Media/Imaging/WeakReferenceCache.cs
h3xds1nz: Didn't really check how 2+ are done, adjusted, don't really have a preference.

src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Media/Imaging/WeakReferenceCache.cs
h3xds1nz: Yeah, should help simplify the code as well.

h3xds1nz:
@ThomasGoulet73 I've made a few more changes, if you get a minute, please review.

src/Microsoft.DotNet.Wpf/src/PresentationCore/MS/internal/WeakReferenceCache.cs
```diff
@@ -0,0 +1,94 @@
...
+            // if entry is already there, exit
+            if (_cacheStore.ContainsKey(key))
+            {
+                return;
+            }
+
+            // if the table has reached the max size, try to see if we can reduce its size
+            if (_cacheStore.Count == MaxCacheSize)
+            {
+                foreach (KeyValuePair<K, WeakReference<V>> item in _cacheStore)
```
lindexi, line 43: I am concerned that in the HotPath, each time a item is added, it will encounter a limit of MaxCacheSize, leading to entering an infinite loop. However, within this loop, no items can be removed because there are no garbage collection (GC) triggers. This will result in severe performance degradation. My suggestion is to check how long the last loop was before attempting to enter another one, so as to avoid potential performance issues within HotPath.

src/Microsoft.DotNet.Wpf/src/PresentationCore/MS/internal/WeakReferenceCache.cs
```diff
@@ -0,0 +1,94 @@
...
+/// </summary>
+/// <remarks>The cache operations are thread-safe.</remarks>
+internal sealed class WeakReferenceCache<K, V>
+    where K : notnull
+    where V : class
+{
+    /// <summary>
+    /// Max number of elements stored in the cache.
+    /// </summary>
+    private const int MaxCacheSize = 300;
```
lindexi, line 22: Since the dictionary's capacity doubles each time, I believe that the 300 number is unreasonable and there's some space wastage.

src/Microsoft.DotNet.Wpf/src/PresentationCore/MS/internal/WeakReferenceCache.cs
h3xds1nz: This is the original design. I don't see an infinite loop here, I do see a bit more expensive enumeration but that's all part of the original design.

I'd prefer to keep it as-is for this PR.

ThomasGoulet73:
> @ThomasGoulet73 I've made a few more changes, if you get a minute, please review.

LGTM now, thanks.

src/Microsoft.DotNet.Wpf/src/PresentationCore/MS/internal/WeakReferenceCache.cs
lindexi: @h3xds1nz Sorry. It use the prime number from https://github.com/dotnet/runtime/blob/e377d6505ff83d59aad6cfea2827c43eba8c8253/src/libraries/System.Private.CoreLib/src/System/Collections/HashHelpers.cs#L31-L39

h3xds1nz: Yeah, as I've said. It has been adjusted, hence resolving.

elachlan:
Your `WeakReferenceCache` seems similar to the solution here (though slightly different): https://github.com/dotnet/runtime/issues/30759

We wanted something similar in winforms, I am not sure it ever got done. Maybe chime in on the issue to see if we can get more traction to implement in runtime?

PR #10544, opened 2025-03-07
Use shared clipboard code

JeremyKuhne:
This PR switches to using the shared clipboard code coming from System.Private.Windows.Core and exposes the new typed and JSON APIs.

Removes all code no longer used.

The PR includes the existing refactoring PR. The final commit is the actual clipboard change.


src/Microsoft.DotNet.Wpf/src/Extensions/PresentationFramework-SystemDrawing/SystemDrawingExtension.cs
```diff
@@ -16,61 +17,30 @@
...
         {
-            IntPtr hMetafile = IntPtr.Zero;
-            Metafile metafile = data as Metafile;
-
-            if (metafile != null)
-            {
-                // Get the Windows handle from the metafile object.
-                hMetafile = metafile.GetHenhmetafile();
-            }
+            Metafile metafile => metafile.GetHenhmetafile(),
```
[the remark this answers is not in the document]
JeremyKuhne: This is correct.

src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Ole/WpfOleServices.cs
```diff
@@ -0,0 +1,191 @@
...
+            if (mediumType == TYMED.TYMED_GDI)
+            {
+                // Get the bitmap from the handle of bitmap.
+                object bitmap = Imaging.CreateBitmapSourceFromHBitmap(
+                    (HBITMAP)(nint)medium.hGlobal,
+                    0,
+                    Int32Rect.Empty,
+                    sizeOptions: null);
+
+                if (bitmap is T t)
```
lindexi, line 128: Should we delete the `medium.hGlobal` here?

src/Microsoft.DotNet.Wpf/tests/UnitTests/PresentationCore.Tests/System/Windows/ClipboardTests.cs
```diff
@@ -103,8 +114,8 @@ public void SetData_EmptyOrWhitespaceFormat_ThrowsArgumentException(string? form
 
     [WpfFact]
     public void SetData_Null_Throws()
-    {
-        Action action = () => Clipboard.SetData("MyData", data: null!);
+        {
```
lindexi, line 117: The error format?

JeremyKuhne: Yes, I'll catch this with the next update I make.

src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/DataObject.DataStore.cs
```diff
@@ -27,71 +25,111 @@ public DataStore()
...
-                bool canAutoConvert = true;
-
-                for (int dataStoreIndex = 0; dataStoreIndex < entries.Length; dataStoreIndex++)
-                {
-                    if (!entries[dataStoreIndex].AutoConvert)
-                    {
-                        canAutoConvert = false;
-                        break;
-                    }
-                }
```
harshit7962: I am not sure if this is an edge case, just pointing it out to get some opinions.

Since this is no longer feasible, is there a case where lets say we do a 
`SetData(format f1, object o1, autoconvert: false)`
and we do it again as in `SetData(same format f1, same or diff object o, autoconvert: true)`

Now according to current implementation I suppose it would have `canAutoConvert` as `false`. But with new implementation, i.e., not having multiple entries, we would have `autoconvert` as `true`, the later entry, and hence the `canAutoConvert` as `true`.

harshit7962 (commented):
Still going through the PR, posting the review to get current queries/nits addressed.

src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/DataObject.DataStore.cs
JeremyKuhne: The new entry is put at the head of the array so I believe it would be found first and match the current behavior. This would be a good test to add to validate that we're either keeping existing behavior here or deliberately changing a corner case.

src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/OleServicesContext.cs
```diff
@@ -13,300 +12,250 @@
...
-                    // Create OleSErvicesContext instance.
-                    oleServicesContext = new OleServicesContext();
+            return oleServicesContext;
+        }
+    }
 
-                    // Save the ole services context into the UIContext data slot.
-                    Thread.SetData(OleServicesContext._threadDataSlot, oleServicesContext);
-                }
+    internal void EnsureThreadState()
```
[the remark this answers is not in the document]
JeremyKuhne: It is at this point. In the final commit the other clipboard methods go away and this helper is all that is called:

``` C#
    internal static void EnsureThreadState()
    {
        // Poke the CurrentOleServicesContext to ensure it is created, this will throw if the current thread is not STA.
        _ = CurrentOleServicesContext;
    }
```

src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/clipboard.cs
```diff
@@ -235,57 +194,40 @@ public static string GetText(TextDataFormat format)
...
     public static void SetData(string format, object data)
     {
-        ArgumentNullException.ThrowIfNull(format);
-
-        if (format == string.Empty)
-        {
-            throw new ArgumentException(SR.DataObject_EmptyFormatNotAllowed);
-        }
-
+        ArgumentException.ThrowIfNullOrEmpty(format);
```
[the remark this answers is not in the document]
JeremyKuhne: It has the same throw behavior, so no. The message will change, however. [ArgumentException.cs](https://source.dot.net/#System.Private.CoreLib/src/libraries/System.Private.CoreLib/src/System/ArgumentException.cs,104)

src/Microsoft.DotNet.Wpf/src/PresentationCore/MS/internal/SystemDrawingHelper.cs
```diff
@@ -1,104 +1,72 @@
...
+        /// <inheritdoc cref="SystemDrawingExtensionMethods.GetHBitmap(object?, out int, out int)"/>
+        internal static HBITMAP GetHBitmap(object? data, out int width, out int height)
         {
-            SystemDrawingExtensionMethods extensions = AssemblyHelper.ExtensionsForSystemDrawing(force: true);
-            if (extensions != null)
+            var extensions = AssemblyHelper.ExtensionsForSystemDrawing(force: true);
+            if (extensions is not null)
             {
-                return extensions.GetHBitmap(data, out width, out height);
+                return (HBITMAP)extensions.GetHBitmap(data, out width, out height);
```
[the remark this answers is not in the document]
JeremyKuhne: Being type safe. As we have definitions for these now, we should use them for clarity and safety.

src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/clipboard.cs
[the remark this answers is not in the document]
harshit7962: Apologies, I thought I had deleted this 😅

src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/DataObject.DataStore.cs
harshit7962: > The new entry is put at the head of the array so I believe it would be found first and match the current behavior

Yeah, (Taking about current implementation) but I guess we are going through older entries as well, in which case at some point our first `SetData`'s `autoconvert` will come into picture and make `canAutoConvert` as `false`. This probably does not make a lot of sense but this is how it is.

I think adding this test case will give more clarity.

harshit7962 (approved):
LGTM

src/Microsoft.DotNet.Wpf/src/Extensions/PresentationFramework-SystemDrawing/SystemDrawingExtension.cs
```diff
@@ -84,54 +54,45 @@ internal override IntPtr GetHBitmap(object data, out int width, out int height)
...
-            // for reference the tag is PropertyTagExifUserComment [0x9286] or 37510 (int)
-            PropertyItem piComment = img.GetPropertyItem(37510);
-            return new MemoryStream(piComment.Value);
+            // Read the GIF header
+            Bitmap bitmap = new(stream);
+
+            // Read the comment as that is where the ISF is stored.
+            // For reference the tag is PropertyTagExifUserComment [0x9286] or 37510 (int)
+            PropertyItem? piComment = bitmap.GetPropertyItem(37510);
+            return new MemoryStream(piComment!.Value!);
```
h3xds1nz, line 83: Could we adjust this as discussed in [other PR](https://github.com/dotnet/wpf/pull/10495/files/a75bfc754a331f466a63e80bc494904a449016f2#r1983519900)?

src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Ole/WpfOleServices.cs
```diff
@@ -0,0 +1,191 @@
...
+using Windows.Win32;
+using Windows.Win32.Graphics.Gdi;
+using Windows.Win32.System.Com;
+using Windows.Win32.System.Ole;
+using Com = Windows.Win32.System.Com;
+using HRESULT = Windows.Win32.Foundation.HRESULT;
+
+namespace System.Windows.Ole;
+
+internal unsafe class WpfOleServices : IOleServices
```
h3xds1nz, line 20: Should we seal this one?

src/Microsoft.DotNet.Wpf/src/Shared/MS/Internal/Invariant.cs
```diff
@@ -68,7 +69,18 @@ internal static void Assert(bool condition)
...
             {
-                FailFast(null, null);
+                FailFast(message: null, detailMessage: null);
+            }
+        }
+
+        /// <summary>
+        ///  Checks <paramref name="value"/> for <see langword="null"/> and shuts down the application if true.
+        /// </summary>
+        internal static void AssertNotNull([NotNull] object value)
```
h3xds1nz, line 79: Is this used anywhere? I couldn't find any usage (then again, just browsing via Github) but given this is a new code, I'd expect it in the diff.

JeremyKuhne:
@h3xds1nz I'll respond to your feedback when I rebase tomorrow

src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Ole/WpfOleServices.cs
```diff
@@ -0,0 +1,191 @@
...
+    {
+        // Image is a special case because we are reading bitmaps directly from the SerializationRecord.
+        return typeof(T).FullName.Equals("System.Drawing.Image");
+    }
+
+    public static bool IsValidTypeForFormat(Type type, string format) => format switch
+    {
+        DataFormatNames.Bitmap or DataFormatNames.BinaryFormatBitmap =>
+            type == typeof(BitmapSource) || type.FullName is "System.Drawing.Bitmap" or "System.Drawing.Image",
+        DataFormatNames.Emf /* TODO: or DataFormatNames.BinaryFormatMetafile */=>
```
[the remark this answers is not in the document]
JeremyKuhne: Removed, had to flow from WinForms- thanks for catching!

src/Microsoft.DotNet.Wpf/src/Shared/MS/Internal/Invariant.cs
JeremyKuhne: It isn't in the final commit. Wanted to leave it here for other asserts to move to with annotating progress.

src/Microsoft.DotNet.Wpf/src/Extensions/PresentationFramework-SystemDrawing/SystemDrawingExtension.cs
JeremyKuhne: Added a comment. Adjusting it would take a little bit of refactoring, want to keep that out of this PR.

src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/clipboard.cs
```diff
@@ -1,768 +1,497 @@
...
+    /// <summary>
+    ///  Return <see langword="true"/> if Clipboard contains the audio data. Otherwise, return <see langword="false"/>.
+    /// </summary>
+    public static bool ContainsAudio() => ContainsDataInternal(DataFormats.WaveAudio);
 
-            string text;
+    /// <summary>
+    ///  Return <see langword="true"/> if Clipboard contains the specified data format. Otherwise, return <see langword="false"/>.
+    /// </summary>
+    public static bool ContainsData(string format) => !string.IsNullOrWhiteSpace(format) && ContainsDataInternal(format);
```
harshit7962, line 34:
```suggestion
    public static bool ContainsData(string format)
    {
        ArgumentException.ThrowIfNullOrEmpty(format);
        return ContainsDataInternal(format);
    }
```
Would return `False` if format is null or empty instead of throwing exception like before.

src/Microsoft.DotNet.Wpf/src/Extensions/PresentationFramework-SystemDrawing/SystemDrawingExtension.cs
h3xds1nz: @JeremyKuhne Sure, could we prefix it with `TODO: ` possibly? Feels appropriate.

harshit7962 (approved):
The test pass on these changes are done, no failures were reported.

siagupta0202 (approved):

PR #10603, opened 2025-03-19
Use HashSet<T> in CommitManager instead of wrapped Dictionary<K, V>

h3xds1nz:
## Description

Swaps use of `Dictionary<K, V>` where only `K` is used under the `Set<T>` wrapper for `HashSet<T>`.

- This decreases memory usage by a substantial amount.
- Also improves performance by further avoiding the double look-up on the original `Add` method.
- `IsInScope` was made static as it has no instance dependencies.
- Members were made `readonly` so they can be treated as JIT constants.
- Seals the class.

## Customer Impact

## Regression

No.

## Testing

Local build.

## Risk

Low.

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/wpf/pull/10603)


dipeshmsft:
@h3xds1nz, I was looking at this code just last week, was thinking of making changes on similar lines, but you got to this first 😄.

PR #10618, opened 2025-03-22
[MilCodeGen] Declare Equals/GetHashCode and property getters for fields as readonly

h3xds1nz:
## Description

Modifiers MilCodeGen to generate `readonly` getters for fields where `SkipProperties="false"`, in case the properties are skipped (`SkipProperties="true"`) they live in the non-generated area and thus were modified manually.
- The `Equals` override uses new syntax and is `readonly`.
- The non-override `Equals ` is now `readonly`.
- `GetHashCode` is `readonly` and all the supporting props (`IsEmpty`, `IsDistinguishedIdentity`) are `readonly` to prevent copies in the particular call chain.

Rest of the members that were not in the impact zone are left untouched and shall be done separately in subsequent PRs.

## Customer Impact

Increased performance/better codegen.

## Regression

No.

## Testing

Local build, most of the generated structs have unit tests already including `Equals`/`GetHashCode` checks.

## Risk

Low. The point to consider is whether we could make any of the members that are to become `readonly` mutate the struct in the future, for me that answer is no. `Equals`/`GetHashCode` should never modify the state, even on mutable structs.

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/WpfGfx/codegen/mcg/generators/ManagedStruct.cs
+++ b/src/Microsoft.DotNet.Wpf/src/WpfGfx/codegen/mcg/generators/ManagedStruct.cs
@@ -101,7 +101,7 @@ namespace MS.Internal.MilCodeGen.Generators
                     /// <returns>
                     /// int - the HashCode for this [[resource.Name]]
                     /// </returns>
-                    public override int GetHashCode()
+                    public override readonly int GetHashCode()
                     {
                         [[getHashCodeBody]]
                     }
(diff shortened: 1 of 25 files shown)


src/Microsoft.DotNet.Wpf/src/WpfGfx/codegen/mcg/generators/ManagedStruct.cs
```diff
@@ -204,15 +204,9 @@ public static bool Equals ([[resource.Name]] [[lowerName]]1, [[resource.Name]] [
...
+                    public override readonly bool Equals(object o)
                     {
-                        if ((null == o) || !(o is [[resource.Name]]))
-                        {
-                            return false;
-                        }
-
-                        [[resource.Name]] value = ([[resource.Name]])o;
-                        return [[resource.Name]].Equals(this,value);
+                        return o is [[resource.Name]] other && [[resource.Name]].Equals(this, other);
```
h3xds1nz, line 209: We could use [[lowerName]] (which is the resource name in lower-camelCase) if desirable instead of `other`.

I'm fine eitherway.

Issue #7536, opened 2023-02-15
`ManagedWndProcTracker.HookUpDefWindowProc` may throw exception on finalizer thread


jhinder:
### Description

During shutdown of a WPF application, the method `MS.Win32.ManagedWndProcTracker.HookUpDefWindowProc` may throw an `System.ComponentModel.Win32Exception` ("Invalid window handle", native error 1400) on the finalizer thread, terminating the application.

### Reproduction Steps

Unfortunately we cannot provide a minimal working example or repro steps, and the affected application is not yet publicly released.

As this is related to the finalizer thread, it is also not reliably reproducible: one of our QA engineers needed about 10 attempts to encounter the crash, on my machine it took more than 200 attempts.

### Expected behavior

No exception is thrown on the finalizer thread, and as a result the runtime is not terminated, and the process can exit normally.

### Actual behavior

`System.ComponentModel.Win32Exception` with this stack trace according to the VS debugger:

```
>	KERNELBASE.dll!RaiseException()	Unknown
 	[Managed to Native Transition]	
 	WindowsBase.dll!MS.Win32.ManagedWndProcTracker.HookUpDefWindowProc(System.IntPtr hwnd)	Unknown
 	WindowsBase.dll!MS.Win32.ManagedWndProcTracker.OnAppDomainProcessExit()	Unknown
 	WindowsBase.dll!MS.Win32.ManagedWndProcTracker.ManagedWndProcTrackerShutDownListener.OnShutDown(object target, object sender, System.EventArgs e)	Unknown
 	WindowsBase.dll!MS.Internal.ShutDownListener.HandleShutDown(object sender, System.EventArgs e)	Unknown
 	System.Private.CoreLib.dll!System.AppContext.OnProcessExit()	Unknown
 	[Native to Managed Transition]	
 	kernel32.dll!BaseThreadInitThunk()	Unknown
 	ntdll.dll!RtlUserThreadStart()	Unknown
```

### Configuration

* .NET SDK 7.0.103
* .NET Desktop Runtime 6.0.13 (x64)
* Target framework `net6.0-windows10.0.17763.0`
* Windows 11 22H2 (x64)

### Other information

This is the statement that causes the exception to be thrown, according to the VS debugger: https://github.com/dotnet/wpf/blob/v6.0.13/src/Microsoft.DotNet.Wpf/src/Shared/MS/Win32/ManagedWndProcTracker.cs#L157

[...]

PR #10631, opened 2025-03-24
Fix: Plausible Exception while closing an Application

harshit7962:
## Description
The `PostMessage` call while closing the application might throw an invalid window handle exception. Ideally, the `Win32`'s `PostMessage` call is not designed to throw an exception; instead, it is meant to return a boolean indicating success or failure. However, the current call is wrapped to throw a `Win32Exception`. Interestingly, the `SetWindowLong` call just above this call is in a try-catch block that ignores the invalid window handle exception. It makes sense to move the `PostMessage` call inside the try-catch block to avoid the same exception being thrown.

The current `PostMessage` wrapper being used is also modified to get the error code as soon as the failure in `IntPostMessage` is encountered.

## Customer Impact
This might be a potential fix to the issue #7536 

## Regression
No

## Testing
Local Build Pass
Tested with a possible failing scenario

## Risk
Low
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/wpf/pull/10631)

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/Shared/MS/Win32/UnsafeNativeMethodsCLR.cs
+++ b/src/Microsoft.DotNet.Wpf/src/Shared/MS/Win32/UnsafeNativeMethodsCLR.cs
@@ -964,7 +964,8 @@ namespace MS.Win32
         {
             if (!IntPostMessage(hwnd, msg, wparam, lparam))
             {
-                throw new Win32Exception(Marshal.GetLastWin32Error());
+                int errorCode = Marshal.GetLastWin32Error();
+                throw new Win32Exception(errorCode);
             }
         }
 
--- a/src/Microsoft.DotNet.Wpf/src/Shared/MS/Win32/UnsafeNativeMethodsCLR.cs
+++ b/src/Microsoft.DotNet.Wpf/src/Shared/MS/Win32/UnsafeNativeMethodsCLR.cs
@@ -964,8 +964,7 @@ namespace MS.Win32
         {
             if (!IntPostMessage(hwnd, msg, wparam, lparam))
             {
-                int errorCode = Marshal.GetLastWin32Error();
-                throw new Win32Exception(errorCode);
+                throw new Win32Exception();
             }
         }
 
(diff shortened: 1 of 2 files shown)


src/Microsoft.DotNet.Wpf/src/Shared/MS/Win32/UnsafeNativeMethodsCLR.cs
```diff
@@ -964,7 +964,8 @@ internal static void PostMessage(HandleRef hwnd, WindowMessage msg, IntPtr wpara
         {
             if (!IntPostMessage(hwnd, msg, wparam, lparam))
             {
-                throw new Win32Exception();
+                int errorCode = Marshal.GetLastWin32Error();
+                throw new Win32Exception(errorCode);
```
h3xds1nz, line 968:
```suggestion
                throw new Win32Exception();
```
This change is unnecessary, the empty ctor already calls `GetLastWin32Error` (technically, its `GetLastPInvokeError`, which is what `GetLastWin32Error` will call).

harshit7962: @h3xds1nz yeah that is the case, but this is added here to avoid any other exception that the empty constructor call might return due to a different failure from any other thread (very less likely though). This follows the practice implemented in the same file. 

**Ref:** https://github.com/dotnet/wpf/blob/3a0442ac70cc6bbb15f45d3b60aa80f6758f234c/src/Microsoft.DotNet.Wpf/src/Shared/MS/Win32/UnsafeNativeMethodsCLR.cs#L396-L409

harshit7962: @h3xds1nz thanks for this piece of information. Makes sense to not explicitly store the error if we are avoiding bad reads. Will go ahead and remove this part.

h3xds1nz: @harshit7962 welcome

pchaurasia14 (approved):

Issue #6264, opened 2022-03-16
WPF ProgressBars with IsIndeterminate = true appear to leak animations, since .NET 4.5


ryanmolden:
* .NET Core Version: (e.g. 3.0 Preview1, or daily build number, use `dotnet --info`): All since .NET 4.5
* Windows version: (`winver`): Any
* Does the bug reproduce also in WPF for .NET Framework 4.8?: Yes
* Is this bug related specifically to tooling in Visual Studio (e.g. XAML Designer, Code editing, etc...)? No

**Problem description:** WPF's ProgressBar uses a well-known named template part (GlowingRectTemplateName which is == PART_GlowRect) in order to control the indeterminate animation. Specifically it wants to stop the animation when the control becomes invisible.

Since aero2.normalcolor.xaml was introduced this no longer works as the template has no PART_GlowRect and instead triggers the animation based on the Indeterminate VisualState.

Specifically see 

Code here: https://github.com/dotnet/wpf/blob/89d172db0b7a192de720c6cfba5e28a1e7d46123/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Controls/ProgressBar.cs#L62

and here: https://github.com/dotnet/wpf/blob/89d172db0b7a192de720c6cfba5e28a1e7d46123/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Controls/ProgressBar.cs#L220

and template here (other templates have similar issues, other than aero.normalcolor.xaml): https://github.com/dotnet/wpf/blob/89d172db0b7a192de720c6cfba5e28a1e7d46123/src/Microsoft.DotNet.Wpf/src/Themes/PresentationFramework.Aero2/Themes/Aero2.NormalColor.xaml#L4428

This means that for stock WPF ProgressBars with IsIndeterminate=True, the animations will run forever regardless of whether the animation is visible. This is esepcially bad on high-refresh rate monitors as the callbacks occur much more frequently and keep what could be an idle system non-idle preventing battery saving mode switches on laptops.
 
**Actual behavior:** Animations run forever
 
**Expected behavior:** Animations stop when the control becomes invisible, as the code appears to facilitate.
 
**Minimal repro:**

MainWindow.xaml

```
<Window x:Class="AnimationTestApp.MainWindow"
        xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
        xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
[...]

PR #10649, opened 2025-03-25
Fix Indeterminate Progress Bar animation with Cache Mode set up

### LINKED ISSUE #8960: Indeterminate progress bar used as BitmapCacheBrush target do not animate
### Description

When using a `BitmapCacheBrush` whose `Target` is a `ProgressBar` with `Indeterminate="True"`, the rendered control does not animate.

### Reproduction Steps

Create a new WPF project that targets either .NET 7 (`net7.0-windows`) or .NET 8 (net8.0-windows).

Add the following code to `MainWindow.xaml`:

```xaml
<Window x:Class="BlankWpf.MainWindow"
        xmlns:local="clr-namespace:BlankWpf"
        mc:Ignorable="d"
    <Window.Resources>
        <ProgressBar x:Key="CachedProgressBar" Width="200" Height="10" IsIndeterminate="True" >
            <ProgressBar.CacheMode>
                <BitmapCache SnapsToDevicePixels="True" />
            </ProgressBar.CacheMode>
        </ProgressBar>

        <BitmapCacheBrush x:Key="CachedProgressBarBrush" Target="{StaticResource CachedProgressBar}" />
    </Window.Resources>

    <Grid>
        <Rectangle Fill="{StaticResource CachedProgressBarBrush}" Width="200" Height="10" />
    </Grid>
</Window>
```

Run the application.

### Expected behavior

An animated, indeterminate progress spinner should be rendered. This is the behavior on .NET 6. 

progress

### Actual behavior

A static, completely filled progress bar is rendered:

image

### Regression?

This is a regression; it used to work in .NET 6.

### Known Workarounds

_No response_

### Impact

My application uses many progress spinner that we synchronize and reduce their overhead with by using the bitmap cache. This bug breaks all of those spinners.

### Configuration

_No response_

### Other information

_No response_

h3xds1nz:
Fixes #8960 

## Description

Adjusts the logic behind transitioning to `Determinate` state to avoid background animation to exclude when a `CacheMode` is set up on the `ProgressBar` and the ProgressBar is not a part of the visual tree (`IsVisible` evaluates to `false`).

## Customer Impact

Customers will not be able to cache their indeterminate progress bars.

## Regression

Yeah, broken in #6266.

## Testing

Local build, repro in #8960 and #6264.

## Risk

Medium, we should have more test cases covering these kinda changes.

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Controls/ProgressBar.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Controls/ProgressBar.cs
@@ -274,7 +274,11 @@ namespace System.Windows.Controls
         {
             // We change the visual state to determinate when not IsVisible for performance reasons.
             // By default, the animation for the Indeterminate state will continue even when the progress bar is not visible.
-            if (!IsIndeterminate || !IsVisible)
+
+            // However, we can only allow ourselves to do that if the ProgressBar is not being cached, since when the UIElement is defined within
+            // resources, IsVisible property will always evaluate to false as the UIElement is not a part of the visual tree at that point. (#8960)
+
+            if (!IsIndeterminate || (!IsVisible && CacheMode is null))
             {
                 VisualStateManager.GoToState(this, VisualStates.StateDeterminate, useTransitions);
             }


src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Controls/ProgressBar.cs
```diff
@@ -274,7 +274,11 @@ internal override void ChangeVisualState(bool useTransitions)
         {
             // We change the visual state to determinate when not IsVisible for performance reasons.
             // By default, the animation for the Indeterminate state will continue even when the progress bar is not visible.
-            if (!IsIndeterminate || !IsVisible)
+
+            // However, we can only allow ourselves to do that if the ProgressBar is not being cached, since when the UIElement is defined within
+            // resources, IsVisible property will always evaluate to false as the UIElement is not a part of the visual tree at that point. (#8960)
+
+            if (!IsIndeterminate || (!IsVisible && CacheMode is null))
```
ThomasGoulet73, line 281: Wouldn't it be better to check if the `ProgressBar` is part of the visual tree instead of checking `CacheMode` ?

Something like this:
```suggestion
            if (!IsIndeterminate || (!IsVisible && VisualTreeHelper.GetParent(this) is not null))
```
<br/>
I haven't tested it but to me it's more precise because the bug does not seem specific to `CacheMode`.

ThomasGoulet73 (commented):
Thanks for fixing this, that was definitely an oversight by me.

src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Controls/ProgressBar.cs
h3xds1nz: @ThomasGoulet73 Maybe I scoped too narrowly indeed, I'll check this but imho should work indeed.

PR #10374, opened 2025-01-30
[StyleCleanUp] Remove invalid global SuppressMessage attributes (IDE0076)

h3xds1nz:
Fixes #10373

## Description

Removes warning suppressions with invalid target or scope.

## Customer Impact

Cleaner codebase for developers.

## Regression

No.

## Testing

Local build.

## Risk

None.

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/wpf/pull/10374)

PR #10683, opened 2025-03-31
[StyleCleanUp] Fix misplaced using directives in CSP (IDE0065)

### LINKED ISSUE #10682: IDE0065: 'using' directive placement
More info: https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/style-rules/ide0065

h3xds1nz:
Fixes #10682

## Description

This fixes `using` placements in CSP which I've overlooked when I've removed it in #10374 as we still don't run `MilCodeGen` during build, hence this fixes that + also corrects the header for IDE0073 for generated files from the last commit.

## Customer Impact

Cleaner codebase for developers, being able to run MilCodeGen.

## Regression

Kinda, MilCodeGen won't work without suppressing warnings now.

## Testing

Local build, codegen run.

## Risk

None.

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/wpf/pull/10683)

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/WpfGfx/tools/csp/CsPrimeParser.cs
+++ b/src/Microsoft.DotNet.Wpf/src/WpfGfx/tools/csp/CsPrimeParser.cs
@@ -11,14 +11,12 @@
 //
 //
 
+using System.Text.RegularExpressions;
+using System.Collections;
+using System.Text;
+
 namespace MS.Internal.Csp
 {
-    using System;
-    using System.IO;
-    using System.Text;
-    using System.Text.RegularExpressions;
-    using System.Collections;
-
     internal sealed class CsPrimeParser
     {
         private struct Position
--- a/src/Microsoft.DotNet.Wpf/src/Shared/MS/Internal/Generated/ToleranceTypeValidation.cs
+++ b/src/Microsoft.DotNet.Wpf/src/Shared/MS/Internal/Generated/ToleranceTypeValidation.cs
@@ -1,6 +1,5 @@
 // Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the MIT license.
-// See the LICENSE file in the project root for more information.
 
 //
 //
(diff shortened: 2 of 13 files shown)


src/Microsoft.DotNet.Wpf/src/WpfGfx/tools/csp/CsPrimeParser.cs
```diff
@@ -11,14 +11,12 @@
 //
 //
 
+using System.Text.RegularExpressions;
+using System.Collections;
```
jizc, line 15: nit: Shouldn't the usings be sorted alphabetically?

h3xds1nz: The day usings are sorted in alphabetically in this repo I'm happy to accept that nit. Currently it is wild west. Neither it was in this exact example you've reviewed.

Also, what is alphabetically? Is it gonna be (dot)

System
System.IO

or

System.IO
System

Are we gonna sort public namespaces first, then internals? Do you sort static usings last or first? How do you sort aliases? Where are they placed?

We don't have an issue for that open yet nor guideline. I don't deem the "sorting" of usings important tbh, 

My IDE settings is up->down based on length, alises first with new line in-between and statics last with new line in-between for example but I don't re-sort then generally, this was just fixer doing its thing based on my settings.

Gonna resolve this, but you may open an issue and we can work something out of that I guess, though this is generally hard to enforce as there's no way (correct me if I'm wrong) to enforce this bit, unlike other stuff.

jizc: The order I'm personally used to is the one that [SA1210](https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/master/documentation/SA1210.md) enforces, but I see now that that rule has been fully disabled in the root `.editorconfig`, so I retract my nit 😅 
https://github.com/dotnet/wpf/blob/4aff730d0aed8d28668f591ca85dd6289b5f23f1/.editorconfig#L1537

siagupta0202 (approved):

siagupta0202:
@h3xds1nz Thank you for your contribution!

h3xds1nz:
@siagupta0202 Thank you for approving :)

PR #10690, opened 2025-04-01
Replace double/float IsNaN + IsInfinity checks with optimized IsFinite

h3xds1nz:
## Description

Several versions ago, `IsFinite` was introduced as an intrinsic, therefore there's no need to perform two separate checks for the same thing, it simplifies the code, uses less bytes and the codegen is also smaller and more efficient.

I've also cleaned up some of the unused util methods in that domain.

### Simple benchmark
| Method    | param | Mean [ns] | Error [ns] | StdDev [ns] | Code Size [B] | Allocated [B] |
|---------- |--------- |----------:|-----------:|------------:|--------------:|--------------:|
| Original | NaN      | 0.0056 ns |  0.0069 ns |   0.0065 ns |          50 B |             - |
| PR__EDIT  | NaN      | 0.0029 ns |  0.0046 ns |   0.0043 ns |          27 B |             - |
| Original | 5        | 0.1519 ns |  0.0082 ns |   0.0073 ns |          50 B |             - |
| PR__EDIT  | 5        | 0.0000 ns |  0.0000 ns |   0.0000 ns |          27 B |             - |
| Original | Infinity | 0.1539 ns |  0.0091 ns |   0.0085 ns |          50 B |             - |
| PR__EDIT  | Infinity | 0.0002 ns |  0.0006 ns |   0.0006 ns |          27 B |             - |

There are a few places I didn't remove these checks, that should be done separately. Those are constructors that emit different messages for `NaN` and `Infinity`. We should have a discussion on how to approach this. As those are all exception cases, ideal approach is probably to create a non-inlinable throw helper that further then checks `IsNaN`/`IsInfinity` in case `IsFinite` already fails and throws the exception with correct messages. Would reduce the ctor complexity for an uncommon case, though it will start to matter a bit less in .NET 10+ as methods with EH frames will now be inlinable.

P.S. That `DoubleValue_Validate` from `DocumentViewer` is the real legacy code gem that should have been preserved I guess.

## Customer Impact

## Regression

No.

## Testing

Local build.

## Risk

Medium, I was doing all of these replacements manually and thus this is rather error-prone where I could invert a condition badly and I didn't notice it. Review carefully this one please, thank you.


src/Microsoft.DotNet.Wpf/src/PresentationCore/MS/internal/AnimatedTypeHelpers.cs
```diff
@@ -662,95 +657,53 @@ internal static bool IsValidAnimationValueMatrix(Matrix value)
...
-        internal static bool IsValidAnimationValueSingle(Single value)
+        internal static bool IsValidAnimationValueSingle(float value)
         {
-            if (IsInvalidDouble(value))
-            {
-                return false;
-            }
-
-            return true;
+            return double.IsFinite(value);
```
jizc, line 686: I'm curious if it wouldn't be a tiny bit better to use `float.IsFinite` here instead?

jizc (commented):
LGTM! Just one comment/question.

src/Microsoft.DotNet.Wpf/src/PresentationCore/MS/internal/AnimatedTypeHelpers.cs
h3xds1nz: It does make sense, doesn't it. Thanks for reviewing btw.

PR #10372, opened 2025-01-30
[StyleCleanUp] Correct file header/license info (IDE0073)

h3xds1nz:
Fixes #10371 

## Description

Puts the file header into correct state, most of the time its removal of the last line, some files didn't have the header at all (ref files).

## Customer Impact

Everyone is informed the project is MIT.

## Regression

No.

## Testing

Local build.

## Risk

None, no IL changes.

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/wpf/pull/10372)

PR #10021, opened 2024-10-30
[Cleanup] Add missing accessibility modifiers in System.Xaml (IDE0040)

h3xds1nz:
## Description

Adds missing accessibility modifiers in `System.Xaml`. All of the edits were done automatically.

## Customer Impact

Cleaner codebase.

## Regression

No.

## Testing

Local build.

## Risk

Minimal.

PR #10702, opened 2025-04-04
[StyleCleanUp] Fix MilCodeGen license header

ThomasGoulet73:
## Description
Most of the generated files were changed in dotnet/wpf#10372, this PR syncs MilCodeGen to the current generated code to avoid unnecessary diffs when running MilCodeGen.

~~This PR also reverts whitespace line changes from dotnet/wpf#10021 that made the generated files out of sync with MilCodeGen.~~ EDIT: Reverted in this PR and fixed in dotnet/wpf#10762.

## Customer Impact
None, code style only.

## Regression
No.

## Testing
Local build + running MilCodeGen

## Risk
Low to none.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/wpf/pull/10702)

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Generated/TextDecoration.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Generated/TextDecoration.cs
@@ -15,8 +15,6 @@ using MS.Utility;
 
 namespace System.Windows
 {
-
-
     public sealed partial class TextDecoration : Animatable
     {
         //------------------------------------------------------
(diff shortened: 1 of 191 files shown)


src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Generated/TextDecoration.cs
```diff
@@ -15,6 +15,8 @@
 
 namespace System.Windows
 {
+
+
```
h3xds1nz, line 19: How about instead of adding the new lines back in (I apologize, I thought I did commit the change) we remove them instead as they serve no purpose and are only a byproduct of the fact that `MilCodeGen` can't avoid the new line at this very moment?

If it is not within your time constraints I'm can put a PR up that will allow for the empty lines not being generated when the embedded C# code evaluates to empty (so the empty new line gets inserted atm).

ThomasGoulet73: I decided to revert the empty lines changes and fix it in a separate PR.

siagupta0202 (approved):

ThomasGoulet73:
Thanks @siagupta0202

PR #10715, opened 2025-04-07
Avoid caching `typeof` result to decrease code size and improve perf

h3xds1nz:
## Description

There are a lot of instances, mostly in generated code, that we cache the result of the `typeof(T)`. This is basically an anti-pattern for a few versions now as `typeof` is a JIT-time constant and the data is stored on the frozen heap.

Previously it was a cool optimization, now it is actually a de-optimization that increases code size and in some cases results in having additional fields stored. What this PR will mostly do is speed up the static init of core objects.

First commit contains generated changes via `MilCodeGen`, second/third are manual replacements which could use a review.

## Customer Impact

Decreased code size, slightly improved performance on startup.

## Regression

No.

## Testing

Local build.

## Risk

Low.

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/wpf/pull/10715)

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/PresentationCore/MS/internal/Ink/InkSerializedFormat/DrawingAttributeSerializer.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationCore/MS/internal/Ink/InkSerializedFormat/DrawingAttributeSerializer.cs
@@ -1,4 +1,4 @@
-// Licensed to the .NET Foundation under one or more agreements.
+﻿// Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the MIT license.
 
 //#define OLD_ISF
@@ -608,7 +608,7 @@ namespace MS.Internal.Ink.InkSerializedFormat
                 using (MemoryStream localStream = new MemoryStream(6)) //reasonable default
                 {
                     Int32 stylusTip = Convert.ToInt32(da.StylusTip, System.Globalization.CultureInfo.InvariantCulture);
-                    System.Runtime.InteropServices.VarEnum type = SerializationHelper.ConvertToVarEnum(PersistenceTypes.StylusTip, true);
+                    System.Runtime.InteropServices.VarEnum type = SerializationHelper.ConvertToVarEnum(typeof(Int32), true);
                     ExtendedPropertySerializer.EncodeAttribute(KnownIds.StylusTip, stylusTip, type, localStream);
 
                     cbData += ExtendedPropertySerializer.EncodeAsISF(KnownIds.StylusTip, localStream.ToArray(), stream, guidList, 0, true);
@@ -680,12 +680,5 @@ namespace MS.Internal.Ink.InkSerializedFormat
 
 
         #endregion // Encoding
-
-        internal static class PersistenceTypes
-        {
-            public static readonly Type StylusTip = typeof(Int32);
-            public static readonly Type IsHollow = typeof(bool);
-            public static readonly Type StylusTipTransform = typeof(string);
-        }
     }
 }
--- a/src/Microsoft.DotNet.Wpf/src/PresentationFramework/MS/Internal/Data/ValueChangedEventManager.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationFramework/MS/Internal/Data/ValueChangedEventManager.cs
@@ -267,13 +267,13 @@ namespace MS.Internal.Data
         {
             get
             {
-                ValueChangedEventManager manager = (ValueChangedEventManager)GetCurrentManager(typeof(StaticPropertyChangedEventManager));
+                ValueChangedEventManager manager = (ValueChangedEventManager)GetCurrentManager(typeof(ValueChangedEventManager));
 
                 // at first use, create and register a new manager
                 if (manager == null)
                 {
                     manager = new ValueChangedEventManager();
-                    SetCurrentManager(typeof(StaticPropertyChangedEventManager), manager);
+                    SetCurrentManager(typeof(ValueChangedEventManager), manager);
                 }
 
                 return manager;
--- a/src/Microsoft.DotNet.Wpf/src/PresentationFramework/MS/Internal/Data/ValueChangedEventManager.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationFramework/MS/Internal/Data/ValueChangedEventManager.cs
@@ -1,4 +1,4 @@
-// Licensed to the .NET Foundation under one or more agreements.
+﻿// Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the MIT license.
 
 //
@@ -267,14 +267,13 @@ namespace MS.Internal.Data
         {
             get
             {
-                Type managerType = typeof(ValueChangedEventManager);
-                ValueChangedEventManager manager = (ValueChangedEventManager)GetCurrentManager(managerType);
+                ValueChangedEventManager manager = (ValueChangedEventManager)GetCurrentManager(typeof(StaticPropertyChangedEventManager));
 
                 // at first use, create and register a new manager
                 if (manager == null)
                 {
                     manager = new ValueChangedEventManager();
-                    SetCurrentManager(managerType, manager);
+                    SetCurrentManager(typeof(StaticPropertyChangedEventManager), manager);
                 }
 
                 return manager;
--- a/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Documents/FlowDocument.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Documents/FlowDocument.cs
@@ -1609,7 +1609,7 @@ namespace System.Windows.Documents
         {
             ArgumentNullException.ThrowIfNull(value);
 
-            if (!TextSchema.IsValidChildOfContainer(typeof(FlowDocument), childType: value.GetType()))
+            if (!TextSchema.IsValidChildOfContainer(parentType: typeof(FlowDocument), childType: value.GetType()))
             {
                 throw new ArgumentException(SR.Format(SR.TextSchema_ChildTypeIsInvalid, nameof(FlowDocument), value.GetType().Name));
             }
--- a/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Documents/FlowDocument.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Documents/FlowDocument.cs
@@ -62,8 +62,6 @@ namespace System.Windows.Documents
     [ContentProperty("Blocks")]
     public class FlowDocument : FrameworkContentElement, IDocumentPaginatorSource, IServiceProvider, IAddChild
     {
-        private static readonly Type _typeofThis = typeof(FlowDocument);
-
         //-------------------------------------------------------------------
         //
         //  Constructors
@@ -83,11 +81,11 @@ namespace System.Windows.Documents
             DependencyProperty[] typographyProperties = Typography.TypographyPropertiesList;
             for (int i = 0; i < typographyProperties.Length; i++)
             {
-                typographyProperties[i].OverrideMetadata(_typeofThis, new FrameworkPropertyMetadata(typographyChanged));
+                typographyProperties[i].OverrideMetadata(typeof(FlowDocument), new FrameworkPropertyMetadata(typographyChanged));
             }
 
-            DefaultStyleKeyProperty.OverrideMetadata(_typeofThis, new FrameworkPropertyMetadata(_typeofThis));
-            FocusableProperty.OverrideMetadata(_typeofThis, new FrameworkPropertyMetadata(true));
+            DefaultStyleKeyProperty.OverrideMetadata(typeof(FlowDocument), new FrameworkPropertyMetadata(typeof(FlowDocument)));
+            FocusableProperty.OverrideMetadata(typeof(FlowDocument), new FrameworkPropertyMetadata(true));
 
             ControlsTraceLogger.AddControl(TelemetryControls.FlowDocument);
         }
@@ -195,8 +193,7 @@ namespace System.Windows.Documents
         /// <summary>
         /// DependencyProperty for <see cref="FontFamily" /> property.
         /// </summary>
-        public static readonly DependencyProperty FontFamilyProperty =
-                TextElement.FontFamilyProperty.AddOwner(_typeofThis);
+        public static readonly DependencyProperty FontFamilyProperty = TextElement.FontFamilyProperty.AddOwner(typeof(FlowDocument));
 
         /// <summary>
         /// The FontFamily property specifies the font family.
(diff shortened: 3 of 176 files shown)


miloush:
For the benefit of others, is there a place where such changes in best patterns can be followed?

src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Documents/FlowDocument.cs
```diff
@@ -1624,9 +1609,9 @@ void IAddChild.AddChild(Object value)
         {
             ArgumentNullException.ThrowIfNull(value);
 
-            if (!TextSchema.IsValidChildOfContainer(/*parentType:*/_typeofThis, /*childType:*/value.GetType()))
+            if (!TextSchema.IsValidChildOfContainer(typeof(FlowDocument), childType: value.GetType()))
```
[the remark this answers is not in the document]
h3xds1nz: Will fix, thanks.

src/Microsoft.DotNet.Wpf/src/PresentationFramework/MS/Internal/Data/ValueChangedEventManager.cs
```diff
@@ -267,14 +267,13 @@ private static ValueChangedEventManager CurrentManager
...
-                Type managerType = typeof(ValueChangedEventManager);
-                ValueChangedEventManager manager = (ValueChangedEventManager)GetCurrentManager(managerType);
+                ValueChangedEventManager manager = (ValueChangedEventManager)GetCurrentManager(typeof(StaticPropertyChangedEventManager));
 
                 // at first use, create and register a new manager
                 if (manager == null)
                 {
                     manager = new ValueChangedEventManager();
-                    SetCurrentManager(managerType, manager);
+                    SetCurrentManager(typeof(StaticPropertyChangedEventManager), manager);
```
miloush, line 276: aren't these two supposed to be ValueChangedEventManager?

h3xds1nz:
> For the benefit of others, is there a place where such changes in best patterns can be followed?

I don't think this was mentioned anywhere per-say (certainly isn't in Stephen's blog post, checked it now for you), so unless you've been following or you're aware of the frozen object heap and the improvements around it (e.g. via runtime PRs and the epic in https://github.com/dotnet/runtime/issues/76151), chances are you will not know about this.

src/Microsoft.DotNet.Wpf/src/PresentationFramework/MS/Internal/Data/ValueChangedEventManager.cs
h3xds1nz: Thank you, that's a pasta oversight.

src/Microsoft.DotNet.Wpf/src/PresentationCore/MS/internal/Ink/InkSerializedFormat/DrawingAttributeSerializer.cs
```diff
@@ -608,7 +608,7 @@ private static void PersistStylusTip(DrawingAttributes da, Stream stream, GuidLi
                 using (MemoryStream localStream = new MemoryStream(6)) //reasonable default
                 {
                     Int32 stylusTip = Convert.ToInt32(da.StylusTip, System.Globalization.CultureInfo.InvariantCulture);
-                    System.Runtime.InteropServices.VarEnum type = SerializationHelper.ConvertToVarEnum(PersistenceTypes.StylusTip, true);
+                    System.Runtime.InteropServices.VarEnum type = SerializationHelper.ConvertToVarEnum(typeof(Int32), true);
```
miloush, line 611: surprised you didn't go for typeof(int) seeing the others like bool and object use keywords

h3xds1nz: I didn't do it because of the variable above, once we're changing the struct types to keywords as a whole I plan to fix that.

miloush:
So, when the pattern is you store typeof(X) in a local variable and then only read it couple of times in the same scope, does this not get optimized away the same way as using typeof(X) directly?

Because if it does, I am mildly unconvinced the changes are worth the benefit, for couple of readability and maintainability reasons: 1) when reading the code, you have to keep checking that the typeof(X) has the correct X in the context 2) you can't see what the type is supposed to be (i.e. the owner type) 3) not that I expect any such development anytime soon, but when writing new controls that have the same DP overrides, you can no longer simply copy and paste the metadata override lines 4) increased risk of introducing copy paste errors, as you already run into yourself.

Don't mind merging it in, just pointing out there are costs to it. Given this is mostly static constructors, the optimization does not even scale with the number of elements in the UI.

Otherwise I reviewed the two manual commits and it looks OK.

h3xds1nz:
Exactly, especially because they're static constructors, this cannot even get tiered as they only get executed once and that's tier0 business.

Here's a sample output. As for the stack loads (this is especially in the Ribbon area), you allocate 8 more bytes on stack, and then load from the stack instead of using an immediate. Which will still be in cache so it doesn't matter much. 

So strictly speaking, when using it on stack only in the static ctors in the non-generated code, it wouldn't wreck my heart to not do it though I can hardly see any development ever being done in the Ribbon area ever, let's be real.

Lastly as a personal preference, I feel using `typeof` inline even in the methods is just more expressive (though yes, as everything, a bit more error phone). But yeah, personal preferences about readibility haha.

```csharp
internal class Test
{
    static readonly DependencyProperty KittyProperty;
    static readonly DependencyProperty KittyProperty2;
    static readonly DependencyProperty KittyProperty3;
    static readonly DependencyProperty KittyProperty4;
    static readonly DependencyProperty KittyProperty5;

    private static readonly Type typeOfHello = typeof(Nullable<>);

    static Test()
    {
        KittyProperty = DependencyProperty.Register(nameof(KittyProperty), typeOfHello, typeof(Test));
        KittyProperty2 = DependencyProperty.Register(nameof(KittyProperty2), typeOfHello, typeof(Test));
        KittyProperty3 = DependencyProperty.Register(nameof(KittyProperty3), typeOfHello, typeof(Test));
        KittyProperty4 = DependencyProperty.Register(nameof(KittyProperty4), typeOfHello, typeof(Test));
        KittyProperty5 = DependencyProperty.Register(nameof(KittyProperty5), typeOfHello, typeof(Test));
    }
}
```
Using `private static readonly Type typeOfHello = typeof(Nullable<>);`:
```asm
G_M000_IG01:                ;; offset=0x0000
       push     rbp
       sub      rsp, 32
       lea      rbp, [rsp+0x20]
 
G_M000_IG02:                ;; offset=0x000A
       mov      rax, 0x272DBFC3AF0
       mov      rcx, 0x23247409E68
       mov      gword ptr [rcx], rax
       mov      rax, 0x23247409E68
       mov      rdx, gword ptr [rax]
       mov      rcx, 0x272DBFEA588
       mov      r8, 0x272DBFEA5B8
       call     [System.Windows.DependencyProperty:Register(System.String,System.Type,System.Type):System.Windows.DependencyProperty]
       mov      rcx, 0x23247409E40
       mov      rdx, rax
       call     CORINFO_HELP_ASSIGN_REF
       mov      rax, 0x23247409E68
       mov      rdx, gword ptr [rax]
       mov      rcx, 0x272DBFEA5E0
       mov      r8, 0x272DBFEA5B8
       call     [System.Windows.DependencyProperty:Register(System.String,System.Type,System.Type):System.Windows.DependencyProperty]
       mov      rcx, 0x23247409E48
       mov      rdx, rax
       call     CORINFO_HELP_ASSIGN_REF
       mov      rax, 0x23247409E68
       mov      rdx, gword ptr [rax]
       mov      rcx, 0x272DBFEA618
       mov      r8, 0x272DBFEA5B8
       call     [System.Windows.DependencyProperty:Register(System.String,System.Type,System.Type):System.Windows.DependencyProperty]
       mov      rcx, 0x23247409E50
       mov      rdx, rax
       call     CORINFO_HELP_ASSIGN_REF
       mov      rax, 0x23247409E68
       mov      rdx, gword ptr [rax]
       mov      rcx, 0x272DBFEA650
       mov      r8, 0x272DBFEA5B8
       call     [System.Windows.DependencyProperty:Register(System.String,System.Type,System.Type):System.Windows.DependencyProperty]
       mov      rcx, 0x23247409E58
       mov      rdx, rax
       call     CORINFO_HELP_ASSIGN_REF
       mov      rax, 0x23247409E68
       mov      rdx, gword ptr [rax]
       mov      rcx, 0x272DBFEA688
       mov      r8, 0x272DBFEA5B8
       call     [System.Windows.DependencyProperty:Register(System.String,System.Type,System.Type):System.Windows.DependencyProperty]
       mov      rcx, 0x23247409E60
       mov      rdx, rax
       call     CORINFO_HELP_ASSIGN_REF
       nop      
 
G_M000_IG03:                ;; offset=0x013F
       add      rsp, 32
       pop      rbp
       ret      
 
; Total bytes of code 325
```

While using `typeof(Nullable<>)` in-line as a parameter.
```assembly
G_M000_IG01:                ;; offset=0x0000
       push     rbp
       sub      rsp, 32
       lea      rbp, [rsp+0x20]
 
G_M000_IG02:                ;; offset=0x000A
       mov      rcx, 0x2F32F48A560
       mov      rdx, 0x2F32F463AF0
       mov      r8, 0x2F32F48A590
       call     [System.Windows.DependencyProperty:Register(System.String,System.Type,System.Type):System.Windows.DependencyProperty]
       mov      rcx, 0x2B29A809E08
       mov      rdx, rax
       call     CORINFO_HELP_ASSIGN_REF
       mov      rcx, 0x2F32F48A5B8
       mov      rdx, 0x2F32F463AF0
       mov      r8, 0x2F32F48A590
       call     [System.Windows.DependencyProperty:Register(System.String,System.Type,System.Type):System.Windows.DependencyProperty]
       mov      rcx, 0x2B29A809E10
       mov      rdx, rax
       call     CORINFO_HELP_ASSIGN_REF
       mov      rcx, 0x2F32F48A5F0
       mov      rdx, 0x2F32F463AF0
       mov      r8, 0x2F32F48A590
       call     [System.Windows.DependencyProperty:Register(System.String,System.Type,System.Type):System.Windows.DependencyProperty]
       mov      rcx, 0x2B29A809E18
       mov      rdx, rax
       call     CORINFO_HELP_ASSIGN_REF
       mov      rcx, 0x2F32F48A628
       mov      rdx, 0x2F32F463AF0
       mov      r8, 0x2F32F48A590
       call     [System.Windows.DependencyProperty:Register(System.String,System.Type,System.Type):System.Windows.DependencyProperty]
       mov      rcx, 0x2B29A809E20
       mov      rdx, rax
       call     CORINFO_HELP_ASSIGN_REF
       mov      rcx, 0x2F32F48A660
       mov      rdx, 0x2F32F463AF0
       mov      r8, 0x2F32F48A590
       call     [System.Windows.DependencyProperty:Register(System.String,System.Type,System.Type):System.Windows.DependencyProperty]
       mov      rcx, 0x2B29A809E28
       mov      rdx, rax
       call     CORINFO_HELP_ASSIGN_REF
       nop      
 
G_M000_IG03:                ;; offset=0x0119
       add      rsp, 32
       pop      rbp
       ret      
 
; Total bytes of code 287
```

oysteinkrog:
> Risk: Low / Regression: No

Hi — apologies for the late drive-by. Bumping into a real regression in this PR (commit 0e965e7 specifically) and wanted to flag it before it propagates further.

## The issue

In `CommandManager.cs`, the inlined `typeof()` for `RequerySuggestedEventManager.CurrentManager` is the **wrong class**:

```csharp
// Before:
Type managerType = typeof(RequerySuggestedEventManager);
RequerySuggestedEventManager manager = (RequerySuggestedEventManager)GetCurrentManager(managerType);
//... new + Set use the same key
SetCurrentManager(managerType, manager);

// After (this PR):
RequerySuggestedEventManager manager =
    (RequerySuggestedEventManager)GetCurrentManager(typeof(CanExecuteChangedEventManager));
//...
SetCurrentManager(typeof(CanExecuteChangedEventManager), manager);
```

`CanExecuteChangedEventManager.CurrentManager` legitimately uses `typeof(CanExecuteChangedEventManager)` for its own dictionary key (and was correct before, still correct after). So now **both** managers register at the same key in `WeakEventManager`'s per-thread table. Whichever runs first wins; the other's getter retrieves the foreign instance and the cast throws.

## Repro

Any `Button` (or other `ButtonBase` / `MenuItem` / `Hyperlink` / `ThumbButtonInfo`) with a `Command` binding, instantiated through XAML template parsing on a fresh thread, will eventually hit:

```
System.InvalidCastException: Unable to cast object of type
  'System.Windows.Input.CanExecuteChangedEventManager' to type
  'System.Windows.Input.CommandManager+RequerySuggestedEventManager'.
   at System.Windows.Input.CommandManager.RequerySuggestedEventManager.get_CurrentManager()
   at System.Windows.Input.CommandManager.add_RequerySuggested(EventHandler value)
   at <ICommand-derived class>.add_CanExecuteChanged(EventHandler value)
   at System.Windows.Input.CanExecuteChangedEventManager.HandlerSink..ctor(...)
   at System.Windows.Input.CanExecuteChangedEventManager.PrivateAddHandler(...)
   at System.Windows.Input.CanExecuteChangedEventManager.AddHandler(ICommand, EventHandler<EventArgs>)
   at System.Windows.Controls.Primitives.ButtonBase.HookCommand(ICommand)
   at System.Windows.Controls.Primitives.ButtonBase.OnCommandChanged(...)
```

It's race-dependent on which manager initializes first, so it surfaces unevenly across builds — JIT-only and crossgen2/R2R can produce different static-init orderings, which is why this might not have shown up in local testing. The R2R'd path I hit consistently lands the `CanExecuteChangedEventManager` instance first, then explodes the moment any commanded button gets templated during startup.

## Suggested fix

Restore `RequerySuggestedEventManager`'s key to `typeof(RequerySuggestedEventManager)` (its own type). Inlined or via `managerType` local — either is fine. The `CanExecuteChangedEventManager.cs` change in this PR is correct as-is and doesn't need to revert.

```csharp
private static RequerySuggestedEventManager CurrentManager
{
    get
    {
        RequerySuggestedEventManager manager =
            (RequerySuggestedEventManager)GetCurrentManager(typeof(RequerySuggestedEventManager));
        if (manager == null)
        {
            manager = new RequerySuggestedEventManager();
            SetCurrentManager(typeof(RequerySuggestedEventManager), manager);
        }
        return manager;
    }
}
```

A regression test would be small: a `WeakEventManager.GetCurrentManager(typeof(RequerySuggestedEventManager))` returning a `CanExecuteChangedEventManager` is the smoking-gun assertion.

Happy to put up a PR if useful, but figured a comment on the original first since this is still open. Thanks for the cleanup work overall — the rest of the typeof inlining looks good.

h3xds1nz:
@oysteinkrog This is not the only one; there's another one in [ValueChangedEventManager.cs](https://github.com/dotnet/wpf/commit/0e965e795ee813ea17bf1bfaf3b8db9a0e0f9a8a#diff-e8ca6a47724c752790e3eaad463bad8c6db7a1dc0d8a69091439f986175e108f), though that one was fixed in https://github.com/dotnet/wpf/pull/10715/commits/dc7d82514e5e1a6985b24fd50095a8aaca4f5b8a already.

I'm not gonna bother updating this PR at the moment as it doesn't really matter since the development in this repo is dead, I'll just draft it instead now.

PR #10720, opened 2025-04-07
Adding more MessageBox buttons and result values

### LINKED ISSUE #9613: [API Proposal]: Make additional MessageBoxButton and MessagBoxResult values available
### Background and motivation

Issue #5795 suggests getting more `MessageBoxButton` and `MessageBoxResult` values available and bringing Wpf repository more in sync with what is available in WinForms repository.

It will also make it easier to migrate from WinForms to WPF that you can practically directly make the same call to show a MessageBox in WPF as in WinForms.

Both repositories are calling into the same `MessageBox` function in Windows, so it should not be a technical problem to handle this.

### API Proposal

Extend `MessageBoxButton` and `MessageBoxResult` enumerations

Add `AbortRetryIgnore`, `RetryCancel` and `CancelTryContinue` to the `MessageBoxButton` enumeration:

```diff
namespace System.Windows
{
    public enum MessageBoxButton
    {
        OK = 0,
        OKCancel = 1,
+       AbortRetryIgnore = 2,
        YesNoCancel = 3,
        YesNo = 4,
+       RetryCancel = 5,
+       CancelTryContinue = 6,
    }
}
```

Also add `Abort`, `Retry`, `Ignore`, `TryAgain` and `Continue` to the `MessageBoxResult` enumeration:

```diff
namespace System.Windows
{
    public enum MessageBoxResult
    {
        None = 0,
        OK = 1,
        Cancel = 2,
+       Abort = 3,
+       Retry = 4,
+       Ignore = 5,
        Yes = 6,
        No = 7,
+       TryAgain = 10,
+       Continue = 11,
    }
}
```

### API Usage

```csharp
// Call the MessageBox.Show with additional parameters

MessageBoxResult result = MessageBox.Show("Operation timed out. What would you like to do?", 
    "My application",
    MessageBoxButton.RetryCancel,
    MessageBoxIcon.Question,
    MessageBoxResult.Retry);
if (result == MessageBoxResult.Retry)
{
    /// Perfom code to retry operation
}
else
{
    // Log cancel request and exit
}
```

### Alternative Designs

_No response_

### Risks

In connection with the extra enumeration values, I don't see a risk, since the values already exist in the underlying Windows API (1).

1. https://learn.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-messagebox

bstordrup:
Fixes #9613 

## Description

The purpose with the pull request is to streamline the `MessageBox.Show` overloads between WinForms and WPF thus making it more intuitive and easier to migrate a WinForms applicatio to WPF.
Also, since (on Windows), both WinForms and WPF call into the same underlying Windows API functions, there is in my opinion only little reason not to streamline them.

## Customer Impact

The customer impact of not taking this fix is that a migration from WinForms to WPF becomes a bit more complex when it comes to using MessageBox.Show with more options than just Yes/No or OK/Cancel. With the extension of both `MessageBoxButton` and `MessageBoxResult` enumerations, a more generous set of options is available for the application after showing `MessageBox.Show (...)` method.

## Regression

This is no regression.

## Testing

I tested the extensions of `MessageBoxButton` and `MessageBoxResult` enumerations with a sample Wpf application targeting a locally build version of Wpf. It simply had a set of `Button` and `TextBlock` controls in `MainWindow`, and each of these buttons uses the `MessageBoxButton` enumeration value as indicated by the button title. The result of `MessageBox.Show` is then assigned to the corresponding `TextBlock` control.

## Risk

The risk of taking this fix is minimal (if not even none). The WinForms repository already contains these extensions, and they are building upon an existing Windows API.

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/MessageBox.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/MessageBox.cs
@@ -17,7 +17,7 @@ namespace System.Windows
     ///       user.
     ///    </para>
     /// </devdoc>
-    public sealed class MessageBox 
+    public sealed class MessageBox
     {
 #if NEVER
         class WindowWin32Window : IWin32Window
@@ -29,7 +29,7 @@ namespace System.Windows
             }
             IntPtr IWin32Window.Handle { get { return window.SourceWindow.Handle; } }
             IntPtr IWin32Window.Parent
-            {   
+            {
                 get
                 {
                     return ((IWin32Window)window).Parent;
@@ -41,38 +41,50 @@ namespace System.Windows
             }
         }
 #endif
-        private const int IDOK             = 1;
-        private const int IDCANCEL         = 2;
-        private const int IDABORT          = 3;
-        private const int IDRETRY          = 4;
-        private const int IDIGNORE         = 5;
-        private const int IDYES            = 6;
-        private const int IDNO             = 7;
-        private const int DEFAULT_BUTTON1  = 0x00000000;
-        private const int DEFAULT_BUTTON2  = 0x00000100;
-        private const int DEFAULT_BUTTON3  = 0x00000200;
+        private const int IDOK = 1;
+        private const int IDCANCEL = 2;
+        private const int IDABORT = 3;
+        private const int IDRETRY = 4;
+        private const int IDIGNORE = 5;
+        private const int IDYES = 6;
+        private const int IDNO = 7;
+        private const int IDTRYAGAIN = 10;
+        private const int IDCONTINUE = 11;
+        private const int DEFAULT_BUTTON1 = 0x00000000;
+        private const int DEFAULT_BUTTON2 = 0x00000100;
+        private const int DEFAULT_BUTTON3 = 0x00000200;
 
         /// <devdoc>
         ///     This constructor is private so people aren't tempted to try and create
         ///     instances of these -- they should just use the static show
         ///     methods.
         /// </devdoc>
-        private MessageBox() 
+        private MessageBox()
         {
         }
 
-        private static MessageBoxResult Win32ToMessageBoxResult(int value) 
+        private static MessageBoxResult Win32ToMessageBoxResult(int value)
         {
-            switch (value) 
+            switch (value)
             {
                 case IDOK:
                     return MessageBoxResult.OK;
                 case IDCANCEL:
                     return MessageBoxResult.Cancel;
+                case IDABORT:
+                    return MessageBoxResult.Abort;
+                case IDRETRY:
+                    return MessageBoxResult.Retry;
+                case IDIGNORE:
+                    return MessageBoxResult.Ignore;
                 case IDYES:
                     return MessageBoxResult.Yes;
                 case IDNO:
                     return MessageBoxResult.No;
+                case IDTRYAGAIN:
+                    return MessageBoxResult.TryAgain;
+                case IDCONTINUE:
+                    return MessageBoxResult.Continue;
                 default:
                     return MessageBoxResult.No;
             }
@@ -85,12 +97,12 @@ namespace System.Windows
         ///    </para>
         /// </devdoc>
         public static MessageBoxResult Show(
-            string messageBoxText, 
-            string caption, 
-            MessageBoxButton button, 
-            MessageBoxImage icon, 
-            MessageBoxResult defaultResult, 
-            MessageBoxOptions options) 
+            string messageBoxText,
+            string caption,
+            MessageBoxButton button,
+            MessageBoxImage icon,
+            MessageBoxResult defaultResult,
+            MessageBoxOptions options)
         {
             return ShowCore(IntPtr.Zero, messageBoxText, caption, button, icon, defaultResult, options);
         }
@@ -101,11 +113,11 @@ namespace System.Windows
         ///    </para>
         /// </devdoc>
         public static MessageBoxResult Show(
-            string messageBoxText, 
-            string caption, 
-            MessageBoxButton button, 
-            MessageBoxImage icon, 
-            MessageBoxResult defaultResult) 
+            string messageBoxText,
+            string caption,
+            MessageBoxButton button,
+            MessageBoxImage icon,
+            MessageBoxResult defaultResult)
         {
             return ShowCore(IntPtr.Zero, messageBoxText, caption, button, icon, defaultResult, 0);
         }
@@ -116,10 +128,10 @@ namespace System.Windows
         ///    </para>
         /// </devdoc>
         public static MessageBoxResult Show(
-            string messageBoxText, 
-            string caption, 
-            MessageBoxButton button, 
-            MessageBoxImage icon) 
+            string messageBoxText,
+            string caption,
+            MessageBoxButton button,
+            MessageBoxImage icon)
         {
             return ShowCore(IntPtr.Zero, messageBoxText, caption, button, icon, 0, 0);
         }
@@ -130,9 +142,9 @@ namespace System.Windows
         ///    </para>
         /// </devdoc>
         public static MessageBoxResult Show(
-            string messageBoxText, 
-            string caption, 
-            MessageBoxButton button) 
+            string messageBoxText,
+            string caption,
+            MessageBoxButton button)
         {
             return ShowCore(IntPtr.Zero, messageBoxText, caption, button, MessageBoxImage.None, 0, 0);
         }
@@ -142,7 +154,7 @@ namespace System.Windows
         ///       Displays a message box with specified text and caption.
         ///    </para>
         /// </devdoc>
-        public static MessageBoxResult Show(string messageBoxText, string caption) 
+        public static MessageBoxResult Show(string messageBoxText, string caption)
         {
             return ShowCore(IntPtr.Zero, messageBoxText, caption, MessageBoxButton.OK, MessageBoxImage.None, 0, 0);
         }
@@ -152,7 +164,7 @@ namespace System.Windows
         ///       Displays a message box with specified text.
         ///    </para>
         /// </devdoc>
-        public static MessageBoxResult Show(string messageBoxText) 
+        public static MessageBoxResult Show(string messageBoxText)
         {
             return ShowCore(IntPtr.Zero, messageBoxText, String.Empty, MessageBoxButton.OK, MessageBoxImage.None, 0, 0);
         }
@@ -165,9 +177,9 @@ namespace System.Windows
         ///       Displays a message box with specified text, caption, and style.
         ///    </para>
         /// </devdoc>
-        /// <ExternalAPI/> 
-        public static MessageBoxResult Show(IWin32Window owner, string messageBoxText, string caption, MessageBoxButton button, MessageBoxImage icon, 
-            MessageBoxResult defaultResult, MessageBoxOptions options) 
+        /// <ExternalAPI/>
+        public static MessageBoxResult Show(IWin32Window owner, string messageBoxText, string caption, MessageBoxButton button, MessageBoxImage icon,
+            MessageBoxResult defaultResult, MessageBoxOptions options)
         {
             return ShowCore(owner, messageBoxText, caption, button, icon, defaultResult, options);
         }
@@ -177,9 +189,9 @@ namespace System.Windows
         ///       Displays a message box with specified text, caption, and style.
         ///    </para>
         /// </devdoc>
-        /// <ExternalAPI/> 
-        public static MessageBoxResult Show(IWin32Window owner, string messageBoxText, string caption, MessageBoxButton button, MessageBoxImage icon, 
-            MessageBoxResult defaultResult) 
+        /// <ExternalAPI/>
+        public static MessageBoxResult Show(IWin32Window owner, string messageBoxText, string caption, MessageBoxButton button, MessageBoxImage icon,
+            MessageBoxResult defaultResult)
         {
             return ShowCore(owner, messageBoxText, caption, button, icon, defaultResult, 0);
         }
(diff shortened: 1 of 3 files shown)


miloush:
Why is there so much whitespace changes involved? This should be trivial to review but these changes make it unnecessary complicated.

miloush:
Or at least have them in a separate commit if intentional. I cannot easily tell whether _src/Microsoft.DotNet.Wpf/cycle-breakers/PresentationFramework/PresentationFramework.cs_ contains any changes or not.

h3xds1nz:
Learning WPF project development by pressing auto format the hard way. 😀

src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/MessageBox.cs
```diff
@@ -459,7 +459,7 @@ private static bool IsValidMessageBoxResult(MessageBoxResult value)
 
         private static bool IsValidMessageBoxOptions(MessageBoxOptions value)
         {
-            int mask = ~((int)MessageBoxOptions.ServiceNotification |
+            int  mask = ~((int)MessageBoxOptions.ServiceNotification |
```
h3xds1nz, line 462: Now there's two spaces instead of one, that's not correct.

bstordrup: Fixed, but the two spaces were in there originally :-)

h3xds1nz: Maybe next year this will be okay 😀

bstordrup:
@h3xds1nz, Added tests for exceptional paths of `MessageBox.ShowCore` method. But please note the comment on the commented out test at the bottom of the `MessageBox.Tests.cs` file.

src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/MessageBox.cs
```diff
@@ -41,38 +41,50 @@ IntPtr IWin32Window.Parent
...
+        private static MessageBoxResult Win32ToMessageBoxResult(int value)
         {
-            switch (value) 
+            switch (value)
             {
                 case IDOK:
                     return MessageBoxResult.OK;
                 case IDCANCEL:
                     return MessageBoxResult.Cancel;
+                case IDABORT:
```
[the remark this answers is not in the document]
bstordrup: Hmmm. `IDABORT` is defined in line 46.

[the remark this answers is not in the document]
h3xds1nz: Don't listen to him.

src/Microsoft.DotNet.Wpf/tests/UnitTests/PresentationFramework.Tests/System/Windows/MessageBox.Tests.cs
```diff
@@ -0,0 +1,202 @@
...
+        thread.SetApartmentState(ApartmentState.STA);
+        thread.Start();
+        thread.Join();
+    }
+
+    [Fact]
+    public void Show_WithOwner_InvalidMessageBoxResult_ThrowsInvalidEnumArgumentException()
+    {
+        // Avoid annoying "The calling thread must be STA, because many UI components require this." exception
+        Thread thread = new Thread(() =>
```
h3xds1nz, line 99: You can use [WpfFact] which will ensure STA Thread with a Dispatcher and avoid these "hacks".

src/Microsoft.DotNet.Wpf/tests/UnitTests/PresentationFramework.Tests/System/Windows/MessageBox.Tests.cs
```diff
@@ -0,0 +1,202 @@
...
+
+public sealed class MessageBoxTests
+{
+    [Fact]
+    public void Show_InvalidMessageBoxButton_ThrowsInvalidEnumArgumentException()
+    {
+        // Arrange
+        string messageBoxText = "Test Message";
+        string caption = "Test Caption";
+        MessageBoxButton invalidButton = (MessageBoxButton)999; // Invalid value
```
h3xds1nz, line 17: In general I'd go with [WpfTheory] here and using Inline-data to fill like 4-5 different invalid values, unless there's exactly 1 specific reason why something should throw, it is better.

src/Microsoft.DotNet.Wpf/tests/UnitTests/PresentationFramework.Tests/System/Windows/MessageBox.Tests.cs
```diff
@@ -0,0 +1,202 @@
...
+    public void Show_InvalidMessageBoxResult_ThrowsInvalidEnumArgumentException()
+    {
+        // Arrange
+        string messageBoxText = "Test Message";
+        string caption = "Test Caption";
+        MessageBoxButton button = MessageBoxButton.OK;
+        MessageBoxResult invalidResult = (MessageBoxResult)999; // Invalid value
+
+        // Act & Assert
+        var exception = Assert.Throws<InvalidEnumArgumentException>(() =>
```
h3xds1nz, line 35: Oddly enough that **Act & Assert** comment feels exactly what GPT would do! Which is perfectly fine, just fine observation. Also in case of `[Fact]` tests I guess you don't need to unnecessarily define the constants in variables and can just inline them.

I'm also not a fan of `var` in general and I believe it should be disallowed.

src/Microsoft.DotNet.Wpf/tests/UnitTests/PresentationFramework.Tests/System/Windows/MessageBox.Tests.cs
```diff
@@ -0,0 +1,202 @@
...
+     * Tried to add this test also to test that an ArgumentException is thrown when passing in either
+     * MessageBoxOptions.ServiceNotification or MessageBoxOptions.DefaultDesktopOnly when also passing in
+     * a Window as the owner.
+     *
+     * However, this is not possible as the MessageBox.Show method passes the Handle property to the ShowCore
+     * method, and the Handle property is null here - which in the ShowCore method results in owner being IntPtr.Zero,
+     * and the intended test for not allowing the above MessageBoxOptions fail in the test.
+     *
+     * It is likely in a real world scenario that it will work because the owner window would be an already created
+     * Window having a proper Handle set.
```
h3xds1nz, line 176: Imho this could go with `Window` -> `WindowInteropHelper.EnsureHandle` under `WpfTheory` to get a hwnd, should work.

src/Microsoft.DotNet.Wpf/tests/UnitTests/PresentationFramework.Tests/System/Windows/MessageBox.Tests.cs
bstordrup: The reason I put them in `WpfFact` is that it is different parameters to `MessageBox.Show` method, and I aim to explicitly test a specific exceptional branch for each in the `MessageBox.ShowCore` method.

h3xds1nz: I do understand and support that but you can try different invalid values for the button, right now I'd think that it only throws an exception when you fill 999. That's a **fact**.

bstordrup: I see your point, but what is the test worth if I pass in 10 different invalid values for MessageBoxButton that all results in an exception thrown, but not being able to test if a valid value for MessageBoxButton does not throw an exception?

Because if an exception is not thrown, then the message box will just show up requireing somebody to click it.

A way to test both invalid and valid methods would be to make the `IsValidMessageBoxXxxx` methods internal and then test these instead.

src/Microsoft.DotNet.Wpf/tests/UnitTests/PresentationFramework.Tests/System/Windows/MessageBox.Tests.cs
bstordrup: Have addressed this, but not yet pushed it to the branch.

src/Microsoft.DotNet.Wpf/tests/UnitTests/PresentationFramework.Tests/System/Windows/MessageBox.Tests.cs
bstordrup: @h3xds1nz, it worked ;-)

Thank you for pointing me to that.

src/Microsoft.DotNet.Wpf/tests/UnitTests/PresentationFramework.Tests/System/Windows/MessageBox.Tests.cs
h3xds1nz: There are actually multiple ways for this to be done.

You may use the window overload and then just queue async task to close the messagebox via WM, since it is easy to find it using owner hwnd+class. Since that's more of an integration though, another way is to use interceptors and mock it but since we don't use them (yet), it might be a bit of an overkill.

Hence why I suggested only testing the failure cases for now.

src/Microsoft.DotNet.Wpf/tests/UnitTests/PresentationFramework.Tests/System/Windows/MessageBox.Tests.cs
h3xds1nz: No worries :)

src/Microsoft.DotNet.Wpf/tests/UnitTests/PresentationFramework.Tests/System/Windows/MessageBox.Tests.cs
```diff
@@ -0,0 +1,165 @@
...
+            MessageBox.Show(new Window(), "Test Message", "Test Caption", invalidButton));
+        Assert.Contains("button", exception.Message);
+    }
+
+    /*
+     * MessageBoxResult range is from 0 to 11 with 8 and 9 not used.
+     * Values outside are illegal as well as values 8 and 9.
+     */
+    public static TheoryData<MessageBoxResult> InvalidMessageBoxResultData =>
+        new TheoryData<MessageBoxResult>
```
h3xds1nz, line 45: Imho in case of non-complicated data like this, it is generally preferred to just use [InlineData] attribute, use the underlying storage type (e.g. `int`) as a param  and then cast it during the method call.

It will be cleaner.

bstordrup: Done

src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/MessageBox.cs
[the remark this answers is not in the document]
bstordrup: Halucinating

src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/MessageBox.cs
```diff
@@ -41,38 +41,50 @@ IntPtr IWin32Window.Parent
...
+                    return MessageBoxResult.Abort;
+                case IDRETRY:
```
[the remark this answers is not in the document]
bstordrup: Halucinating

src/Microsoft.DotNet.Wpf/tests/UnitTests/PresentationFramework.Tests/System/Windows/MessageBox.Tests.cs
```diff
@@ -0,0 +1,179 @@
...
+using System.ComponentModel;
+using System.Windows.Interop;
+
+namespace System.Windows;
+
+public sealed class MessageBoxTests
+{
+    /*
+     * MessageBoxButton range is from 0x00000000 to 0x00000006
+     * Values outside are illegal.
```
h3xds1nz, line 13: @bstordrup Just a nit, comments in general in C# but mostly in this repo (unless in-line comment is required) are using the `//` syntax. Otherwise LGTM, no further comments.

bstordrup: Resolved

src/Microsoft.DotNet.Wpf/tests/UnitTests/PresentationFramework.Tests/System/Windows/MessageBox.Tests.cs
```diff
@@ -0,0 +1,179 @@
...
+     */
+    [WpfTheory]
+    [InlineData(0xFFFFFFFF)]
+    [InlineData(0x00000007)]
+    public void Show_InvalidMessageBoxButton_ThrowsInvalidEnumArgumentException(uint invalidButton)
+    {
+        InvalidEnumArgumentException exception = Assert.Throws<InvalidEnumArgumentException>(() =>
+            MessageBox.Show("Test Message", "Test Caption", (MessageBoxButton) invalidButton));
```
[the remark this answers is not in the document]
bstordrup: Resolved

siagupta0202:
@bstordrup Thank you for taking up this API proposal, seeing it through to completion, and also contributing the accompanying tests 😊

P.S. @h3xds1nz Thank you for your detailed and thorough review of the proposal!

bstordrup:
@siagupta0202 my pleasure 😊

h3xds1nz:
@siagupta0202 happy to :)

miloush:
Who is in charge of XML documentation of these?

dipeshmsft:
@miloush, we are waiting of this PR now : https://github.com/dotnet/dotnet-api-docs/pull/11295#pullrequestreview-2840237795

Once this is done, I will make the updates.

Issue #10721, opened 2025-04-07
WindowsDesktop redistributes many libraries already carried by MSBuild


ericstj:
### Description

PresentationBuildTasks redistributes many runtime libraries:
https://github.com/dotnet/wpf/blob/72af9a57bbfb4aac9bdc554d860ab320ac73c587/src/Microsoft.DotNet.Wpf/src/PresentationBuildTasks/PresentationBuildTasks.csproj#L53-L62

All of these have a version provided by MSBuild.

WPF can reference the version provided by MSBuild and avoid shipping these.  The one tradeoff is that it won't be able to benefit from new API in these libraries right away - but that's most likely an acceptable tradeoff to avoid redistributing these libraries.

### Reproduction Steps

Examine `C:\Program Files\dotnet\sdk\10.0.100-preview.2.25164.34\Sdks\Microsoft.NET.Sdk.WindowsDesktop\tools\net472`

### Expected behavior

No binaries included that overlap with those provided by MSBuild.

### Actual behavior

Many binaries redistributed with newer versions than MSBuild.

### Regression?

No

### Known Workarounds

No

### Impact

Added complexity in servicing / size on disk

### Configuration

All

PR #10722, opened 2025-04-07
Don't redistribute inbox msbuild dependencies

ViktorHofer:
Fixes https://github.com/dotnet/wpf/issues/10721

System.CodeDom is only used on .NETCoreApp and its live version is available in the SDK directory. Therefore, we shouldn't need to pin it to an older version.

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/PresentationBuildTasks/PresentationBuildTasks.csproj
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationBuildTasks/PresentationBuildTasks.csproj
@@ -301,8 +301,8 @@
     <PackageReference Include="Microsoft.Build.Framework" Version="$(MicrosoftBuildFrameworkPackageVersion)" IncludeAssets="compile" />
     <PackageReference Include="Microsoft.Build.Utilities.Core" Version="$(MicrosoftBuildUtilitiesCorePackageVersion)" IncludeAssets="compile" />
     <PackageReference Include="System.Reflection.MetadataLoadContext" Version="$(SystemReflectionMetadataLoadContextVersion)" IncludeAssets="compile" />
-    <!-- Use minimum supported version as this component runs inside MSBuild.  -->
-    <PackageReference Include="System.CodeDom" Version="$(SystemCodeDomToolsetPackageVersion)" />
+    <!-- System.CodeDom live version is provided by the SDK. Don't redistribute it. -->
+    <PackageReference Include="System.CodeDom" Version="$(SystemCodeDomToolsetPackageVersion)" Condition="'$(TargetFrameworkIdentifier)' == '.NETCoreApp'" IncludeAssets="compile" />
   </ItemGroup>
 
 </Project>
--- a/src/Microsoft.DotNet.Wpf/src/PresentationBuildTasks/PresentationBuildTasks.csproj
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationBuildTasks/PresentationBuildTasks.csproj
@@ -300,9 +300,14 @@
   <ItemGroup>
     <PackageReference Include="Microsoft.Build.Framework" Version="$(MicrosoftBuildFrameworkPackageVersion)" IncludeAssets="compile" />
     <PackageReference Include="Microsoft.Build.Utilities.Core" Version="$(MicrosoftBuildUtilitiesCorePackageVersion)" IncludeAssets="compile" />
-    <PackageReference Include="System.Reflection.MetadataLoadContext" Version="$(SystemReflectionMetadataLoadContextVersion)" IncludeAssets="compile" />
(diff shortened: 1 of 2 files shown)


src/Microsoft.DotNet.Wpf/src/PresentationBuildTasks/PresentationBuildTasks.csproj
```diff
@@ -308,14 +298,11 @@
...
-    <PackageReference Include="Microsoft.Build.Utilities.Core" Version="$(MicrosoftBuildUtilitiesCorePackageVersion)" />
-    <PackageReference Include="System.Reflection.MetadataLoadContext" Version="$(SystemReflectionMetadataLoadContextVersion)" />
-
-    <!--
-      Provide specific/old versions for PresentationBuildTasks which needs to run inside MSBuild
-    -->
-    <PackageReference Include="System.CodeDom" Version="$(SystemCodeDomPackageVersionForPresentationBuildTasks)" />
+    <PackageReference Include="Microsoft.Build.Framework" Version="$(MicrosoftBuildFrameworkPackageVersion)" IncludeAssets="compile" />
+    <PackageReference Include="Microsoft.Build.Utilities.Core" Version="$(MicrosoftBuildUtilitiesCorePackageVersion)" IncludeAssets="compile" />
+    <PackageReference Include="System.Reflection.MetadataLoadContext" Version="$(SystemReflectionMetadataLoadContextVersion)" IncludeAssets="compile" />
```
ericstj, line 303: This needs to be a toolset version

src/Microsoft.DotNet.Wpf/src/PresentationBuildTasks/PresentationBuildTasks.csproj
```diff
@@ -308,14 +298,16 @@
...
+    <PackageReference Include="Microsoft.Build.Framework" Version="$(MicrosoftBuildFrameworkPackageVersion)" IncludeAssets="compile" />
+    <PackageReference Include="Microsoft.Build.Utilities.Core" Version="$(MicrosoftBuildUtilitiesCorePackageVersion)" IncludeAssets="compile" />
 
-    <!--
-      Provide specific/old versions for PresentationBuildTasks which needs to run inside MSBuild
-    -->
-    <PackageReference Include="System.CodeDom" Version="$(SystemCodeDomPackageVersionForPresentationBuildTasks)" />
+    <!-- Reference (don't redistribute) the live version of SRM on .NETCoreApp and fallback to the toolset version for .NET Framework. -->
+    <PackageReference Include="System.Reflection.MetadataLoadContext" IncludeAssets="compile">
+      <Version Condition="'$(TargetFrameworkIdentifier)' != '.NETFramework'">$(SystemReflectionMetadataLoadContextVersion)</Version>
```
ericstj, line 306: You could use the toolset version in both cases.   I suspect it needs to be available for source build regardless.

src/Microsoft.DotNet.Wpf/src/PresentationBuildTasks/PresentationBuildTasks.csproj
```diff
@@ -308,14 +298,16 @@
   </ItemGroup>
```
ericstj, line 298:
```suggestion
    <Reference Include="System.CodeDom" />
  </ItemGroup>
```

Not sure if you need this or not?

ericstj (approved):

src/Microsoft.DotNet.Wpf/src/PresentationBuildTasks/PresentationBuildTasks.csproj
ViktorHofer: Not needed on netfx

src/Microsoft.DotNet.Wpf/src/PresentationBuildTasks/PresentationBuildTasks.csproj
ViktorHofer: I'm following the dotnet/sdk pattern which uses live versions by default (via CPM) and downgrade only when targeting .NET Framework. As this repo doesn't use CPM, the pattern here looks differently.

> I suspect it needs to be available for source build regardless.

The windows UI stack isn't part of source-build.

ViktorHofer:
@ericstj I had to push another commit because of a typo. Can you please approve again?

src/Microsoft.DotNet.Wpf/src/PresentationBuildTasks/PresentationBuildTasks.csproj
ericstj: Sorry, I was just trying to guess at why you might have wanted to reference the latest on core.

The SDK actually needs to distribute live versions itself, while tasks could stay on older versions.

ericstj (approved):

mmitche (approved):

PR #10732, opened 2025-04-07
Fix DataObject SetData methods inaccordance to previous implementation

harshit7962:
## Description
These changes are in accordance to the original API contract of certain `SetData` methods in `DataObject` class. The original behaviour was altered a bit in form of exceptions not being thrown as part of the shared clipboard code for winforms and wpf as in the PR #10544. These changes brings back the expected behaviour that respects WPFs original implementation.
 

## Customer Impact
Avoids breaking changes in subsequent preview release of .NET 10.

## Regression
Yes

## Testing
Local Build Pass, unit testing and verified the same with sample applications.

## Risk
Low

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/wpf/pull/10732)

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/dataobject.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/dataobject.cs
@@ -225,19 +225,30 @@ public sealed unsafe partial class DataObject :
     /// <summary>
     ///  Stores the specified data and its associated format in this instance.
     /// </summary>
-    public void SetData(string format, object? data) => _innerData.SetData(format, data);
+    public void SetData(string format, object? data)
+    {
+        ArgumentNullException.ThrowIfNull(data, nameof(data));
+        _innerData.SetData(format, data);
+    }
 
     /// <summary>
     ///  Stores the specified data and its associated class type in this instance.
     /// </summary>
-    public void SetData(Type format, object? data) => _innerData.SetData(format, data);
+    public void SetData(Type format, object? data)
+    {
+        ArgumentNullException.ThrowIfNull(data, nameof(data));
+        _innerData.SetData(format, data);
+    }
 
     /// <summary>
     ///  Stores the specified data and its associated format in this instance, using the automatic conversion parameter
     ///  to specify whether the data can be converted to another format.
     /// </summary>
-    public void SetData(string format, object? data, bool autoConvert) =>
+    public void SetData(string format, object? data, bool autoConvert)
+    {
+        ArgumentNullException.ThrowIfNull(data, nameof(data));
         _innerData.SetData(format, autoConvert, data);
+    }
 
     // WinForms and WPF have these defined in a different order.
     void IDataObjectInternal.SetData(string format, bool autoConvert, object? data) => SetData(format, data, autoConvert);
--- a/src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/dataobject.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/dataobject.cs
@@ -227,7 +227,8 @@ public sealed unsafe partial class DataObject :
     /// </summary>
     public void SetData(string format, object? data)
     {
-        ArgumentNullException.ThrowIfNull(data, nameof(data));
+        ArgumentNullException.ThrowIfNull(data);
+
         _innerData.SetData(format, data);
     }
 
@@ -236,7 +237,8 @@ public sealed unsafe partial class DataObject :
     /// </summary>
     public void SetData(Type format, object? data)
     {
-        ArgumentNullException.ThrowIfNull(data, nameof(data));
+        ArgumentNullException.ThrowIfNull(data);
+
         _innerData.SetData(format, data);
     }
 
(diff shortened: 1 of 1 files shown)


src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/dataobject.cs
```diff
@@ -225,19 +225,30 @@ private bool TryGetDataInternal<T>(
     /// <summary>
     ///  Stores the specified data and its associated format in this instance.
     /// </summary>
-    public void SetData(string format, object? data) => _innerData.SetData(format, data);
+    public void SetData(string format, object? data)
```
h3xds1nz, line 228: Eh, I was hoping this would indeed be removed as a "sort-of" breaking change, it always felt weird in WPF.

Anyways, in that case, `data` cannot be marked as nullable, that applies to all methods.

And I think in general you'd want an empty line between a throw method and the next statement.

harshit7962: @h3xds1nz it seems that we can proceed with the WinForms' behavior and document this as a behavior change. However, the intention here is to maintain as much backward compatibility as possible, given that the upside is not very significant. Please let me know if you have a different opinion on this.

Will go ahead and change the later mentioned comments.

h3xds1nz: @harshit7962 There's currently a note in the tests (written by Jeremy it seems) .

"Setting null in WinForms is allowed, but really should be blocked."

So I guess my point of view it depends on what WF will do. The unified behaviour is what matters the most imho, after all, we're sharing the implementation now. This change (not throw where you previously would) cannot really break anyone imho, you should have been checking the data you put in, at least all my Clipboard calls always do in WPF.

Personally, I always told myself "why can't you just let me put `null`", I want to clear the clipboard but why do I need to explicitly call `Clear` to do it. Sure, it may present subtle bugs (where you're filling `null` and you don't realize I guess) but imho it was just inconvenience over anything.

JeremyKuhne: @harshit7962 @h3xds1nz If we just internally called Clear then that would be different, but we don't. 

> it may present subtle bugs

That's the big worry with allowing it, as well as changing behavior. I do want to see it blocked in WinForms as well, but with the scope of the other changes this time I felt it was better to do another "consistency pass" in .NET 11.

h3xds1nz: Yeah alright, if that's the case, then it is fine to keep WPF as it was, throwing on `null`.

src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/dataobject.cs
```diff
@@ -225,19 +225,33 @@ private bool TryGetDataInternal<T>(
...
+    {
+        ArgumentNullException.ThrowIfNull(data, nameof(data));
```
pchaurasia14, line 230: `nameof(data)` can be omitted.

src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/dataobject.cs
```diff
@@ -225,19 +225,33 @@ private bool TryGetDataInternal<T>(
...
+        _innerData.SetData(format, data);
+    }
 
     /// <summary>
     ///  Stores the specified data and its associated class type in this instance.
     /// </summary>
-    public void SetData(Type format, object? data) => _innerData.SetData(format, data);
+    public void SetData(Type format, object? data)
+    {
+        ArgumentNullException.ThrowIfNull(data, nameof(data));
```
pchaurasia14, line 240: `nameof(data)` can be omitted.

src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/dataobject.cs
```diff
@@ -225,19 +225,33 @@ private bool TryGetDataInternal<T>(
...
+    }
 
     /// <summary>
     ///  Stores the specified data and its associated format in this instance, using the automatic conversion parameter
     ///  to specify whether the data can be converted to another format.
     /// </summary>
-    public void SetData(string format, object? data, bool autoConvert) =>
+    public void SetData(string format, object? data, bool autoConvert)
+    {
+        ArgumentNullException.ThrowIfNull(data, nameof(data));
```
pchaurasia14, line 251: `nameof(data)` can be omitted.

pchaurasia14 (changes requested):

src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/dataobject.cs
harshit7962: Addressed.

pchaurasia14 (approved):

h3xds1nz:
I'll just say, outside a review, that with the decision to keep throwing `null`, currently it's really unfortunate that the `data` param is annotated as nullable because of the parent implementation but we actually throw on `null`.

Like the if I'm new to the API and I see a nullable param, I'm happy to pass `null` just to find out the hard way. This should be solved before release of .NET 10 because it completely eliminates the benefit of nullable reference annotations.

harshit7962:
@h3xds1nz yeah that's true. Adding the nullability kind of decreases the readability of the method, but for now this is probably the cost of having a shared implementation. We will try to work it out when WF gets the similar behaviour.

PR #10762, opened 2025-04-12
 [StyleCleanUp] Fix MilCodeGen empty lines

ThomasGoulet73:
## Description
Fixes empty lines added by MilCodeGen.

This PR also syncs MilCodeGen with the empty line changes from dotnet/wpf#10021.

## Customer Impact
None, whitespace changes only.

## Regression
No.

## Testing
Local build + ran MilCodeGen.

## Risk
None, whitespace changes only.

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/WpfGfx/codegen/mcg/generators/ManagedResource.cs
+++ b/src/Microsoft.DotNet.Wpf/src/WpfGfx/codegen/mcg/generators/ManagedResource.cs
@@ -159,13 +159,35 @@ namespace MS.Internal.MilCodeGen.Generators
                                 );
                     }
 
-                    csFile.WriteBlock(
+                    csFile.Write(
                         [[inline]]
 
                             namespace [[resource.ManagedNamespace]]
                             {
-                                [[Helpers.CollectionHelper.WriteCollectionSummary(resource)]]
-                                [[attributes]]
+                        [[/inline]]
+                    );
+
+                    string collectionSummary = Helpers.CollectionHelper.WriteCollectionSummary(resource);
+                    if (!string.IsNullOrEmpty(collectionSummary))
+                    {
+                        csFile.Write(
+                            [[inline]]
+                                    [[collectionSummary]]
+                            [[/inline]]
+                            );
+                    }
+
+                    if (!string.IsNullOrEmpty(attributes))
+                    {
+                        csFile.Write(
+                            [[inline]]
+                                    [[attributes]]
+                            [[/inline]]
+                            );
+                    }
+
+                    csFile.WriteBlock(
+                        [[inline]]
                                 [[WriteClassDeclaration(resource.Name, !resource.IsValueType, modifiers, extends)]]
                                 {
                                     [[Helpers.ManagedStyle.WriteSection("Public Methods")]]
--- a/src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Generated/TextDecorationCollection.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Generated/TextDecorationCollection.cs
@@ -18,7 +18,6 @@ namespace System.Windows
     /// <summary>
     /// A collection of TextDecoration objects.
     /// </summary>
-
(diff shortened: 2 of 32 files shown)


src/Microsoft.DotNet.Wpf/src/WpfGfx/codegen/mcg/generators/ManagedResource.cs
```diff
@@ -159,13 +159,35 @@ public override void Go()
...
 
                             namespace [[resource.ManagedNamespace]]
                             {
-                                [[Helpers.CollectionHelper.WriteCollectionSummary(resource)]]
-                                [[attributes]]
+                        [[/inline]]
+                    );
+
+                    string collectionSummary = Helpers.CollectionHelper.WriteCollectionSummary(resource);
+                    if (!string.IsNullOrEmpty(collectionSummary))
```
h3xds1nz, line 171: Imho this does only complicate the templating files and it is not a generic solution that could aid with tons of other new lines that are currently present (e.g. empty ctors, conditionally generated methods, etc.)

There's at least 10 empty lines in collection files that don't have to be there. And then there are other templates.

A more generic-based approach by e.g. returning `NULL` from generator methods instead of `string.Empty` when no content is produced / having the attributes variable be `NULL`, provided to the code sink which would then propagate NULLs as "don't insert `\n`" is a way cleaner solution that also wouldn't require any templating changes, keeping the whole thing more readable.

dipeshmsft: @h3xds1nz although I get your point, but this seems a reasonable solution for now. 
@ThomasGoulet73 comments ?

h3xds1nz: @dipeshmsft As I've said in https://github.com/dotnet/wpf/pull/10702#discussion_r2029407692 I'm happy to do it in case it is not within Thomas's time constraints currently.

As I've mentioned, there are other (and more overall) empty lines that are being generated with the same root cause.

dipeshmsft: > As I've mentioned, there are other (and more overall) empty lines that are being generated with the same root cause.

Yeah, there are around 4-5 or more similar fixes that will be needed for this.

ThomasGoulet73: Sorry, I didn't see @h3xds1nz's comment.

I had the same tought as @h3xds1nz and had a quick look if it would be feasible to skip lines with [[XYZ]] parameters but it wasn't really straightforward and I don't plan on doing it right now. This PR is mainly to sync the generated code with the code generators because currently running MilCodeGen unmodified on main creates diffs.

Also, the Csp interpreter seem like an "obsolete" tool for code generation IMO. I'd much rather we invest on replacing it with a Roslyn-based generator (Whether source generator or custom runnable project) instead of improving Csp. The Roslyn-based generator would probably use StringBuilders for generating the code so we would still need to move conditionally generated code to separate if blocks.

All that to say that I think moving conditional generated code to a manual if block isn't a big problem. It does make the code slightly more complex but it avoids having to improve Csp and maybe help in the future if/when we move to a Roslyn-based generator.

dipeshmsft: Thanks for the insights @ThomasGoulet73. Given the information IMO, we should move forward with the PR. @h3xds1nz  if you are planning to work on fixing Csp, and if it is straigtforward enough, I am more than willing to take it and review the PR.

h3xds1nz: @dipeshmsft https://github.com/h3xds1nz/wpf/commit/e0a49c5f454b0aa3e041b73ae4c871d58fb2d505

dipeshmsft: @h3xds1nz it looks promising, can you create a PR for the same.

PR #10789, opened 2025-04-21
Removing unsupported templates

pchaurasia14:
Removes unsupported templates (netcoreapp3.0, netcoreapp3.1, net5.0, net7.0)

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/wpf/pull/10789)


singhashish-wpf (approved):

h3xds1nz:
Why are we then leaving .NET 6 in when its out of support for 6 months already?

pchaurasia14:
> Why are we then leaving .NET 6 in when its out of support for 6 months already?

We're keeping them for now until they go out-of-support (similar to other .NET repos like Winforms)

PR #10845, opened 2025-05-12
Minimal support for `ObservableCollection<T>` range actions

### LINKED ISSUE #52: Improved INotifyCollectionChanged Handling

Sergio0694:
Fixes #52
Supersedes #9568, #6097

## Description

This PR is needed to unblock #https://github.com/dotnet/runtime/issues/18087.
To keep work/risk to a minimum, **this PR only unblocks WPF by avoiding crashes**.
That is, it treats all range operations as collection resets, and nothing else.
This can be further improved later, but it's sufficient to at least unblock upstream .NET work.

## Customer Impact

WPF apps with code using any new range APIs (https://github.com/dotnet/runtime/issues/18087) will not crash.

## Regression

No.

## Testing

I'll need help with this.

## Risk

Minimal. Changes are minimal and scoped, and only affect code paths that were previously throwing already.


miloush:
Looks like a reasonable compromise to me. Did you test it gets you unblocked?

Sergio0694:
@miloush I'm not really familiar with the WPF test infrastructure, I was hoping for some guidance/help on that part 😅

src/Microsoft.DotNet.Wpf/src/PresentationFramework/MS/Internal/Data/CompositeCollectionView.cs
```diff
@@ -1459,6 +1466,15 @@ private void ValidateCollectionChangedEventArgs(NotifyCollectionChangedEventArgs
...
+        private bool IsCollectionChangedRangeAction(NotifyCollectionChangedEventArgs e)
+        {
+            return
+                (e.Action == NotifyCollectionChangedAction.Add && e.NewItems.Count > 1) ||
+                (e.Action == NotifyCollectionChangedAction.Remove && e.OldItems.Count > 1) ||
+                (e.Action == NotifyCollectionChangedAction.Replace && (e.OldItems.Count > 1 || e.NewItems.Count > 1)) ||
+                (e.Action == NotifyCollectionChangedAction.Move && (e.OldItems.Count > 1 || e.NewItems.Count > 1));
+        }
+
         /// <summary>
```
Symbai, line 1478: You are creating 4x the same method. Cant this be made static and moved into a helper class or something like that instead of duplicating code all over?

h3xds1nz: All it needs is `private protected` on `CollectionView` as its a base class for all of them.

h3xds1nz:
@Sergio0694 I think its best if you get in touch with the team internally, a lot of the tests aren't in this repo, especially the functional ones. There are also CV tests in https://github.com/dotnet/wpf-test but afaik that's not everything.

michael-hawker:
Thanks @h3xds1nz for the heads up, we've reached out to them directly too! Just bootstrapping the process here with the PR to lead the conversation in the open.

PR #10856, opened 2025-05-18
[MilCodeGen] Allow emitting empty-lines via NULL, including conditionals

h3xds1nz:
Supersedes #10762 

## Description

Removes most of the empty lines that are present due to generating methods returning `string.Empty` as they're not applicable and interpreter then just emitting empty lines by simply using a magic string for `null` in the parser.

The added whitespaces are mostly wanted (e.g. newline after '}' is something that's enforced by StyleCop).

The only changes that matter (all those files are at the bottom):

- This allows for the NULL support
     - CsPrimeRuntime
     - CsPrimeParser
     - CodeSink

- These are the templating changes (note that "WriteParse" has no real changes, its just a whitespace move).
     - CollectionHelper
     - CodeGenHelpers
     - ManagedStruct
     - ManagedResource

## Customer Impact

None, whitespace changes only.

## Regression

No.

## Testing

Local build.

## Risk

None, there are only whitespace changes.


h3xds1nz:
@dipeshmsft I've done mostly everything that I could find.

PR #10980, opened 2025-07-15
Removed separate MinHeight resources for text based controls

dipeshmsft:
## Description
This PR reverts a workaround that I had to introduce while fixing the layout of text controls ( DatePicker, TextBox, RichTextBox, PasswordBox, etc ). These all controls use a shared resource TextControlThemeMinHeight, which needed to be changed from 24 to 32. However, in each PR I only fixed a single style, I had to introduce individual resources for these controls at the time.

In this PR, I remove the extra resources and their references are replaced with TextControlThemeMinHeight resource.

## Customer Impact
Anyone modifying this resource will see the uniform behavior for all text based controls.

## Regression
N/A

## Testing
Local app testing

## Risk
Minimal. Minor issues in content alignment may come up.

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/wpf/pull/10980)

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/Themes/PresentationFramework.Fluent/Styles/RichTextBox.xaml
+++ b/src/Microsoft.DotNet.Wpf/src/Themes/PresentationFramework.Fluent/Styles/RichTextBox.xaml
@@ -12,10 +12,6 @@
     
     <!-- Deprecated RichTextBox Resources ( Used in .NET 9 ) -->
     <Thickness x:Key="RichTextBoxAccentBorderThemeThickness">0,0,0,1</Thickness>
-    
-    <!-- Instead of RichTextBoxThemeMinHeight, we need to use TextControlThemeMinHeight, but modifying that right now
-            will affect the other styles as well. Once the remaining styles using it are fixed, will remove this resource -->
-    <system:Double x:Key="RichTextBoxThemeMinHeight">32</system:Double>
 
     <Style x:Key="DefaultRichTextBoxStyle" TargetType="{x:Type RichTextBox}">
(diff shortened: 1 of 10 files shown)


src/Microsoft.DotNet.Wpf/src/Themes/PresentationFramework.Fluent/Styles/RichTextBox.xaml
```diff
@@ -30,7 +26,7 @@
         <Setter Property="HorizontalContentAlignment" Value="Left" />
         <Setter Property="VerticalContentAlignment" Value="Top" />
         <Setter Property="Cursor" Value="IBeam" />
-        <Setter Property="MinHeight" Value="{DynamicResource RichTextBoxThemeMinHeight}" />
```
harshit7962, line 33: Why are we doing this change? Basically, the granularity of the min height in controls is moving to a common resource. Isn't it better to have this resource (since it is present anyways) used specific to control?

harshit7962 (approved):
Apart from the sole query, the changes looks fine.

src/Microsoft.DotNet.Wpf/src/Themes/PresentationFramework.Fluent/Styles/RichTextBox.xaml
dipeshmsft: In WinUI, for all the text based controls we have a single resource and I am trying to maintain that behavior. 
So that, modifying one resource uniformly modifies the behavior at all places.

PR #10998, opened 2025-07-22
Use Environment.ProcessPath instead of Process.GetCurrentProcess().Pr…

NeverMorewd:
Fixes # 

Main PR 

## Description

Replace Process.GetCurrentProcess().ProcessName with Path.GetFileNameWithoutExtension(Environment.ProcessPath) for better performance and cleaner code.

## Customer Impact

- Reduced system call overhead
- No temporary Process object creation

###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/wpf/pull/10998)

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/Shared/MS/Win32/ManagedWndProcTracker.cs
+++ b/src/Microsoft.DotNet.Wpf/src/Shared/MS/Win32/ManagedWndProcTracker.cs
@@ -203,7 +203,7 @@ namespace MS.Win32
         {
             string msg = String.Format("BEGIN: {0:X} -- Setting DWP, process = {1} ({2}) {3}",
                    hwnd,
-                   System.Diagnostics.Process.GetCurrentProcess().ProcessName,
+                   Path.GetFileNameWithoutExtension(Environment.ProcessPath),
                    fromWhere,
                    System.Environment.NewLine);
 
@@ -214,7 +214,7 @@ namespace MS.Win32
         {
             string msg = String.Format("END:   {0:X} -- Setting DWP, process = {1} ({2}) {3}",
                    hwnd,
-                   System.Diagnostics.Process.GetCurrentProcess().ProcessName,
+                   Path.GetFileNameWithoutExtension(Environment.ProcessPath),
                    fromWhere,
                    System.Environment.NewLine);
 


src/Microsoft.DotNet.Wpf/src/Shared/MS/Win32/ManagedWndProcTracker.cs
```diff
@@ -203,7 +203,7 @@ internal static void LogStartHWND(IntPtr hwnd, string fromWhere)
         {
             string msg = String.Format("BEGIN: {0:X} -- Setting DWP, process = {1} ({2}) {3}",
                    hwnd,
-                   System.Diagnostics.Process.GetCurrentProcess().ProcessName,
+                   Path.GetFileNameWithoutExtension(Environment.ProcessPath),
```
lindexi, line 206: `Environment.ProcessPath` will call the `Kernel32.GetModuleFileNameW`, see https://learn.microsoft.com/en-us/windows/win32/api/libloaderapi/nf-libloaderapi-getmodulefilenamew

` System.Diagnostics.Process.GetCurrentProcess().ProcessName` will call `Kernel32.QueryFullProcessImageNameW`, see https://learn.microsoft.com/en-us/windows/win32/api/libloaderapi/nf-libloaderapi-getmodulefilenamew

I do not think this is the same behavior. But I believe it is the same behavior in the vast majority of cases.

miloush:
I agree this is a compatibility risk.

lindexi:
@miloush This is the edge module code, and it only effect the log message. So that I think it is ok.

miloush:
Which should be discussed in the issue description, together with how it was tested. You have to have a special build with LOGGING set, so the impact of this change is questionable.

lindexi:
@miloush I think it is the debug code.

jizc:
Just want to point out that the contents of the `Log` method that `msg` is sent to has been commented out.

https://github.com/dotnet/wpf/blob/725fc8adca9ba66f4318d95bacd85a8eb8645e21/src/Microsoft.DotNet.Wpf/src/Shared/MS/Win32/ManagedWndProcTracker.cs#L224-L235

src/Microsoft.DotNet.Wpf/src/Shared/MS/Win32/ManagedWndProcTracker.cs
NeverMorewd: Process.GetCurrentProcess() without `using` will left a Process instance after each calling.
In certain scenarios with high call volume, this pr will make sense.

Symbai: > But I believe it is the same behavior in the vast majority of cases.

For WPF both .NET APIs return the same result. There is no difference. Only for other .NET scenarios (not WPF related: framework dependend when running from dotnet.exe, browser etc) there is a difference between Environment.ProcessPath and ProcessName. Yes, they call different winapis but for WPF the result is the same in all cases.

miloush: One difference is when the file is renamed while the process is running. But the discussion is moot as per the PR discussion.

PR #11124, opened 2025-09-17
Pass the HwndSource parameters directly

lindexi:
## Description

The RegisterHwndForInput only be called by `HwndStylusInputProvider..ctor`. And the `HwndStylusInputProvider..ctor` will pass the `HwndSource source` to RegisterHwndForInput method. So that it unnecessary to use the PresentationSource in RegisterHwndForInput.

## Customer Impact

None.

## Testing

Only CI

## Risk

Low. I do not change any behavior.


h3xds1nz (approved):
Makes sense

PR #11184, opened 2025-10-14
Remove auditSources from NuGet.config

akoeplinger:
The api.nuget.org source will clash with the CFSClean network isolation policy and the same data is provided by AzDO now.

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/wpf/pull/11184)


akoeplinger:
you'll also need to backport this to release/10.0 to fix the SFI item

PR #11229, opened 2025-11-04
Update branding to 9.0.12

vseanreesermsft:
###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/wpf/pull/11229)

### DIFF
--- a/eng/Versions.props
+++ b/eng/Versions.props
@@ -3,7 +3,7 @@
   <PropertyGroup>
     <MajorVersion>9</MajorVersion>
     <MinorVersion>0</MinorVersion>
-    <PatchVersion>11</PatchVersion>
+    <PatchVersion>12</PatchVersion>
     <PreReleaseVersionLabel>rtm</PreReleaseVersionLabel>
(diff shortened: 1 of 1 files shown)


eng/Versions.props
```diff
@@ -3,7 +3,7 @@
   <PropertyGroup>
     <MajorVersion>9</MajorVersion>
     <MinorVersion>0</MinorVersion>
-    <PatchVersion>11</PatchVersion>
+    <PatchVersion>12</PatchVersion>
     <PreReleaseVersionLabel>rtm</PreReleaseVersionLabel>
```
AndriySvyryd, line 7: Shouldn't this be servicing?

pranav-gupta-msft: @vseanreesermsft , can you please update the PreReleaseVersaionLabel ?

akoeplinger: yeah. in this case it doesn't matter since wpf doesn't ship any packages

Issue #11257, opened 2025-11-16
ListCollectionView generating bad NewStartingIndex on LiveSorting


nalywa:
When using a `ListCollectionView` with live sorting enabled, occasionally the `CollectionChanged` event will emit a `NotifyCollectionChangedAction.Move` notification where the `NotifyCollectionChangedEventArgs.NewStartingIndex` property doesn't match the actual index of the item in the collection found with `ListCollectionView.IndexOf(object item)`.

If the collection view is used as an items source for a WPF component such as `ListBox`, this can result in incorrect sorting behavior or duplicate items appearing.

This seems to only occur for large collections (hundreds of items) where most of them have been marked for re-sort, presumably enough to meet `ListCollectionView.LiveSortingDensityThreshold` and trigger the `LiveShapingList.RestoreLiveSortingByInsertionSort(...)` code path.

I've attached a failing unit test demonstrating the problem. The expected behavior is that observing `CollectionChanged` events would be enough to keep an external collection in sync with the `ListCollectionView` items.

[ListCollectionViewTests.cs.txt](https://github.com/user-attachments/files/23568895/ListCollectionViewTests.cs.txt)

PR #11258, opened 2025-11-16
Fix post-sort index tracking in RBTree insertion sort

nalywa:
When using insertion sort to restore ListCollectionView sort order when live sorting is in effect, ensure that the new index (after sort) is resolved as a global index within the entire tree rather than a local index within a RBNode subtree.

Fixes [# 11257](https://github.com/dotnet/wpf/issues/11257)

## Description

When using a ListCollectionView with live sorting enabled, occasionally the CollectionChanged event will emit a NotifyCollectionChangedAction.Move notification where the NotifyCollectionChangedEventArgs.NewStartingIndex property doesn't match the actual index of the item in the collection found with ListCollectionView.IndexOf(object item).

This change addresses the bug responsible for the observed incorrect behavior, fixing a code path where a local subtree index would be provided as the NewStartingIndex instead of the global index within the live shaping tree.

## Customer Impact

If the collection view is used as an items source for a WPF component such as ListBox, this can result in incorrect sorting behavior or duplicate items appearing.

## Regression

I believe this bug was carried over from the .NET Framework implementation.

## Testing

In addition to the unit test ensuring the `NotifyCollectionChangedEventArgs` is consistent with both the source and observed collection, this has been tested in a real WPF application to confirm that this addresses the incorrect UI behavior experienced before.

## Risk

I believe this to be low risk. There are no structural changes to the code. There is no performance cost beyond simple integer arithmetic. Both the incorrect behavior and fix are easily understandable to those who are familiar with the domain. The change is covered with a new unit test validating behavior.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/wpf/pull/11258)

### DIFF
--- a/src/Microsoft.DotNet.Wpf/tests/UnitTests/PresentationFramework.Tests/System/Windows/Data/ListCollectionViewTests.cs
+++ b/src/Microsoft.DotNet.Wpf/tests/UnitTests/PresentationFramework.Tests/System/Windows/Data/ListCollectionViewTests.cs
@@ -12,7 +12,7 @@ namespace System.Windows.Data;
 
 public sealed class ListCollectionViewTests
 {
-    public sealed class TestItem : INotifyPropertyChanged
+    private sealed class TestItem : INotifyPropertyChanged
     {
         public int Value
         {


nalywa:
@dotnet-policy-service agree

lindexi (approved):
LGTM

src/Microsoft.DotNet.Wpf/tests/UnitTests/PresentationFramework.Tests/System/Windows/Data/ListCollectionViewTests.cs
```diff
@@ -0,0 +1,79 @@
...
+using System.ComponentModel;
+using System.Linq;
+using System.Threading.Tasks;
+using System.Windows.Threading;
+
+namespace System.Windows.Data;
+
+public sealed class ListCollectionViewTests
+{
+    public sealed class TestItem : INotifyPropertyChanged
```
IAmTheCShark: Im not entirely sure what the guidlines are, but is public required here?

h3xds1nz: `file` would be a better choice.

h3xds1nz (approved):
LGTM

src/Microsoft.DotNet.Wpf/tests/UnitTests/PresentationFramework.Tests/System/Windows/Data/ListCollectionViewTests.cs
nalywa: Restricted to `private`

PR #11284, opened 2025-11-25
Enable retry mechanism is Fluent Crashes

harshit7962:
## Description
The following changes aim to fix certain repetitive crashes due to win32 DWM API. The idea here is to re-try extending content into client area which might result in a crash for about 2 times and if that still results in an exception, we log the same.

**cc:/** @himgoyalmicro 

## Customer Impact

## Regression
Yes

## Testing
Local Build Pass
Sample Application Testing
Tested with a runtime application

## Risk
Low, things are very much in parity with previous impl.

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/wpf/pull/11284)

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Standard/NativeMethods.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Standard/NativeMethods.cs
@@ -11,6 +11,7 @@ using System.Runtime.InteropServices.ComTypes;
 using System.Text;
 using System.Threading;
 using Microsoft.Win32.SafeHandles;
+using MS.Internal;
 
 // Some COM interfaces and Win32 structures are already declared in the framework.
 // Interesting ones to remember in System.Runtime.InteropServices.ComTypes are:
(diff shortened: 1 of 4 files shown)


src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Standard/NativeMethods.cs
```diff
@@ -2511,6 +2512,32 @@ public static IntPtr CreateWindowEx(
         [DllImport("dwmapi.dll", EntryPoint = "DwmGetColorizationColor", PreserveSig = true)]
         private static extern HRESULT _DwmGetColorizationColor(out uint pcrColorization, [Out, MarshalAs(UnmanagedType.Bool)] out bool pfOpaqueBlend);
 
+        public static int DwmExtendFrameIntoClientArea(IntPtr hwnd, ref MARGINS pMarInset)
+        {
+            int retryCount = 2;
```
pranav-gupta-msft: Is there no flag which we can add to enable/disable this functionality ? (VelocityKey, Containment ? )

harshit7962: Yeah it makes sense. We can use `AppContextSwitch` for containment. Will make the changes.

harshit7962: Done.

src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Standard/NativeMethods.cs
```diff
@@ -2511,6 +2512,32 @@ public static IntPtr CreateWindowEx(
...
+                {
+                    // DWM composition may not be available or the operation failed
+                    // Return the error code rather than crashing the application
+                    retryCount--;
+                    if(retryCount == 0)
+                    {
+                        return ex.HResult;
+                    }
+                    System.Diagnostics.Trace.WriteLine($"DwmExtendFrameIntoClientArea: retrying after COMException (HResult={ex.HResult})");
+                    Thread.Sleep(100);
```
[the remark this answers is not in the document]
harshit7962: Keeping 100ms for now, will decrease it if we see issues in preview builds.

Issue #11235, opened 2025-11-06
New MC3080 error when building existing .NET 8-targetting WPF code with .NET 10 SDK rc.2


cyanite:
This is difficult to describe in details, because it's in a rather large code base, and has not been reduced, but it happens when building existing code using the .NET 10 SDK (-rc2), but without any changes to the actual code, so still targeting .NET 8.

The error is:

    ...\StyleDictionary.xaml(306,180): error MC3080: The attached property 'VersatileListView.IsGroupButton' cannot be set because it does not have an accessible Set method accessor. [...Core.Windows.Controls.csproj]

This property is defined in xaml like this:

    <ToggleButton x:Name="toggleButton" Grid.Column="1" IsTabStop="False" FocusVisualStyle="{StaticResource MiniToggleButtonFocus}" Template="{StaticResource MiniToggleButton}" local:VersatileListView.IsGroupButton="{Binding RelativeSource={RelativeSource FindAncestor, AncestorType=GridViewColumnHeader}, Path=Column}" Margin="0,0,5,0">

Where I assume (I am nowhere near an WPF expert) that the binding is "local:VersatileListView.IsGroupButton".

In the code, there is a getter and setter method like this:

    internal static void SetIsGroupButton(DependencyObject obj, bool value)
    {
      obj.SetValue(IsGroupButtonProperty, value);
    }

    internal static bool GetIsGroupButton(DependencyObject obj)
    {
      return (bool)obj.GetValue(IsGroupButtonProperty);
    }

Note that both are internal. If both, or even just the Set-method, are changed to public, the bug disappears, and of course we can, and will, do that in the code base. The problem is that we have a large code base with many branches and many supported releases, so this will be somewhat complicated. And I have not experienced similar regressions when building in the past, with newer SDKs.

The error happens in the MarkupCompilePass2 target, in the identically named MarkupCompilePass2 task from `C:\Program Files\dotnet\sdk\10.0.100-rc.2.25502.107\Sdks\Microsoft.NET.Sdk.WindowsDesktop\tools\net10.0\PresentationBuildTasks.dll`

Are regressions like this known or generally accepted, perhaps in some exotic cases, or how does it work?

Also: I can try to reduce this, if need be.

pomianowski:
Same here in WPF UI repo:  
https://github.com/lepoco/wpfui

```
[...]

PR #11289, opened 2025-11-26
Fix .Net 10 SDK build issue when using internal members

ThomasGoulet73:
Fixes dotnet/wpf#11235

## Description
The issue was because dotnet/wpf#9739 introduced the new method `ReflectionUtils.GetAssemblyPartialName` and updated several places to use it. The problem is that `ReflectionUtils.GetAssemblyPartialName` returns a `ReadOnlySpan<char>` when built using .Net 10 while in the .Net Framework build it returns a `string`. The problem is that comparison were not changed to replace `==` with `Equals` or something else and `==` on a `ReadOnlySpan<>` only compares the reference not the content (See [docs](https://learn.microsoft.com/en-us/dotnet/api/system.readonlyspan-1.op_equality?view=net-10.0#remarks)). The means that even if 2 `ReadOnlySpan<char>` had the same content the `==` comparison could return false if the 2 spans don't point to the same memory address.

I fixed the issue by replacing `==` with [MemoryExtensions.Equals](https://learn.microsoft.com/en-us/dotnet/api/system.memoryextensions.equals?view=net-10.0#system-memoryextensions-equals(system-readonlyspan((system-char))-system-readonlyspan((system-char))-system-stringcomparison)) on .Net 10 and [String.Equals](https://learn.microsoft.com/en-us/dotnet/api/system.string.equals?view=net-10.0#system-string-equals(system-string-system-stringcomparison)) on .Net Framework. For .Net Framework, the behavior remains the same but on .Net 10 the comparison now compares the content instead of the memory address.

This PR should probably be backported to .Net 10.

## Customer Impact
Currently, projects that use internal members from a XAML file will fail to build.

## Regression
Yes, it was introduced in the .Net 10 SDK by this PR: dotnet/wpf#9739

## Testing
Tested locally using a custom build of `PresentationBuildTasks`.

## Risk
Low.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/wpf/pull/11289)

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Markup/XamlTypeMapper.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Markup/XamlTypeMapper.cs
@@ -2481,7 +2481,7 @@ namespace System.Windows.Markup
 
         private static bool IsInternalAllowedOnType(Type type)
         {
-            bool isInternalAllowed = ReflectionHelper.LocalAssemblyName == ReflectionUtils.GetAssemblyPartialName(type.Assembly) ||
+            bool isInternalAllowed = ReflectionUtils.GetAssemblyPartialName(type.Assembly).Equals(ReflectionHelper.LocalAssemblyName, StringComparison.Ordinal) ||
                                      IsFriendAssembly(type.Assembly);
             _hasInternals = _hasInternals || isInternalAllowed;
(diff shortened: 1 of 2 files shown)


dipeshmsft:
Thanks @ThomasGoulet73 for the fix. We have ran the test pass and validated the scenario.

PR #11290, opened 2025-11-26
Enable assetless builds and clean up YAMLs

mmitche:
- azure-pipelines1.yml no longer used.
- PR/CI and non-PR scenarios were split into two yamls, so no need to have two eng/pipeline* yamls. Instead, put logic into root yamls.
- Make official build assetless. Keep option for Loc.
- Clean up PR yaml. Ensure that helix testing is actually running on internal PRs.

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/wpf/pull/11290)

### DIFF
--- a/azure-pipelines-pr.yml
+++ b/azure-pipelines-pr.yml
@@ -34,7 +34,7 @@ stages:
   - template: /eng/common/templates/jobs/jobs.yml@self
     parameters:
       enablePublishBuildArtifacts: true
-      enablePublishTestResults: false # tests run in helix
+      enablePublishTestResults: false
       enablePublishBuildAssets: false
       enablePublishUsingPipelines: false
       enableTelemetry: true


azure-pipelines-pr.yml
```diff
@@ -59,34 +23,133 @@ pr:
...
 - stage: build
   displayName: Build 
   jobs:
-  - template: /eng/pipeline-pr.yml
+  - template: /eng/common/templates/jobs/jobs.yml@self
     parameters:
-      ${{ if and(ne(variables['System.TeamProject'], 'public'), notin(variables['Build.Reason'], 'PullRequest')) }}:
-        runAsPublic: false
+      enablePublishBuildArtifacts: true
+      enablePublishTestResults: false # tests run in helix
```
ericstj: This says test run in helix but below I see TRX files being uploaded?  Which is it?

azure-pipelines.yml
```diff
@@ -46,27 +28,31 @@ extends:
...
\ No newline at end of file
+          enablePublishBuildArtifacts: true
+          enablePublishBuildAssets: false
+          enablePublishTestResults: false
+          publishAssetsImmediately: true
+          isAssetlessBuild: true
+          enableTelemetry: true
+          sourceIndexParams:
+            condition: eq(variables['Build.SourceBranch'], 'refs/heads/main')
+            binlogPath: artifacts/log/Debug/x86/Build.binlog
```
ericstj, line 50: Is this going to work?  Who actually does the build for source index to produce this binlog.  Might need to specify the build command to run.  cc @radical

azure-pipelines.yml
```diff
@@ -1,23 +1,3 @@
...
-  value: WINDOWSDESKTOP
-- ${{ if or(startswith(variables['Build.SourceBranch'], 'refs/heads/release/'), startswith(variables['Build.SourceBranch'], 'refs/heads/internal/release/'), eq(variables['Build.Reason'], 'Manual')) }}:
-  - name: PostBuildSign
-    value: false
-- ${{ else }}:
-  - name: PostBuildSign
-    value: true
-- ${{ if and(ne(variables['System.TeamProject'], 'public'), notin(variables['Build.Reason'], 'PullRequest')) }}:
-  - group: DotNet-Wpf-SDLValidation-Params
 trigger:
```
ericstj, line 3: Probably a comment is needed here to indicate that the official build does no actual building of real shipping assets.  We rely on VMR to do that.

azure-pipelines.yml
```diff
@@ -46,27 +28,31 @@ extends:
         configFile: '$(Build.SourcesDirectory)/eng/tsaoptions.json'
```
ericstj: Do we still want SDL checks to run here, or are they completely replaced with VMR?

azure-pipelines.yml
ericstj: I see -- it will run the default command -- https://github.com/dotnet/arcade/blob/e29f11a331730bd4f51024ba83fa067d2ab80ca9/eng/common/core-templates/job/source-index-stage1.yml#L3

Is this correct for WPF?  I would have thought they might need `_InternalRuntimeDownloadArgs` and `_PlatformArgs`

azure-pipelines-pr.yml
mmitche: I think tests run locally...not in Helix

azure-pipelines.yml
mmitche: On main we wouldn't need the internal runtime, and platform..maybe it just chooses the default?

harshit7962:
Hi @mmitche, checking on the PR status. Can you please look at the pointers above. Thanks.

azure-pipelines-pr.yml
dipeshmsft: Right, as far as I know, there was an effort to run tests on Helix, but it wasn't completed.

azure-pipelines.yml
mmitche: Nope. Removed.

mmitche:
Test https://dev.azure.com/dnceng/internal/_build/results?buildId=2856468&view=results

mmitche:
@harshit7962 updated.

mmitche:
https://dev.azure.com/dnceng/internal/_build/results?buildId=2856474&view=results

PR #11333, opened 2025-12-18
Revert D3DCompiler redist workaround

dipeshmsft:
## Description
Reverting the D3DCompiler redist workaround to allow pulling in newer version of D3DCompiler from Windows SDK. This will fix the API Scan issue we are facing with D3DCompiler

## Customer Impact
--

## Regression
--

## Testing
--

## Risk
--

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/wpf/pull/11333)

### DIFF
--- a/eng/WpfArcadeSdk/tools/Wpf.Cpp.props
+++ b/eng/WpfArcadeSdk/tools/Wpf.Cpp.props
@@ -3,8 +3,10 @@
     <VCProjectVersion>15.0</VCProjectVersion>
     <CharacterSet>Unicode</CharacterSet>
 
+    <!-- v143 = Visual Studio 2022 -->
     <PlatformToolset>v143</PlatformToolset>
 
+    <!-- Windows 11 SDK -->
     <WindowsTargetPlatformVersion>10.0.26100.0</WindowsTargetPlatformVersion>
 
     <ConfigurationType Condition="'$(ConfigurationType)'==''">DynamicLibrary</ConfigurationType>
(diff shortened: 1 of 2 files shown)


himgoyalmicro (commented):
How are we ensuring that we are picking the latest D3DCompiler_47 dll?

dipeshmsft:
> How are we ensuring that we are picking the latest D3DCompiler dll?

For this I will download the produced assets and check the version of D3DCompiler there. Most probably it should be the new one.

eng/WpfArcadeSdk/tools/Wpf.Cpp.props
```diff
@@ -3,11 +3,9 @@
     <VCProjectVersion>15.0</VCProjectVersion>
     <CharacterSet>Unicode</CharacterSet>
 
-    <!-- v143 = Visual Studio 2022 -->
     <PlatformToolset>v143</PlatformToolset>
```
harshit7962, line 7: Should we also bump this up to 145?
Probably not related and hence wasn't updated, but checking nonetheless.

dipeshmsft:
After downloading the published asset, it seems like the Windows SDK version on the pipeline is not the most recent. 

image

I don't think this will solve the D3DCompiler API Scan issue with this. Although the changes seem to work in the sense that the D3DCompiler being redist is newer than what it used to be.

image

eng/WpfArcadeSdk/tools/Wpf.Cpp.props
dipeshmsft: We are blocked on that due to unavailability of C++ ATL for ARM64 component in the new images. Hence, I was trying to find a way to fix this issue independent of WPF not building on VS2026 in CI

PR #11364, opened 2026-01-17
Stop preventing inlining of `Matrix.Identity` by making `Matrix.s_identity` readonly.

hexawyz:
## Description

While trying to vectorize the WPF `Vector`/`Point`/`Matrix` code with limited success, I stumbled on some low-hanging fruits that could bring noticeable performance improvements to code using `Matrix`.
(See here for my tentatives: https://github.com/hexawyz/WpfVectorBenchmark/tree/master/WpfVectorBenchmark/A)

This first change is very simple but does allow the JIT to inline `Matrix.Identity`, which was previously not the case. This might have been intentional at the time, I wouldn't know, but on the few simple benchmarks i was running, this improved performance by 10% to 23%.
Also, if this ever intentional, this would have been done at a time where support for SIMD in the runtime was inexistant, so making the code more "correct" here should delegate the decision making to the JIT.

## Regression

This won't break any code, but it will change the performance characteristics of the code. While I would expect the change to be mostly positive, there could be scenarios where it induces a performance regression.

## Testing

Benchmark

````
| Method                           | Mean      | Error     | StdDev    | Code Size |
|--------------------------------- |----------:|----------:|----------:|----------:|
| Rotate_Ref                       | 21.586 ns | 0.1903 ns | 0.1780 ns |   1,115 B |
| Rotate_A                         | 15.334 ns | 0.0559 ns | 0.0496 ns |   1,124 B |
| TranslateRotateTranslate_Ref     | 11.076 ns | 0.0367 ns | 0.0325 ns |   1,409 B |
| TranslateRotateTranslate_A       |  9.640 ns | 0.0257 ns | 0.0201 ns |   1,162 B |
````

````csharp
private static Vector_Ref Direction_Ref
{
	[MethodImpl(MethodImplOptions.NoInlining)]
	get => new(7, 5);
}

private static Vector_A Direction_A
{
	[MethodImpl(MethodImplOptions.NoInlining)]
	get => new(7, 5);
}

[Benchmark]
public Vector_Ref Rotate_Ref()
{
	var v = Direction_Ref;
	var m = Matrix_Ref.Identity;
	m.Rotate(55);
	return m.Transform(v);
}

[Benchmark]
public Vector_A Rotate_A()
{
	var v = Direction_A;
	var m = Matrix_A.Identity;
	m.Rotate(55);
	return m.Transform(v);
}

[Benchmark]
public Vector_Ref TranslateRotateTranslate_Ref()
{
	var v = Direction_Ref;
[...]

PR #11365, opened 2026-01-17
Rework/simplify the Matrix Multiplication algorithm.

hexawyz:
## Description

Similar to my other PR #11364 , this seeks to improve performance of Matrix operations.
While attempting to vectorize Matrix code to improve performance, I tried various approaches at changing the shape of the code, when it occured to me that the structure of the code itself could be made more straightforward.
While this change in itself did little to help with performance in the vectorization case (it changed nothing), it does have a much greater impact when applied to the current non-vectorized code.

The approach I took here was:

- Remove local variables (avoid hinting the runtime at using the stack)
- Remove reliance on multiple conditional jumps, in favor of a single switch
- Pack the switch cases densely (the previous algorithm was leaving large gaps between cases)
- Rely on case fallback to reduce code duplication
- Do not change the computation logic itself (i.e. for M11, M12, M21, …)

As a consequence, the final assembly code ends up significantly smaller.
However, benchmarking showed that the JIT now very much liked inlining the matrix multiplication code, even in the absence of any specific hint. This may or may not be a good thing. This does mean, though, that the performance is increased in this instance.

## Testing

If I force the method to not be inlined, it seems that there could be a very slight performance regression:

````
| Method             | Mean     | Error    | StdDev   | Code Size |
|------------------- |---------:|---------:|---------:|----------:|
| MultiplyMatrix_Ref | 13.19 ns | 0.073 ns | 0.065 ns |   1,191 B |
| MultiplyMatrix_A   | 13.44 ns | 0.029 ns | 0.024 ns |     961 B |
````

However, without any hint, the performance gains can be huge:

````
| Method                       | Mean      | Error     | StdDev    | Code Size |
|----------------------------- |----------:|----------:|----------:|----------:|
| Rotate_Ref                   | 20.237 ns | 0.0669 ns | 0.0626 ns |   1,463 B |
| Rotate_A                     |  1.023 ns | 0.0138 ns | 0.0129 ns |     149 B |
| TranslateRotateTranslate_Ref | 11.288 ns | 0.0307 ns | 0.0287 ns |   1,757 B |
| TranslateRotateTranslate_A   |  1.796 ns | 0.0227 ns | 0.0212 ns |     195 B |
| MultiplyMatrix_Ref           | 13.068 ns | 0.0335 ns | 0.0313 ns |   1,191 B |
| MultiplyMatrix_A             |  4.940 ns | 0.0187 ns | 0.0146 ns |     719 B |
````

Benchmark code: https://github.com/hexawyz/WpfVectorBenchmark/blob/6513f1eb2cf4a6beb797df5ccc4f515a376c57d8/WpfVectorBenchmark/Benchmarks.cs

## Risk

It will change the performance characteristics of matrix multiplication. Identity special cases will be marginally slower due to losing their special treatment, but the reduction in the number of conditional jumps should be helpful overall.
Performance gains from the method becoming inlinable can be impressive, but this could lead to code size increasing significantly in some scenarios. It will largely depend on common usage patterns of each program.

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/wpf/pull/11365)

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/Shared/MS/Internal/MatrixUtil.cs
+++ b/src/Microsoft.DotNet.Wpf/src/Shared/MS/Internal/MatrixUtil.cs
@@ -1,4 +1,4 @@
-// Licensed to the .NET Foundation under one or more agreements.
+﻿// Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the MIT license.
 
 // Description: This file contains the implementation of MatrixUtil, which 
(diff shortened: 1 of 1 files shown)


hexawyz:
@dotnet-policy-service agree

PR #11381, opened 2026-01-22
Add DirectWriteForwarder unit tests

AaronRobinsonMSFT:
Using [Visual Studio's Code Coverage tool](https://learn.microsoft.com/visualstudio/test/microsoft-code-coverage-console-tool?view=visualstudio), reported code coverage is **71.7 %**.

Created test suite for the DirectWriteForwarder project:

Test coverage includes:
- Font, FontFace, FontFamily, FontCollection classes
- All enums and structs (FontWeight, FontStretch, FontStyle, etc.)
- FontMetrics and GlyphMetrics calculations
- DWriteMatrix transforms (identity, scale, rotation)
- LocalizedStrings IDictionary implementation
- FontFile, FontSource, FontSourceFactory infrastructure
- FontFileLoader, FontFileStream, FontFileEnumerator
- ItemProps, Span, and ClassificationUtility for text analysis
- TrueTypeSubsetter.ComputeSubset with unsafe code
- Unsafe pointer APIs (TryGetFontTable, GetGlyphIndices, GetGlyphMetrics)
- TextAnalyzer integration tests (Itemize, GetGlyphs, GetGlyphPlacements)
- Edge cases and error handling (boundary conditions, invalid inputs)

Infrastructure changes:
- Added InternalsVisibleTo for DirectWriteForwarder.Tests in DirectWriteForwarder
- Added InternalsVisibleTo for DirectWriteForwarder.Tests in PresentationCore
- Created test project targeting net10.0-windows with xUnit and FluentAssertions
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/wpf/pull/11381)

### DIFF
--- a/src/Microsoft.DotNet.Wpf/tests/UnitTests/DirectWriteForwarder.Tests/LocalizedStringsTests.cs
+++ b/src/Microsoft.DotNet.Wpf/tests/UnitTests/DirectWriteForwarder.Tests/LocalizedStringsTests.cs
@@ -14,7 +14,7 @@ public class LocalizedStringsTests
     public void Count_ShouldBeGreaterThanZero()
     {
         var localizedStrings = TestHelpers.GetArialLocalizedStringsOrSkip();
-        
+
         localizedStrings.Count.Should().BeGreaterThan(0);
     }
 
@@ -22,9 +22,9 @@ public class LocalizedStringsTests
     public void Keys_ShouldContainCultureInfo()
     {
         var localizedStrings = TestHelpers.GetArialLocalizedStringsOrSkip();
-        
+
         var keys = localizedStrings.Keys;
-        
+
         keys.Should().NotBeEmpty();
         keys.Should().AllBeOfType<CultureInfo>();
     }
@@ -33,9 +33,9 @@ public class LocalizedStringsTests
     public void Values_ShouldContainStrings()
     {
         var localizedStrings = TestHelpers.GetArialLocalizedStringsOrSkip();
-        
+
         var values = localizedStrings.Values;
-        
+
         values.Should().NotBeEmpty();
         values.Should().AllBeOfType<string>();
     }
@@ -44,12 +44,12 @@ public class LocalizedStringsTests
     public void Indexer_WithValidCulture_ShouldReturnName()
     {
         var localizedStrings = TestHelpers.GetArialLocalizedStringsOrSkip();
-        
+
         // Get the first available culture
         var firstCulture = TestHelpers.GetFirstCultureOrSkip(localizedStrings);
-        
+
         var name = localizedStrings[firstCulture];
-        
+
         name.Should().NotBeNullOrEmpty();
     }
 
@@ -57,35 +57,21 @@ public class LocalizedStringsTests
     public void ContainsKey_WithExistingCulture_ShouldReturnTrue()
     {
         var localizedStrings = TestHelpers.GetArialLocalizedStringsOrSkip();
-        
-        var firstCulture = TestHelpers.GetFirstCultureOrSkip(localizedStrings);
-        
-        localizedStrings.ContainsKey(firstCulture).Should().BeTrue();
-    }
 
-    [Fact]
-    public void ContainsKey_WithNonExistingCulture_ShouldReturnFalse()
-    {
-        var localizedStrings = TestHelpers.GetArialLocalizedStringsOrSkip();
-        
-        // Use a culture that's unlikely to be in the font
-        var rareCulture = new CultureInfo("zu-ZA"); // Zulu
-        
-        // This may or may not contain it, but test that it doesn't throw
-        var result = localizedStrings.ContainsKey(rareCulture);
-        // Just verify it returns a boolean (doesn't throw)
-        result.Should().Be(result); // Self-comparison just verifies no exception
+        var firstCulture = TestHelpers.GetFirstCultureOrSkip(localizedStrings);
+
+        localizedStrings.ContainsKey(firstCulture).Should().BeTrue();
     }
 
     [Fact]
     public void TryGetValue_WithExistingCulture_ShouldReturnTrueAndValue()
     {
         var localizedStrings = TestHelpers.GetArialLocalizedStringsOrSkip();
-        
+
         var firstCulture = TestHelpers.GetFirstCultureOrSkip(localizedStrings);
-        
+
         bool result = localizedStrings.TryGetValue(firstCulture, out string value);
-        
+
         result.Should().BeTrue();
         value.Should().NotBeNullOrEmpty();
     }
@@ -94,9 +80,9 @@ public class LocalizedStringsTests
     public void GetEnumerator_ShouldEnumerateAllPairs()
     {
         var localizedStrings = TestHelpers.GetArialLocalizedStringsOrSkip();
-        
+
         var pairs = localizedStrings.ToList();
-        
+
         pairs.Should().NotBeEmpty();
         pairs.Should().AllSatisfy(kvp =>
         {
@@ -109,10 +95,10 @@ public class LocalizedStringsTests
     public void EnglishName_ShouldExist()
     {
         var localizedStrings = TestHelpers.GetArialLocalizedStringsOrSkip();
-        
+
         // Most fonts should have an English (US) name
         var englishCulture = new CultureInfo("en-US");
-        
+
         if (localizedStrings.ContainsKey(englishCulture))
         {
             var englishName = localizedStrings[englishCulture];
@@ -124,7 +110,7 @@ public class LocalizedStringsTests
     public void IsReadOnly_ShouldBeTrue()
     {
         var localizedStrings = TestHelpers.GetArialLocalizedStringsOrSkip();
-        
+
         localizedStrings.IsReadOnly.Should().BeTrue();
     }
 
@@ -132,9 +118,9 @@ public class LocalizedStringsTests
     public void Add_ShouldThrowNotSupportedException()
     {
         var localizedStrings = TestHelpers.GetArialLocalizedStringsOrSkip();
-        
+
         Action act = () => localizedStrings.Add(CultureInfo.InvariantCulture, "Test");
-        
+
         act.Should().Throw<NotSupportedException>();
     }
 
@@ -142,11 +128,11 @@ public class LocalizedStringsTests
     public void Remove_ShouldThrowNotSupportedException()
     {
         var localizedStrings = TestHelpers.GetArialLocalizedStringsOrSkip();
-        
+
         var firstCulture = TestHelpers.GetFirstCultureOrSkip(localizedStrings);
-        
+
         Action act = () => localizedStrings.Remove(firstCulture);
-        
+
         act.Should().Throw<NotSupportedException>();
     }
 
@@ -154,9 +140,9 @@ public class LocalizedStringsTests
     public void Clear_ShouldThrowNotSupportedException()
     {
         var localizedStrings = TestHelpers.GetArialLocalizedStringsOrSkip();
-        
+
         Action act = () => localizedStrings.Clear();
-        
+
         act.Should().Throw<NotSupportedException>();
     }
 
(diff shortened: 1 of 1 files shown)


AaronRobinsonMSFT:
I'm currently done with this PR. The highlevel is the tests appear to be passing on all legs, look for `DirectWriteForwarder.Tests`, the reported code coverage is 71.1 %, and the only product change is adding the `InternalsVisibleTo` attribute to the `DirectWriteForwarder` and `PresentationCore` for the `DirectWriteForwarder.Tests` project.

/cc @dipeshmsft @JeremyKuhne @agocke @jkoritzinsky

src/Microsoft.DotNet.Wpf/tests/UnitTests/DirectWriteForwarder.Tests/TrueTypeSubsetterTests.cs
```diff
@@ -0,0 +1,167 @@
...
+                IndexOutOfRangeException? caught = null;
+                try
+                {
+                    TrueTypeSubsetter.ComputeSubset(pFontData, fontData.Length, arialUri, 0, glyphArray);
+                }
+                catch (IndexOutOfRangeException ex)
+                {
+                    caught = ex;
+                }
+                Assert.NotNull(caught);
```
jkoritzinsky: Convert to `Assert.Throws`?

src/Microsoft.DotNet.Wpf/tests/UnitTests/DirectWriteForwarder.Tests/TrueTypeSubsetterTests.cs
```diff
@@ -0,0 +1,167 @@
...
+        {
+            fixed (byte* pFontData = fontData)
+            {
+                var result = TrueTypeSubsetter.ComputeSubset(pFontData, fontData.Length, arialUri, 0, glyphArray);
+
+                result.Should().NotBeNull();
+                result.Should().NotBeEmpty();
+            }
+        }
+    }
```
jkoritzinsky, line 205: This feels like it's not really validating the results of the test, just that there's data returned at all.

src/Microsoft.DotNet.Wpf/tests/UnitTests/DirectWriteForwarder.Tests/DWriteTypeConverterTests.cs
```diff
@@ -0,0 +1,841 @@
...
+    [InlineData(7)]  // Expanded
+    [InlineData(8)]  // ExtraExpanded
+    [InlineData(9)]  // UltraExpanded
+    public void Convert_FontStretch_ToNative_ShouldSucceed(int stretchValue)
+    {
+        var arialFamily = TestHelpers.GetArialFamilyOrSkip();
+
+        var stretch = (FontStretch)stretchValue;
+        var matchingFonts = arialFamily.GetMatchingFonts(FontWeight.Normal, stretch, FontStyle.Normal);
+        matchingFonts.Should().NotBeNull();
```
jkoritzinsky, line 96: Could we have this test validate that the correct matching font for a given weight, stretch, and style is returned instead of "any font"? (multiple tests)

src/Microsoft.DotNet.Wpf/tests/UnitTests/DirectWriteForwarder.Tests/DWriteTypeConverterTests.cs
```diff
@@ -0,0 +1,841 @@
...
+    public void Convert_InformationalStringID_ToNative_ShouldSucceed(int stringIdValue)
+    {
+        var arialFamily = TestHelpers.GetArialFamilyOrSkip();
+        Assert.SkipUnless(arialFamily.Count > 0, "Arial font family has no fonts");
+
+        var font = arialFamily[0u];
+        var stringId = (InformationalStringID)stringIdValue;
+
+        // GetInformationalStrings calls DWriteTypeConverter.Convert(InformationalStringID)
+        _ = font.GetInformationalStrings(stringId, out _);
```
jkoritzinsky: Have the expected informational strings for a given font hard-coded in the test?

ThomasGoulet73:
@AaronRobinsonMSFT I'm a bit curious with the background for this PR, is it because you're planning on working on DirectWriteForwarder or is it just to increase code coverage (Maybe to increase testing of C++/CLI for the .Net interop team) ?

I ask this because I have open PRs (dotnet/wpf#9902 and dotnet/wpf#10523, though they've been open for a while) and more code locally to migrate DirectWriteForwarder to C# and I just want to make sure I'm not stepping on anybody's toes.

In any case, this PR would help a lot with migrating DirectWriteForwarder to C# so that's very nice.

src/Microsoft.DotNet.Wpf/tests/UnitTests/DirectWriteForwarder.Tests/TrueTypeSubsetterTests.cs
AaronRobinsonMSFT: Can't because of pinning.

src/Microsoft.DotNet.Wpf/tests/UnitTests/DirectWriteForwarder.Tests/DWriteTypeConverterTests.cs
AaronRobinsonMSFT: So that is possible, but not really the point for this PR or library at present. This is about strict code coverage and that is satisfied sicne much of the APIs here are about wrapping native APIs. The correctness angle is important for sure, but not precisely the point given this is wrapping calls to `IDWriteFontFamily`.

I'll go through this and see what APIs are doing more logic compared to just forwarding.

src/Microsoft.DotNet.Wpf/tests/UnitTests/DirectWriteForwarder.Tests/TrueTypeSubsetterTests.cs
AaronRobinsonMSFT: Agree. This can be made more useful.

src/Microsoft.DotNet.Wpf/tests/UnitTests/DirectWriteForwarder.Tests/DWriteTypeConverterTests.cs
jkotas: If you are not validating the values returned by the APIs being tested, it means that the wrapper can be mixing up or corrupting the values being forwarded and the tests won't notice...

AaronRobinsonMSFT:
> more code locally to migrate DirectWriteForwarder to C#

@ThomasGoulet73 This is indeed the long term goal, but we need tests. That is the purpose/beginning of this PR process. We need measurable metrics in order to have confidence in any port to C# and the cost of doing it manually is non-zero. The current suite of WPF integration and unit tests is helpful but not sufficient for something as low level and sensitive as the `DirectWriteForwarder` assembly. Regardless if/when moving toward C# or not, the tests help us have confidence in changes so that is a win I'm assuming we can all agree on.

> In any case, this PR would help a lot with migrating DirectWriteForwarder to C# so that's very nice.

Yep.

src/Microsoft.DotNet.Wpf/tests/UnitTests/DirectWriteForwarder.Tests/DWriteTypeConverterTests.cs
AaronRobinsonMSFT: I'm not disagreeing that the tests here are incomplete or not as robust as needed. I'm simply pointing out that the metric being optimized for in this pass in coverage. The boiler plate is now here and additional testing can now be added in a more targeted fashion. That is not the point of this particular pass, more needs to be done.

ThomasGoulet73:
> This is indeed the long term goal, but we need tests. That is the purpose/beginning of this PR process. We need measurable metrics in order to have confidence in any port to C# and the cost of doing it manually is non-zero. The current suite of WPF integration and unit tests is helpful but not sufficient for something as low level and sensitive as the `DirectWriteForwarder` assembly. Regardless if/when moving toward C# or not, the tests help us have confidence in changes so that is a win I'm assuming we can all agree on.

That's good news, thanks

src/Microsoft.DotNet.Wpf/tests/UnitTests/DirectWriteForwarder.Tests/TrueTypeSubsetterTests.cs
jkoritzinsky: Could we wrap the whole fixed block in the Throws callback?

src/Microsoft.DotNet.Wpf/tests/UnitTests/DirectWriteForwarder.Tests/DWriteTypeConverterTests.cs
jkoritzinsky: Coverage without validation feels to me like false coverage. The number looks good but it doesn't actually give us confidence for changes. I'd rather not have these tests that don't provide real validation in this first pass and add them in a second PR.

AaronRobinsonMSFT: This isn't about a traditional coverage metric and "false coverage" isn't the argument here. The goal of this is to have something to build on top of. Understanding how to exercise the code is the point and the course coverage metric is the goal. If this was a code base that is having active development, I fully agree with this concern and false coverage is a concern. This conversation is missing the context of this repo and intent of the work.

I will take another stab and expanding these tests, but the only thing that is going to happen is increase the change and make for more to review and validate. We then play the probability game of seeing tests that are potentially insufficient. Regardless of anything that is checked in here, this is about boilerplate and these tests will need to be re-examined as code changes because everything here is new.

src/Microsoft.DotNet.Wpf/tests/UnitTests/DirectWriteForwarder.Tests/TrueTypeSubsetterTests.cs
AaronRobinsonMSFT: I suppose, what does that accomplish?

src/Microsoft.DotNet.Wpf/tests/UnitTests/DirectWriteForwarder.Tests/TrueTypeSubsetterTests.cs
AaronRobinsonMSFT: Done.

src/Microsoft.DotNet.Wpf/tests/UnitTests/DirectWriteForwarder.Tests/TrueTypeSubsetterTests.cs
jkoritzinsky: I think it's cleaner to read a test that's validating that an exception is thrown via `Assert.Throws` than having to mentally reconstruct that when it's done manually.

src/Microsoft.DotNet.Wpf/tests/UnitTests/DirectWriteForwarder.Tests/DWriteTypeConverterTests.cs
jkoritzinsky: If this is just intended to be "tests that exercise the paths that we can later expand with actual validation," sure. I wouldn't feel much more comfortable with making changes to DirectWriteForwarder with this PR than without given the quality of the tests in this PR, but if that's a non-goal for this PR, then it's fine.

jkoritzinsky (dismissed):
Looks fine enough for the goal of this PR.

Would like to see the comment from the Copilot review applied before this is merged though.

PR #11417, opened 2026-02-03
Prevent crash propagation from DWM

himgoyalmicro:
## Description

It’s possible for DWM to crash or exit unexpectedly, and when that happens, `DwmExtendFrameIntoClientArea` returns a failure. 

This change adds a safety check to prevent that failure from propagating and causing a crash in WPF and reverting the app to the previous behavior. 

We will only crash if we encounter Invalid Args exception. 

## Customer Impact

## Regression

N/A

## Testing

Local Build Pass
Sample Application Testing
Tested with a runtime application

## Risk

Low, things are very much in parity with previous impl.

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/wpf/pull/11417)

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Standard/NativeMethods.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Standard/NativeMethods.cs
@@ -2520,18 +2520,19 @@ namespace Standard
             {
                 return true;
             }
+            
             if(FrameworkAppContextSwitches.DisableDWMCrashContainment)
             {
                 // Original behavior: just call the native method and let any exceptions propagate.
                 HRESULT.ThrowLastError();
             }
 
-            // Crash containment behavior: catch and handle DWM errors gracefully.
-            // Check if DWM composition is disabled and log it
+            // Crash only if the arguments are invalid.
             if(hr == HRESULT.E_INVALIDARG)
             {
                 HRESULT.ThrowLastError();
             }
+
             return false;
         }
 


src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Standard/NativeMethods.cs
```diff
@@ -2513,40 +2513,26 @@ public static IntPtr CreateWindowEx(
...
-                        System.Diagnostics.Trace.WriteLine($"DwmExtendFrameIntoClientArea: retrying after COMException (HResult={ex.HResult})");
-                        Thread.Sleep(100);
-                    }
-                }
+                HRESULT.ThrowLastError();
             }
-            
-            
-            return 0;
+            return false;
```
harshit7962, line 2536: nit: new line

src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Standard/NativeMethods.cs
```diff
@@ -2513,40 +2513,26 @@ public static IntPtr CreateWindowEx(
...
             {
                 // Original behavior: just call the native method and let any exceptions propagate.
-                return _DwmExtendFrameIntoClientArea(hwnd, ref pMarInset);
-            } 
-            else
+                HRESULT.ThrowLastError();
+            }
+
+            // Crash containment behavior: catch and handle DWM errors gracefully.
+            // Check if DWM composition is disabled and log it
```
harshit7962: Is this valid anymore?

harshit7962 (dismissed):

src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Standard/NativeMethods.cs
himgoyalmicro: Fixed

src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Standard/NativeMethods.cs
himgoyalmicro: Updated

mw-ea:
Is there a chance this can be backported to .NET 10?
It's affecting our users under https://github.com/dotnet/wpf/issues/11440, I'm sure we're not the only ones.

PR #11431, opened 2026-02-08
Fix Slider snapping precision issue (#8819)

### LINKED ISSUE #8819: WPF - slider precision is not correct.
### Description

Hello,

we use slider with this configuration
```xaml
        <Slider
            Grid.Row="3"
            Grid.Column="1"
            IsSnapToTickEnabled="True"
            TickFrequency="0.1"
            Minimum="{Binding Gamma.Minimum}"
            Maximum="{Binding Gamma.Maximum}"
            Value="{Binding Gamma.Value}"/>
```
Gamma.Minimum is 0.5.
Gamma.Maximum is 4.
Gamma.Value is bounded.
All properties are `double`.
Problem is that when Value is `1.9` and i move slider very little then value does not change to `2` or `1.8` but it changes to `1.9000000000000001` which is not correct. Value should be `2`.

I have tried to debug it and I think that I have found problem at this line https://github.com/dotnet/wpf/blob/2fdb74d7b11465f36246b7caf06fa00e9c6d8454/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Controls/Slider.cs#L1135
When `* TickFrequency` is used value result of multiplication brings precision problem.

Part of the code from the link above.
```csharp
                else if (DoubleUtil.GreaterThanZero(TickFrequency))
                {
                    previous = Minimum + (Math.Round(((value - Minimum) / TickFrequency)) * TickFrequency);
                    next = Math.Min(Maximum, previous + TickFrequency);
                }
```

### Reproduction Steps

Add Slider with binding to double value

```xaml
        <Slider
            Grid.Row="3"
            Grid.Column="1"
            IsSnapToTickEnabled="True"
            TickFrequency="0.1"
            Minimum="0.5"
            Maximum="4"
            Value="{Binding Value}"/>
```

Set value to `1.9` from code.
Try to move slidre a little. See problem.

### Expected behavior

- Value should be set to `2.0`

### Actual behavior

Value is set to `1.9000000000000001`.

### Regression?

_No response_

### Known Workarounds

_No response_

### Impact

_No response_

### Configuration

_No response_

### Other information

_No response_

sachPundir:
Fixes #8819

## Description

This PR fixes floating-point precision issues in Slider snapping when TickFrequency is used.

Due to double precision limitations, calculations such as:

19 * 0.1 → 1.9000000000000001

can produce small rounding errors that affect snapping behavior.

This change ensures the final snapped value is rounded to a stable precision to avoid floating-point artifacts while preserving existing Slider logic.

Rounding is applied only to the final snapped value and does not modify Minimum, Maximum, or TickFrequency. This ensures that real precision coming from Min/Max offsets is preserved.

For example:

Minimum = 0.5001
TickFrequency = 0.1

continues to correctly produce values like 1.9001.

The intent is to remove IEEE floating-point noise, not to change logical tick precision.

## Customer Impact

Improves numerical stability of Slider snapping when fractional TickFrequency values are used, resulting in more predictable snapped values.

## Regression

No — this addresses a long-standing floating-point precision behavior rather than a recently introduced regression.

## Testing

- Reproduced the issue using a minimal WPF sample with fractional TickFrequency values (e.g., 0.1, 0.112).
- Confirmed incorrect snapping behavior prior to the fix due to floating-point precision artifacts.
- Verified corrected snapping behavior after the change across multiple ranges and tick frequencies.
- Tested slider movement in both increasing and decreasing directions.
- Tested with non-zero Minimum values to ensure correct offset handling.
- Tested with non-zero Minimum values to ensure correct offset handling.
- Validated behavior using values with multiple decimal places in a sample application to ensure stable snapping results.
- Tested values close to Minimum and Maximum limits.
- Ran the full WPF unit test suite locally — all tests pass.
- Confirmed no behavioral changes for integer TickFrequency scenarios.

## Risk

Low.

The change only affects the final snapped value calculation and does not alter core Slider logic or public APIs.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/wpf/pull/11431)

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Controls/Slider.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Controls/Slider.cs
@@ -1181,18 +1181,31 @@ namespace System.Windows.Controls
                 }
                 else if (DoubleUtil.GreaterThanZero(TickFrequency))
                 {
-                    previous = Minimum + (Math.Round(((value - Minimum) / TickFrequency)) * TickFrequency);
+                    try
+                    {
+                        decimal dMin = (decimal)Minimum;
+                        decimal dFreq = (decimal)TickFrequency;
+                        decimal dValue = (decimal)value;
+
+                        // Round to nearest tick index, then reconstruct in decimal to avoid
+                        // the floating-point noise that occurs when multiplying back in double
+                        // (e.g. 19 * 0.1 = 1.9000000000000001). Decimal arithmetic is exact
+                        // for the base-10 fractions commonly used as TickFrequency values.
+                        decimal index = Math.Round((dValue - dMin) / dFreq);
+                        previous = (double)(dMin + index * dFreq);
+                    }
+                    catch (OverflowException)
+                    {
+                        // Fallback for values outside decimal's range (approx. ±7.9e28),
+                        // or for TickFrequency values not representable in decimal.
+                        previous = Minimum + (Math.Round(((value - Minimum) / TickFrequency)) * TickFrequency);
+                    }
                     next = Math.Min(Maximum, previous + TickFrequency);
                 }
 
                 // Choose the closest value between previous and next. If tie, snap to 'next'.
                 value = DoubleUtil.GreaterThanOrClose(value, (previous + next) * 0.5) ? next : previous;
             }
-
-            // Normalize floating-point rounding errors introduced by
-            // repeated tick arithmetic (e.g. 0.1 + 0.2 ≠ 0.3 exactly in IEEE-754).
-            // This prevents visual artifacts like 1.1312000000000002.
-            value = Math.Round(value, 10);
 
             return value;
         }
--- a/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Controls/Slider.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Controls/Slider.cs
@@ -1189,6 +1189,11 @@ namespace System.Windows.Controls
                 value = DoubleUtil.GreaterThanOrClose(value, (previous + next) * 0.5) ? next : previous;
             }
 
+            // Normalize floating-point rounding errors introduced by
+            // repeated tick arithmetic (e.g. 0.1 + 0.2 ≠ 0.3 exactly in IEEE-754).
+            // This prevents visual artifacts like 1.1312000000000002.
+            value = Math.Round(value, 10);
+
             return value;
         }
 


miloush:
I don't think that's a correct approach, it will use wrong values if the tick frequency has higher precision, and the rounding number seems arbitrary.

sachPundir:
@miloush, Thanks for the clarification — you're right that snapping must remain strictly “closest tick” and not bias toward the next one.

My intent was only to remove IEEE noise when the computed value is already on a tick (e.g., 1.9000000000000001).

I agree arbitrary rounding could affect higher-precision TickFrequency.

I’ll explore a solution based on tick-index rounding or DoubleUtil.AreClose so logical tick precision is preserved.

sachPundir:
Thanks for the feedback — that makes sense.

I’ll add unit tests to cover snapping behavior and edge cases so we can validate the expected behavior more precisely before finalizing the fix.

sachPundir:
I ran additional edge-case tests beyond the original unit tests to validate the behavior across different tick magnitudes.

Observations:

✅ The change improves decimal-grid snapping for common cases (e.g., 0.1, 0.0001) by eliminating visible floating-point artifacts such as 1.9000000000000001.
✅ Standard scenarios (integer ticks, fractional ticks, negative ranges, boundary cases) continue to behave correctly.
❌ However, the normalization step introduces precision loss for very small tick frequencies (e.g., 1e-12), where valid snapped values can collapse to 0 due to rounding.

Example:

- TickFrequency = 1e-12
- Expected: 4e-12
- Actual (with normalization): 0

Because of this regression, the current approach is not safe for all tick magnitudes. I am refining the solution to preserve small-tick precision while still addressing decimal-grid floating artifacts.

I’ll follow up with an updated approach shortly.

sachPundir:
@dotnet-policy-service agree

sachPundir:
I’ve updated the snapping reconstruction step to use decimal arithmetic when rebuilding the snapped tick value.

The rounding of the tick index remains unchanged. Previously, reconstructing the snapped value using double multiplication (e.g. 19 * 0.1) could introduce binary floating-point representation artifacts such as 1.9000000000000001.

This change keeps the existing snapping logic intact and replaces only the reconstruction step with decimal arithmetic. This eliminates representation noise for common base-10 tick frequencies (0.1, 0.01, 0.0001, etc.) while preserving full precision for high-precision tick frequencies (e.g. 1e-12).

A double-based fallback is retained for values outside the decimal range to avoid overflow.

Existing snapping behavior is unchanged apart from stabilizing the reconstructed value.

sachPundir:
@dipeshmsft @miloush This PR fixes the slider snapping precision issue caused by floating point rounding when calculating tick values.

I verified the behavior using a Slider with TickFrequency and confirmed the corrected snapping logic resolves the precision problem.

Could you please review it when you have time? Happy to update the PR if any changes or additional tests are required.

sachPundir:
Hi @dipeshmsft,

I’ve rebased this PR on the latest main and re-verified the fix.

This addresses floating-point precision artifacts in Slider snapping.

Could you please take a look when convenient?

miloush:
To be honest I am not entirely convinced this needs fixing. It's an inherent nature of floating point values that some operations result in artifacts. Whoever is consuming the values should be prepared to deal with them if needed.

The price of decimal conversions and operations and exceptions does not feel like worth the benefit to me.

sachPundir:
@miloush Thanks again for the thoughtful feedback.

After exploring multiple approaches and testing edge cases, I agree that this behavior is inherent to floating-point arithmetic and not something that should be addressed at the framework level.

The trade-offs in complexity and performance don’t justify the benefit here.

I’ll close this PR.

Really appreciate the guidance — this was a valuable learning experience.

Issue #696, opened 2019-05-16
Full temporary folder will crash cursor initialization


lindexi:
* .NET Core Version: 3.0 Preview5
* Windows version: windows 10 1903
* Does the bug reproduce also in WPF for .NET Framework 4.8?: Yes

 
 
 **Problem description:**

The cursor will crash at initialization when the temporary folder full.
 
 **Actual behavior:** 

I set a cursor from a resource and use this code.

```
var uri = new Uri("pack://application:,,,/Foo.cur");
var resource = Application.GetResourceStream(uri);
Cursor = new Cursor(resource.Stream); 
```

I can find the cursor will crash at initialization when the temporary folder full.

    System.IO.__Error.WinIOError(Int32 errorCode, String maybeFullPath)
    System.IO.Directory.InternalCreateDirectory(String fullPath, String path, Object dirSecurityObj, Boolean checkHost)
    System.IO.Directory.InternalCreateDirectoryHelper(String path, Boolean checkHost)
    System.IO.Directory.CreateDirectory(String path)
    System.IO.FileHelper.CreateAndOpenTemporaryFile(String& filePath, FileAccess fileAccess, FileOptions fileOptions, String extension, String subFolder)
    System.Windows.Input.Cursor.LoadFromStream(Stream cursorStream)
    System.Windows.Input.Cursor..ctor(Stream cursorStream, Boolean scaleWithDpi)
    System.Windows.Input.Cursor..ctor(Stream cursorStream)
    FawlalnejajerelaWhallgemcurkear.MainWindow..ctor()

 **Expected behavior:**
 
The cursor can be set.

 **Minimal repro:**
 
Run this code when the temporary folder full.

            var uri = new Uri("pack://application:,,,/Text.cur");
            var resource = Application.GetResourceStream(uri);
            Cursor = new Cursor(resource.Stream);

And the other way is set a can not visit folder as the temp folder. And you can see the code in https://github.com/lindexi/lindexi_gd/tree/8e346e750fe2075e1366a2a098d63af3b50db177/FawlalnejajerelaWhallgemcurkear that I set a folder name as `D:\lindexi\不存在文件` that is an unauthorized access folder.

**Reason**

The Cursor.LoadFromStream will generate a temporary file based on the memory stream. But the [Path.GetTempFileName Method](https://docs.microsoft.com/en-us/dotnet/api/system.io.path.gettempfilename?wt.mc_id=MVP ) limit the number of the files is 65535 and it will throw IOException when overcount.

[...]

PR #11455, opened 2026-02-17
Avoid calling GetTempFileName in BitmapDownload

### LINKED ISSUE #259: Avoid calling Path.GetTempFileName
`System.Windows.Input.Cursor.LoadFromStream` may throw IOException.
`DevDiv\FX\Rel\Net471Rel1Last\wpf\src\Core\CSharp\System\Windows\Input\Cursor.cs` calls `Path.GetTempFileName` which will throw if the `%TEMP%` directory contains too many temp files (65k).

Reported by customer as [VS Feedback issue](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/537936)

apoorvdarshan:
## Summary

- Replace the Win32 `GetTempFileName` P/Invoke in `BitmapDownload.BeginDownload` with `Path.GetRandomFileName()` + `CREATE_NEW` in a retry loop, avoiding `IOException` when the folder contains 65k+ temp files
- Add `CREATE_NEW` native constant and remove the now-unused `GetTempFileName` P/Invoke wrapper
- Follows the same pattern already used by `FileHelper.CreateAndOpenTemporaryFile` in the codebase

Fixes #259

## Test plan

- [ ] Verify bitmap download from URI still works (temp file created with `DELETE_ON_CLOSE` semantics)
- [ ] Verify no regressions in `BitmapDecoder` async download path
- [ ] Confirm removed `GetTempFileName` P/Invoke has no other callers in production code
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/wpf/pull/11455)

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Media/Imaging/BitmapDownload.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Media/Imaging/BitmapDownload.cs
@@ -9,7 +9,6 @@ using System.Windows.Threading;
 using MS.Internal;
 using System.Net;
 using System.Net.Cache;
-using System.Text;
 using MS.Win32;
 using Microsoft.Win32.SafeHandles;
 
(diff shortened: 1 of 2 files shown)


h3xds1nz:
A better solution would be to directly use `Path.GetTempFileName` whose implementation on .NET 8+ already solved this problem.

apoorvdarshan:
Thanks for the suggestion @h3xds1nz! You're right — Path.GetTempFileName on .NET 8+ already handles this internally with random filenames and retry logic. I've simplified the PR to just replace the Win32 P/Invoke with Path.GetTempFileName() directly.

h3xds1nz:
@apoorvdarshan It can still throw an exception, so for correctness, it should be under the exception handler; though it's very unlikely anyone's gonna be that lucky in their lifetime (but exceptional cases, hey).

Other issue is that this implementation now changes the observable behaviour as the files will be now stored in `%temp%`, so that should be noted.

apoorvdarshan:
@h3xds1nz  Good catch, moved it inside the try block. And yeah the location changes but since it's DELETE_ON_CLOSE it shouldn't really matter in practice.

lindexi:
Reference: https://github.com/dotnet/wpf/issues/696

PR #11461, opened 2026-02-18
TextBoxView: fix crash during large selection with pending background layout

### LINKED ISSUE #11460: [TextBoxView] Crash during large selection while background layout is incomplete
# [TextBoxView] Crash during large selection while background layout is incomplete

## Summary
`TextBoxView.GetLineIndexFromOffset` can fail during large-text selection when layout is still in time-sliced background progress, causing a crash/fail-fast path instead of graceful handling.

## Observed behavior
- Selecting text in very large `TextBox` content (for example, very long lines / high offsets) can trigger a consistency/assert failure.
- The failure happens when selection/highlight processing requests an offset beyond the currently measured line range.

## Repro notes (public)
1. Create/open a WPF app view with a `TextBox` containing a very large amount of text (for example, one very long line or total offsets in 100K+ range).
2. Trigger selection operations while layout is still being processed incrementally (IME/search/focus interactions can increase likelihood).
3. Observe crash/fail-fast behavior from `TextBoxView.GetLineIndexFromOffset` path.

## Impact
- Affects reliability of text editing/selection in large-document scenarios.
- More likely in interactive flows where layout and selection/focus updates interleave.

## Expected behavior
Selection/highlight should remain stable and never crash, even when background layout has not measured all lines yet.

## Actual behavior
The lookup assumes the requested offset is always inside measured `_lineMetrics`, and can fail when it is temporarily outside that range.

## Suspected root cause
`TextBoxView` performs time-sliced layout, so `_lineMetrics` may be only partially populated. During that window, selection/highlight can request an offset greater than the last measured line `EndOffset`, and the search path fails.

## Proposed fix direction
- Add defensive clamping in `GetLineIndexFromOffset` for offsets beyond currently measured range.
- Return the last measured line index when background layout is pending.
- Keep correctness checks for unexpected conditions while avoiding crash in expected incremental-layout windows.
- Remove redundant preconditions that incorrectly assume full containment during transitional layout states.

etvorun:
## Summary
Fixes a `TextBoxView` crash path that can occur during large-text selection when layout is still in background progress.

## What changed
- Added defensive clamping in `GetLineIndexFromOffset` when the requested offset is beyond the currently measured region.
- Return the last measured line index in this transitional background-layout window.
- Removed fragile `Invariant.Assert(Contains(position))` checks from affected text-view operations that can be reached during transitional layout timing.

## Why
`TextBoxView` uses incremental/time-sliced layout. During that window, selection/highlight paths can request offsets beyond the measured `_lineMetrics` range. The defensive path prevents assertion/fail-fast and preserves expected behavior.

## Validation
- Ported patch applied to `TextBoxView.cs` only.
- File diagnostics checked locally (`No errors found` for the modified file).

Fixes #11460

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/wpf/pull/11461)

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/PresentationFramework/MS/Internal/documents/TextBoxView.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationFramework/MS/Internal/documents/TextBoxView.cs
@@ -576,7 +576,6 @@ namespace System.Windows.Controls
             Rect rect;
 
             Invariant.Assert(this.IsLayoutValid);
-            Invariant.Assert(Contains(position));
 
             int offset = position.Offset;
             if (offset > 0 && position.LogicalDirection == LogicalDirection.Backward)
@@ -717,7 +716,6 @@ namespace System.Windows.Controls
         ITextPointer ITextView.GetPositionAtNextLine(ITextPointer position, double suggestedX, int count, out double newSuggestedX, out int linesMoved)
         {
             Invariant.Assert(this.IsLayoutValid);
-            Invariant.Assert(Contains(position));
 
             newSuggestedX = suggestedX;
             int lineIndex = GetLineIndexFromPosition(position);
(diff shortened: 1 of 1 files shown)


gpaskaleva-msft:
@etvorun When you get a chance, could you please rebase this PR on the latest main branch? That will help us move it forward. Thanks!

PR #11465, opened 2026-02-18
WpfGfx: Mitigate RemoveChild crash

### LINKED ISSUE #11464: [WpfGfx] RemoveChild should tolerate already-detached children
# [WpfGfx] RemoveChild should tolerate already-detached children

## Summary
A crash can occur when visual tree removal processes a child that has already been detached from its parent.

## Observed behavior
- In rare timing/order scenarios, `CMilVisual::RemoveChild` receives a child that is no longer present in the parent collection.
- The current path treats this as invalid and fails, which can terminate the process.

## Repro notes
1. Use a rendering-heavy visual tree scenario with frequent attach/detach operations.
2. Trigger composition command processing where a remove command can arrive after the child has already been detached.
3. Observe failure in the `RemoveChild` path when the child is not found in the parent list.

## Impact
- Improves reliability in visual tree mutation paths under composition pressure.
- Reduces crash risk for apps with dynamic visuals and frequent child removal.

## Expected behavior
Removing an already-detached child should be treated as a safe no-op.

## Actual behavior
The code reports invalid argument and fails when the child is absent from the parent collection.

## Suspected root cause
A remove request can target a child that has already been detached, but the method currently treats this state as a hard error rather than an idempotent operation.

## Proposed fix direction
- In `CMilVisual::RemoveChild`, when removal from `m_rgpChildren` fails, check whether `pChild->GetParent()` is `NULL`.
- If already detached, return success and treat the operation as no-op.
- Keep existing invalid-argument behavior for genuinely inconsistent states where the child still reports a parent.

etvorun:
## Summary
Mitigates a crash in visual child removal by making `CMilVisual::RemoveChild` tolerant of already-detached children.

## What changed
- Updated `src/Microsoft.DotNet.Wpf/src/WpfGfx/core/resources/node.cpp` in `CMilVisual::RemoveChild`.
- Added a guard for the case where `m_rgpChildren.Remove(pChild)` fails but `pChild->GetParent()` is `NULL`.
- Return `S_OK` for this detached-child case to make the operation idempotent.

## Why
In practice, child removal can occasionally be requested for a child that is already detached. Treating this as a hard invalid-argument failure can crash the process. Returning success for the already-detached case preserves stability while keeping strict failure behavior for other invalid states.

## Validation
- Manual validation only, per repository workflow.
- No local build or test commands were run.

Fixes #11464

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/wpf/pull/11465)

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/WpfGfx/core/resources/node.cpp
+++ b/src/Microsoft.DotNet.Wpf/src/WpfGfx/core/resources/node.cpp
@@ -320,6 +320,14 @@ CMilVisual::RemoveChild(
 
     if (!m_rgpChildren.Remove(pChild))
     {
+        // We've seen a handful of Watson dumps where incoming child is already detached
+        // from parent and about to be deleted by the next MilCmdChannelDeleteResource
+        // command. There are no known repros. For now we'll pretend we processed it.
+        if (pChild->GetParent() == NULL)
+        {
+            return S_OK;
+        }
+
         IFC(E_INVALIDARG);
     }
 


mmanolova-msft:
@etvorun, could you please rebase the PR with latest changes from the `main` branch? There is a build issue that should be resolved by the recent changes there.

PR #11472, opened 2026-02-20
HwRender: fix ARM64 crash when GPU is suspended during Present (DXGI_ERROR_DEVICE_REMOVED)

### LINKED ISSUE #11471: HwRender: WPF crashes on ARM64 when GPU is suspended during frame presentation (DXGI_ERROR_DEVICE_REMOVED)
# [HwRender] WPF crashes on ARM64 when GPU is suspended during frame presentation

## Summary
On ARM64 devices, WPF crashes when the GPU is suspended (e.g. during power management transitions) while the compositor thread is in the middle of calling `IDirect3DSwapChain9::Present`. The process terminates with a fatal exception rather than recovering gracefully.

## Observed behavior
- WPF application crashes on ARM64 hardware when the GPU is suspended or power-cycled during rendering.
- Fault bucket signature: `XAML_887a0005_wpfgfx_cor3.dll!VS_FatalError`
- `IDirect3DSwapChain9::Present` returns `0x887A0005` (`DXGI_ERROR_DEVICE_REMOVED`).
- The crash is ARM64-specific; x64 is not affected.

## Repro notes
1. Run a WPF application on ARM64 hardware (e.g. Snapdragon-based Windows device) with continuous rendering activity (animations, video, or rapid UI updates).
2. Trigger a GPU power-management event: sleep/resume the device, lock the screen for an extended period, or physically remove/disable the display adapter.
3. Observe the WPF application crash rather than recovering and repainting after the GPU comes back.

## Impact
- Affects any WPF application running on ARM64 hardware (Snapdragon X Elite/Plus, other Windows-on-ARM devices).
- Crash is deterministic when a GPU suspension event occurs during an active Present call.
- Hit count indicates this affects a significant number of ARM64 users in production.
- Issue originally observed in Visual Studio integration scenario.

## Expected behavior
When `IDirect3DSwapChain9::Present` returns an error indicating the GPU was lost or removed, WPF should:
1. Recognize the error as a device-lost condition.
2. Release GPU resources via the existing `MarkUnusable()` path.
3. Skip the current frame and notify the UI thread (`RENDERING_STATUS_DEVICE_LOST`).
4. Automatically recreate GPU resources and repaint the window when the GPU resumes.

No crash, no zombie window.

## Actual behavior
`PresentWithD3D` only checks for `S_OK`, `S_PRESENT_MODE_CHANGED`, and `S_PRESENT_OCCLUDED`. Any other HRESULT falls through to `MIL_THR`, which treats `0x887A0005` as a fatal error, triggering an unrecoverable exception and terminating the process.

## Suspected root cause
On ARM64, the D3D9-over-DXGI compatibility layer surfaces `DXGI_ERROR_DEVICE_REMOVED` (`0x887A0005`) directly when the GPU is suspended or removed, instead of translating it to the legacy D3D9 equivalent `D3DERR_DEVICELOST`. WPF's `PresentWithD3D` does not recognize this DXGI code and passes it unchanged to `MIL_THR`, which classifies it as fatal. `HandlePresentFailure` — which contains the correct device-lost recovery logic — is never reached.

This is an ARM64-specific behavior difference in the D3D9-over-DXGI translation layer.

## Proposed fix direction
- Define `DXGI_ERROR_DEVICE_REMOVED` locally in `wgx_error.h` (as a stable `#define`) to avoid pulling in `dxgi.h`.
- In `PresentWithD3D` (`d3ddevice.cpp`): add an `else if (hr == DXGI_ERROR_DEVICE_REMOVED)` branch — placed **before** the `MIL_THR` call — that converts the error to `D3DERR_DEVICELOST`. This feeds the existing device-lost recovery path.
- In `HandlePresentFailure` (`d3ddevice.cpp`): add `DXGI_ERROR_DEVICE_REMOVED` to the device-lost condition block as defense-in-depth for other callers (e.g. `PresentWithGDI`).

etvorun:
## Summary

On ARM64 devices, WPF crashes when the GPU is suspended during an active `IDirect3DSwapChain9::Present` call. The D3D9-over-DXGI compatibility layer on ARM64 surfaces `DXGI_ERROR_DEVICE_REMOVED` (`0x887A0005`) directly instead of translating it to `D3DERR_DEVICELOST`. WPF's `PresentWithD3D` did not recognize this code; it fell through to `MIL_THR`, which treats `0x887A0005` as fatal and crashes the process. `HandlePresentFailure` — and the device-lost recovery path — was never reached.

## What changed

**`src/Microsoft.DotNet.Wpf/src/WpfGfx/include/wgx_error.h`**
- Added a local `#define DXGI_ERROR_DEVICE_REMOVED ((HRESULT)0x887A0005L)` guarded by `#ifndef`, avoiding a new dependency on `dxgi.h`. The value is a stable DirectX ABI constant.

**`src/Microsoft.DotNet.Wpf/src/WpfGfx/core/hw/d3ddevice.cpp`**
- `PresentWithD3D`: new `else if (hr == DXGI_ERROR_DEVICE_REMOVED)` branch placed **before** the `MIL_THR` call, converting the error to `D3DERR_DEVICELOST` so the existing device-lost recovery path handles it.
- `HandlePresentFailure`: added `DXGI_ERROR_DEVICE_REMOVED` to the device-lost `if` condition, as defense-in-depth for other callers such as `PresentWithGDI`.

## Why

The fix is aligned with the previously validated internal source fix behavior. The change is surgical: no new recovery logic is introduced; WPF reuses the existing TDR / device-lost path (`MarkUnusable()` → `WGXERR_DISPLAYSTATEINVALID` → `RENDERING_STATUS_DEVICE_LOST` → resource recreation on GPU resume).

### Recovery sequence

| State | What happens |
|---|---|
| `Present` returns `DXGI_ERROR_DEVICE_REMOVED` | Converted to `D3DERR_DEVICELOST`; `HandlePresentFailure` calls `MarkUnusable()`, invalidates GPU resources, notifies device manager |
| Same frame — render thread | `CRenderTargetManager::HandlePresentErrors` swallows `WGXERR_DISPLAYSTATEINVALID`; fires `RENDERING_STATUS_DEVICE_LOST` to UI thread. No crash, no zombie. |
| Subsequent frames (GPU suspended) | `UpdateDisplayState` returns `WGXERR_DISPLAYSTATEINVALID`; render passes are skipped |
| GPU comes back | `UpdateDisplayState` succeeds; `NotifyTierChange` fires; render targets recreated; window repainted |

## Validation

Validated manually on ARM64 hardware by triggering GPU suspension events while a WPF application is actively rendering. The application recovered without crashing after applying this fix. No automated test infrastructure changes are included.

Fixes #11471

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/wpf/pull/11472)

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/WpfGfx/core/hw/d3ddevice.cpp
+++ b/src/Microsoft.DotNet.Wpf/src/WpfGfx/core/hw/d3ddevice.cpp
@@ -3378,6 +3378,17 @@ CD3DDeviceLevel1::PresentWithD3D(
 
         hr = S_OK;
     }
+    else if (hr == DXGI_ERROR_DEVICE_REMOVED)
+    {
+        //
+        // On ARM64, the D3D9-over-DXGI compatibility layer surfaces
+        // DXGI_ERROR_DEVICE_REMOVED directly when the GPU is suspended or
+        // removed, instead of mapping it to D3DERR_DEVICELOST.
+        // Convert here so HandlePresentFailure can recognize and handle it
+        // gracefully rather than letting MIL_THR trigger a fatal crash.
+        //
+        hr = D3DERR_DEVICELOST;
+    }
 
     //
     // !!! Critical Note: After this point hr may not be S_OK.  Make sure not
@@ -3477,9 +3488,11 @@ CD3DDeviceLevel1::HandlePresentFailure(
         hr == D3DERR_DEVICEHUNG ||  // Hw Adapter timed out and has
                                     // been reset by the OS (LH Only)
                                     //
-        hr == D3DERR_DEVICEREMOVED  // Hw Adapter has been removed (LH
-                                    // Only)
-                                    //
+        hr == D3DERR_DEVICEREMOVED ||  // Hw Adapter has been removed (LH
+                                       // Only)
+                                       //
+        hr == DXGI_ERROR_DEVICE_REMOVED // surfaced directly by D3D9-over-DXGI
+                                        // on ARM64 when GPU is suspended or removed
         )
     {
         hr = WGXERR_DISPLAYSTATEINVALID;
--- a/src/Microsoft.DotNet.Wpf/src/WpfGfx/include/wgx_error.h
+++ b/src/Microsoft.DotNet.Wpf/src/WpfGfx/include/wgx_error.h
@@ -12,6 +12,16 @@
 #pragma once
 
 /*=========================================================================*\
+    DXGI error codes
+    Defined here to avoid a dependency on dxgi.h, which is not included
+    in WpfGfx. These values are stable ABI constants defined by DirectX.
+\*=========================================================================*/
+
+#ifndef DXGI_ERROR_DEVICE_REMOVED
+#define DXGI_ERROR_DEVICE_REMOVED   ((HRESULT)0x887A0005L)
+#endif
+
+/*=========================================================================*\
     MIL Status Codes
 \*=========================================================================*/
 


lindexi (approved):
LGTM

amarinov-msft:
@etvorun
Just two potential improvements:
Your fix only covers `Present`, not the other D3D9 call sites that share the same root cause. The same reasoning applies to every other D3D9 call:
- `CheckDeviceState` checks `D3DERR_DEVICELOST/DEVICEHUNG/DEVICEREMOVED` but not `DXGI_ERROR_DEVICE_REMOVED`. This is the API whose whole job is polling device state; if it returns the raw DXGI code on `ARM64`, device-loss recovery is missed.
-Several calls ( `CreateTexture`, `GetRenderTargetData` , etc.) do `MIL_THR(hr)` then route through `HandleDIE(hr)`, and `HandleDIE` only maps `D3DERR_DRIVERINTERNALERROR` . A raw `DXGI_ERROR_DEVICE_REMOVED` from any of these propagates unconverted.
If suspension is caught mid-frame by a draw/resource call or by `CheckDeviceState`, the crash can still occur. The most robust fix would normalize `DXGI_ERROR_DEVICE_REMOVED`  to `D3DERR_DEVICELOST` centrally (e.g., in `HandleDIE` / the `MIL_THR` normalization) rather than per-site. At minimum, `CheckDeviceState` should get the same branch.

PR #11509, opened 2026-03-07
Fix screen readers reading the wrong line on blank lines in textboxes (closes #7190)

### LINKED ISSUE #7190: Screen-reader reads lines twice in a TextBox / TextBlock / RichTextBox
* .NET Core Version: 7.0.0-rc.1.22426.10
* Windows version: 22H2
* Is this bug related specifically to tooling in Visual Studio? No

 **Problem description:**
 When reviewing line by line a multi-line TextBox with a screen-reader (Narrator or NVDA), the screen-reader will always read a line of text twice as long as there is a blank line above said line of text. The user will either think that the line of text is present twice or that the text navigation is broken.

 **Actual behavior:**
 When using down arrow to go line by line in a TextBox, Narrator / NVDA will read some lines of text twice, even though the line is there only once.

 **Expected behavior:**
 Just like any other text edit, the screen-reader only reads the current line of text once when cursor moves down or up by one line.

 **Minimal repro:**
1. Setup the WPF project
Option 1: Download [TextBoxUIABugRepro.zip](https://github.com/dotnet/wpf/files/9761599/TextBoxUIABugRepro.zip) and open the .sln file in Visual Studio.
Option 2: Create a new WPF project, and in MainWindow.xaml.cs add the onInitialize override with the following code:

```
protected override void OnInitialized(EventArgs e)
        {
            base.OnInitialized(e);
            var tb = new TextBox();
            tb.AcceptsReturn = true;
            var layout = new StackPanel();
            layout.Children.Add(tb);
            this.Content = layout;
            tb.Text = "Hello, this is a test\r\n\r\nThis is a second line of text, you will notice that if you move focus to the line above this line, you will also hear this line, this makes Narrator read it twice.\r\n\r\nAgain, you will hear this third line twice";
            tb.Focus();
        }
```

2. Start debugging
3. Put your text cursor at the very beginning of the textbox (Ctrl+Home).
4. Start Narrator (Ctrl+Win+Enter, or from Start menu). You can also use NVDA.
5. With focus in the textbox, press the down arrow
6. Notice that when you are on line 2, which is blank, Narrator / NVDA will read the third line of text.
7. Press the down arrow again
8. This time Narrator and NVDA read line 3 correctly, reading line 3 for the second time.
9. Keep pressing down and up arrow to notice how Narrator / NVDA will read the line of text below a blank line instead of reading "blank".

 **Additional information**
I might be wrong in this section, I just thought since I did more investigation I would put it here in case it's useful.

This seems like a UIA provider bug.

When debugging WPF, I noticed that in [TextRangeAdapter.cs](https://github.com/dotnet/wpf/blob/main/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Controls/TextRangeAdaptor.cs) --> line 522: `private void ExpandToEnclosingUnit(TextUnit unit, bool expandStart, bool expandEnd)`
- When cursor is on line 2, _start is right at the start of line 2, however _end is at the start of line 3. When the line unit expansion is done, the text range now covers both lines 2 and 3.
- This is the symptom however, the problem must come from how _start and _end state is managed. The expansion is doing the right thing here, except that _end should not be on line 3, but on line 2.

Other notes:
- This repro uses Windows style line endings (CrLf, \r\n), however the issue repros no matter what line ending used (\r,\n,\r\n).
- This issue is not limited to TextBox, but also TextBlock, RichTextBox, and I'm guessing anything using the TextRangeAdapter or at least TextView.
- Although this is a WPF bug and not a NVDA bug, this issue was filed in 2011 on NVDA's repo, here it is for reference: [NVDA reads text in a standard WPF text box incorrectly](https://github.com/nvaccess/nvda/issues/1412)

trypsynth:
Fixes #7190

## Description

When navigating a TextBox or RichTextBox line-by-line with a screen reader (e.g. Narrator, NVDA), blank lines caused the next line's content to be announced instead of "blank", and that next line would then be announced again on the following navigation.

Root cause: In TextRangeAdaptor.ExpandToEnclosingUnit(TextUnit.Line), blank lines produce a degenerate lineRange where Start == End (ContentLength == 0). The condition lineRange.Contains(_end) fails when _end has already advanced past the blank line, so snapEndPosition stays true and the expandEnd block runs, bleeding _end into the next non-blank line. The resulting range spans both the blank line and the next line, so GetText() returns the next line's content.

Fix: add an additional guard clause for the degenerate case: When lineRange.Start == lineRange.End (blank line), snapEndPosition is set to false and _end is snapped to lineRange.End (same position), preventing expandEnd from running. GetText() returns "" and screen readers correctly announce "blank".

Covers TextBox (TextBoxView.GetLineRange returns [offset, offset] for blank lines) and RichTextBox (TextDocumentView returns a degenerate segment for empty paragraphs).

## Customer Impact

Incredibly inconsistent and confusing behavior for screen reader users in WPF text boxes.

## Regression

No, this has existed since at least 2011, when the issue was first raised against NVDA.

## Testing

Built WPF, coppied in the required DLLs into a new wpf app created with `dotnet new`, added a text box with blank lines and content, and confirmed it reads correctly with NVDA, my screen reader.

## Risk

None

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/wpf/pull/11509)

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Controls/TextRangeAdaptor.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationFramework/System/Windows/Controls/TextRangeAdaptor.cs
@@ -607,8 +607,11 @@ namespace MS.Internal.Automation
                                     _start = lineRange.Start.CreatePointer();
                                 }
                                 // If this line contains also end position, move it to the
-                                // end of this line.
-                                if (lineRange.Contains(_end))
+                                // end of this line. Also handle blank/empty lines where
+                                // lineRange.Start == lineRange.End (ContentLength == 0):
+                                // snap _end to the blank line boundary so expandEnd doesn't
+                                // bleed into the next line.
+                                if (lineRange.Contains(_end) || lineRange.Start.CompareTo(lineRange.End) == 0)
                                 {
                                     snapEndPosition = false;
                                     if (_end.CompareTo(lineRange.End) != 0)


trypsynth:
Just quickly pinging @dipeshmsft as this has been sitting in the queue for a while, although I bet you're a busy person. Would love to get this merged though so WPF can be more accessible! Cheers!

trypsynth:
Hellooo? Is WPF still having community PRs reviewed?

PR #11521, opened 2026-03-11
Fix an InvalidOperationException when WeakEventTable.Purge is called …

### LINKED ISSUE #6452: WeakEventManager lock issue

kmaddock:
…with EnableWeakEventMemoryImprovements enabled

Fixes #6452

## Description

If EnableWeakEventMemoryImprovements is true, the WeakEventTable.Purge enumerates through _dataTable calling WeakEventManager.Purge on each list:
```
IDictionaryEnumerator ide = _dataTable.GetEnumerator() as IDictionaryEnumerator;
while (ide.MoveNext())
{
    ...
    foundDirt |= key.Manager.PurgeInternal(source, ide.Value, purgeAll);
    ...
}
```

WeakEventManager.PurgeInternal redirects to .Purge, which will check the ListenerList for concurrent readers, and clone and update Table (_dataTable) if concurrent readers exist:
```
if (ListenerList.PrepareForWriting(ref list) && source != null)
{
    Table[this, source] = list;
}

``` 

Writing to _dataTable invalidates the enumerator in WeakEventTable, which will throw an InvalidOperationException.

This PR exits WeakEventManager.Purge if EnableWeakEventMemoryImprovements is true, the list purge will be retried next time WeakEventTable is purged, 

## Customer Impact

Our App throws InvalidOperationException and crashes.

## Regression

No

## Testing

I reprod with the test case attached to issue #6452 and confirmed that this PR fixes it.

## Risk

It seems pretty low risk to me

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/wpf/pull/11521)


vinnarayana-msft:
@kmaddock  Thank you for this contribution and for identifying the root cause. The analysis that `Table[this, source] = list` invalidates the enumerator in `WeakEventTable.Purge` is correct, and the early return approach does resolve the crash.

One area worth exploring: since the return occurs before `list.Purge()` executes, dead entries in the listener list are not removed during the current purge cycle and are deferred to the next timer tick. In scenarios involving high-frequency event delivery where concurrent readers (`_users > 0`) are consistently present during purge, this could result in repeated deferral of cleanup for the same entry. Additionally, the `removeList` check and `StopListening` call are also skipped for that cycle.

An alternative approach that prevents the crash while preserving full cleanup in a single cycle would be to defer the table **write** rather than the entire purge operation. This can be done by extending the existing `_toRemove` pattern in `WeakEventTable.cs`:

1. Introduce a `_toUpdate` list alongside the existing `_toRemove` list.
2. Modify the indexer setter to check `_inPurge` — if true, enqueue the update to `_toUpdate` rather than writing directly to `_dataTable`.
3. After enumeration completes, apply the entries from `_toUpdate` before processing `_toRemove`.

```csharp
// WeakEventTable.cs — indexer setter
set
{
    EventKey key = new EventKey(manager, source, true);
    if (!_inPurge)
    {
        _dataTable[key] = value;
    }
    else
    {
        _toUpdate.Add(new KeyValuePair<EventKey, object>(key, value));
    }
}
```

With this approach, `list.Purge()`, the `removeList` evaluation, and `StopListening` all continue to execute normally — only the `_dataTable` mutation is deferred. The change is confined entirely to `WeakEventTable.cs`, requiring no modifications to `WeakEventManager.cs`.

Please evaluate this approach and would be happy to discuss this further.

kmaddock:
@dotnet-policy-service agree

kmaddock:
I have updated the PR to add a _toUpdate list and preserve full cleanup within a cycle.

I have no opinion on what the correct thing to do is, they both fix the bug

src/Microsoft.DotNet.Wpf/src/WindowsBase/MS/Internal/WeakEventTable.cs
```diff
@@ -355,6 +363,16 @@ private bool Purge(bool purgeAll)
                     _toRemove.TrimExcess();
                 }
 
+                if (_toUpdate.Count > 0)
```
vinnarayana-msft: Thanks for adopting the deferred-update approach — the implementation looks good overall!

One issue I noticed: the `_toUpdate` block is applied **after** `_toRemove`, but it should be **before**. This matters when the same entry ends up in both lists, which happens when a cloned list becomes empty after `list.Purge()`:

1. `PrepareForWriting` clones the list → `Table[this, source] = clone` → added to `_toUpdate`
2. `list.Purge()` removes all dead refs from the clone
3. `list.IsEmpty == true` → `Table.Remove(this, source)` → added to `_toRemove`

With the current ordering (remove then update), the entry is correctly removed from `_dataTable`, but then immediately re-inserted by `_toUpdate` with the empty clone — creating a zombie entry that will never be cleaned up.

If `_toUpdate` is applied first, the entry gets updated with the clean clone, and then `_toRemove` removes it — which is the correct outcome.

I think we are good to merge the pr after this change

Suggested fix: move the `_toUpdate` block above the `_toRemove` / `purgeAll` block.

vinnarayana-msft (dismissed):
LGTM

vinnarayana-msft:
@kmaddock we need to do an internal testing before merging this pr. Can you rebase your pr with latest changes so we can perform the testing and can merge the pr post that.

vinnarayana-msft (approved):
LGTM

harshit7962 (approved):
Changes LGTM

vinnarayana-msft:
Thank you so much @kmaddock for your contribution and we really appreciate you for putting your time and effort for helping us in identifying and resolving this issue. I am merging this pr now.

kmaddock:
Thanks!

When will a new version(s) be released? Will it be released to patch (.NET 10) versions?

Camios:
@vinnarayana-msft / @harshit7962
We're currently on dotnet 10 LTS, and we have numerous customers getting this problem and need a fix. 
I don't know where to look, how do we tell if and when a fix (like this) will make it to a particular dotnet version?

PR #11575, opened 2026-04-13
Remove /clr:pure from WPF C++/CLI projects

vinnarayana-msft:
## Description

MSVC 14.51 (VS 18.3) is removing support for `/clr:pure` when used with `/clr:netcore`. WPF is the only known consumer of `/clr:pure`. This PR migrates the two affected C++/CLI projects (`DirectWriteForwarder.vcxproj` and `System.Printing.vcxproj`) from `/clr:pure` to standard `/clr:netcore` (IJW mixed mode).

**Changes:**
- Changed `CLRSupport` and `CompileAsManaged` from `pure` to `NetCore` in both vcxproj files
- Updated `Wpf.Cpp.props` to remove the `pure` to `NetCore` conversion logic in `ManagedCxx` conditions
- Removed `ijwhost.dll` from packaging exclusions in both vcxproj files — mixed-mode assemblies require this .NET runtime bootstrap shim to load correctly
- Fixed ARM64 build break in `intsafe_private_copy.h` where `_M_ARM64` was missing from an architecture detection `#if` condition (previously masked by `_M_CEE_PURE`)

## Customer Impact

Without this fix, WPF will fail to build with MSVC 14.51+ (VS 18.3) since `/clr:pure` + `/clr:netcore` will no longer be a valid combination. This is a build-time breaking change from the MSVC toolset.

## Regression

No. This is a proactive change to address an upcoming MSVC toolset deprecation. No existing behavior is regressed — the output DLLs function identically at runtime.

## Testing

| Test | Result |
|------|--------|
| x64 build | ✅ Pass |
| ARM64 build compilation | ✅ Pass |
| WPF Gallery app run (x64) | ✅ Pass |

## Risk

**Low-Medium.** The change is minimal (4 files) and surgical — it only affects the two C++/CLI projects in WPF. The runtime behavior of the DLLs is functionally equivalent; the difference is the compilation mode (pure MSIL → IJW mixed mode). The primary risk is potential performance impact from losing ahead-of-time MSIL compilation.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/wpf/pull/11575)


miloush:
Can we have some link/reference to the deprecation of the combination?

vinnarayana-msft:
> Can we have some link/reference to the deprecation of the combination?
@miloush 

Please refer to this [page ](https://learn.microsoft.com/en-us/cpp/build/reference/clr-common-language-runtime-compilation?view=msvc-170)
This page says:

/clr:pure is deprecated. The option is removed in Visual Studio 2017 and later. We recommend that you port code that must be pure MSIL to C#.

The driver also gives a deprecation warning if /clr:pure is used:

cl : Command line warning D9035 : option 'clr:pure' has been deprecated and will be removed in a future release

vinnarayana-msft:
> @vinnarayana-msft please read the following Contributor License Agreement(CLA). If you agree with the CLA, please reply with the following information.
> 
> ```
> @dotnet-policy-service agree [company="{your company}"]
> ```
> 
> > Options:
> > 
> > * (default - no company specified) I have sole ownership of intellectual property rights to my Submissions and I am not making Submissions in the course of work for my employer.
> > 
> > ```
> > @dotnet-policy-service agree
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > 
> > * (when company given) I am making Submissions in the course of work for my employer (or my employer has intellectual property rights in my Submissions by contract or applicable law). I have permission from my employer to make Submissions and enter into this Agreement on behalf of my employer. By signing below, the defined term “You” includes me and my employer.
> > 
> > ```
> > @dotnet-policy-service agree company="Microsoft"
> > ```
> 
> Contributor License Agreement

@dotnet-policy-service agree company="Microsoft"

src/Microsoft.DotNet.Wpf/src/DirectWriteForwarder/CPP/TrueTypeSubsetter/TtfDelta/intsafe_private_copy.h
```diff
@@ -19,7 +19,7 @@ typedef unsigned __int64    ULONGLONG;
 //
 #if defined(MIDL_PASS) || defined(RC_INVOKED) || defined(_M_CEE_PURE) \
     || defined(_68K_) || defined(_MPPC_) \
-    || defined(_M_IA64) || defined(_M_AMD64)
+    || defined(_M_IA64) || defined(_M_AMD64) || defined(_M_ARM64)
```
dipeshmsft, line 22: What is the purpose of this change ?

eng/WpfArcadeSdk/tools/Wpf.Cpp.props
dipeshmsft: nit: We can probably remove CLRSupport==pure here.

src/Microsoft.DotNet.Wpf/src/DirectWriteForwarder/DirectWriteForwarder.vcxproj
```diff
@@ -92,12 +92,7 @@
   <ItemGroup>
     <AdditionalPackageReference Include="$(SystemIOPackagingPackage)" Version="$(SystemIOPackagingVersion)" />
   </ItemGroup>
-  <!--
-    Ignore ijwhost.dll during packaging
-  -->
-  <ItemGroup>
```
harshit7962, line 98: Just double checking if `ijwhost` is not duplicated? Since the runs have passed, I am supposing it fine.

harshit7962 (dismissed):

src/Microsoft.DotNet.Wpf/src/DirectWriteForwarder/CPP/TrueTypeSubsetter/TtfDelta/intsafe_private_copy.h
vinnarayana-msft: With /clr:pure, _M_CEE_PURE was always defined, so the #if on line 22 was true for all architectures including ARM64. After switching to /clr:netcore, _M_CEE_PURE is no longer defined, so ARM64 builds would fall through to #error Must define a target architecture (line 34). Adding _M_ARM64 fixes this

eng/WpfArcadeSdk/tools/Wpf.Cpp.props
vinnarayana-msft: Good catch! Addressed in commit 2aad7eea4 — removed the dead CLRSupport=='pure' condition from DllRenameClause and updated the comment to reference /clr:netcore

src/Microsoft.DotNet.Wpf/src/DirectWriteForwarder/DirectWriteForwarder.vcxproj
vinnarayana-msft: The original code had an explicit <FileNamesExcludedFromPackaging Include="ijwhost.dll" /> — this means the SDK was already producing `ijwhost.dll` during build, but WPF was actively suppressing it from the package because /clr:pure didn't need it.

Our change simply removed that exclusion, allowing the single SDK-produced ijwhost.dll to flow through to the package. There's no second source adding another copy — the ResolvedIjwHostPack in RuntimeFrameworkReference.targets is the standard .NET SDK mechanism that resolves it once per build.

dipeshmsft (dismissed):

himgoyalmicro (approved):

dipeshmsft (approved):

PR #11621, opened 2026-05-05
DWM Crash Fix Backport

himgoyalmicro:
Main PR: https://github.com/dotnet/wpf/pull/11417

## Description

It’s possible for DWM to crash or exit unexpectedly, and when that happens, DwmExtendFrameIntoClientArea returns a failure.

This change adds a safety check to prevent that failure from propagating and causing a crash in WPF and reverting the app to the previous behavior.

We will only crash if we encounter Invalid Args exception.

## Customer Impact

## Regression
NA

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/wpf/pull/11621)


harshit7962 (approved):
LGTM

pranav-gupta-msft (approved):
Backporting Fix Only

PR #11641, opened 2026-05-14
Fix FatalExecutionEngineError on insert at start of soft-wrapped line in TextBox (#11481)

### LINKED ISSUE #11481: Crash in TextBoxView.cs related to wrapping.
### Description

I created this small application so that I could report the problem (see the reproduction steps). The callstack is different in this particular case to the crash that I am having in my real application, but the origin is the same:

- Text in a TextBox that consists of a single space followed by a long word so that the text wrapping results in a top line with a single space and 2 more lines.
- Type a space at the beginning of the 3rd line so that the text wrapping will now result in 2 lines, the 1st line consisting of a space followed by the characters that were previously on the 2nd line, and the 2nd line consisting of the characters that were previously on the 3rd line.

The crashes are always in TextBoxView.cs.

What seems to happen is that the space gets positioned at the end of the 2nd line, so we have a 1st line with a single space, a 2nd line with o's then a space, then a 3rd line with the remaining o's. If the wrapping algorithm was applied to the whole text then we would have a 1st line with a space following by o's, then a 2nd line with the remaining o's. When something happens that uses the wrapping algorithm on the beginning of the text, the result does not match the current line metrics causing a crash.

Is it possible to have a fix for this problem?

Thanks.

### Reproduction Steps

Create a WPF application with this MainWindow:

`
<Window x:Class="MyTestApp.MainWindow"
        xmlns:local="clr-namespace:MyTestApp"
        mc:Ignorable="d"
        Title="MainWindow" Height="200" Width="200">
    <Grid>
        <Grid.RowDefinitions>
            <RowDefinition Height="Auto" />
        </Grid.RowDefinitions>
        <TextBox
            x:Name="MyTextBox"
            Grid.Row="1"
            HorizontalAlignment="Stretch"
            VerticalAlignment="Stretch"
            VerticalContentAlignment="Top"
            FontSize="14.0"
            TextAlignment="Left"
            TextWrapping="Wrap">
        </TextBox>
    </Grid>
</Window>
`

1. Run application (and DO NOT resize the window).
2. Paste the text that is within the quotes here " oooooooooooooooooooooooo". (The result is 3 lines: first line has a single space, the second has lots of o characters and the last has 3 o characters.
3. Place the cursor at the start of the third line and type a single space.
4. Move the mouse cursor around the window (specifically over the first line).
5. Application crashes with a FatalExecutionEngineError.

### Expected behavior

No crash.

### Actual behavior

In our application, the problem is a hard crash and our application simply disappears with no message.

### Regression?

_No response_

### Known Workarounds

We have no good workaround. It may seem that the workaround would be to not allow users to enter a second space in the TextBox, but the bug occurs in other circumstances i.e. if we have a mix of Arabic and Latin text, the crash can occur by entering a Latin character within the Arabic text.

### Impact

_No response_

### Configuration

_No response_

### Other information

For your information, this particular TextBox is used to enter a character set for the creation of a font file. So in this case it is normal to have a long 'word' without spaces in between.

wnvko-msft:
Fixes #11481 

## Description

A `TextBox` with `TextWrapping="Wrap"` whose content includes a single long token that soft-wraps across multiple visual lines could crash the host process with a `FatalExecutionEngineError` when the user typed a character at the start of a soft-wrap continuation line.

`TextBoxView.IncrementalMeasureLinesAfterInsert` only remeasured starting at the directly inserted line, which left the preceding soft-wrap `LineRecord` entries pointing at stale character offsets. A subsequent measure pass read those stale records and corrupted internal state, producing the engine error.

The fix walks back over preceding soft-wrap predecessors (lines whose `Length == ContentLength`, i.e. lines that wrapped because text ran out of width rather than because of a hard break) before remeasuring, so the entire wrapped paragraph is re-formatted as a unit. The per-line work is extracted into a new `FormatIncrementalLine` helper that updates the line record and invalidates the cached visual when the line's offset or length changed.

## Customer Impact

Without this fix, any WPF application that hosts a wrapping `TextBox` containing long unbreakable tokens (URLs, identifiers, generated strings, certain non-Latin scripts, etc.) is exposed to a hard process crash during normal text entry. `FatalExecutionEngineError` cannot be caught by user code, so the application terminates immediately with no opportunity for the user to save work.

## Regression

No. The defective logic in `IncrementalMeasureLinesAfterInsert` has been present in `TextBoxView` since the original incremental-measure implementation; this is not introduced by a recent release.

## Testing

- Manual repro from issue #11481 confirmed to crash before the change and to insert/wrap correctly after the change: a `TextBox` with `TextWrapping="Wrap"`, `Width="60"`, `FontFamily="Segoe UI"`, `FontSize="12"`, and `Text=" oooooooooooooooooooooooo"` — place the caret at the start of the second visual line and press space.
- Exercised additional offsets across the wrapped run (start, middle, end of each soft-wrap line) - no crashes; line metrics remain consistent with the displayed text.
- Full PresentationFramework build under `Debug|x64` passes; existing `PresentationFramework.Tests` suite passes.
- An automated regression test was not added: the only available test host for this code (`PresentationFramework.Tests`, xunit.stafact / `[WpfFact]`) provides an STA thread but does not pump a `Dispatcher`, and the bug requires a fully realized `TextBoxView` (template applied, hosted in a window, layout driven by the dispatcher) to reproduce. Attempts to drive the scenario in that host either degenerated into no-op smoke tests or hung the test runner.

## Risk

Low.
- Scope is confined to one method (`IncrementalMeasureLinesAfterInsert`) plus a small, extracted helper (`FormatIncrementalLine`) in `TextBoxView`. No public API changes, no behavior changes for the non-soft-wrap path.
- In the common case (insertion on a line that is not a soft-wrap continuation), the walk-back loop exits immediately on the first iteration, so the post-fix code path is equivalent to the previous code path.
- When walk-back does occur, it is bounded by the existing `_lineMetrics` list and terminates as soon as a non-soft-wrap predecessor is found (or the first line is reached). It cannot run beyond the current paragraph.
- The new helper preserves the existing line-format / line-record-update sequence and only adds an explicit visual invalidation when the line's offset or length actually changed, which mirrors the implicit invalidation behavior the old inline code relied on.

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/wpf/pull/11641)

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/PresentationFramework/MS/Internal/documents/TextBoxView.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationFramework/MS/Internal/documents/TextBoxView.cs
@@ -2346,43 +2346,45 @@ namespace System.Windows.Controls
                 }
             }
 
+            // Walk back through soft-wrap predecessors (lines without a hard break:
+            // Length == ContentLength) so the whole wrapped paragraph is re-formatted.
+            int firstAffectedLineIndex = lineIndex;
+            if (firstAffectedLineIndex > 0)
+            {
+                firstAffectedLineIndex--;
+                while (firstAffectedLineIndex > 0 &&
+                       _lineMetrics[firstAffectedLineIndex - 1].Length == _lineMetrics[firstAffectedLineIndex - 1].ContentLength)
+                {
+                    firstAffectedLineIndex--;
+                }
+            }
+
             TextBoxLine line = new TextBoxLine(this);
-            int lineOffset;
+            int lineOffset = _lineMetrics[firstAffectedLineIndex].Offset;
             bool endOfParagraph = false;
 
-            // We need to re-format the previous line, because if someone inserted
-            // a hard break, the first directly affected line might now be shorter
-            // and mergeable with its predecessor.
-            if (lineIndex > 0) // we can skip this if line wrap is disabled.
+            // Re-format each line from firstAffectedLineIndex through lineIndex,
+            // absorbing any successor lines fully covered by the new line.
+            int idx = firstAffectedLineIndex;
+            while (idx <= lineIndex && !endOfParagraph)
             {
-                FormatFirstIncrementalLine(lineIndex - 1, constraintWidth, lineProperties, line, out lineOffset, out endOfParagraph);
-            }
-            else
-            {
-                lineOffset = _lineMetrics[lineIndex].Offset;
-            }
+                FormatIncrementalLine(idx, constraintWidth, lineProperties, line, ref lineOffset, out endOfParagraph);
 
-            // Format the line directly affected by the change.
-            // If endOfParagraph == true, then the line was absorbed into its
-            // predessor (because its new content is thinner, or because the
-            // TextWrapping property changed).
-            if (!endOfParagraph)
-            {
-                using (line)
+                while (idx + 1 < _lineMetrics.Count && lineOffset >= _lineMetrics[idx + 1].EndOffset)
                 {
-                    line.Format(lineOffset, constraintWidth, constraintWidth, lineProperties, _cache.TextRunCache, _cache.TextFormatter);
-
-                    _lineMetrics[lineIndex] = new LineRecord(lineOffset, line);
-
-                    lineOffset += line.Length;
-                    endOfParagraph = line.EndOfParagraph;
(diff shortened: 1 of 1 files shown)


wnvko-msft:
@dotnet-policy-service agree company="Microsoft"

mmanolova-msft (approved):

Issue #10676, opened 2025-03-30
Unnecessary delegate/closure allocations in UIElementAutomationPeer


h3xds1nz:
_Note that I will submit a PR in the coming days for this issue, so its more of an FYI._

### Description

`GetChildrenCore` and `GetRootAutomationPeer` both allocate a new delegate and closure each time they execute with the call to `iterate` function that the callback is passed into. Given that both capture a single local parameter, we can easily just modify the function signature and pass the parameter by `ref`, saving easily 220 MB in this case.

### Showcase

![Image](https://github.com/user-attachments/assets/b1e7d42c-d972-4343-a77b-4239d16c748c)

### Repro

Note: This repro is specifically crafted to highlight an issue like this.

https://github.com/h3xds1nz/UIAPerformancePitfalls

PR #11657, opened 2026-05-21
Disconnect stale automation peers from UIA to fix memory leak in virtualized ItemsControls

### LINKED ISSUE #11337: WPF leaks ElementProxy instances when UI Automation is used
### Description

```
   ...
```

### Reproduction Steps

Start 
   WPFUI.exe
Then start
  WPFAutomationClient.exe

Image

### Expected behavior

No leaks in WPFUI process. 

### Actual behavior

```
   {
       {
           {
           }
       }
   }
```

### Regression?

No it is there since .NET 4.8

### Known Workarounds

```
```

```
WPFUI.exe -disconnect    
```

Image

### Impact

### Other information

```

```

amarinov-msft:
Fixes dotnet#11337

## Description
When a UIA client (screen reader, test-automation, inspector) repeatedly enumerates a virtualized `ItemsControl` (e.g. a large `DataGrid`), WPF accumulates two kinds of never-released objects, producing an unbounded leak:

 1. **AutomationPeers** — `ElementProxy` held its peer with a **strong** reference, and the UIA CCW held the proxy, so every peer surfaced to a client (and the data row / visual sub-tree it roots) was pinned forever.
 2. **`ElementProxy` CCWs** — even once a peer is no longer referenced, the proxy itself stays rooted by its UIA COM-callable wrapper (CCW), dragging its `WeakReference`, runtime-id and name strings with it.

This PR addresses both:

- **Weakly-held peers.** `ElementProxy` now holds `UIElement`/`ContentElement`/ `UIElement3D` peers - and newly **data-item** peers - via `WeakReference`, so UIA references no longer pin recycled/virtualized peers. Data-item behavior is gated by the `Switch.System.Windows.Automation.Peers.UseStrongReferenceForItemAutomationPeers` AppContext switch (default `false` = weak; set `true` to restore legacy strong refs).
- **Keep-alive window.** Because a weakly-held data-item peer could otherwise be collected *in the middle of* a live UIA traversal (surfacing `ElementNotAvailableException`), each peer surfaced to or touched by UIA is strong-rooted in a small rotating bucket for a bounded time window (one-to-two windows ≈ 9–18 s) via a `Background`-priority    `DispatcherTimer`, then released. Retention is bounded by the in-flight working set, not the data-set size.
- **Proxy/CCW disconnect.** A second `Background`-priority `DispatcherTimer` sweep (~2 s cadence, capped at 5000 disconnects/tick) calls the new `UiaDisconnectProvider` on `ElementProxy` instances **whose weak peer has already been collected**, releasing the CCW so the proxy and everything it roots can be reclaimed. Gated by `Switch.System.Windows.Automation.Peers.DisableUiaProviderDisconnect` (default `false` = sweep enabled). Both timers self-stop when there is nothing to track.

A new public API, `AutomationInteropProvider.DisconnectProvider(IRawElementProviderSimple)`, is added (with matching reference-assembly entry) to expose `UiaDisconnectProvider` to the bridge.

 UIA clients may observe `ElementNotAvailableException` when accessing an element whose peer has been collected/disconnected — standard UIA contract behavior that well-behaved clients already handle.

## Customer Impact
WPF applications that expose a virtualized `ItemsControl` (DataGrid, ListView, etc.) to UI Automation experience an unbounded memory leak (≈260 MB/min in the customer's 200k-row DataGrid repro), ending in `OutOfMemoryException` and crashes in long-running sessions. Any app with accessibility active (assistive technology or test automation present) is affected. The customer is currently held on a workaround.

## Regression

No.

## Testing

Controlled customer repro (200k-row DataGrid + a UIA client driving `FindAll` ~1×/s), memory sampled after forced Gen2 GC:

- **Baseline (no fix): ** ≈ +260 MB/min; `ElementProxy` and data-item peers grow without bound.
- **Peer fix only (weak refs + keep-alive): ** AutomationPeers bounded (e.g. `DataGridCellItemAutomationPeer` flat/declining), but `ElementProxy` CCWs still grew (~3.7× over 15 min).
- **Full fix (peer fix + proxy disconnect sweep): ** `ElementProxy` bounded over a 12-min run (2.5k ->3.6k, not growing); process working set flat (≈614–892 MB); no OOM.
- Kill-switch paths verified: setting either `AppContext` switch restores the prior behavior.

## Risk

**Moderate, well-contained, and fully reversible via two AppContext kill-switches.** 
The risk areas are:

1. **Weak data-item peers (behavioral default change).** Data-item peers are now held weakly by default. Any code that implicitly relied on a peer staying alive purely because UIA had seen it could see it collected earlier. Mitigated by the keep-alive window and fully reversible via `UseStrongReferenceForItemAutomationPeers=true`.
2. **Keep-alive retention window.** Peers are deliberately retained for ≈9–18 s after their last UIA touch. This is the main new behavioral trade-off: a small, bounded increase in peak retention in exchange for safety against mid-walk collection. The 9 s window is sized around 2x the measured worst-case gap (around 4.4 s under forced-Gen2 GC) between an element's surface-time touch and its read-back touch; pathological client timing slower than the window could still surface `ElementNotAvailableException`.
3. **New public API surface.** `AutomationInteropProvider.DisconnectProvider` is additive (plus a ref-assembly entry) and needs API-review sign-off; it is a thin wrapper over the documented `UiaDisconnectProvider`.
4. **Background disconnect sweep.**  Runs on the UI `Dispatcher` at `Background` priority and only disconnects proxies whose peer is already collected (which already throw ENA), so it cannot disturb a walk of a live element. Work is capped per tick (5000). Failed-disconnect handling. `AutomationInteropProvider.DisconnectProvider` can occasionally fail transiently (e.g. re-entrant input-sync, or a proxy whose runtime id was never cached because `GetRuntimeId` was never called on it). `ElementProxy.Disconnect()` returns a success/failure result, and the sweep re-queues a failed proxy for a later tick - its CCW is a GC root, so silently dropping it would leak the proxy. A per-proxy attempt counter ( `failedAttempts`) caps this at `MaxDisconnectAttempts` (8): genuine transients clear within 1–2 ticks and never approach the cap, while a proxy that fails every time is abandoned after 8 tries (leaking only the lightweight shell) so the timer can still self-stop instead of spinning forever. Reversible via `DisableUiaProviderDisconnect=true`.
 5. **ENA timing.** Clients may now receive `ElementNotAvailableException` during steady state (not only on tree changes). Conformant clients (Narrator, NVDA, JAWS) handle this; non-conformant tools that don't catch it could surface new errors.
 6. A client that retrieves an element for an unrealized item via `FindItemByProperty` and then calls `VirtualizedItemPattern.Realize` after the keep-alive window (~10–18 s) will get  `ElementNotAvailableException`, since the peer was collected and its proxy disconnected. This is within the UIA ENA contract but changes the pattern's previously indefinite realizability. Kill switch `DisableUiaProviderDisconnect` restores the old behavior.
 7. WPF runtime ids are `{7, ProcessId, GetHashCode()}`, so identity-hash collisions are possible. Because `UiaDisconnectProvider` matches by runtime id, a dead proxy's cached id that collides with a live peer's could invalidate a client's view of a healthy element. Pre-existing weakness; low probability, but the disconnect sweep is the first path to exercise it at scale.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/wpf/pull/11657)

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/PresentationCore/MS/internal/Automation/ElementProxy.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationCore/MS/internal/Automation/ElementProxy.cs
@@ -8,6 +8,7 @@
 //
 //
 
+using System.Collections.Generic;
 using System.Windows;
 using System.Windows.Automation;
 using System.Windows.Automation.Provider;
@@ -38,8 +39,13 @@ namespace MS.Internal.Automation
         // private ctor - the Wrap() pseudo-ctor is used instead.
         private ElementProxy(AutomationPeer peer)
         {
-            if ((AutomationInteropReferenceType == ReferenceType.Weak) && 
-                (peer is UIElementAutomationPeer || peer is ContentElementAutomationPeer || peer is UIElement3DAutomationPeer))
+            // Weak-reference peers whose lifetime is owned elsewhere (the visual tree element, or the parent
+            // ItemsControl/Calendar peer that tracks data-item peers in its own weak storage) so UIA client
+            // references don't pin recycled/virtualized peers - and the controls they root - in memory.
+            // Data-item peers (IsDataItemAutomationPeer) are gated by an opt-out switch.
+            if ((AutomationInteropReferenceType == ReferenceType.Weak) &&
+                (peer is UIElementAutomationPeer || peer is ContentElementAutomationPeer || peer is UIElement3DAutomationPeer ||
+                 (peer.IsDataItemAutomationPeer() && !CoreAppContextSwitches.UseStrongReferenceForItemAutomationPeers)))
             {
                 _peer = new WeakReference(peer);
             }
@@ -275,6 +281,11 @@ namespace MS.Internal.Automation
                         if(peer.IsDataItemAutomationPeer())
                         {
                             peer.AddToParentProxyWeakRefCache();
+
+                            // Root the peer for the duration of the in-flight traversal: it is being surfaced to UIA
+                            // here but is only weakly held by its proxy, so without this it could be collected in the
+                            // window between being returned to UIA and UIA's first call back on it.
+                            PeerKeepAlive.KeepAlive(peer);
                         }
                     }
                 }
@@ -292,6 +303,17 @@ namespace MS.Internal.Automation
                 if (_peer is WeakReference)
                 {
                     AutomationPeer peer = (AutomationPeer)((WeakReference)_peer).Target;
+
+                    // A data-item peer is rooted only by its parent ItemsControl/Calendar peer, whose
+                    // _dataChildren entry and wrapper-peer EventsSource link both vanish in one layout pass when
+                    // the row virtualizes out - so a GC mid-walk could collect it and surface
+                    // ElementNotAvailableException. Park a short-lived strong root (see PeerKeepAlive)
+                    // that outlives the walk but is released once the peer stops being touched.
+                    if (peer != null && peer.IsDataItemAutomationPeer())
+                    {
+                        PeerKeepAlive.KeepAlive(peer);
+                    }
+
                     return peer;
                 }
                 else
@@ -499,6 +521,72 @@ namespace MS.Internal.Automation
 
             return StaticWrap(root, peer);
         }
+
+        #region data-item peer keep-alive
+
+        // Bounds the lifetime of weakly-referenced data-item peers so an in-flight UIA traversal cannot observe one
+        // being collected mid-walk, without reintroducing the unbounded leak the weak reference exists to fix.
+        // Each peer surfaced to UIA (at StaticWrap) or touched on a UIA callback (the Peer getter) is stored in a
+        // strong-rooted "current" bucket; a Background-priority DispatcherTimer rotates the buckets once per window
+        // (drop the oldest, promote "current" to "previous").  A fixed time window is deliberate rather than a
+        // dispatcher-idle callback: under non-concurrent GC, blocking collection pauses leave the dispatcher
+        // transiently idle mid-walk, which an idle-driven rotation would mistake for "walk finished".  The retained
+        // set is bounded by the in-flight working set, not by the data-set size.
+        private static class PeerKeepAlive
+        {
+            // A peer stays rooted until it has gone untouched for at least one full window and at most two (9-18 s).
+            // The lower bound is sized from measurement: under forced-Gen2 GC stress a single FindAll + readback
+            // iteration peaks at ~4.4 s (walk-dominated, stretched by GC pauses) - the worst-case gap between an
+            // element's surface-time touch and its readback touch.  A 9 s window is ~2x that, so the guaranteed
+            // minimum survival (one window) covers it even under blocking GC, while still releasing peers the
+            // client has stopped touching within seconds (bounding the retained set to the in-flight working set).
+            private static readonly TimeSpan Window = TimeSpan.FromSeconds(9);
+            private static readonly object _lock = new object();
+            private static HashSet<AutomationPeer> _current = new HashSet<AutomationPeer>();
+            private static HashSet<AutomationPeer> _previous = new HashSet<AutomationPeer>();
+            private static DispatcherTimer _timer;
+
+            internal static void KeepAlive(AutomationPeer peer)
+            {
+                Dispatcher dispatcher = peer.Dispatcher;
+                if (dispatcher == null)
+                {
+                    return;
+                }
+
+                lock (_lock)
+                {
+                    _current.Add(peer);
+                    if (_timer == null)
+                    {
+                        // The constructor associates the timer with the supplied dispatcher and starts it, so this
+                        // is safe to call from the UIA worker thread that drives the proxy.
+                        _timer = new DispatcherTimer(Window, DispatcherPriority.Background, OnTick, dispatcher);
+                    }
+                }
+            }
+
+            private static void OnTick(object sender, EventArgs e)
+            {
+                lock (_lock)
+                {
+                    // Drop the oldest bucket and promote the current one.
+                    HashSet<AutomationPeer> recycled = _previous;
+                    recycled.Clear();
+                    _previous = _current;
+                    _current = recycled;
+
+                    // Nothing left to keep alive - stop ticking until the next peer is surfaced.
+                    if (_previous.Count == 0)
+                    {
+                        _timer.Stop();
+                        _timer = null;
+                    }
+                }
+            }
+        }
+
+        #endregion data-item peer keep-alive
 
         #region disable switch for ElementProxy weak reference fix
 
--- a/src/Microsoft.DotNet.Wpf/src/PresentationCore/MS/internal/Automation/ElementProxy.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationCore/MS/internal/Automation/ElementProxy.cs
@@ -299,12 +299,9 @@ namespace MS.Internal.Automation
                             peer.AddToParentProxyWeakRefCache();
 
                             // Root the peer for the in-flight traversal: it is only weakly held by its proxy and
-                            // could be collected between being returned to UIA and its first callback. Skip when
-                            // peers are strongly referenced - the proxy already roots them.
-                            if (!CoreAppContextSwitches.UseStrongReferenceForItemAutomationPeers)
-                            {
-                                PeerKeepAlive.KeepAlive(peer);
-                            }
+                            // could be collected between being returned to UIA and its first callback. KeepAlive
+                            // self-guards on the switch / registry-key / data-item eligibility.
+                            PeerKeepAlive.KeepAlive(peer);
                         }
                     }
                 }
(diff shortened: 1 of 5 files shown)


src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Automation/Peers/AutomationPeer.cs
```diff
@@ -1812,6 +1829,49 @@ private IRawElementProviderSimple ProviderFromPeerNoDelegation(AutomationPeer pe
...
+                // Sever proxy→peer reference immediately so the peer and its data items become GC-eligible without waiting for the deferred call.
+                proxy.ClearPeer();
+
+                // UiaDisconnectProvider must NOT be called during a UIA callback (e.g. during Navigate/FindAll handling).
+                // UpdateChildrenInternal is invoked from within UIA callbacks, so defer the actual COM disconnect to a separate dispatcher operation.
+                Dispatcher dispatcher = peer.Dispatcher;
+                if (dispatcher != null && !dispatcher.HasShutdownStarted)
+                {
+                    dispatcher.BeginInvoke(DispatcherPriority.Background,
+                        new Action(() => UiaDisconnectProvider(proxy)));
```
h3xds1nz: This creates a closure unnecessarily and a new action everytime. Ideally use the parameter variant, it's an object you're passing so no boxing here.

src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Automation/Peers/AutomationPeer.cs
```diff
@@ -1812,6 +1829,49 @@ private IRawElementProviderSimple ProviderFromPeerNoDelegation(AutomationPeer pe
...
+            WeakReference proxyWeakRef = peer._elementProxyWeakReference;
+            if (proxyWeakRef?.Target is ElementProxy proxy)
+            {
+                // Sever proxy→peer reference immediately so the peer and its data items become GC-eligible without waiting for the deferred call.
+                proxy.ClearPeer();
+
+                // UiaDisconnectProvider must NOT be called during a UIA callback (e.g. during Navigate/FindAll handling).
+                // UpdateChildrenInternal is invoked from within UIA callbacks, so defer the actual COM disconnect to a separate dispatcher operation.
+                Dispatcher dispatcher = peer.Dispatcher;
+                if (dispatcher != null && !dispatcher.HasShutdownStarted)
```
h3xds1nz: I'm not sure this is even required/true; the UIA callbacks should already run on the `Dispatcher` synchronously iirc, so that check seems redundant. If invoked from proxy directly, this would already throw if it was the case.

src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Automation/Peers/AutomationPeer.cs
```diff
@@ -1812,6 +1829,49 @@ private IRawElementProviderSimple ProviderFromPeerNoDelegation(AutomationPeer pe
...
+        /// </summary>
+        /// <remarks>
+        /// This method intentionally does NOT recursively disconnect children.
+        /// In virtualized controls, a removed item peer's cached _children may reference container peers (e.g. DataGridCellAutomationPeer) that have
+        /// been recycled and are now serving new items. Disconnecting those would break accessibility on currently visible elements.
+        /// Children are disconnected naturally when their own parent's UpdateChildrenInternal runs and detects them as removed.
+        /// </remarks>
+        private static void DisconnectPeerFromUia(AutomationPeer peer)
+        {
+            if (peer == null)
```
h3xds1nz: This check is redundant since you're already checking in the calling function.

src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Automation/Peers/AutomationPeer.cs
```diff
@@ -1812,6 +1829,49 @@ private IRawElementProviderSimple ProviderFromPeerNoDelegation(AutomationPeer pe
...
+            peer._elementProxyWeakReference = null;
+
+            // Break the peer→children chain.  Without this, a disconnected ItemAutomationPeer still roots cached cell peers via _children,
+            // which in turn root DataGridCell/DataGridRow containers via _owner, preventing the entire visual sub-tree from being GC`d.
+            peer._children = null;
+            peer._childrenValid = false;
+        }
+
+        [DllImport("UIAutomationCore.dll", EntryPoint = "UiaDisconnectProvider", CharSet = CharSet.Unicode)]
+        private static extern int UiaDisconnectProvider(IRawElementProviderSimple provider);
```
h3xds1nz: This shouldn't be defined in this file.

src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Automation/Peers/AutomationPeer.cs
```diff
@@ -1937,6 +1993,22 @@ internal void UpdateChildrenInternal(int invalidateLimit)
             //calls for "bulk" notification, use per-child notification, otherwise use "bulk"
             int removedCount = (hs == null ? 0 : hs.Count);
 
+            // Disconnect removed children from UIA so the client-side releases its
+            // COM references to the ElementProxy CCWs. Without this the CCW ref count
+            // never drops to 0, which prevents the managed peer from being GC`d.
+            // This must happen regardless of StructureChanged event registration.
+            if (removedCount > 0)
+            {
+                foreach (AutomationPeer removedChild in hs)
```
h3xds1nz: This is just a sigh for another PR _but this whole thing should be turned into a simple merge walk to differentiate old and new and avoid the HashSet allocations altogether. Sometimes this is tens of MBs for something that could have been a simple loop._

The least that can be done here is to repeat this loop over the HashSet only once; by either doing all the StructureChanged stuff or not.

src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Automation/Peers/AutomationPeer.cs
```diff
@@ -1495,6 +1497,21 @@ private void EnsureChildren()
...
+                // The _elementProxyWeakReference gate keeps this allocation-free on the common path where no UIA client has walked this subtree.
+                if (oldChildren != null)
+                {
+                    for (int i = 0; i < oldChildren.Count; i++)
+                    {
+                        AutomationPeer old = oldChildren[i];
+                        if (old._elementProxyWeakReference == null)
+                            continue;
+
+                        if (_children == null || !_children.Contains(old))
```
h3xds1nz: This basically O(n*m), if this is a top automation peer, it's gonna be pretty heavy.

h3xds1nz (commented):
@amarinov-msft I'm really worried about perf with these changes. UIA implementation itself is already a really heavy mess (things like #10676 and more), and this will further increase the weight (an async `DispatcherOperation` for every disconnected peer), further enumeration over the intermediate collections.

Now, I don't have an exact solution here since in our wpf fork I've reworked this extensively but I think it's worth a further investigation and **profiling** to fix this.

amarinov-msft:
@h3xds1nz thank you for your review. I tried addressing your comments. While the fundamental question about this issue is resolved (performance vs behavior), please check the updates when you have the opportunity and let me know if they are in the direction that you had in mind.

src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Automation/Peers/AutomationPeer.cs
```diff
@@ -1812,6 +1843,43 @@ private IRawElementProviderSimple ProviderFromPeerNoDelegation(AutomationPeer pe
...
+
+            peer._elementProxyWeakReference = null;
+
+            // Break the peer→children chain.  Without this, a disconnected ItemAutomationPeer still roots cached cell peers via _children,
+            // which in turn root DataGridCell/DataGridRow containers via _owner, preventing the entire visual sub-tree from being GC`d.
+            peer._children = null;
+            peer._childrenValid = false;
+        }
+
+        private static readonly DispatcherOperationCallback _disconnectProxyCallback = static arg =>
```
h3xds1nz: You should be able to use `SendOrPostCallback` which doesn't return but is special-cased.

https://github.com/dotnet/wpf/blob/c161dc985dbe19b6c8d19f7ab774ca1c90ca4995/src/Microsoft.DotNet.Wpf/src/WindowsBase/MS/Internal/Threading/ExceptionWrapper.cs#L92

src/Microsoft.DotNet.Wpf/src/PresentationCore/MS/internal/Automation/ElementProxy.cs
```diff
@@ -534,6 +535,31 @@ internal static ReferenceType AutomationInteropReferenceType
...
+        /// can be garbage collected. Must NOT be invoked from inside a UIA callback;
+        /// callers defer this to a dispatcher operation.
+        /// </summary>
+        internal void Disconnect()
+        {
+            UiaDisconnectProvider(this);
+            _peer = null;
+        }
+
+        [DllImport(DllImport.UIAutomationCore, EntryPoint = "UiaDisconnectProvider", CharSet = CharSet.Unicode)]
```
h3xds1nz: When I've mentioned it before, I meant that the common practice is to put it in specialized file; in this case that would be https://github.com/dotnet/wpf/blob/main/src/Microsoft.DotNet.Wpf/src/UIAutomation/UIAutomationProvider/MS/Internal/Automation/UiaCoreProviderApi.cs ... the provider library would have been loaded at this point anyway.

h3xds1nz (commented):
@amarinov-msft Just a few nits, I guess this is acceptable at the current codebase state to resolve the issue.

amarinov-msft:
@h3xds1nz thank you for your review and effort! However, while testing this - I found a regression introduced by my changes where visible peers are being treated as stale and disconnected. The disconnect-based approach in the PR seems to be fundamentally incompatible with how UIA traverses peers. We probably need to change the approach completely.

h3xds1nz:
@amarinov-msft No worries; I unfortunately cannot test this as I'm unable to replicate the memory leak with the repro whatsoever. Plus WPFAutomationClient report seems invalid, given that there's literally no GC pressure, just forcing a GC manually clears out everything (not that this fix would be an attempt to solve that).

PR #11684, opened 2026-06-02
Text Rendering: Add DirectWrite color glyph (COLR v0) rendering support

### LINKED ISSUE #91: Add support for Color Emoji support
Currently when you render emojis in WPF, they are rendered black'n'white, and not rendering with the correct emoji color.
It would be good if we could update the text rasterizer renderers to render with the correct color.
There's a proof of concept implementation here: https://github.com/samhocevar/emoji.wpf, so perhaps @samhocevar would be willing to contribute? (I'm guessing this would be done on the DirectWrite level though)

etvorun:
## Summary

Add DirectWrite color glyph (COLR v0) rendering support to WPF and fix emoji font fallback so that emoji characters render in full color instead of monochrome outlines.

### Before and after

Text like the following now renders in full color instead of monochrome outlines:

| Category | Example text |
|---|---|
| Faces | Hello 😀😃😄😁😆 World |
| Gestures | Great job 👍👏🙌 |
| Animals | Nature 🐶🐱🦊🐻🐼 |
| Colored circles (U+1F7E0–1F7EB) | 🟠🟡🟢🟣🟤🔴🔵🟥🟧🟨🟩🟦🟪🟫 |
| Star / circle (U+2B50, U+2B55) | ⭐⭕ |
| Heavy equals sign (U+1F7F0) | 🟰 |
| Extended range (U+1FA00–1FAFF) | 🫠🫡🫢🫣🫤🫥🫨 |

**Before:** Emoji show as monochrome outlines or empty boxes.
**After:** Emoji render with full COLR v0 palette colors.

### Color glyph rendering
- **`GlyphTypeface.IsColorFont`** — New property that detects whether a font has a COLR OpenType table (lazy-cached).
- **`GlyphRun.TryBuildColorGlyphDrawing()`** — Calls `IDWriteFactory2::TranslateColorGlyphRun` to decompose a glyph run into per-layer sub-runs, each with a CPAL palette color. Builds a frozen `DrawingGroup` of `GeometryDrawing` layers.
- **`DrawingDrawingContext` / `RenderDataDrawingContext`** — Intercept `DrawGlyphRun` to render color layers when available, falling through to the existing monochrome path for non-color fonts.
- **`Factory.TranslateColorGlyphRun()`** — Managed wrapper around `IDWriteFactory2::TranslateColorGlyphRun` with input validation and proper COM lifetime management. Queries for `IDWriteFactory2` at factory creation with graceful fallback on pre-Windows 8.1.
- **`FontFace.HasColorGlyphs()`** — Native helper that probes for the COLR OpenType table.
- **New interop types** — `IDWriteColorGlyphRunEnumerator`, `DWRITE_COLOR_GLYPH_RUN`, `ColorGlyphLayer` for managed/native interop.

### Font fallback fixes
- Extended emoji Unicode ranges in all three `FontFamilyMap` sections of `GlobalUserInterface.CompositeFont` from `1F900-1F9FF` to `1F900-1FAFF` (adds Symbols and Pictographs Extended-A coverage).
- Added explicit Segoe UI Emoji entries for default-presentation emoji code points (U+1F7E0–1F7EB colored circles/squares, U+1F7F0 heavy equals sign, U+2B50 star, U+2B55 hollow red circle) that were previously mapped to text-priority font families.

### Tests
- Added `ColorEmojiTests` covering `IsColorFont` detection, color glyph layer decomposition, and font fallback for previously missing emoji code points.

## Why

WPF renders all emoji as monochrome outlines because the glyph pipeline produces single-channel coverage bitmaps with no per-layer color support. Additionally, newer Unicode emoji ranges are not mapped to Segoe UI Emoji in the composite font, causing them to show as empty boxes.

This fix decomposes COLR v0 color glyphs via DirectWrite into individually colored geometry layers, and extends font fallback to cover all standard emoji Unicode ranges.

### Limitations
- COLR v0 only (IDWriteFactory2). COLRv1/SVG/CBDT color fonts would need IDWriteFactory4+.
- Color layers use geometry-based rendering (no ClearType), which is acceptable for emoji at typical display sizes.

### Validation
- Verified locally that emoji render in full color with the fix applied.
- All PresentationCore.Tests pass (976 total: 947 passed, 29 skipped for pre-existing reasons, 0 failures).

Fixes #91

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/wpf/pull/11684)

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/PresentationCore/MS/internal/Interop/DWrite/IDWriteColorGlyphRunEnumerator.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationCore/MS/internal/Interop/DWrite/IDWriteColorGlyphRunEnumerator.cs
@@ -1,4 +1,4 @@
-// Licensed to the .NET Foundation under one or more agreements.
+﻿// Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the MIT license.
 
 using System.Runtime.CompilerServices;
--- a/src/Microsoft.DotNet.Wpf/src/PresentationCore/MS/internal/Interop/DWrite/IDWriteColorGlyphRunEnumerator.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationCore/MS/internal/Interop/DWrite/IDWriteColorGlyphRunEnumerator.cs
@@ -1,0 +1,31 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System.Runtime.CompilerServices;
+
+namespace MS.Internal.Interop.DWrite
+{
+    /// <summary>
+    /// Managed interop for IDWriteColorGlyphRunEnumerator COM interface.
+    /// Vtable layout: IUnknown (0-2), MoveNext (3), GetCurrentRun (4).
+    /// </summary>
+    internal unsafe struct IDWriteColorGlyphRunEnumerator
+    {
+        public void** lpVtbl;
+
+        public uint Release()
+        {
+            return ((delegate* unmanaged<IDWriteColorGlyphRunEnumerator*, uint>)(lpVtbl[2]))((IDWriteColorGlyphRunEnumerator*)Unsafe.AsPointer(ref this));
+        }
+
+        public int MoveNext(int* hasRun)
+        {
+            return ((delegate* unmanaged<IDWriteColorGlyphRunEnumerator*, int*, int>)(lpVtbl[3]))((IDWriteColorGlyphRunEnumerator*)Unsafe.AsPointer(ref this), hasRun);
+        }
+
+        public int GetCurrentRun(DWRITE_COLOR_GLYPH_RUN** colorGlyphRun)
+        {
+            return ((delegate* unmanaged<IDWriteColorGlyphRunEnumerator*, DWRITE_COLOR_GLYPH_RUN**, int>)(lpVtbl[4]))((IDWriteColorGlyphRunEnumerator*)Unsafe.AsPointer(ref this), colorGlyphRun);
+        }
+    }
+}
--- a/src/Microsoft.DotNet.Wpf/src/PresentationCore/MS/internal/Text/TextInterface/Factory.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationCore/MS/internal/Text/TextInterface/Factory.cs
@@ -41,7 +41,7 @@ namespace MS.Internal.Text.TextInterface
         /// Used for COLR v0 color glyph support (TranslateColorGlyphRun).
         /// Available on Windows 8.1+.
         /// </summary>
-        private void* _pFactory2;
+        private IDWriteFactory2* _pFactory2;
 
         /// <summary>
         /// Constructs a factory object.
@@ -66,7 +66,7 @@ namespace MS.Internal.Text.TextInterface
             void* pFactory2 = null;
             int qiHr = _factory.Value->QueryInterface(&iidFactory2, &pFactory2);
             if (qiHr == 0 && pFactory2 != null)
-                _pFactory2 = pFactory2;
+                _pFactory2 = (IDWriteFactory2*)pFactory2;
 
             _wpfFontFileLoader = new FontFileLoader(fontSourceFactory);
             _wpfFontCollectionLoader = new FontCollectionLoader(
@@ -420,16 +420,16 @@ namespace MS.Internal.Text.TextInterface
                         glyphRun.isSideways = isSideways ? 1 : 0;
                         glyphRun.bidiLevel = bidiLevel;
 
-                        void* pEnum = null;
-
-                        // IDWriteFactory2::TranslateColorGlyphRun is vtable slot 28.
-                        void** factory2Vtbl = *(void***)_pFactory2;
-                        int hr = ((delegate* unmanaged<void*, float, float, DWRITE_GLYPH_RUN*, void*, int, void*, uint, void**, int>)(factory2Vtbl[28]))(
-                            _pFactory2,
-                            baselineOriginX, baselineOriginY,
-                            &glyphRun, null,
-                            (int)measuringMode, null,
-                            0, &pEnum);
+                        IDWriteColorGlyphRunEnumerator* pEnum = null;
+                        int hr = _pFactory2->TranslateColorGlyphRun(
+                            baselineOriginX,
+                            baselineOriginY,
+                            &glyphRun,
+                            null,
+                            (int)measuringMode,
+                            null,
+                            0,
+                            &pEnum);
 
                         // Release the AddRef'd font face now that the DWrite call is complete.
                         ((IDWriteFontFace*)glyphRun.fontFace)->Release();
@@ -565,7 +565,7 @@ namespace MS.Internal.Text.TextInterface
                 // Release the IDWriteFactory2 interface obtained via QueryInterface.
                 if (_managedFactory._pFactory2 != null)
                 {
-                    ((IDWriteFactory*)_managedFactory._pFactory2)->Release();
+                    _managedFactory._pFactory2->Release();
                     _managedFactory._pFactory2 = null;
                 }
 
--- a/src/Microsoft.DotNet.Wpf/src/PresentationCore/MS/internal/Text/TextInterface/Factory.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationCore/MS/internal/Text/TextInterface/Factory.cs
@@ -1,6 +1,7 @@
 // Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the MIT license.
 
+using System.Collections.Generic;
 using System.Runtime.InteropServices;
 using MS.Internal.Interop;
 using MS.Internal.Interop.DWrite;
@@ -36,6 +37,13 @@ namespace MS.Internal.Text.TextInterface
         private readonly IFontSourceFactory _fontSourceFactory;
 
         /// <summary>
+        /// Pointer to the IDWriteFactory2 interface, or null if not available.
+        /// Used for COLR v0 color glyph support (TranslateColorGlyphRun).
+        /// Available on Windows 8.1+.
+        /// </summary>
+        private void* _pFactory2;
+
+        /// <summary>
         /// Constructs a factory object.
         /// </summary>
         /// <param name="factoryType">Identifies whether the factory object will be shared or isolated.</param>
@@ -49,6 +57,16 @@ namespace MS.Internal.Text.TextInterface
         private Factory(FactoryType factoryType, IFontSourceCollectionFactory fontSourceCollectionFactory, IFontSourceFactory fontSourceFactory)
         {
             Initialize(factoryType);
+
+            // QueryInterface for IDWriteFactory2, which provides TranslateColorGlyphRun
+            // for COLR v0 color glyph decomposition. Available on Windows 8.1+; on older
+            // versions the QI returns E_NOINTERFACE and _pFactory2 stays null, which
+            // disables color glyph rendering gracefully.
+            Guid iidFactory2 = new Guid(0x0439fc60, 0xca44, 0x4994, 0x8d, 0xee, 0x3a, 0x9a, 0xf7, 0xb7, 0x32, 0xec);
+            void* pFactory2 = null;
+            int qiHr = _factory.Value->QueryInterface(&iidFactory2, &pFactory2);
+            if (qiHr == 0 && pFactory2 != null)
+                _pFactory2 = pFactory2;
 
             _wpfFontFileLoader = new FontFileLoader(fontSourceFactory);
             _wpfFontCollectionLoader = new FontCollectionLoader(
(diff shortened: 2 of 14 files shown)


etvorun:
Pinging participants from #91 who may be interested in reviewing this implementation:

@dotMorten (original issue author), @mikedn, @be5invis, @fdwr (DirectWrite expertise), @samhocevar (Emoji.Wpf author), @fcharlie, @pchaurasia14, @Bart-Verdonck, @alexinea, @gusmanb, @Lorymi, @virzak, @XamDR, @gusmanb


This PR adds native COLR v0 color glyph support directly in WPF's rendering pipeline via `IDWriteFactory2::TranslateColorGlyphRun`, along with font fallback fixes for missing emoji Unicode ranges. Feedback and review are very welcome!

miloush:
I think this is fine for Visual Studio but not for WPF itself. We should have a proper model for color glyph runs, not a render-time hack. If there is an interest to have this interim solution, it should at least be opt-in.

src/Microsoft.DotNet.Wpf/src/PresentationCore/MS/internal/Interop/DWrite/IDWriteColorGlyphRunEnumerator.cs
```diff
@@ -0,0 +1,31 @@
...
+    /// Managed interop for IDWriteColorGlyphRunEnumerator COM interface.
+    /// Vtable layout: IUnknown (0-2), MoveNext (3), GetCurrentRun (4).
+    /// </summary>
+    internal unsafe struct IDWriteColorGlyphRunEnumerator
+    {
+        public void** lpVtbl;
+
+        public uint Release()
+        {
+            return ((delegate* unmanaged<IDWriteColorGlyphRunEnumerator*, uint>)(lpVtbl[2]))((IDWriteColorGlyphRunEnumerator*)Unsafe.AsPointer(ref this));
```
fdwr, line 18: Can't C# call COM interfaces directly, rather than going through manual vtable offsets? I'm not that familiar with the approach (C++ is more my area), but I see some [promising leads here](https://learn.microsoft.com/en-us/dotnet/standard/native-interop/comwrappers-source-generation) that look a bit cleaner. e.g.

```
[GeneratedComInterface]
[Guid("3faca0d2-e7f1-4e9c-82a6-404fd6e0aab8")]
internal partial interface IFoo
{
    void Method(int i);
}
```

If not, then the vtable is publicly published and is guaranteed stable in any case.

src/Microsoft.DotNet.Wpf/src/PresentationCore/MS/internal/Text/TextInterface/Factory.cs
```diff
@@ -330,6 +348,179 @@ internal TextAnalyzer CreateTextAnalyzer()
...
+                        void* pEnum = null;
+
+                        // IDWriteFactory2::TranslateColorGlyphRun is vtable slot 28.
+                        void** factory2Vtbl = *(void***)_pFactory2;
+                        int hr = ((delegate* unmanaged<void*, float, float, DWRITE_GLYPH_RUN*, void*, int, void*, uint, void**, int>)(factory2Vtbl[28]))(
+                            _pFactory2,
+                            baselineOriginX, baselineOriginY,
+                            &glyphRun, null,
+                            (int)measuringMode, null,
+                            0, &pEnum);
```
fdwr: (nit) Does WPF not normally split parameters onto separate lines? I'm not sure, just noticing that pattern elsewhere like below:

```c#
                    var colorLayers = factory.TranslateColorGlyphRun(
                        fontFace,
                        (float)_renderingEmSize,
                        glyphIndices,
                        glyphAdvances,
                        glyphOffsets,
                        (uint)glyphCount,
                        IsSideways,
                        (uint)_bidiLevel,
                        (float)_baselineOrigin.X,
                        (float)_baselineOrigin.Y,
                        _textFormattingMode == TextFormattingMode.Display ? 1u : 0u  // GDI_CLASSIC = 1, NATURAL = 0
                        );
```

src/Microsoft.DotNet.Wpf/src/PresentationCore/MS/internal/Text/TextInterface/Factory.cs
```diff
@@ -330,6 +348,179 @@ internal TextAnalyzer CreateTextAnalyzer()
...
+                {
+                    // Convert interleaved float pairs to DWRITE_GLYPH_OFFSET structs.
+                    DWRITE_GLYPH_OFFSET* pOffsets = null;
+                    DWRITE_GLYPH_OFFSET* allocatedOffsets = null;
+
+                    try
+                    {
+                        if (glyphOffsets != null && glyphCount <= (uint)(glyphOffsets.Length / 2))
+                        {
+                            allocatedOffsets = (DWRITE_GLYPH_OFFSET*)Marshal.AllocHGlobal(
```
fdwr, line 399: Can we just use `fixed` and reinterpret the data (e.g. `short* shortPointer = (short*)bytePointer`) rather than allocate and copy? Though, I suppose you may need to label the function `unsafe` then. 🤷‍♂️

src/Microsoft.DotNet.Wpf/src/PresentationCore/MS/internal/Text/TextInterface/Factory.cs
```diff
@@ -330,6 +348,179 @@ internal TextAnalyzer CreateTextAnalyzer()
...
+                return layers.Count > 0 ? layers : null;
+            }
+            finally
+            {
+                if (pColorEnum != null)
+                    pColorEnum->Release();
+            }
+        }
+
+        private static float Clamp01(float v) => v < 0f ? 0f : v > 1f ? 1f : v;
```
fdwr, line 522: `=> Math.Clamp(v, 0, 1f)`?

src/Microsoft.DotNet.Wpf/src/PresentationCore/MS/internal/Text/TextInterface/Factory.cs
```diff
@@ -330,6 +348,179 @@ internal TextAnalyzer CreateTextAnalyzer()
...
+                    layer.ColorG = Clamp01(pColorRun->runColor.g);
+                    layer.ColorB = Clamp01(pColorRun->runColor.b);
+                    layer.ColorA = Clamp01(pColorRun->runColor.a);
+
+                    uint layerGlyphCount = pColorRun->glyphRun.glyphCount;
+
+                    // Deep-copy glyph data (pointer is only valid until next MoveNext).
+                    layer.GlyphIndices = new ushort[layerGlyphCount];
+                    for (uint i = 0; i < layerGlyphCount; i++)
+                        layer.GlyphIndices[i] = pColorRun->glyphRun.glyphIndices[i];
```
fdwr, line 487: Consider single line option (avoids repeating `i` 5x):

`Marshal.Copy(pColorRun->glyphRun.glyphIndices, layer.GlyphIndices, 0, layerGlyphCount);`

src/Microsoft.DotNet.Wpf/src/PresentationCore/MS/internal/Text/TextInterface/Factory.cs
```diff
@@ -330,6 +348,179 @@ internal TextAnalyzer CreateTextAnalyzer()
...
+                        layer.GlyphOffsets = new float[layerGlyphCount * 2];
+                        for (uint i = 0; i < layerGlyphCount; i++)
+                        {
+                            layer.GlyphOffsets[i * 2] = pColorRun->glyphRun.glyphOffsets[i].advanceOffset;
+                            layer.GlyphOffsets[i * 2 + 1] = pColorRun->glyphRun.glyphOffsets[i].ascenderOffset;
+                        }
+                    }
+                    else
+                    {
+                        layer.GlyphOffsets = null;
```
fdwr, line 507: Isn't null the default state of `GlyphOffsets` anyway when the `ColorGlyphLayer` is initialized?

src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Media/GlyphRun.cs
```diff
@@ -1597,6 +1598,228 @@ public Geometry BuildGeometry()
...
+        /// </summary>
+        private Geometry BuildLayerGeometry(
+            ushort[] layerGlyphIndices,
+            float[] layerGlyphAdvances,
+            float[] layerGlyphOffsets,
+            float baselineOriginX,
+            float baselineOriginY)
+        {
+            GeometryGroup layerGeometry = null;
+            double accAdvance = 0;
```
fdwr: What does "ACC" stand for? Typically this is named `advanceWidth` or `penX` or just `x`.

fdwr: Ah, I'm guessing you meant `accumulatedWidth`. If so, spelling out is nice.

src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Media/GlyphRun.cs
```diff
@@ -2464,6 +2687,8 @@ private enum GlyphRunFlags : byte
         private XmlLanguage         _language;
         private string              _deviceFontName;
         private object              _inkBoundingBox;    // Used when CacheInkBounds is on
+        private Drawing             _cachedColorDrawing;   // Cached result of TryBuildColorGlyphDrawing
+        private Brush               _cachedColorForeground; // Foreground brush used for _cachedColorDrawing
         private TextFormattingMode      _textFormattingMode;
```
fdwr:
```suggestion
        private TextFormattingMode  _textFormattingMode;
```
(existing misalignment nit)

src/Microsoft.DotNet.Wpf/tests/UnitTests/PresentationCore.Tests/System/Windows/Media/ColorEmojiTests.cs
```diff
@@ -0,0 +1,165 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+namespace System.Windows.Media;
+
+/// <summary>
+/// Tests for COLR v0 color emoji rendering support:
```
fdwr, line 7: Alas, there's no gradient support in v0 (need v1 for that, like [Chromium's Skia](https://source.chromium.org/chromium/chromium/src/+/main:third_party/skia/src/ports/SkScalerContext_win_dw.cpp;l=690?q=TranslateColorGlyphRun&ss=chromium%2Fchromium%2Fsrc)), but infinitely more color than none! :)

fdwr (approved):
Thanks ET - this is cool. Looks functionally correct, with minor feedback.

Regarding Jan's concern, is there a way to render monochromatically if desired, or is this always-on if the font has color in it? For UI controls like labels and text boxes, opting in by default makes sense, but for low level glyph rendering, there might be cases where we want to enable users to configure this. For Direct2D's [`DrawTextLayout`](https://learn.microsoft.com/en-us/windows/win32/api/d2d1/nf-d2d1-id2d1rendertarget-drawtextlayout), we didn't automatically enable it at the rendering layers because people could be doing custom highlight selection logic that wasn't prepared for full color yet, but we made it really easy to enable (just pass `D2D1_DRAW_TEXT_OPTIONS_ENABLE_COLOR_FONT`). Maybe the `DrawingDrawingContext` (what a weird name) and `RenderDataDrawingContext` could have a property.

src/Microsoft.DotNet.Wpf/src/PresentationCore/MS/internal/Interop/DWrite/IDWriteColorGlyphRunEnumerator.cs
etvorun: This matches pattern in other PresentationCore/MS/internal/Interop/DWrite/IDWriteXxx.cs files. Someone would need to do a separate pass to make it more COM friendly

miloush: There seems to be a desire to move away from COM interfaces to source-generated vtables because the former is not trimmable.

etvorun:
> Thanks ET - this is cool. Looks functionally correct, with minor feedback.
> 
> Regarding Jan's concern, is there a way to render monochromatically if desired, or is this always-on if the font has color in it? For UI controls like labels and text boxes, opting in by default makes sense, but for low level glyph rendering, there might be cases where we want to enable users to configure this. For Direct2D's [`DrawTextLayout`](https://learn.microsoft.com/en-us/windows/win32/api/d2d1/nf-d2d1-id2d1rendertarget-drawtextlayout), we didn't automatically enable it at the rendering layers because people could be doing custom highlight selection logic that wasn't prepared for full color yet, but we made it really easy to enable (just pass `D2D1_DRAW_TEXT_OPTIONS_ENABLE_COLOR_FONT`). Maybe the `DrawingDrawingContext` (what a weird name) and `RenderDataDrawingContext` could have a property.

This PR a port of a fix used by Visual Studio 2026 (release 18.8). We can treat it as a starting point and improve upon it. If you can recommend specific changes to this PR I can include them :)

miloush:
The issue #91 you referenced discussed some options. Introducing `TextOptions.EnableColorFont` sounds like a pretty good compromise.

src/Microsoft.DotNet.Wpf/src/PresentationCore/MS/internal/Interop/DWrite/IDWriteColorGlyphRunEnumerator.cs
fdwr: > This matches pattern in other PresentationCore/MS/internal/Interop/DWrite/IDWriteXxx.cs files

Okie dokie.

fdwr (approved):
LGTM sir 🫡. Like you say, it's a great starting point to improve upon. I'll defer to the WPF-side folks for policy decisions like whether to opt-in explicitly or opt-in by default like UWP does. (now we'll be able to use color emoji in [Windows PIX](https://devblogs.microsoft.com/pix/download/) which uses WPF 😉)

davidxuang:
> * COLR v0 only (IDWriteFactory2). COLRv1/SVG/CBDT color fonts would need IDWriteFactory4+.

This is a great improvement! Will this also work for COLRv1 fonts with COLRv0 fallback entries (`BaseGlyphRecord` and `LayerRecord`)?

PR #11717, opened 2026-06-12
Break the reference chain so old containers can be garbage collected.

### LINKED ISSUE #11337: WPF leaks ElementProxy instances when UI Automation is used

gpaskaleva-msft:
Fixes https://github.com/dotnet/wpf/issues/11337

## Description

**ElementProxy.cs:**
 - `_peer` made non-readonly + `ClearPeer()` method — allows `DisconnectPeerFromUia` to sever the strong proxy → peer reference immediately (for non-UIElement peers like `DataGridItemAutomationPeer`)

**AutomationPeer.cs:**

 1. `EnsureChildren()` disconnect logic — detects removed peers when `_children` is refreshed during UIA traversal and calls `DisconnectPeerFromUia` on them.
 2. `DisconnectPeerFromUia` changes:
 - `proxy.ClearPeer()` — severs proxy→peer strong reference
 - `Dispatcher.BeginInvoke` for `UiaDisconnectProvider` — defers COM disconnect to outside UIA callback context
 - `peer._children = null; peer._childrenValid = false;` — the critical fix that breaks the chain from disconnected peer → cached cell peers → old DataGridCell/Row containers

## Regression

No

## Testing

 Test Environment

   - Repro app (WPFUI): DataGrid with 200k rows, ItemsSource rebound every 3 seconds
   - UIA client (WPFAutomationClient): Calls FindAll(TreeScope.Descendants, TrueCondition) every 1 second
   - Runtime: Custom-built PresentationCore.dll deployed to local .NET 11 Preview 5 SDK

  Tests Performed
   | # | Fix Applied | Result | Duration |
   |---|---|---|---|
   | 1 | PR #11657 only (EnsureChildren disconnect) | ❌ Memory grew to ~2500 MB | 20 min |
   | 2 | + ClearPeer() on ElementProxy | ❌ No improvement | ~5 min |
   | 3 | + Deferred UiaDisconnectProvider via Dispatcher.BeginInvoke | ❌ Still ~6-9 MB/cycle growth | ~5 min |
   | 4 | + Diagnostic counters (verified disconnects fire: 45/cycle, refreshes accelerating) | ❌ 911→1019 MB in 14 cycles | ~1 min |
   | 5 | + `peer._children = null; peer._childrenValid = false;` in DisconnectPeerFromUia | ✅ **Stable 449-534 MB** | 3+ min |

###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/wpf/pull/11717)

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/PresentationCore/MS/internal/Automation/ElementProxy.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationCore/MS/internal/Automation/ElementProxy.cs
@@ -540,9 +540,22 @@ namespace MS.Internal.Automation
         //
         //------------------------------------------------------
  
+        #region Internal Methods for Disconnect
+
+        // Called during UIA disconnect to sever the strong reference from this
+        // proxy to the automation peer.  This allows the peer (and its associated
+        // data items) to be garbage collected even if UIA/client still holds a
+        // COM reference to this CCW temporarily.
+        internal void ClearPeer()
+        {
+            _peer = null;
+        }
+
+        #endregion Internal Methods for Disconnect
+
         #region Private Fields
 
-        private readonly object _peer;
+        private object _peer;
 
         #endregion Private Fields
     }
--- a/src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Automation/Peers/AutomationPeer.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Automation/Peers/AutomationPeer.cs
@@ -1484,6 +1484,7 @@ namespace System.Windows.Automation.Peers
             // UpdateSubtree is not called on it yet.
             if (!_childrenValid || _ancestorsInvalid)
             {
+                List<AutomationPeer> oldChildren = _children;
                 _children = GetChildrenCore();
                 if (_children != null)
                 {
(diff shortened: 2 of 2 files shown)


amarinov-msft (dismissed):
The changes are looking good. We just need to merge https://github.com/dotnet/wpf/pull/11723/ here before merging this PR.

h3xds1nz:
Not gonna lie, this adds quite some overhead to already massively heavy invocation this recursive method call is.

amarinov-msft:
> Not gonna lie, this adds quite some overhead to already massively heavy invocation this recursive method call is.

@h3xds1nz Can you take a look at the [original PR](https://github.com/dotnet/wpf/pull/11657)  . The fix there is identical - only the hot-path cost differs, Trying to minimize the changes inside `EnsureChildren()`, we just capture `oldChildren` and add one gated loop that does nothing unless a peer was both client-exposed and removed. That's the smallest possible addition that still finds every leaked peer. I will close this PR (and the related one, introducing the AppContext switch). If you have any considerations - you can comment in #11657

PR #11723, opened 2026-06-15
Add AppContext switch to guard automation peer disconnect behavior

### LINKED ISSUE #11337: WPF leaks ElementProxy instances when UI Automation is used

amarinov-msft:
Fixes #11337

Main PR #11717

## Description

Adds an AppContext compatibility switch (`Switch.System.Windows.Automation.Peers.UseLegacyAutomationPeerDisconnect`) to guard the new UIA disconnect behavior introduced in #11717 .

When false (default): removed automation peers are disconnected via `UiaDisconnectProvider`, fixing the memory leak where CCW ref counts never drop to zero.

When true: legacy behavior is preserved - removed peers are not disconnected, maintaining the old element lifetime semantics for consumers that depend on it.

## Customer Impact

 Without this switch, the disconnect fix in #11717 unconditionally changes observable UIA behavior: removed peers get UiaDisconnectProvider'd and their proxy _peer is cleared, so clients still holding cached elements receive `ElementNotAvailableException` instead of data. Test automation harnesses and AT products that depend on the old element lifetime would break with no escape hatch.

## Regression

No

## Testing

 - Verified no new C# compilation errors introduced by the changes.
 - Functional validation of the disconnect behavior itself is covered by #11717's testing.

## Risk

 Low. The switch is purely additive infrastructure - it defines a new AppContext switch and guards the #11717 call sites behind a property check.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/wpf/pull/11723)

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/PresentationCore/MS/internal/CoreAppContextSwitches.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationCore/MS/internal/CoreAppContextSwitches.cs
@@ -177,6 +177,24 @@ namespace MS.Internal
 
         #endregion
 
+        #region UseLegacyAutomationPeerDisconnect
+
+        /// <summary>
+        /// Switch to opt-out of the automation peer disconnect behavior.
+        /// When true, removed automation peers are NOT disconnected from UIA (legacy behavior).
+        /// When false (default), removed automation peers are disconnected via UiaDisconnectProvider.
+        /// </summary>
+        public static bool UseLegacyAutomationPeerDisconnect
+        {
+            [MethodImpl(MethodImplOptions.AggressiveInlining)]
+            get
+            {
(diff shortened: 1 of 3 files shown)


gpaskaleva-msft (approved):

amarinov-msft:
Closing the PR as this will be addressed in #11657

PR #11763, opened 2026-07-08
Add CodeQL comments to clarify trusted input handling in ResourcesGen…

gpaskaleva-msft:
Adds a suppression comment for CodeQL finding SM00414.

After investigation, the reported data flows were determined to be build-time only and originate from trusted MSBuild inputs (ITaskItem.ItemSpec values such as ResourceFiles, OutputResourcesFile, and _markupFiles). These values are supplied by the project file and build items authored by the developer performing the build, not by untrusted external users or runtime application input.

This area can be re-evaluated in the future if these tasks begin accepting untrusted or externally supplied input.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/wpf/pull/11763)

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/PresentationBuildTasks/Microsoft/Build/Tasks/Windows/ResourcesGenerator.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationBuildTasks/Microsoft/Build/Tasks/Windows/ResourcesGenerator.cs
@@ -290,7 +290,7 @@ namespace Microsoft.Build.Tasks.Windows
 
                 strFileName = inputFile.ItemSpec;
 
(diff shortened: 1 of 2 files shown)


himgoyalmicro (dismissed):
LGTM

harshit7962:
The comments by copilot are not fully valid, but I think we should truncate the reasoning that we have provided. It should be a minimum of 25 chars as per the rules but if copilot suggest that analyzer will find it difficult to read, then we should consider truncating it.

himgoyalmicro (approved):
LGTM

vinnarayana-msft (approved):

gpaskaleva-msft:
@dotnet-policy-service agree

Issue #11774, opened 2026-07-11
DragLeave constructs DragEventArgs with coordinates relative to the new target instead of the event target


mosyy098:
### Description

In WPF's drag-and-drop, when the mouse moves from an [AllowDrop](https://learn.microsoft.com/en-us/dotnet/api/system.windows.uielement.allowdrop?view=windowsdesktop-10.0) element to another element, [DragLeave](https://learn.microsoft.com/en-us/dotnet/api/system.windows.uielement.dragleave?view=windowsdesktop-10.0) event will be raised for the previous element. However, the [DragEventArgs](https://learn.microsoft.com/en-us/dotnet/api/system.windows.drageventargs?view=windowsdesktop-10.0) passed contains Point that not calculated based on the event target (element which mouse left).
Case A: When the new element (which mouse gonna enter) is `AllowDrop="true"`, that Point is calculated based on the new element.
Case B: If the new element is `AllowDrop='false'`, that Point will be raw point in screen coordinates.

This bug causes [DragEventArgs.GetPosition(IInputElement)](https://learn.microsoft.com/en-us/dotnet/api/system.windows.drageventargs.getposition?view=windowsdesktop-10.0) to return meaningless Point when you get such a bad DragEventArgs. Since the `DragEventArgs._dropPoint` stored is not in the coordinate system of `DragEventArgs._target` that used to perform transformation.

Root cause：
https://github.com/dotnet/wpf/blob/6ebec6211a69e7d86e24116b29b9e5d571f8f9e3/src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/DragDrop.cs#L955-L1035

As you can see, there are two way to raise a `DragDrop.DragLeaveEvent`, located in line 978 and line 1018 separately.
Both of them used variable `_lastTarget` and `targetPoint`.
However, `targetPoint` was calced with `target` by calling `GetCurrentTarget` in line 963, which means `targetPoint` is associated with `target`.
And that explains why Case A happened.

Then take a look at the function `GetCurrentTarget`.
https://github.com/dotnet/wpf/blob/6ebec6211a69e7d86e24116b29b9e5d571f8f9e3/src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/DragDrop.cs#L1288-L1366

[...]

PR #11781, opened 2026-07-14
Solution (#11774): DragLeave constructs DragEventArgs with coordinates relative to t

TFGSUMIT:
Title: Fix `DragLeave` event coordinates issue

Description:
This pull request fixes an issue where the `DragLeave` event constructs `DragEventArgs` with coordinates relative to the new target instead of the event target. The issue is fixed by introducing a new method `TransformPointToTarget` that transforms the target point to the coordinate system of the event target. This method is used in the `RaiseDragEvent` method to ensure that the coordinates in the `DragEventArgs` are correct.

Changes:
* Introduced `TransformPointToTarget` method in `DragDrop` class
* Updated `RaiseDragEvent` method to use `TransformPointToTarget` method
* Updated `DragEventArgs` class to use transformed point

Testing instructions:
1. Run the application with the `DragLeave` event handler attached.
2. Verify that the coordinates in the `DragEventArgs` are correct in the `DragLeave` event handler.
3. Verify that the coordinates are relative to the event target, not the new target.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/wpf/pull/11781)

### DIFF
--- a/path/to/DragDrop.cs
+++ b/path/to/DragDrop.cs
@@ -1,0 +1,38 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+using System;
+using System.Windows;
+using System.Windows.Input;
+
+namespace Microsoft.DotNet.Wpf.PresentationCore
+{
+    public class DragDrop
+    {
+        // ... existing code ...
+
+        private static Point TransformPointToTarget(Point targetPoint, DependencyObject target)
+        {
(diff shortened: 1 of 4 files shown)


h3xds1nz:
Auto generated AI slop PR without an actual solution. Plus it looks like the tokens ran out.

Aaand there's #11800; closing.

PR #11800, opened 2026-07-27
DragDrop: construct DragEventArgs correctly when raising DragLeave event (closes #11774)

### LINKED ISSUE #11774: DragLeave constructs DragEventArgs with coordinates relative to the new target instead of the event target

redgreenlight802:
Fixes #11774

## Issue summary

`DragEventArgs` is constructed with incorrect point when raising `DragLeave` event.

## What Changed

1. `GetCurrentTarget` now return client point instead of target-related point.
2. Add function `TranslateClientPoint` to translate client point obtained from `GetCurrentTarget` to specified target point.
3. Modified `OleDragEnter`/`OleDragOver`/`OleDrop` to adapt to the change of `GetCurrentTarget`. They will now call `TranslateClientPoint` before using `RaiseDragEvent` to raise a drag-drop event.

## Why

We need both `target`-related point and `_lastTarget`-related point at the same time.
So the minimal change might be separating the client point translate logic and only do that when the element used to raise `DragLeave` is determined.

Why not just using `InputElement.TranslatePoint(targetPoint, target, _lastTarget)`:
In issue case B, `targetPoint` is a raw client point, not `target`-related. The translate logic in `TranslateClientPoint` is needed.

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/wpf/pull/11800)

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/DragDrop.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/DragDrop.cs
@@ -912,7 +912,7 @@ namespace System.Windows
         int UnsafeNativeMethods.IOleDropTarget.OleDragEnter(object data, int dragDropKeyStates, long point, ref int effects)
         {
             DependencyObject target;
-            Point targetPoint;
+            Point clientPoint;
 
             // Get the data object and immediately return if there isn't the data object or no available data.
             _dataObject = GetDataObject(data);
@@ -925,7 +925,7 @@ namespace System.Windows
             }
 
             // Get the current target from the mouse drag point that is based on screen.
-            target = GetCurrentTarget(point, out targetPoint);
+            target = GetCurrentTarget(point, out clientPoint);
 
             // Set the last target element with the current target.
             _lastTarget = target;
@@ -938,7 +938,7 @@ namespace System.Windows
                     dragDropKeyStates,
                     ref effects,
                     target,
-                    targetPoint);
+                    TranslateClientPoint(clientPoint, target));
             }
             else
             {
@@ -955,12 +955,12 @@ namespace System.Windows
         int UnsafeNativeMethods.IOleDropTarget.OleDragOver(int dragDropKeyStates, long point, ref int effects)
         {
             DependencyObject target;
-            Point targetPoint;
+            Point clientPoint;
 
             Invariant.Assert(_dataObject != null);
 
             // Get the current target from the mouse drag point that is based on screen.
-            target = GetCurrentTarget(point, out targetPoint);
+            target = GetCurrentTarget(point, out clientPoint);
 
             // Raise DragOver event to the target to get DragDrop effect status from the target.
             if (target != null)
@@ -980,7 +980,7 @@ namespace System.Windows
                                 dragDropKeyStates,
                                 ref effects,
                                 _lastTarget,
-                                targetPoint);
+                                TranslateClientPoint(clientPoint, _lastTarget));
                         }
 
                         // Raise DragEnter event to the new target.
@@ -989,7 +989,7 @@ namespace System.Windows
                             dragDropKeyStates,
                             ref effects,
                             target,
-                            targetPoint);
+                            TranslateClientPoint(clientPoint, target));
                     }
                     finally
                     {
@@ -1005,7 +1005,7 @@ namespace System.Windows
                         dragDropKeyStates,
                         ref effects,
                         target,
-                        targetPoint);
+                        TranslateClientPoint(clientPoint, target));
                 }
             }
             else
@@ -1020,7 +1020,7 @@ namespace System.Windows
                             dragDropKeyStates,
                             ref effects,
                             _lastTarget,
-                            targetPoint);
+                            TranslateClientPoint(clientPoint, _lastTarget));
                     }
                 }
                 finally
@@ -1074,7 +1074,7 @@ namespace System.Windows
         {
             IDataObject dataObject;
             DependencyObject target;
-            Point targetPoint;
+            Point clientPoint;
 
             // Get the data object and then immediately return fail if there isn't the proper data.
             dataObject = GetDataObject(data);
@@ -1089,7 +1089,7 @@ namespace System.Windows
             _lastTarget = null;
 
             // Get the current target from the screen mouse point.
-            target = GetCurrentTarget(point, out targetPoint);
+            target = GetCurrentTarget(point, out clientPoint);
 
             // Raise Drop event to the target element.
             if (target != null)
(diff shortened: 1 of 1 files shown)


redgreenlight802:
@dotnet-policy-service agree

PR #11816, opened 2026-08-04
Fix Nullable<T> detection inside MetadataLoadContext to emit proper BamlRecordType

### LINKED ISSUE #11815: PresentationBuildTasks emits `BamlRecordType.Property` (5) instead of `PropertyWithConverter` (36) for `Nullable<T>` properties
### Description

Two `PresentationBuildTasks` compile the same XAML differently. For a `Nullable<T>` property, the in-box .NET Framework one (4.8, from the GAC) writes `BamlRecordType.PropertyWithConverter` (36), while the one shipped in the .NET SDK writes `BamlRecordType.Property` (5). Non-nullable properties are identical (record 36) in both.

Record 36 carries a trailing `converterTypeId`; record 5 does not. WPF's loader reads both, so it is invisible at runtime, but it breaks tools that read or rewrite compiled BAML.

The record type is decided by **project style** (SDK-style vs non-SDK), **not** by `TargetFramework` and **not** by the build runner: an SDK-style `net472` project yields record 5 via both `dotnet build` and `MSBuild.exe`.

### Reproduction Steps

The attached solution (**[dxSample.zip](https://github.com/user-attachments/files/30696634/dxSample.zip)**) compiles one two-line XAML twice and prints the record type each compiler emitted.

`NullableProbe.xaml`:

```xml
<SolidColorBrush x:Key="plain" Color="#FF112233" />   <!-- Color   (struct)    -> control -->
<ColorAnimation  x:Key="anim"  To="#FF445566" />       <!-- Color?  (Nullable<T>) -> bug    -->
```

Run `build.cmd`, which:

1. builds `Repro.Legacy` (non-SDK, `TargetFrameworkVersion v4.7.2`) with `MSBuild.exe` → in-box `PresentationBuildTasks 4.0.0.0` from the GAC;
2. builds `Repro.Sdk` (SDK-style, `net8.0-windows`) with `dotnet build` → `PresentationBuildTasks` from the .NET SDK;
3. `Check` reads the compiled `.baml` out of each assembly's `*.g.resources` and reports the record type byte for each value.

### Expected behavior

Both compilers emit `PropertyWithConverter` (36) for the `Nullable<T>` property, as the in-box compiler does:

| property | type | .NET Framework | .NET SDK |
|---|---|---|---|
| `SolidColorBrush.Color` | `Color` | 36 | 36 |
| `ColorAnimation.To` | `Color?` | 36 | **36** |

### Actual behavior

The SDK compiler writes plain `Property` (5) for the nullable property:

| property | type | .NET Framework | .NET SDK |
|---|---|---|---|
| `SolidColorBrush.Color` | `Color` | 36 | 36 |
| `ColorAnimation.To` | `Color?` | 36 | **5** |

This affects every `Nullable<T>` property, not just colors (`DoubleAnimation.To`, `double?`, behaves the same).

`Check` also demonstrates the root cause directly, resolving `ColorAnimation.To` in a `MetadataLoadContext`:
```
GetGenericTypeDefinition() == typeof(Nullable<>) : False   (should be True)
IsNullableType(...)                              : False
runtime typeof(Nullable<int>) IsNullableType     : True    (sanity check)
```

### Regression?

No. The behavior differs between the .NET Framework and the .NET (SDK) WPF compilers.

### Known Workarounds

None. Keeping projects non-SDK-style avoids it, but is not viable long-term.

### Impact

This affects **every** `Nullable<T>` property and breaks any tool that reads or rewrites compiled BAML — localizers, theme/palette generators, and other post-build BAML processors — because the record layout no longer matches what the .NET Framework compiler produced. 

In DevExpress WPF it silently corrupts runtime theming, producing wrong colors on .NET while the identical markup is correct on .NET Framework. The divergence is also a migration hazard: the .NET Framework build is correct only because it is still a non-SDK project, so moving those projects to SDK-style would regress it the same way.

### Configuration

* .NET SDK 10.0.302 (reproduces with the SDK's `tools\net472` task, i.e. independent of `TargetFramework`, and independent of `dotnet build` vs `MSBuild.exe` — what matters is SDK-style vs non-SDK project)
* In-box `PresentationBuildTasks` 4.8.9032.0
* Windows 11

### Possible cause

`ReflectionHelper.IsNullableType` (`Shared/System/Windows/Markup/ReflectionHelper.cs`) compares by `Type` identity:

```csharp
internal static bool IsNullableType(Type type) =>
    type.IsGenericType && type.GetGenericTypeDefinition() == typeof(Nullable<>);
```

Under the SDK compiler `type` comes from a reflection-only context, so `GetGenericTypeDefinition()` returns that context's `Nullable<>`, not the runtime `typeof(Nullable<>)`, and `==` is `false` (see part 2 of `Check`). The `PBTCOMPILER` branch in `GetKnownConverterTypeFromType` (`BamlMapTable.cs`) then never returns `NullableConverter`, and `WriteProperty` (`BamlRecordWriter.cs`) falls through to the plain record 5. The 4.8 compiler runs on .NET Framework against the same WPF assemblies it loaded (runtime types), so there the comparison is `true` and record 36 is written.

h3xds1nz:
Fixes #11815 

## Description

Fixes detection of Nullable<T> types on SDK-style builds where a new API for detecting `Nullable<T>` types was added (https://github.com/dotnet/runtime/issues/124216). `GetNullableUnderlyingType` checks the `IsGenericType` for us.

## Customer Impact

Well, the customer impact depends on how many years does it take for a PR to land in dotnet/wpf.

## Regression

Regression for SDK-style projects. Otherwise this has been the case since `MetadataLoadContext` is available (which is the time when WPF was opensourced).

## Testing

The repro included in the original issue. Using `dotnet build` for any TFM after swapping PBT in SDK.

Also verified builds in Visual Studio which use the net472 build of PBT.

## Risk

There shouldn't be any, though I'm unsure at the moment if we can surface a non-in-box `Type` to check on inside WPF code so that such code would need to override this new method to not receive a random `NotSupportedException`. Eitherway, all inbox types ship with overrides and if a custom type override wants to work with nullables, they should override it.

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/wpf/pull/11816)

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/Shared/System/Windows/Markup/ReflectionHelper.cs
+++ b/src/Microsoft.DotNet.Wpf/src/Shared/System/Windows/Markup/ReflectionHelper.cs
@@ -46,7 +46,9 @@ namespace System.Xaml
 #if PBTCOMPILER
         // System.Reflection.MetadataLoadContext instance
         private static MetadataLoadContext _metadataLoadContext = null;
-
+#if !NET11_0_OR_GREATER
+        private static Type s_nullableType = null;
+#endif
         // MetadataLoadContext Assembly cache
         private static Dictionary<string, Assembly> _cachedMetadataLoadContextAssemblies = null;
         private static Dictionary<string, Assembly> _cachedMetadataLoadContextAssembliesByNameNoExtension = null;
@@ -122,7 +124,10 @@ namespace System.Xaml
 #if NET11_0_OR_GREATER
             return type.GetNullableUnderlyingType() is not null;
 #else
-            return type.IsGenericType && (type.GetGenericTypeDefinition() == typeof(Nullable<>));
+            s_nullableType ??= _metadataLoadContext.CoreAssembly?.GetType("System.Nullable`1");
+            Debug.Assert(s_nullableType is not null, "System.Nullable`1 type not found in MetadataLoadContext");
+
+            return type.IsGenericType && s_nullableType.Equals(type.GetGenericTypeDefinition());
 #endif
         }
 
--- a/src/Microsoft.DotNet.Wpf/src/Shared/System/Windows/Markup/ReflectionHelper.cs
+++ b/src/Microsoft.DotNet.Wpf/src/Shared/System/Windows/Markup/ReflectionHelper.cs
@@ -1,4 +1,4 @@
-// Licensed to the .NET Foundation under one or more agreements.
+﻿// Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the MIT license.
 
 #nullable disable
@@ -119,7 +119,11 @@ namespace System.Xaml
 
         internal static bool IsNullableType(Type type)
         {
-            return (type.IsGenericType && (type.GetGenericTypeDefinition() == typeof(Nullable<>)));
+#if NET11_0_OR_GREATER
+            return type.GetNullableUnderlyingType() is not null;
+#else
+            return type.IsGenericType && (type.GetGenericTypeDefinition() == typeof(Nullable<>));
+#endif
         }
 
         internal static bool IsInternalType(Type type)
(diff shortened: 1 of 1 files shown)


lindexi:
Oh, my experiment failed and I left some junk on your PR — turns out my WpfRuntime project hasn't been upgraded to .NET 11 yet. I'm still maintaining the older version and haven't added multi-targeting support. So when building, anything outside `#if PBTCOMPILER` can't find `_metadataLoadContext` and the build just breaks.

h3xds1nz:
@lindexi No worries, yeah this assumes that the `ReflectionHelper` is only built for .NET 11 or in PBT context for NETFX; I don't see a reason to complicate this since it doesn't count on any backporting.

PR #11837, opened 2026-08-12
Restoring autoConvert behavior in Clipboard APIs

Shyam-Gupta:
Fixes one issue in https://github.com/dotnet/winforms/issues/14322

## Description

Till .NET 9.0, `autoConvert` was set to true only for `FileDrop` and `Bitmap` scenarios. However, since .NET 10, it is being set to true for `GetAudioStream`, `GetFileDropList`, `GetImage`, and `GetText` APIs inside the private method `GetTypedDataIfAvailable`. This causes issues with GetText API with mojibake characters as described in bug https://github.com/dotnet/winforms/issues/14322

## Customer Impact

There is no way for the app to know whether the clipboard contents were corrupted or if there is any problem with the API as it didn't get the desired data.

## Regression

Yes - from .NET 9.0

## Testing

Created a WPF app to write unicode data on the clipboard and another app to read it. Before the changes, the reader app returned several hits for mojibake characters. After fix, there were none.

## Risk

This should be a very low risk change as this behavior was working fine in 9.0 and previous versions.
 
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/wpf/pull/11837)

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/clipboard.cs
+++ b/src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/clipboard.cs
@@ -487,7 +487,7 @@ public static class Clipboard
             return false;
         }
 
-        return dataObject.TryGetData(format, out data);
+        return dataObject.TryGetData(format, autoConvert: false, out data);
     }
 
     /// <inheritdoc cref="DataObject.SetDataAsJson{T}(string, T)"/>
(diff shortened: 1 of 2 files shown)


Shyam-Gupta:
Requesting @JeremyKuhne for review.

JeremyKuhne (approved):

Shyam-Gupta:
For reference, similar commit for WinForms: https://github.com/dotnet/winforms/pull/14893

cc: @dipeshmsft, @pranav-gupta-msft

dipeshmsft (approved):

harshit7962 (approved):
LGTM

PR #11864, opened 2026-08-24
Fix Binding errors when Ribbon is outside of RibbonWindow

### LINKED ISSUE #11860: Binding errors when using a Ribbon outside of a RibbonWindow
### Description

When using the Ribbon control in a regular Window (or any Window that is not a RibbonWindow) using the default template, there are two Binding errors in the output.

```
System.Windows.Data Error: 4 : Cannot find source for binding with reference 'RelativeSource FindAncestor, AncestorType='System.Windows.Controls.Ribbon.RibbonWindow', AncestorLevel='1''. BindingExpression:Path=WindowState; DataItem=null; target element is 'Ribbon' (Name=''); target property is 'NoTarget' (type 'Object')
System.Windows.Data Error: 4 : Cannot find source for binding with reference 'RelativeSource FindAncestor, AncestorType='System.Windows.Controls.Ribbon.RibbonWindow', AncestorLevel='1''. BindingExpression:Path=IsActive; DataItem=null; target element is 'Ribbon' (Name=''); target property is 'NoTarget' (type 'Object')
```

These two errors appear harmless but can fill up the logs for the applications that keep track of Binding errors, like our applications do.

These errors do not appear when using the Ribbon in a RibbonWindow, but from my understanding using a Ribbon outside of a RibbonWindow still seems to be a valid scenario.

### Reproduction Steps

Just use a Ribbon in a regular Window like this:

```
<Window x:Class="WpfApp1.MainWindow"
        xmlns:local="clr-namespace:WpfApp1"
    <Grid>
        <!-- Ribbon that causes the error -->
        <Ribbon />
    </Grid>
</Window>
```

### Expected behavior

As far as I can tell using a Ribbon outside of a RibbonWindow is a supported scenario, so I would expect to not have any error.

### Actual behavior

There are two binding errors in the output:
```
System.Windows.Data Error: 4 : Cannot find source for binding with reference 'RelativeSource FindAncestor, AncestorType='System.Windows.Controls.Ribbon.RibbonWindow', AncestorLevel='1''. BindingExpression:Path=WindowState; DataItem=null; target element is 'Ribbon' (Name=''); target property is 'NoTarget' (type 'Object')
System.Windows.Data Error: 4 : Cannot find source for binding with reference 'RelativeSource FindAncestor, AncestorType='System.Windows.Controls.Ribbon.RibbonWindow', AncestorLevel='1''. BindingExpression:Path=IsActive; DataItem=null; target element is 'Ribbon' (Name=''); target property is 'NoTarget' (type 'Object')
```

### Regression?

Not that I am aware of. I only tested on .NET 8 and 10 but from the looks of it, I don't think it is a regression.

### Known Workarounds

Either:
 - Overwrite the ControlTemplate for the Ribbon with a fixed one (the errors come from the default template, see "Other information")
 - Ignore the errors

### Impact

No noticeable impact for the end user, but it can fill up the logs for the applications that keep track of all the Binding errors.

### Configuration

.NET 10.0.111, on Windows 11 x64

But I don't think it really matters for this issue.

### Other information

The errors seem to come from the default template for the Ribbon, there are two conditions that try to bind on the parent RibbonWindow even when there is no parent RibbonWindow [there](https://github.com/dotnet/wpf/blob/1cfc37f708f91ff4556bd25af414546c446f3a16/src/Microsoft.DotNet.Wpf/src/System.Windows.Controls.Ribbon/Themes/Generic.xaml#L2844) and [there](https://github.com/dotnet/wpf/blob/1cfc37f708f91ff4556bd25af414546c446f3a16/src/Microsoft.DotNet.Wpf/src/System.Windows.Controls.Ribbon/Themes/Generic.xaml#L2903).

We already have a fix internally and we intend to open a PR shortly to propose our fix.

vchaillou-arpege:
Fixes #11860

## Description

The Binding errors happened when using the default template for the Ribbon. It has two conditions that may try to bind to the parent RibbonWindow but it did not check if there was a parent RibbonWindow before, so the two Bindings ended up failing if the Ribbon was outside of a RibbonWindow.

This fix ensures that the Ribbon is hosted in a RibbonWindow before binding to the parent RibbonWidow by having a condition that binds to `IsHostedInRibbonWindow` first. if `IsHostedInRibbonWindow == false`, the following conditions are not evaluated.

FYI, this was already done this way [there](https://github.com/dotnet/wpf/blob/1cfc37f708f91ff4556bd25af414546c446f3a16/src/Microsoft.DotNet.Wpf/src/System.Windows.Controls.Ribbon/Themes/Generic.xaml#L2742) (and [there](https://github.com/dotnet/wpf/blob/1cfc37f708f91ff4556bd25af414546c446f3a16/src/Microsoft.DotNet.Wpf/src/System.Windows.Controls.Ribbon/Themes/XAML/Ribbon.xaml#L228)).

## Customer Impact

Without this PR, there will still be the Binding errors as described in #11860.

## Regression

No (or at least not that I am aware of).

## Testing

For now, I only tested with the simple case linked in #11860.

## Risk

I believe that there are no real risks given the changes I made but correct me if I am wrong.

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/wpf/pull/11864)

### DIFF
--- a/src/Microsoft.DotNet.Wpf/src/System.Windows.Controls.Ribbon/Themes/Generic.xaml
+++ b/src/Microsoft.DotNet.Wpf/src/System.Windows.Controls.Ribbon/Themes/Generic.xaml
@@ -2841,13 +2841,13 @@ To automatically copy the files, set the environment variable
                     </Trigger>
                     <MultiDataTrigger>
                         <MultiDataTrigger.Conditions>
-                            <Condition Binding="{Binding RelativeSource={RelativeSource Mode=FindAncestor, AncestorType=ribbon:RibbonWindow}, Path=IsActive}" Value="False"/>
 
                             <Condition Binding="{Binding Path=(SystemParameters.IsGlassEnabled)}" Value="False"/>
 
 
 
                             <Condition Binding="{Binding RelativeSource={RelativeSource Self}, Path=IsHostedInRibbonWindow}" Value="True" />
+                            <Condition Binding="{Binding RelativeSource={RelativeSource Mode=FindAncestor, AncestorType=ribbon:RibbonWindow}, Path=IsActive}" Value="False"/>
                         </MultiDataTrigger.Conditions>
                         <Setter TargetName="PART_TitleHost" Property="TextElement.Foreground" Value="{DynamicResource {x:Static SystemColors.InactiveCaptionTextBrushKey}}" />
                     </MultiDataTrigger>
@@ -2900,6 +2900,7 @@ To automatically copy the files, set the environment variable
 
 
 
+                            <Condition Binding="{Binding RelativeSource={RelativeSource Self}, Path=IsHostedInRibbonWindow}" Value="True" />
                             <Condition Binding="{Binding RelativeSource={RelativeSource Mode=FindAncestor, AncestorType=ribbon:RibbonWindow}, Path=WindowState}" Value="Maximized"/>
                         </MultiDataTrigger.Conditions>
                         <Setter TargetName="QatTopHost" Property="Margin" Value="0"/>
--- a/src/Microsoft.DotNet.Wpf/src/System.Windows.Controls.Ribbon/Themes/Generic.xaml.sdk
+++ b/src/Microsoft.DotNet.Wpf/src/System.Windows.Controls.Ribbon/Themes/Generic.xaml.sdk
@@ -2717,13 +2717,13 @@ Theme Styles For Windows Presentation Foundation Version 
                     </Trigger>
                     <MultiDataTrigger>
                         <MultiDataTrigger.Conditions>
-                            <Condition Binding="{Binding RelativeSource={RelativeSource Mode=FindAncestor, AncestorType=ribbon:RibbonWindow}, Path=IsActive}" Value="False"/>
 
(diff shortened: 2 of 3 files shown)


vchaillou-arpege:
@dotnet-policy-service agree company="Arpège"

miloush:
`RibbonWindow` inherits `Windows` which is where the `IsActive` and `WindowState` properties are defined, why does the binding just not look for a `Window`?

vchaillou-arpege:
> `RibbonWindow` inherits `Windows` which is where the `IsActive` and `WindowState` properties are defined, why does the binding just not look for a `Window`?

I guess this is because the `Ribbon` behaves slightly differently inside and outside of `RibbonWindow`. The two triggers that this PR fixes set properties on two elements that are placed differently in a `RibbonWindow` (`PART_TitleHost` and `QatTopHost` which is part of the Quick Access Toolbar) so I assume that these conditions are only valid inside `RibbonWindows` specifically. At least, that is the current behavior.

In a regular `Window`:
image

In a `RibbonWindow`:
image

Since this PR is solely about fixing Binding errors, I would rather not add any unneccessary change or any behavior change. So unless this is a requirement for merging, I would prefer not to touch these Bindings and keep the changes as minimal as possible.

